PR #5, opened 2022-08-03
Closure bufferization

chelini:


chelini:
I think we should get rid of the closure before bufferization.

PR #8, opened 2022-08-21
block conv

chelini:


chelini:
For some reason collapsing on the filter does now work as `inversePermutation` cannot be computed. Not sure why.

PR #11, opened 2022-09-08
BF16 support for matmul ops

KavithaTipturMadhu:


lib/Standalone/ConvertTppToXsmm.cpp
```diff
@@ -67,8 +67,16 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
         rewriter.getContext(), ArrayRef<int64_t>{m, n, k, lda, ldb, ldc});
     xsmm::TernaryKindAttr attr = xsmm::TernaryKindAttr::get(
         matmulOp.getContext(), xsmm::TernaryKind::MATMUL);
+    xsmm::DataTypeAttr dtype;
+    if(memrefC.getElementType().isBF16()){
+      dtype = xsmm::DataTypeAttr::get(
+        matmulOp.getContext(), xsmm::DataType::BF16);
+    }else{
+      dtype = xsmm::DataTypeAttr::get(
```
chelini, line 75: Can you please assert if the element type is not a `f32` here?

lib/Standalone/ConvertTppToXsmm.cpp
```diff
@@ -115,8 +123,17 @@ struct ConvertTppBrgemmOp : public OpRewritePattern<BrgemmOp> {
         rewriter.getContext(), ArrayRef<int64_t>{m, n, k, lda, ldb, ldc});
     xsmm::TernaryKindAttr attr = xsmm::TernaryKindAttr::get(
         brgemmOp.getContext(), xsmm::TernaryKind::BRGEMM);
+    xsmm::DataTypeAttr dtype;
+    if(memrefC.getElementType().isBF16()){
+      dtype = xsmm::DataTypeAttr::get(
+        brgemmOp.getContext(), xsmm::DataType::BF16);
+    }else{
```
chelini, line 130: see above

lib/Standalone/ConvertXsmmToFunc.cpp
```diff
@@ -65,6 +64,9 @@ static SmallVector<Value, 4> getMemRefOperands(OpBuilder &b, Location loc,
 static LogicalResult buildInvokeCall(Location loc, std::string funcName,
                                      Operation *op, PatternRewriter &rewriter) {
   funcName = "xsmm_" + funcName + "_invoke";
+  if(op->getOperand(1).getType().cast<MemRefType>().getElementType().isBF16()){
```
chelini, line 67: please drop trivial braces. MLIR does not uses it for single stmt. Let's follow a similar approach.

lib/Standalone/ConvertXsmmToFunc.cpp
```diff
@@ -179,7 +181,11 @@ struct ConvertTernaryDispatch : public OpRewritePattern<TernaryDispatchOp> {
                                 PatternRewriter &rewriter) const override {
     Location loc = dispatchOp.getLoc();
     std::string kindAsString = stringifyEnum(dispatchOp.getKind()).str();
+    std::string typeAsString = stringifyEnum(dispatchOp.getDataType()).str();
     kindAsString = "xsmm_" + kindAsString + "_dispatch";
+    if(typeAsString == "bf16"){
```
chelini, line 186: see above

standalone-rt/XsmmRunnerUtils.cpp
```diff
@@ -19,25 +19,73 @@ extern "C" void _mlir_ciface_xsmm_matmul_invoke(int64_t funcAddr,
...
-  // std::cout << "matrix B: \n";
-  // printMemRefMetaData(std::cout, DynamicMemRefType<float>(*B));
-  // std::cout << "\n";
-  // std::cout << "matrix C: \n";
-  // printMemRefMetaData(std::cout, DynamicMemRefType<float>(*C));
-  // std::cout << "\n";
-  // std::cout << "funcAddr: " << funcAddr << "\n";
-
+/*   std::cout << "matrix A: \n";
+   printMemRefMetaData(std::cout, DynamicMemRefType<void>(*A));
```
chelini, line 23: Can we put it back to `float` instead of `void`?

standalone-rt/XsmmRunnerUtils.cpp
```diff
@@ -19,25 +19,73 @@ extern "C" void _mlir_ciface_xsmm_matmul_invoke(int64_t funcAddr,
...
+  sgemm.gemm = reinterpret_cast<libxsmm_gemmfunction>(funcAddr);
+  sgemm.gemm(&gemm_param);
+}
+
+extern "C" void _mlir_ciface_xsmm_matmul_invoke_bf16(int64_t funcAddr,
+                                                UnrankedMemRefType<bf16> *A,
+                                                UnrankedMemRefType<bf16> *B,
+                                                UnrankedMemRefType<bf16> *C) {
+/*   std::cout << "matrix A: \n";
+   printMemRefMetaData(std::cout, DynamicMemRefType<void>(*A));
```
chelini, line 71: Also here why do we use `void`?

test/Integration/matmul_bf16.mlir
```diff
@@ -0,0 +1,72 @@
...
+        %1 = arith.addf %c, %0 : bf16
+        linalg.yield %1 : bf16
+    } -> tensor<4x4xbf16>
+    return %D : tensor<4x4xbf16>
+  }
+
+  func.func @entry() {
+    %c0 = arith.constant 0 : index
+
+    // Initialize various matrices, dense for stress testing,
```
chelini, line 30: Please fix the comment we are not doing sparse computation.

test/Integration/matmul_bf16.mlir
```diff
@@ -0,0 +1,72 @@
...
+    %0 = call @matmultpp(%da, %db, %C)
+       : (tensor<4x8xbf16>, tensor<8x4xbf16>, tensor<4x4xbf16>) -> tensor<4x4xbf16>
+
+    //
+    // CHECK:    ( ( 388, 426, 462, 500),
+    // CHECK-SAME: ( 396, 434, 472, 510 ),
+    // CHECK-SAME: ( 406, 444, 484, 520 ),
+    // CHECK-SAME: ( 414, 454, 492, 532 ) )
+    //
+    // %m0 = bufferization.to_memref %0 : memref<4x4xbf16>
```
chelini, line 60: Also drop commented code if not used.

chelini (commented):
Overall ok. Two main issues:
1. Please run `git-clang-format HEAD~1` to properly format the code
2. If we pass an `int` to represent the type into the dispatch/invoke can we avoid code duplication?

KavithaTipturMadhu:
> 2. If we pass an int to represent the type into the dispatch/invoke can we avoid code duplication?
We can't pass DynamicMemref<void> type to invoke as there's a base address+offset computation which can't be done on void pointers. We could use the same non-void base pointer type and add an offset based on input but that's ugly. We can use an extra int argument only in dispatch but then it becomes inconsistent with the invoke signature. What would you recommend?

chelini:
Ok thanks. For the time being les't keep it as it is. Later we need to find better solution and use `void` pointers.

chelini (approved):

PR #13, opened 2022-09-08
fix: Overloading of xsmm invoke calls doesn't work

KavithaTipturMadhu:


include/Standalone/Dialect/Xsmm/XsmmOps.td
```diff
@@ -46,6 +46,15 @@ def Xsmm_TernaryOp : Xsmm_Op<"ternary", []> {
     }
     operand_iterator arg_operand_begin() { return operand_begin(); }
     operand_iterator arg_operand_end() { return operand_end(); }
+
+    std::string getOperandType(){
+    	auto operand = std::next(arg_operand_begin());
+        Type type = (*operand).getType().isa<MemRefType>()?(*operand).getType().cast<MemRefType>().getElementType(): (*operand).getType();
+        if(type.isBF16())
+		return "bf16";
+	assert(type.isF32());
```
chelini, line 55: Please add a message on the assert: `assert(types.isF32() && "expect bf16 or f32")`

lib/Standalone/ConvertXsmmToFunc.cpp
```diff
@@ -19,6 +19,7 @@ using namespace mlir::xsmm;
 
 #define GEN_PASS_CLASSES
 #include "Standalone/Passes.h.inc"
+#include <iostream>
```
chelini, line 22: Drop this.

lib/Standalone/ConvertXsmmToFunc.cpp
```diff
@@ -96,8 +96,15 @@ struct ConvertTernaryXsmmOp : public OpRewritePattern<TernaryOp> {
 
   LogicalResult matchAndRewrite(TernaryOp ternaryOp,
                                 PatternRewriter &rewriter) const override {
+    std::string funcName =
+        "xsmm_" + stringifyEnum(ternaryOp.getCallee()).str() + "invoke";
+    if (ternaryOp.getOperandType() == "bf16")
+      funcName = funcName + "_bf16";
+    else
+      assert(ternaryOp.getOperandType() == "f32";
```
chelini, line 104: missing a parenthesis and a message.

chelini: Why for bf16 we add a `_bf16` while for f32 only `f32` should we add `_f32`? (to be consistent)

chelini (approved):
Pleas run `git-clang-format HEAD~1`

lib/Standalone/ConvertXsmmToFunc.cpp
KavithaTipturMadhu: do you mean the xsmm call should have an _f32 suffix? The above is just a type check, and there are no underscores. Only when composing the function call name, we are appending _bf16.

KavithaTipturMadhu:
> Pleas run `git-clang-format HEAD~1`

Is there some way to run git clang format on .td files? There don't seem to be any changes after running this.

chelini:
No there is not way (unfortunately) to format this. We need to do manually.

lib/Standalone/ConvertXsmmToFunc.cpp
chelini: Yeah, I would make a _f32 suffix too. Just to be consistent.

chelini (approved):
Thanks. Much better! Please run clang-format and fix remaining comments. Feel free to merge after that! Cool.

PR #20, opened 2022-09-19
Update cmake script to be buildable within IREE

nicolasvasilache:


rengolin:
@hfp @chelini This are the changes needed to build the sandbox inside or IREE. How can we test this in the CI before merging?

PR #22, opened 2022-09-20
Revisit the TPP build tree to work both standalone and with IREE

nicolasvasilache:
This revision refactors the cmake build tree to be more in line with upstream best practices.
A few issues surfaced when trying to integrate this deeply with IREE.

Local testing with both IREE and standalone TPP is successful with these changes.


chelini (approved):
Tested on CI and my machine and it works.

PR #24, opened 2022-09-26
3 dimensional input for packed bf16

KavithaTipturMadhu:
Even round-trippability seems broken with this, I see no errors or outputs on running standalone-opt on the committed mlir file.

### DIFF
--- a/lib/Standalone/ConvertTppToLoops.cpp
+++ b/lib/Standalone/ConvertTppToLoops.cpp
@@ -220,6 +220,8 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
     ArrayRef<int64_t> shapeC = matmulOp.getMatrixCType().getShape();
     ArrayRef<int64_t> shapeA =
         matmulOp.getMatrixA().getType().cast<MemRefType>().getShape();
+    if (shapeA.size() == 3)
+      matmulOp.emitError("Packed BF16 loops unsupported");
     Value i = rewriter.create<arith::ConstantIndexOp>(loc, shapeC[0]);
     Value j = rewriter.create<arith::ConstantIndexOp>(loc, shapeC[1]);
     Value k = rewriter.create<arith::ConstantIndexOp>(loc, shapeA[1]);
--- a/lib/Standalone/ConvertTppToXsmm.cpp
+++ b/lib/Standalone/ConvertTppToXsmm.cpp
@@ -52,6 +52,12 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
     if (failed(ldaDim))
       return failure();
     int64_t lda = *ldaDim;
+    if (memrefA.getElementType().isBF16() && memrefA.getShape().size() == 3) {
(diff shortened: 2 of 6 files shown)


lib/Standalone/ConvertTppToXsmm.cpp
```diff
@@ -47,11 +47,17 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
     MemRefType memrefB = matmulOp.getMatrixBType();
     int64_t m = memrefC.getShape()[0];
     int64_t n = memrefC.getShape()[1];
-    int64_t k = memrefA.getShape()[1];
+    int64_t k = memrefB.getShape()[0];
```
chelini, line 50: why this change?

lib/Standalone/ConvertTppToXsmm.cpp
```diff
@@ -47,11 +47,17 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
...
     auto ldaDim = getLeadingDim(memrefA);
     if (failed(ldaDim))
       return failure();
     int64_t lda = *ldaDim;
+    if (memrefA.getShape().size() == 3) {
```
chelini, line 55: Should we also check the element type in the memref?

lib/Standalone/Dialect/Tpp/TppOps.cpp
```diff
@@ -56,14 +56,15 @@ LogicalResult MatmulOp::verify() {
   MemRefType a = getMatrixA().getType().cast<MemRefType>();
   MemRefType b = getMatrixB().getType().cast<MemRefType>();
   MemRefType c = getMatrixC().getType().cast<MemRefType>();
-  if ((a.getShape().size() != 2) || (b.getShape().size() != 2) ||
-      (c.getShape().size() != 2))
-    return failure();
+  if ((a.getShape().size() != 2 && a.getShape().size() != 3) ||
```
chelini, line 59: Would you mind also adding constraints on the element type?

lib/Standalone/ConvertTppToLoops.cpp
```diff
@@ -220,6 +220,8 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
     ArrayRef<int64_t> shapeC = matmulOp.getMatrixCType().getShape();
     ArrayRef<int64_t> shapeA =
         matmulOp.getMatrixA().getType().cast<MemRefType>().getShape();
+    if (shapeA.size() == 3)
+      matmulOp.emitError("Packed BF16 loops unsupported");
```
chelini, line 224: Can you please add a test to check also this exit? thx/

rengolin:
For some reason, on my x86_64 machine, the two bf16 tests are `UNSUPPORTED`. I had to remove the unsupported line from the tests to run (and they pass).

lib/Standalone/ConvertTppToXsmm.cpp
KavithaTipturMadhu: This is a stray change, I will revert this.

PR #31, opened 2022-09-28
Use more descriptive name in `MapConvToMatmulOp` (NFC)

chelini:
filter_height_pos, and filter_width_pos represent the filter height and weight index, respectively. If the filter height and width are not one, we need to consider a sliding window when mapping to matmul; thus, we need the position of the height and width in the filter.


nicolasvasilache (approved):
thanks! much clearer

PR #40, opened 2022-10-07
Add BF16 test dir and check

rengolin:


test/lit.cfg.py
```diff
@@ -42,10 +42,8 @@
 # excludes: A list of directories to exclude from the testsuite. The 'Inputs'
 # subdirectories contain auxiliary inputs for various tests in their parent
 # directories.
-config.excludes = ['Inputs', 'Examples', 'CMakeLists.txt', 'README.txt', 'LICENSE.txt']
```
chelini, line 45: Why this line is removed?

chelini (approved):

test/lit.cfg.py
rengolin: AFAICS, this line is unused because the the `line.suffixes` above states `mlir` only, and there are no MLIR files on those directories. All tests pass without this line, and even the `stdx-ops.mlir` below work, so I'm not sure how much effect that was having.

Issue #45, opened 2022-10-07
Standalone/stdx-ops.mlir crash


chelini:
```
FAIL: STANDALONE_OPT :: Standalone/stdx-ops.mlir (64 of 64)
******************** TEST 'STANDALONE_OPT :: Standalone/stdx-ops.mlir' FAILED ********************
Script:
--
: 'RUN: at line 1';   /home/lorenzo/tpp-sandbox/build/bin/standalone-opt /home/lorenzo/tpp-sandbox/test/Standalone/stdx-ops.mlir | /home/lorenzo/llvm-project/build/bin/FileCheck /home/lorenzo/tpp-sandbox/test/Standalone/stdx-ops.mlir
--
Exit Code: 2

Command Output (stderr):
--
realloc(): invalid pointer
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: /home/lorenzo/tpp-sandbox/build/bin/standalone-opt /home/lorenzo/tpp-sandbox/test/Standalone/stdx-ops.mlir
1.	MLIR Parser: custom op parser 'func.func'
2.	MLIR Parser: custom op parser 'stdx.closure'
3.	MLIR Parser: custom op parser 'stdx.yield'
 #0 0x000055cb0c56a79a llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/lorenzo/llvm-project/llvm/lib/Support/Unix/Signals.inc:569:11
 #1 0x000055cb0c56a94b PrintStackTraceSignalHandler(void*) /home/lorenzo/llvm-project/llvm/lib/Support/Unix/Signals.inc:636:1
 #2 0x000055cb0c568f96 llvm::sys::RunSignalHandlers() /home/lorenzo/llvm-project/llvm/lib/Support/Signals.cpp:104:5
 #3 0x000055cb0c56b075 SignalHandler(int) /home/lorenzo/llvm-project/llvm/lib/Support/Unix/Signals.inc:407:1
 #4 0x00007f44396af520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #5 0x00007f4439703a7c __pthread_kill_implementation ./nptl/pthread_kill.c:44:76
 #6 0x00007f4439703a7c __pthread_kill_internal ./nptl/pthread_kill.c:78:10
 #7 0x00007f4439703a7c pthread_kill ./nptl/pthread_kill.c:89:10
[...]

PR #46, opened 2022-10-07
exclude test

chelini:


rengolin (approved):

test/lit.cfg.py
```diff
@@ -42,7 +42,7 @@
 # excludes: A list of directories to exclude from the testsuite. The 'Inputs'
 # subdirectories contain auxiliary inputs for various tests in their parent
 # directories.
-config.excludes = [ ]
+config.excludes = [ "stdx-ops.mlir" ]
```
rengolin, line 45: can you add a comment to #45 ?

PR #50, opened 2022-10-10
Generalize pack propagation for elementwise ops

chelini:


lib/Standalone/ToBlockLayoutAndBack.cpp
```diff
@@ -417,138 +433,161 @@ struct BlockConv2DNchwFchwLayout
 };
 
 //===----------------------------------------------------------------------===//
-// SinkBlockLayout
+// PropagateThrElementWiseOp
```
rengolin, line 436: Convoluted name, no comments. Would be nice to have a comment like matmul above.

rengolin (approved):

PR #55, opened 2022-10-11
Improve `transform.pack`

chelini:


hfp:
I will build the missing LLVM version and then just retry the above test.

chelini:
I still don't know how to pass diagnostics across patternRewriter and transform. Not sure if there is a mechanism.

rengolin:
@chelini you just added a new LLVM (#53) which isn't the version in this string. Are you sure this is the version you want?

If so, then we might need another `bump_llvm` PR to account for the code differences.

rengolin:
Weird, the LLVM build in Buildkite used a different LLVM version, how did the `bump_llvm` PR pass without having the right version built?

rengolin (approved):

PR #60, opened 2022-10-12
Pack Conv2DNhwcHwcfOp

chelini:


lib/Standalone/ToBlockLayoutAndBack.cpp
```diff
@@ -209,6 +167,102 @@ static LogicalResult BlockOpPreconditions(linalg::LinalgOp linalgOp) {
...
+      AffineMap::get(/*dims=*/9, /*symbols=*/0, {p1, p2, p3, p4, p5}, ctx);
+  AffineMap mapImg = AffineMap::get(/*dims=*/9, /*symbols=*/0,
+                                    {p1, r1, p3 + r2, p4 + r3, r4}, ctx);
+  AffineMap mapFil =
+      AffineMap::get(/*dims=*/9, /*symbols=*/0, {p2, r1, r2, r3, r4, p5}, ctx);
+  linalg::GenericOp replacementOp = rewriter.create<linalg::GenericOp>(
+      loc, packedOutput.getType(), packedInputs, ValueRange{packedOutput},
+      ArrayRef<AffineMap>{mapImg, mapFil, mapOut},
+      ArrayRef<StringRef>{
+          getParallelIteratorTypeName(), getParallelIteratorTypeName(),
```
rengolin, line 248: this could be cached, at least for code simplicity?

lib/Standalone/ToBlockLayoutAndBack.cpp
```diff
@@ -209,6 +167,102 @@ static LogicalResult BlockOpPreconditions(linalg::LinalgOp linalgOp) {
...
+  Location loc = convOp.getLoc();
+  MLIRContext *ctx = convOp.getContext();
+  SmallVector<Value, 2> reshapedInputTensors;
+
+  SmallVector<Value> inputOperands = convOp.getInputOperands();
+  SmallVector<Value> outputOperands = convOp.getOutputOperands();
+
+  // pack the image and the filter.
+  Value image = inputOperands[0];
+  Value packedImage = toPackLayoutNKPQk(loc, image, tiles[0], rewriter);
```
rengolin, line 224: These don't `replaceAllUses` because the shapes of the following linalgs that need the same tensor don't have the same shape, right?

This should be easy to do a pass later on to de-duplicate the calls.

test/Standalone/convolution2d_nhwc_hwcf.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: standalone-opt -pack-conv2DNhwcHwcf="block-factors=32,32" %s | FileCheck %s
```
rengolin, line 1: Would also be nice to have a second RUN line that goes all the way to TPP and see if the right GEMM call is being made.

rengolin (commented):
Have you discusses the `outDimPerm` idea with IREE folks? Are they on-board with it?

lib/Standalone/ToBlockLayoutAndBack.cpp
chelini: Yes, I will clean up tomorrow.

test/Standalone/convolution2d_nhwc_hwcf.mlir
chelini: I don't have a mapping for this conv.

chelini: We dont' have for the blocked form: https://github.com/plaidml/tpp-sandbox/blob/dc33e62e9c76f01692bfa8f4c5942ee00e356f7f/lib/Standalone/DecomposeConvsToMatmulOrBrgemm.cpp#L793

rengolin (approved):

PR #62, opened 2022-10-13
Drop return for `mapConvToGemm` as it was never used

chelini:
- An early exits in `mapConvToGemm`

- Add more tests

### DIFF
--- a/lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
+++ b/lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
@@ -124,7 +124,7 @@ static bool isConvLike(linalg::LinalgOp linalgOp) {
 }
 
 static bool isInBound(unsigned filterRank, unsigned rPos, unsigned sPos) {
-  if (rPos >= filterRank || sPos >= filterRank)
+  return !(rPos >= filterRank || sPos >= filterRank);
     return false;
   return true;
 }
--- a/lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
+++ b/lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
@@ -125,8 +125,6 @@ static bool isConvLike(linalg::LinalgOp linalgOp) {
 
 static bool isInBound(unsigned filterRank, unsigned rPos, unsigned sPos) {
   return !(rPos >= filterRank || sPos >= filterRank);
-    return false;
-  return true;
 }
 
 FailureOr<linalg::MatmulOp>
--- a/lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
+++ b/lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
@@ -109,24 +109,53 @@ static bool checkMappingToMatmul(linalg::LinalgOp linalgOp) {
   return match;
 }
 
-static bool preConditionsMapConvToGemm(linalg::LinalgOp linalgOp) {
+// Check if the operation looks like a convolution.
+static bool isConvLike(linalg::LinalgOp linalgOp) {
   if (linalgOp.getNumInputs() != 2)
     return false;
   if (linalgOp.getNumOutputs() != 1)
     return false;
-  if (linalgOp.hasBufferSemantics())
+  if (llvm::any_of(linalgOp.getInputAndOutputOperands(),
+                   [](OpOperand *operand) {
+                     return !operand->get().getType().isa<RankedTensorType>();
+                   }))
     return false;
-  return checkMappingToMatmul(linalgOp);
+  return tpp::hasMatmulBody(linalgOp);
+}
+
+static bool isInBound(unsigned filterRank, unsigned rPos, unsigned sPos) {
+  if (rPos >= filterRank || sPos >= filterRank)
+    return false;
+  return true;
 }
 
 FailureOr<linalg::MatmulOp>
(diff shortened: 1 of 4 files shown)


lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
```diff
@@ -109,24 +109,53 @@ static bool checkMappingToMatmul(linalg::LinalgOp linalgOp) {
...
+                   [](OpOperand *operand) {
+                     return !operand->get().getType().isa<RankedTensorType>();
+                   }))
     return false;
-  return checkMappingToMatmul(linalgOp);
+  return tpp::hasMatmulBody(linalgOp);
+}
+
+static bool isInBound(unsigned filterRank, unsigned rPos, unsigned sPos) {
+  if (rPos >= filterRank || sPos >= filterRank)
```
rengolin, line 127:
```suggestion
  return !(rPos >= filterRank || sPos >= filterRank);
```

test/Standalone/transform/transform-convolutions.mlir
```diff
@@ -7,7 +7,7 @@ transform.with_pdl_patterns {
       %0 = transform.structured.match ops{["linalg.conv_2d_nhwc_hwcf"]} in %arg1
       %1 = transform.structured.generalize %0
       %2 = transform.structured.interchange %1 { iterator_interchange = [0, 1, 4, 5, 2, 3, 6] }
-      %3 = transform.structured.map_conv_to_matmul %2 (filter_height_pos = 0, filter_width_pos = 1)
+      transform.structured.map_conv_to_matmul %2 (filter_height_pos = 0, filter_width_pos = 1)
```
rengolin, line 10: You don't need to yield anything, right? The result of that operation will be automatically replaced in the original region automatically, I imagine.

test/Standalone/transform/transform-convolutions.mlir
```diff
@@ -61,3 +61,90 @@ func.func @conv(%i: tensor<14x512x28x28xf32>, %f: tensor<1024x512x1x1xf32>,
...
+
+// -----
+
+transform.with_pdl_patterns {
+^bb0(%arg0: !pdl.operation):
+  sequence %arg0 failures(propagate) {
+    ^bb0(%arg1: !pdl.operation):
+      %0 = transform.structured.match ops{["linalg.conv_2d_nhwc_hwcf"]} in %arg1
+      %1 = transform.structured.generalize %0
+      // expected-error @below {{Could not map to matmul}}
```
rengolin, line 73: Interesting, I imagine that isn't `ConvLike` because it's not in `generic+matmul`, which is ok, but the error may be confusing. What is the error message here? Can we CHECK for it, just to make clear on the tests what they're testing?

chelini: Here the problem is that you did not interchange, so you don't have an inner [parallel, parallel, reduction]. What would you like to check?

test/Standalone/transform/transform-convolutions.mlir
chelini: Yes, it will be replaced. It is unclear what we want to pass from this op (loops, or matmul). For now, I prefer to not return anything. In the future this may change

lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
```diff
@@ -109,24 +109,53 @@ static bool checkMappingToMatmul(linalg::LinalgOp linalgOp) {
...
+                   }))
     return false;
-  return checkMappingToMatmul(linalgOp);
+  return tpp::hasMatmulBody(linalgOp);
+}
+
+static bool isInBound(unsigned filterRank, unsigned rPos, unsigned sPos) {
+  return !(rPos >= filterRank || sPos >= filterRank);
+    return false;
+  return true;
```
rengolin, line 129: Sorry, my bad, I should have selected all lines:
```suggestion
  return !(rPos >= filterRank || sPos >= filterRank);
```

test/Standalone/transform/transform-convolutions.mlir
rengolin: It's not clear from each test what the problem is. Either a comment before the test explaining what's wrong with it, or some check (CHECK or expected-error) with a more clear error message.

If we don't have a clear error message (why it fails), then perhaps we should have?

lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
```diff
@@ -109,24 +109,53 @@ static bool checkMappingToMatmul(linalg::LinalgOp linalgOp) {
   return match;
 }
 
-static bool preConditionsMapConvToGemm(linalg::LinalgOp linalgOp) {
+// Check if the operation looks like a convolution.
+static bool isConvLike(linalg::LinalgOp linalgOp) {
```
rengolin, line 113: For example, this here could return `rewriter.notifyMatchFailure()` directly with a more fine-grained error message?

chelini: Yeah, unclear to me if we can connect this notifyMatchFailure with the transform diagnostic. The error you see comes from the transform not the rewriter.

rengolin: Right, then perhaps a comment for now, per test, as to why things don't work would be fine

PR #66, opened 2022-10-14
bump-llvm 62efe4567b697116610c152fa08c7083aa8df797

chelini:
- Note that we upgrade right before: https://github.com/llvm/llvm-project/commit/1b99f3a224afcab4a08f3b5b32ece63aa22cae0c This commit introduces more copies/allocs on `mlp-packed.mlir`.
- Note: https://github.com/llvm/llvm-project/commit/86771d0b65ee13242f89b8dfdf3c66f738eae4e5 breaks the test `mlp-packed.mlir` as we cannot move inner layers out of `arg14`. It looks that `affine.for` does not implement `ConditionallySpeculatable` yet. I hope this is a transitive state.


chelini:
@rengolin our test `mlp-packed` broke due to these commits. This: https://github.com/llvm/llvm-project/commit/86771d0b65ee13242f89b8dfdf3c66f738eae4e5 prevent LICM to hoist out the inner layers. I hope to be a transitive state as in the commit message they mention missing implementation for `scf.for` hooks. I did not add https://github.com/llvm/llvm-project/commit/1b99f3a224afcab4a08f3b5b32ece63aa22cae0c this yet but also here we break the test as more allocs/copies are generated.

rengolin:
The LICM patch doesn't seem to affect all loops, just those that were relying on `NoSideEffect`. AFAICS, only the XSMM ops use that. If that's the problem, can we just apply the recommendation on the commit message (replace with `Pure`)?

The bufferization patch seems to intentionally create out-of-place buffers to make it more consistent. Is there a cleanup pass that we can run after that (canonicalization?) that will undo the effects?

Regarding tests, let's not disable any tests unless there's a clear problem with our code. If this is an upstream change affecting our code generation, then we either change our code to match or fix the problem upstream, before merging with the next LLVM version.

chelini:
> The LICM patch doesn't seem to affect all loops, just those that were relying on `NoSideEffect`. AFAICS, only the XSMM ops use that. If that's the problem, can we just apply the recommendation on the commit message (replace with `Pure`)?
> 
> The bufferization patch seems to intentionally create out-of-place buffers to make it more consistent. Is there a cleanup pass that we can run after that (canonicalization?) that will undo the effects?
> 
> Regarding tests, let's not disable any tests unless there's a clear problem with our code. If this is an upstream change affecting our code generation, then we either change our code to match or fix the problem upstream, before merging with the next LLVM version.

XSMM and our ops are already adjusted, and I am not aware of other bufferization passes that can help here.

rengolin:
Ok, let's talk to Sanjoy about the implementation of `scf.for` which is in his plans (according to the commit). The best approach is to only rebase LLVM when we have both changes in. Disabling the test means we'll forget to re-enable or even ask the question later on, and carrying local patches is a non-starter.

On the bufferisation issue, what do you mean "the IR is not ideal". Does that break our passes?

My read of the commit message is that this commit should not change the semantics (or shapes) of the tensors, just have an extra local copy that can "easily be elided later on", just to avoid a mechanical issue (order of the pass). If that's correct, then why do the shapes change in our IR?

rengolin:
Sanjoy is working on adding `Speculatable` support for `scf.for` with compiler-time known bounds, which is good enough for us, since we don't yet work with dynamic shapes. Should be landing in a week-ish.

chelini:
> Sanjoy is working on adding `Speculatable` support for `scf.for` with compiler-time known bounds, which is good enough for us, since we don't yet work with dynamic shapes. Should be landing in a week-ish.

Cool thanks!

rengolin:
you got your other PR in here, I think

rengolin:
Better to rebase and update this branch, so you can merge from the PR and we can run the actual tests without conflict (especially if the other branch merges first).

test/TPP/mlp-packed.mlir
```diff
@@ -1,5 +1,7 @@
 // RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -pack-matmul="block-factors=32,32,32" -canonicalize -tile-consumer-and-fuse-producers="tile-sizes=1,0,0,0" -canonicalize -tile-consumer-and-fuse-producers="tile-sizes=1,0,0" -canonicalize -loop-invariant-code-motion -canonicalize -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize -map-linalg-to-tpp -convert-linalg-to-tpp="use-parallel-loops=false" -map-to-brgemm | FileCheck %s
 
+// TODO: re-enable me.
+// XFAIL: *
```
rengolin, line 4: I don't think we should disable tests to allow a merge. We should fix LLVM upstream and rebase on the new commit.

rengolin (changes requested):

build_tools/llvm_version.txt
```diff
@@ -1 +1 @@
-e750c41ec1accc25f12e46fd64e52cd1495cfc58
+c1864ab9534080ee77a016bca24dd9a318bc6d7e
```
rengolin, line 1: rename the PR with the new SHA, or it will be confusing when we see that as a commit message

test/TPP/transform/transform-convolutions.mlir
```diff
@@ -122,11 +121,10 @@ transform.sequence failures(propagate) {
   ^bb0(%arg1: !pdl.operation):
     %0 = transform.structured.match ops{["linalg.matmul"]} in %arg1
     %1 = transform.structured.generalize %0
+    // expected-error @below {{Could not map to matmul}}
```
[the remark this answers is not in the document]
chelini: I was not checking the diagnostic. LLVM bump does not affect here.

test/TPP/mlp-packed.mlir
chelini: re-enabled

Issue #65, opened 2022-10-14
BF16 brgemm crash


KavithaTipturMadhu:
BRGEMM throws a segfault when used with bf16 operands. The same example works just fine when the datatype is f32. The segfault is inside libxsmm's call.

KavithaTipturMadhu:
```// RUN:standalone-opt %s -linalg-ext-to-loops -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -convert-vector-to-llvm -convert-func-to-llvm -convert-memref-to-llvm -canonicalize -sparse-compiler |\
// RUN: mlir-cpu-runner \
// RUN:  -e entry -entry-point-result=void  \
// RUN: -shared-libs=%llvmlirdir/libmlir_c_runner_utils%shlibext,%standalonelibdir/libstandalone_c_runner_utils%shlibext | \
// RUN: FileCheck %s
//

 module @predict_function {
  func.func @entry(){
    %arg0 = memref.alloc(): memref<128x256xbf16>
    %arg1 = memref.alloc(): memref<256x512xbf16>
    %arg2 = memref.alloc(): memref<512xbf16> 
    %arg3 = memref.alloc(): memref<128x512xbf16>
    %f0 = arith.constant 1.0:bf16
    linalg.fill ins(%f0:bf16) outs(%arg0:memref<128x256xbf16>)
    linalg.fill ins(%f0:bf16) outs(%arg1:memref<256x512xbf16>)
    linalg.fill ins(%f0:bf16) outs(%arg2:memref<512xbf16>)
    linalg.fill ins(%f0:bf16) outs(%arg3:memref<128x512xbf16>)
    %c4 = arith.constant 4 : index
    %c0 = arith.constant 0 : index
    %c16 = arith.constant 16 : index
    %c1 = arith.constant 1 : index
    tpp.identity ins(%arg2 : memref<512xbf16>) out(%arg3 : memref<128x512xbf16>)
    %alloc = memref.alloc() {alignment = 128 : i64} : memref<32x64x4x4xbf16>
    linalgx.pack %arg0 inner_dims_pos = [0, 1] inner_tiles = [4, 4] into %alloc : (memref<128x256xbf16> memref<32x64x4x4xbf16>)
    %alloc_0 = memref.alloc() {alignment = 128 : i64} : memref<128x64x4x4xbf16>
[...]

PR #67, opened 2022-10-14
Fix: Lda and num batches incorrect for brgemm with packed layout

KavithaTipturMadhu:


lib/Standalone/ConvertTppToXsmm.cpp
```diff
@@ -115,13 +115,13 @@ struct ConvertTppBrgemmOp : public OpRewritePattern<BrgemmOp> {
     if (failed(ldaDim))
       return failure();
     int64_t lda = *ldaDim;
-    if (memrefA.getElementType().isBF16() && memrefA.getShape().size() == 3) {
-      auto divLdaDim = getLeadingDim(memrefA, 1);
+    if (memrefA.getElementType().isBF16() && memrefA.getShape().size() == 4) {
```
rengolin, line 118: It's unclear to me what's going on here. Can you add a comment to explain what are those dimensions and what `b` is in the end?

chelini, line 118: Can you please add a comment to document what this if handles?

lib/Standalone/ConvertTppToXsmm.cpp
```diff
@@ -115,13 +115,13 @@ struct ConvertTppBrgemmOp : public OpRewritePattern<BrgemmOp> {
...
+      auto divLdaDim = getLeadingDim(memrefA, 2);
       if (failed(divLdaDim))
         return failure();
       lda = lda / (*divLdaDim);
+      b = b * memrefA.getShape()[3];
```
chelini, line 123: what is `b` here?

test/BF16/mlp-single-layer-blocked-bf16.mlir
```diff
@@ -42,7 +42,7 @@
         %f1 = arith.extf %v0:vector<4x4xbf16> to vector<4x4xf32>
         
        //
-       // CHECK:( ( 129, 129, 129, 129 ), ( 129, 129, 129, 129 ), ( 129, 129, 129, 129 ), ( 129, 129, 129, 129 ) )
+       // CHECK:( ( 256, 256, 256, 256 ), ( 256, 256, 256, 256 ), ( 256, 256, 256, 256 ), ( 256, 256, 256, 256 ) )
```
rengolin, line 45: Is this different because the previous value was being "cut in half" by just walking through half of the tensor?

lib/Standalone/ConvertTppToXsmm.cpp
rengolin: Duplicate of @chelini comment, ignore it.

test/BF16/mlp-single-layer-blocked-bf16.mlir
KavithaTipturMadhu: This is different because batch size being considered was using the 0th dimension of packed A tensor, ignoring the last dimension (of size 2 in packed layout).

lib/Standalone/ConvertTppToXsmm.cpp
KavithaTipturMadhu: b corresponds to batch size. Instead of initializing b and updating it later, I will initialize b to take the 0th dimension of memrefB, much cleaner.

chelini: I would spell it out. `b` is too generic

rengolin (approved):

rengolin:
@KavithaTipturMadhu, if you add the words "Fixes #65" to the commit message, Github will automatically close the issue associated with it when you merge.

PR #68, opened 2022-10-17
Add support for 1d/3d/4d memref

chelini:


rengolin (commented):
accidental delete of `vec_2d_compare`, as broken in the test. Either keep `vec_Nd_compare` for all dimensions or rewrite the benchmark drivers to match (all three use it).

chelini:
> 

We want to remove `*_compare` functions as we want to use libxsmm utils for that. I will check why it fails.

rengolin:
It's because we use `vec_f2d_compare` directly on the benchmarks. All three drivers do that, we'd need to remove from them before merge.

rengolin (approved):
Once you make it work, it should be fine to keep it for experiments, even if we're going to use the `tpp-opt` soon.

rengolin:
This will require rebase and probably re-work to fix the failing test. Do you want to keep this in a branch instead?

chelini:
ok.

PR #69, opened 2022-10-17
Use the same mapping pattern to rewrite a conv to matmul (NFC)

chelini:
This NFC unifies `DecomposeConvsToMatmulOrBrgemm.cpp` and `MapConv2DNhwcHwcfToMatmul.cpp` to use the same pattern when rewriting a convolution to matmul. `MapConv2DNhwcHwcfToMatmul` contains the actual pattern (`MapPackedConvToMatmul`) that does the job (we will rename the file in a follow-up).


include/Standalone/Transforms.h
```diff
@@ -29,10 +29,23 @@ namespace linalgx {
...
+  MapConvToMatmul() = delete;
+  MapConvToMatmul(int64_t rPos, int64_t sPos) : rPos(rPos), sPos(sPos){};
+  FailureOr<linalg::MatmulOp> doIt(RewriterBase &rewriter,
+                                   linalg::LinalgOp linalgOp);
+
+  int64_t getRPos() { return rPos; };
+  int64_t getSPos() { return sPos; };
+
+private:
+  int64_t rPos;
```
rengolin, line 46: isn't R and S always contiguous? Isn't the C position what really matters here? Perhaps this should have a single argument, and derive the others from it.

include/Standalone/Transforms.h
```diff
@@ -29,10 +29,23 @@ namespace linalgx {
...
-FailureOr<linalg::MatmulOp> mapConvToGemm(RewriterBase &rewriter,
-                                          linalg::LinalgOp linalgOp,
-                                          int64_t rPos, int64_t sPos);
+// Map a convolution to a matmul operation.
+// [N][P][Q][K] += [N][H][W][C] * [R][S][C][K]
+// [N][K’][P][Q][k] += [N][C’][H][W][c] * [K’][C’][R][S][c][k] (blocked)
+class MapConvToMatmul {
+public:
+  MapConvToMatmul() = delete;
+  MapConvToMatmul(int64_t rPos, int64_t sPos) : rPos(rPos), sPos(sPos){};
```
rengolin, line 38: IIUC, you're creating this structure amd calling `doIt` straight after on all cases. Why do you need to cache the arguments and not just export a function?

lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
```diff
@@ -55,38 +56,89 @@ static bool hasFilterWithRandSEqualOne(OpOperand *filter, unsigned i,
...
+getSlicedConvOperand(OpBuilder &builder, OpOperand *operand,
+                     linalg::LinalgOp linalgOp, ValueRange ivs,
+                     ValueRange valuesToUse, ArrayRef<int64_t> rAndSPos = {}) {
+  Location loc = linalgOp.getLoc();
+  FailureOr<SmallVector<Value>> involvedDimForOperand =
+      utils::getInvolvedLocalDimsForOperand(
+          builder, loc, operand, linalgOp.getMatchingIndexingMap(operand), ivs);
+  if (failed(involvedDimForOperand))
+    return failure();
+  return getSlicedConvOperandImpl(builder, linalgOp, operand,
```
rengolin, line 110: If you're not calling this from multiple sources, why is this not inlined?

lib/Standalone/TransformUtils.cpp
```diff
@@ -86,16 +86,15 @@ getExpectedResultMemRefShape(SmallVector<OpFoldResult> sizes,
   return targetShape;
 }
 
-static Value getSliceOperand(OpBuilder &builder, linalg::LinalgOp linalgOp,
-                             Value operand,
-                             const SmallVector<OpFoldResult> &offsets,
-                             const SmallVector<OpFoldResult> &sizes,
-                             const SmallVector<OpFoldResult> &strides,
-                             unsigned desiredResultRank) {
+Value getSliceOperand(OpBuilder &builder, linalg::LinalgOp linalgOp,
```
rengolin, line 89: If this isn't static anymore, perhaps we should make it return `FailureOr` like the other one... Who know what will call this now...

lib/Standalone/TransformUtils.cpp
```diff
@@ -119,15 +118,14 @@ static Value getSliceOperand(OpBuilder &builder, linalg::LinalgOp linalgOp,
...
+  assert(extractOperation->getNumResults() == 1 && "expect single result");
   return extractOperation->getResult(0);
 }
 
-static Value
-getSliceOperandImpl(OpBuilder &builder, linalg::LinalgOp linalgOp,
-                    OpOperand *operand, ValueRange ivs, ValueRange valuesToUse,
-                    unsigned desiredResultRank,
-                    const SmallVector<OpFoldResult> &sizesToUse = {}) {
+static Value getSliceOperandImpl(OpBuilder &builder, linalg::LinalgOp linalgOp,
```
rengolin, line 125: This doesn't seem to be called anymore...

include/Standalone/Transforms.h
```diff
@@ -29,10 +29,23 @@ namespace linalgx {
 FailureOr<SmallVector<Value>> mapToBRGEMMOp(RewriterBase &rewriter,
                                             linalg::LinalgOp linalgOp);
 
-//
-FailureOr<linalg::MatmulOp> mapConvToGemm(RewriterBase &rewriter,
-                                          linalg::LinalgOp linalgOp,
-                                          int64_t rPos, int64_t sPos);
+// Map a convolution to a matmul operation.
+// [N][P][Q][K] += [N][H][W][C] * [R][S][C][K]
+// [N][K’][P][Q][k] += [N][C’][H][W][c] * [K’][C’][R][S][c][k] (blocked)
```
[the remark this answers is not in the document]
chelini: yes but should already be in the block form: [N][K’][P][Q][k] += [N][C’][H][W][c] * [K’][C’][R][S][c][k]

include/Standalone/Transforms.h
chelini: Originally was a function: `FailureOr<linalg::MatmulOp>(RewriterBase &rewriter, linalg::LinalgOp linalgOp, int64_t rPos, int64_t sPos)`. But I did not like carrying around `rPos` and `sPos` so I make a struct. Do you think it is better to revert as before?

include/Standalone/Transforms.h
chelini: Why the channel? I am interested in the filter width and height as I need them to understand the matmul sizes. They are contiguous so we could specify only `Rpos` or `sPos`.

lib/Standalone/MapConv2DNhwcHwcfToMatmul.cpp
chelini: you mean `inline`? or you expect to be squashed with the caller?

lib/Standalone/TransformUtils.cpp
chelini: It is called by `getSliceOperand`

include/Standalone/Transforms.h
rengolin: Perhaps pass both and just have an assert that `s = r+1`?

lib/Standalone/TransformUtils.cpp
rengolin: Ignore me, Github folding code

lib/Standalone/TransformUtils.cpp
rengolin: perhaps merge the two, then and force the caller to prepare the arguments every time

chelini: Added a TODO. Will do later

include/Standalone/Transforms.h
chelini: done

include/Standalone/Transforms.h
chelini: Will do in a follow-up

include/Standalone/Transforms.h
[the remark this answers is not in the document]
chelini: Looks better now?

rengolin (approved):

PR #70, opened 2022-10-17
Check dialect - wip

KavithaTipturMadhu:
@chelini I tried adding a new check dialect for test operations but the dialect doesn't seem to be getting registered. I don't see it in the available dialects list. Can you please take a look to see what I may have missed?


chelini:
> 

You need to add the dialect here: https://github.com/plaidml/tpp-sandbox/blob/cb2488f06dc57ea8c6eb3dc857b62347d6b5aa2e/standalone-opt/standalone-opt.cpp#L35

include/Standalone/Dialect/Check/CheckOps.td
```diff
@@ -0,0 +1,60 @@
...
+    any non-zero integer.
+
+    Issues a non-fatal failure if the verification fails.
+
+    ```mlir
+    check.expect_true(%arg0) : i32
+    ```
+  }];
+
+  let arguments = (ins AnySignlessInteger:$operand);
```
rengolin, line 33: Do we want to restrict this to booleans (`i1`)?

include/Standalone/Dialect/Check/CheckOps.td
```diff
@@ -0,0 +1,60 @@
...
+
+  let assemblyFormat = "`(` $operand `)` attr-dict `:` type($operand)";
+}
+
+def CHECK_ExpectAlmostEqOp :
+    Op<CHECK_Dialect, "expect_almost_eq", [SameTypeOperands]> {
+  let summary = [{Checks that the operands are almost equal}];
+  let description = [{
+    Verifies that the tensor operands with float elements are
+    almost equal to within an implementation-defined "reasonable" tolerance.
```
rengolin, line 43: I think what's _reasonable_ should be a part of this op, via an attribute. We can start with a simple fp-comparison and a delta (which is the attribute), but we could upgrade this into more complex attributes to handle statistical analysis later.

chelini: Agree. I think TOSA has something along this line.

include/TPP/Dialect/Check/CheckOps.h
```diff
@@ -0,0 +1,20 @@
+//===- XsmmOps.h - Xsmm dialect ops -----------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CHECK_STANDALONE_OPS_H
```
chelini, line 9: Do use STANDALONE. We have TPP now

include/TPP/Dialect/Check/CheckOps.h
```diff
@@ -0,0 +1,20 @@
+//===- XsmmOps.h - Xsmm dialect ops -----------------------------*- C++ -*-===//
```
chelini, line 1: Please update.

include/TPP/Dialect/Check/CheckOps.td
```diff
@@ -0,0 +1,60 @@
...
+    any non-zero integer.
+
+    Issues a non-fatal failure if the verification fails.
+
+    ```mlir
+    check.expect_true(%arg0) : i1
+    ```
+  }];
+
+  let arguments = (ins AnyI1:$operand);
```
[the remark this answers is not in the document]
KavithaTipturMadhu: IIUC, expect_true op needs to check only an I1 unlike the iree version that takes i32. Is this right?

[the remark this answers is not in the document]
chelini: Agree. Just wondering what the prefix `Any` means in this context.

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+using namespace mlir;
+using namespace mlir::check;
+
+#define GEN_PASS_CLASSES
+#include "Standalone/Passes.h.inc"
+#include <iostream>
+namespace {
+
+static func::CallOp buildExpectTrueCall(Location loc, Operation *op,
```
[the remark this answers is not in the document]
KavithaTipturMadhu: you mean add lower calls in xsmm conversion pass? Parts of the code is similar between xsmm and check lowering, but ops are different, and there are no return values from check ops, so things like replaceUsesWith won't work.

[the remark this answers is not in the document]
chelini: no no, here are good. Just wondering if we can share some code, but maybe this is not the case. Just ignore the comment for now

include/TPP/Dialect/Check/CheckOps.td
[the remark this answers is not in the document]
rengolin: Probably just `i1` would do.

include/TPP/Dialect/Check/CheckOps.td
```diff
@@ -0,0 +1,60 @@
...
+
+  let assemblyFormat = "`(` $operand `)` attr-dict `:` type($operand)";
+}
+
+def CHECK_ExpectAlmostEqOp :
```
rengolin, line 38: I don't like the word "almost". It doesn't convey what we're trying to do. In the LLVM test suite / benchmarks we use `fp-comp`. We may get into a state that each number isn't even remotely similar to the next, but the overall distribution is, so this is just a special type of comparison.

We'll probably need to compare integers at some point (INT8), so this also can't be called `FPComp`. I don't know a good name, but "almost" isn't it.

lib/TPP/Dialect/Check/CheckDialect.cpp
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020 The IREE Authors
```
rengolin, line 1: We cannot attribute copyright to other authors. This line has to go.

Keep them all like the LLVM ones, since we want these files to be in LLVM soon,

include/TPP/Dialect/Check/CheckDialect.h
```diff
@@ -0,0 +1,24 @@
+// Copyright 2020 The IREE Authors
```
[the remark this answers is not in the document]
rengolin: I don't think it's necessary. We can say this in the commit message, in the RFC discussion, etc.

Let's keep those like the LLVM ones, because even if IREE wanted to upstream them, they'd have to drop their copyright comments.

include/TPP/Dialect/Check/CheckOps.td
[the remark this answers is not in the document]
KavithaTipturMadhu: ```AnyI``` refers to any integer regardless of its signedness semantics, whereas ```I``` refers to signless integer of specific width. I guess using ```I<width>``` makes more sense here.

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,154 @@
...
+  auto call = rewriter.create<func::CallOp>(loc, fnName.getValue(), None,
+                                            op->getOperands());
+  return call;
+}
+
+static SmallVector<Value> getMemRefOperands(OpBuilder &b, Location loc,
+                                            ValueRange operands) {
+  SmallVector<Value> res;
+  res.reserve(operands.size());
+  for (Value op : operands) {
```
chelini, line 55: I would switch the else with the if here. Write something like:
``` 
if (auto memrefType = op.getType().dyn_cast<MemRefType>()) {
   /* do else stmts*/
}
else {
  /* do if stmts */
}
```

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,154 @@
...
+    Location loc = expectTrueOp.getLoc();
+    std::string kindAsString = "expect_true";
+
+    FlatSymbolRefAttr fnName =
+        SymbolRefAttr::get(rewriter.getContext(), kindAsString);
+
+    ModuleOp module = expectTrueOp->getParentOfType<ModuleOp>();
+
+    buildExpectTrueCall(loc, expectTrueOp, module, fnName, rewriter);
+    expectTrueOp.erase();
```
chelini, line 110: mm, I think you want to notify the rewriter. `rewriter.eraseOp(exepectedTrueOp)`.

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,154 @@
...
+    Location loc = almostEqualsOp.getLoc();
+    std::string kindAsString = "expect_almost_equals";
+
+    FlatSymbolRefAttr fnName =
+        SymbolRefAttr::get(rewriter.getContext(), kindAsString);
+
+    ModuleOp module = almostEqualsOp->getParentOfType<ModuleOp>();
+
+    buildExpectAlmostEqualsCall(loc, almostEqualsOp, module, fnName, rewriter);
+    almostEqualsOp.erase();
```
chelini, line 130: same here. Have a look at: https://mlir.llvm.org/docs/PatternRewriter/#restrictions

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,154 @@
...
+    return success();
+  }
+};
+
+struct ConvertCheckToFunc : public ConvertCheckToFuncBase<ConvertCheckToFunc> {
+  ConvertCheckToFunc() = default;
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    tpp::populateCheckToFuncPatterns(patterns);
```
chelini, line 139: mlir::tpp:: populateCheckToFuncPatterns

include/TPP/Dialect/Check/CheckOps.td
rengolin: How about `expect_within_range`?

KavithaTipturMadhu: That name seems to suggest there would be a range minimum and maximum that are passed as arguments to the func. Shall we decide on the name later and get the PR reviewed for now?

KavithaTipturMadhu:
@chelini I will create new issues for all the changes we have discussed so far with the PR. Shall we review it in its current state and merge for now?

include/TPP/Dialect/Check/CheckOps.td
```diff
@@ -0,0 +1,61 @@
...
+    Op<CHECK_Dialect, "expect_almost_eq",[TypesMatchWith<"Operand types match",
+                   "lhs", "rhs",
+                   "$_self.cast<ShapedType>()">]> {
+  let summary = [{Checks that the operands are almost equal}];
+  let description = [{
+    Verifies that the tensor operands with float elements are
+    almost equal to within an implementation-defined "reasonable" tolerance.
+
+    Issues a non-fatal failure if the verification fails.
```
chelini, line 45: If we assert this is not the case, I believe.

lib/TPP/Dialect/Check/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+  SmallVector<OpResult> getAliasingOpResult(Operation *op, OpOperand &opOperand,
+                                            const AnalysisState &state) const {
+    return {};
+  }
+
+  BufferRelation bufferRelation(Operation *op, OpResult opResult,
+                                const AnalysisState &state) const {
+    return BufferRelation::Equivalent;
+  }
```
chelini, line 50: nit add space.

lib/TPP/Dialect/Check/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,130 @@
...
```
chelini, line 92: same here.

lib/TPP/Dialect/Check/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+add_mlir_dialect_library(TPPCheckDialect
+  # Ops and dialects
```
chelini, line 2: can you please align the comment with the file names?

lib/TPP/Dialect/Check/CheckOps.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/OpImplementation.h"
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir {
+namespace check {
+namespace {
+template <typename SrcOp, typename DstOp>
+struct ExpandAttributeToConst : public OpRewritePattern<SrcOp> {
```
chelini, line 12: Why this pattern?

tpp-rt/CheckRunnerUtils.h
```diff
@@ -0,0 +1,28 @@
...
+// retargetable, including on targets without a C++ runtime.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TPP_EXECUTIONENGINE_CRUNNERUTILS_H
+#define TPP_EXECUTIONENGINE_CRUNNERUTILS_H
+
+#include "mlir/ExecutionEngine/RunnerUtils.h"
+
+// TODO: here we want to have dispatch/invoke only for unary/binary and ternary.
```
chelini, line 20: this comment is not related, please remove it.

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,172 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Dialect/Check/CheckOps.h"
+#include "TPP/Passes.h"
+#include "TPP/Transforms.h"
```
chelini, line 11: not needed.

lib/TPP/ConvertCheckToFunc.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
```
chelini, line 14: not needed

chelini (approved):
@KavithaTipturMadhu thanks! Looks good but I would like to add the following tests:
1. We need to check that the IR roundtrip properly (`expect_true` is not tested), please add a test case similar to this one: https://github.com/plaidml/tpp-sandbox/blob/main/test/TPP/tpp-ops.mlir
2. Please add a smaller test case that exercise only the lowerings of the ops to the func dialect.

lib/TPP/ConvertCheckToFunc.cpp
KavithaTipturMadhu: Method populateCheckToFuncPatterns is defined in TPP/Transforms.h. Do you recommend moving it elsewhere?

chelini: Ah, then it is fine

lib/TPP/ConvertCheckToFunc.cpp
KavithaTipturMadhu: I can only remove Arith header from this list. There are dependences on other headers that I cant get rid of.

PR #75, opened 2022-10-18
Handle strided conv when mapping to Matmul

chelini:


test/Standalone/transform/transform-convolutions.mlir
```diff
@@ -11,7 +11,7 @@ transform.with_pdl_patterns {
   }
 }
 
-func.func @conv2d_1x56x56x64_3x3x64x64_pad(%arg0: tensor<1x56x56x64xf32>, 
+func.func @main(%arg0: tensor<1x56x56x64xf32>,
```
rengolin, line 14: Why renaming all functions to `main`? It may make it harder to debug if something goes wrong...

rengolin (approved):
LGTM with perhaps keeping the previous names?

test/Standalone/transform/transform-convolutions.mlir
chelini: It does not affect diagnostic, but no hard feeling if we want to keep the old names (they looked to long to write a test case for).

rengolin: Perhaps call it `kernel`? Or `conv2d` for the convs, etc?

The main problem is that you'll have naming issues if you have to use the same arguments (mangling) then you'd have `kernel_1`, `kernel_2` or whatever for the variants.

You don't need to use the whole sizes (they're not that relevant), so perhaps just name them `conv_pad`, `conv_stride2`, etc would be more meaningful without leading to name clashes.

_Also, for the record, `main` is a dangerous name for a generic function... :)_

Issue #74, opened 2022-10-17
Add run/benchmark options to `tpp-opt` / `tpp-runner`


rengolin:
The `tpp-opt` tool works with MLIR and could very well JIT-compile and run (replacing the call to `mlir-cpu-runner`) and benchmark. Alternatively, we can create a new `tpp-cpu-runner` from `mlir-cpu-runner` like `tpp-opt` was from `mlir-opt`. Whatever is easier.

Either way, we'll need to:
* Link with the right libraries (libxsmm, libtpp_c_runtime, openmp, etc)
* Create a wrapper function in MLIR to prepare the input tensors
* Warm-up libxsmm calls, time calls as close as possible, avoid tensor preparation

Future stuff:
* Use the `check` dialect to test MLIR execution
* Use the `perf` dialect to time inside kernels

rengolin:
([1b99e8ba487b](https://reviews.llvm.org/rG1b99e8ba487bb2759c81ecd925b709830bdd7aa9)) allows me to use the options without hacks. We should get to that commit when possible before merging the next fix I'm working on.

PR #84, opened 2022-10-22
Create TPP-RUN program

rengolin:
This program is a replacement for mlir-cpu-runner for a small number of cases, where we want to quickly call a MLIR method with a wrapper function that initializes the arguments and prints the output.

In addition, on benchmark mode, we also want to time the execution of the kernel (but not the tensor setup, JIT compile times, etc).

For everything else, you should use mlir-cpu-runner and equivalent.

Tests were amended to use the right tool using the right command line options for tpp-run as well as update options for tpp-opt.

For now, the features available are:
 * Reads MLIR file in `func` dialect (not `LLVM` dialect)
 * Detects / uses entry point passed as argument (or single function)
 * Creates a wrapper function that scans arguments and return values
 * Initializes arguments, prepare return values and calls the function
 * Prints the return value (whether it was returned or inplace)

What's still missing:
 * Expecting IR to be on a particular shape (memref) isn't very robust
 * Initializing arguments with splat(1.0), need to be random
 * Needs random seed to allow for predictable / comparable runs
 * Has no timer functionality yet
 * Doesn't try to figure out libxsmm/tpp library paths automaticaly

Bugs:
 * Inplace returns are still printing the original (arg) value
 * Can't update cmd-line args, needs hack, can't hav additional args

Future changes that may need MLIR (JitRunner) changes:
 * We need to access the cmd-line options
 * We need to time just the kernel execution, not setup, not JIT

Partially Implements #74

### DIFF
--- a/test/BF16/matmul-bf16.mlir
+++ b/test/BF16/matmul-bf16.mlir
@@ -1,5 +1,5 @@
 // RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf | \
-// RUN: tpp-run \
+// RUN: mlir-cpu-runner \
 // RUN:  -e entry -entry-point-result=void  \
 // RUN: -shared-libs=%llvmlirdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext | \
 // RUN: FileCheck %s
--- a/test/BF16/matmul-bf16.mlir
+++ b/test/BF16/matmul-bf16.mlir
@@ -1,5 +1,5 @@
 // RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf | \
-// RUN: mlir-cpu-runner \
+// RUN: tpp-run \
 // RUN:  -e entry -entry-point-result=void  \
 // RUN: -shared-libs=%llvmlirdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext | \
 // RUN: FileCheck %s
--- a/test/BF16/matmul-bf16.mlir
+++ b/test/BF16/matmul-bf16.mlir
@@ -1,5 +1,5 @@
-// RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -convert-vector-to-llvm -convert-func-to-llvm -convert-memref-to-llvm -canonicalize -reconcile-unrealized-casts | \
-// RUN: mlir-cpu-runner \
+// RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf | \
+// RUN: tpp-run \
 // RUN:  -e entry -entry-point-result=void  \
 // RUN: -shared-libs=%llvmlirdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext | \
 // RUN: FileCheck %s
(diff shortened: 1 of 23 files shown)


tpp-run/tpp-run.cpp
```diff
@@ -0,0 +1,281 @@
+//===- tpp-run.cpp - TPP CPU Execution Driver------------------------------===//
+//
+// Main entry point to a command line utility that executes an MLIR file on the
+// CPU by translating MLIR to LLVM IR before JIT-compiling and executing the
+// latter. Handles TPP/LIBXSMM include/library paths as well as benchmarking
+// modes, with warmups, measurements, output comparison, etc.
+//
+//===----------------------------------------------------------------------===//
+
+#include <iostream>
```
chelini, line 10: please drop.

tpp-run/tpp-run.cpp
```diff
@@ -0,0 +1,281 @@
...
+// Lowers to LLVM Dialect
+static LogicalResult lowerToLLVMDialect(ModuleOp module) {
+  // Minimal passes to make it work
+  // We don't want TPP passes here, as that's the job of tpp-opt
+  // The IR here should be free of TPP/XSMM or any TPP extensions
+  PassManager passManager(module.getContext());
+  applyPassManagerCLOptions(passManager);
+
+  // Bufferization, if needed
+  passManager.addNestedPass<func::FuncOp>(createTensorBufferizePass());
```
chelini, line 67: We should prefer one shot bufferization instead of partial ones.

tpp-run/tpp-run.cpp
```diff
@@ -0,0 +1,281 @@
...
+  passManager.addPass(createConvertVectorToLLVMPass());
+  passManager.addPass(createConvertFuncToLLVMPass());
+  passManager.addPass(createMemRefToLLVMConversionPass());
+  passManager.addNestedPass<func::FuncOp>(createArithToLLVMConversionPass());
+  passManager.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  passManager.addPass(createReconcileUnrealizedCastsPass());
+
+  auto result = passManager.run(module);
+  if (failed(result)) {
+    std::cerr << "ERROR: Failed to lower module to LLVM dialect" << std::endl;
```
chelini, line 88: llvm::errs()

KavithaTipturMadhu (approved):
Lgtm

test/BF16/mlp-single-layer-bf16.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -linalg-ext-to-loops -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -convert-vector-to-llvm -convert-func-to-llvm -convert-memref-to-llvm -canonicalize -sparse-compiler|\
+// RUN: tpp-opt %s -linalg-ext-to-loops -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -sparse-compiler |\
```
chelini, line 1: Can we drop `sparse-compiler`?

test/BF16/matmul-bf16.mlir
```diff
@@ -1,5 +1,5 @@
-// RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -convert-vector-to-llvm -convert-func-to-llvm -convert-memref-to-llvm -canonicalize -reconcile-unrealized-casts | \
-// RUN: mlir-cpu-runner \
+// RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf | \
```
chelini, line 1: I generally don't understand what is the entry point of `tpp-run`. What is the reasoning of leaving this two conversions outside `-convert-vector-to-scf -convert-scf-to-cf`?

tpp-run/tpp-run.cpp
rengolin: sorry, left over.

test/BF16/matmul-bf16.mlir
rengolin: I tried to keep only LLVM lowering passes in `tpp-run`. this can be moved around, but I don't want to assume what kind of IR we'll be creating with `tpp-opt`.

test/BF16/mlp-single-layer-bf16.mlir
rengolin: Maybe? I don't know why it was chosen here... But these aren't the only examples, so I'm guessing these were copy from some others.

tpp-run/tpp-run.cpp
rengolin: Bufferization should happen *before* TPP passes, so these here are _just_ for the code that we generate in the wrapper call, and it doesn't need to be (and doesn't quite work when it is) global.

test/BF16/mlp-single-layer-bf16.mlir
chelini: It was probably missing a conversion. We should remove it, but we can do it later.

test/BF16/matmul-bf16.mlir
chelini: Ok please document it.

tpp-run/tpp-run.cpp
chelini: mm, let's see. As long as we do not introduce extra allocs is fine.

test/BF16/mlp-single-layer-bf16.mlir
rengolin: #87

tpp-run/tpp-run.cpp
rengolin: Even if it does, it's ok, because the kernel code would have been bufferized *before* this pass is done. These passes *only* bufferize the wrapper code, which isn't in the critical path and will not be in the timed portion, so we don't really care what happens there.

chelini (approved):

PR #89, opened 2022-10-27
Allow outer-dims-perm when propagating through padOp

chelini:


chelini:
@rengolin and @KavithaTipturMadhu this test fail because there is a bug in upstream which is not fixed. But we need to more the llvm version forward. But https://github.com/plaidml/tpp-sandbox/pull/66 is currently blocked. How do we proceed?

rengolin:
Can we fix the bug upstream?

PR #92, opened 2022-10-30
Bump llvm 5e12c183920b506f83b2a10205b4a26b17ca9115

chelini:


chelini:
@rengolin Let's merge this. MLP is disabled for now but I know what is going on. Fix upstream is difficult and takes time, see: https://github.com/chelini/tpp-sandbox/commit/62abbef2640457c7dc77e46400a3e4afabebd925

rengolin:
A few things are still not clear to me:

**1. What exactly is that test checking?**

It's a large IR being CHECK'ed for a large and complicated output. Expecting the IR to remain the same over multiple versions of LLVM is not stable.

We should not disable tests because they stop passing after a change, we should either fix it upstream or fix the test. The reasons in chelini@62abbef are far too complicated to quickly "fix" upstream, and perhaps are just showing how fragile and broken this test is.

The only difference between `mlp-all.mlir` and `mlp-packed.mlir` is that the dimension of one tensor is 1024 instead of 1000. If it's testing some padding, it shouldn't need 4 layers, so that's a start to simplify things. Would that at least expose the problem more cleanly?

But if we're now missing the transform that allows us to use `linalgx.pack`, then we're missing an important part of the work we've done so far and will be moving backwards, not forwards. Ignoring it is not wise.

We will have this problem again multiple times, that is the price of working downstream (and why we need to upstream things as soon as possible). So, disabling a test and pushing the upstream fixes down the line will *not* help us, because it will just pile up with the next set of problems that will surely come again.

**2. What is the pressure for updating the LLVM version?**

We should keep up-to-date with LLVM as much as we can, I thoroughly agree, but not at the detriment of stability, and definitely not disabling tests or regressing in functionality to do so.

Is this because we need some functionality in the new delta? If so, what functionality and why is it more important than recognising relayouts in input IR?

Or is this because IREE is moving on and they need us to keep up? If so, how soon can we see a demonstration (with clear reproducible instructions) on how to use the sandbox in IREE? Without it, I can't really justify regressing functionality over it.

If we need to stall at some version of LLVM until we get our stuff together, and have time to fix the problems upstream, it will be more stable than rushing with this commit and not being able to fix (or at least not at a great cost) later on.

chelini:
> A few things are still not clear to me:
> 
> **1. What exactly is that test checking?**
> 
> It's a large IR being CHECK'ed for a large and complicated output. Expecting the IR to remain the same over multiple versions of LLVM is not stable.
> 
> We should not disable tests because they stop passing after a change, we should either fix it upstream or fix the test. The reasons in [chelini/tpp-sandbox@62abbef](https://github.com/chelini/tpp-sandbox/commit/62abbef) are far too complicated to quickly "fix" upstream, and perhaps are just showing how fragile and broken this test is.
> 
> The only difference between `mlp-all.mlir` and `mlp-packed.mlir` is that the dimension of one tensor is 1024 instead of 1000. If it's testing some padding, it shouldn't need 4 layers, so that's a start to simplify things. Would that at least expose the problem more cleanly?

`mlp-all.mlir` exercises the mapping only there is no packing nor fusion nor tiling. `mlp-packed.mlir` exercises all of the transformations we want. A single layer is not enough as we need to fuse across all the layers. In fact, in `mlp-single-layer-packed.mlir` we do see the issue. We can reduce from 4 layers to 2 or 3, but I don't think it will be an important IR simplification.

> 
> But if we're now missing the transform that allows us to use `linalgx.pack`, then we're missing an important part of the work we've done so far and will be moving backwards, not forwards. Ignoring it is not wise.

We need two things: 1) A fusion call-back. I have an implementation for it here: https://github.com/chelini/llvm-project/commit/f2a09adbe37de17306b50dea3b1892ce6867f782 but it is unclear if this form will end-up into main (see this discussion: https://discord.com/channels/689900678990135345/689932666505920596/1035585559684464733). 2) Bufferization is now more strict and the fix is either to revert this commit: https://github.com/llvm/llvm-project/commit/1b99f3a224afcab4a08f3b5b32ece63aa22cae0c or fix it upstream. I don't have cycles to do that now as I want to make progresses on other fronts.

> 
> We will have this problem again multiple times, that is the price of working downstream (and why we need to upstream things as soon as possible). So, disabling a test and pushing the upstream fixes down the line will _not_ help us, because it will just pile up with the next set of problems that will surely come again.

Agree but we need to set priorities. For me now the priority is to make progresses on the pack transformations and how it propagates. 

> 
> **2. What is the pressure for updating the LLVM version?**
> 
> We should keep up-to-date with LLVM as much as we can, I thoroughly agree, but not at the detriment of stability, and definitely not disabling tests or regressing in functionality to do so.
> 
> Is this because we need some functionality in the new delta? If so, what functionality and why is it more important than recognising relayouts in input IR?

To clarify, we are not missing "recongnising relayouts in input IR" this works perfectly fine. The problem are the missing fusion call-back and the extra copies generated by bufferization. Regarding LLVM we need to move forward otherwise our propagation will not work (and pack/unpack propagation is what I am working on now). There was a bug in MLIR which is now fixed: https://github.com/llvm/llvm-project/commit/9bcac22be559842cf488db469964fce37c2035a3. Without this commit our propagation pass will assert. 

> 
> Or is this because IREE is moving on and they need us to keep up? If so, how soon can we see a demonstration (with clear reproducible instructions) on how to use the sandbox in IREE? Without it, I can't really justify regressing functionality over it.

It is not IREE specific.

> 
> If we need to stall at some version of LLVM until we get our stuff together, and have time to fix the problems upstream, it will be more stable than rushing with this commit and not being able to fix (or at least not at a great cost) later on.

Again we need to understand our priority here.

KavithaTipturMadhu (approved):

rengolin:
> `mlp-all.mlir` exercises the mapping only there is no packing nor fusion nor tiling. `mlp-packed.mlir` exercises all of the transformations we want.

Is it possible to have more tests, checking each individual step (pack/fuse/tile) instead of nothing and everything?

> A single layer is not enough as we need to fuse across all the layers. In fact, in `mlp-single-layer-packed.mlir` we do see the issue. We can reduce from 4 layers to 2 or 3, but I don't think it will be an important IR simplification.

Ack.

>1) A fusion call-back. I have an implementation for it here: [chelini/llvm-project@f2a09ad](https://github.com/chelini/llvm-project/commit/f2a09adbe37de17306b50dea3b1892ce6867f782) but it is unclear if this form will end-up into main (see this discussion: https://discord.com/channels/689900678990135345/689932666505920596/1035585559684464733).

Don't take one comment in discord as gospel. Adapt the call back as suggested and push a review. Without one, we'll never know what other people think. Not every developer is on discord and no one reads everything that goes in there, even if they do.

> 2) Bufferization is now more strict and the fix is either to revert this commit: [llvm/llvm-project@1b99f3a](https://github.com/llvm/llvm-project/commit/1b99f3a224afcab4a08f3b5b32ece63aa22cae0c) or fix it upstream.

It's been 17 days and without an actual fault, reverting won't be meaningful. We need to fix upstream.

> I don't have cycles to do that now as I want to make progresses on other fronts.

We can't progress in other fronts by dropping working code. We need to move cycles to fix upstream, then we come back to continue our work.

> Agree but we need to set priorities. For me now the priority is to make progresses on the pack transformations and how it propagates.

Not at the cost of dropping functionality. The priority is to give only positive steps, and sometimes that means fixing upstream code that isn't ours.


> The problem are the missing fusion call-back and the extra copies generated by bufferization.

Can we cope with it by running a local cleanup pass to merge the copies? If this is as bad as I think it is, such a cleanup would probably be relevant to other people upstream as well... At least until the original pass stops creating useless copies on its own.

> Regarding LLVM we need to move forward otherwise our propagation will not work (and pack/unpack propagation is what I am working on now). There was a bug in MLIR which is now fixed: [llvm/llvm-project@9bcac22](https://github.com/llvm/llvm-project/commit/9bcac22be559842cf488db469964fce37c2035a3). Without this commit our propagation pass will assert. This bump is not even enough to get there, so expect a follow-up bump.

This does not look to me like a stable move.

rengolin (changes requested):
I'm still not comfortable with the XFAIL. We need to make forward progress by fixing this upstream.

chelini:
> 



> ###



> > `mlp-all.mlir` exercises the mapping only there is no packing nor fusion nor tiling. `mlp-packed.mlir` exercises all of the transformations we want.
> 
> Is it possible to have more tests, checking each individual step (pack/fuse/tile) instead of nothing and everything?

There are already tests for this, see for example: `transform-pack.mlir`. Here the test is exercising all the patterns together.

> 
> > A single layer is not enough as we need to fuse across all the layers. In fact, in `mlp-single-layer-packed.mlir` we do see the issue. We can reduce from 4 layers to 2 or 3, but I don't think it will be an important IR simplification.
> 
> Ack.
> 
> > 1. A fusion call-back. I have an implementation for it here: [chelini/llvm-project@f2a09ad](https://github.com/chelini/llvm-project/commit/f2a09adbe37de17306b50dea3b1892ce6867f782) but it is unclear if this form will end-up into main (see this discussion: https://discord.com/channels/689900678990135345/689932666505920596/1035585559684464733).
> 
> Don't take one comment in discord as gospel. Adapt the call back as suggested and push a review. Without one, we'll never know what other people think. Not every developer is on discord and no one reads everything that goes in there, even if they do.
> 
> > 2. Bufferization is now more strict and the fix is either to revert this commit: [llvm/llvm-project@1b99f3a](https://github.com/llvm/llvm-project/commit/1b99f3a224afcab4a08f3b5b32ece63aa22cae0c) or fix it upstream.
> 
> It's been 17 days and without an actual fault, reverting won't be meaningful. We need to fix upstream.

Ok let me clarify, maybe I explain myself wrongly. The commits did not break anything it just introduce performance regression in the MLP example we have. We have now copies that should not be there and we cannot fuse as before by relying on LICM. The IR is correct it is just sub-optimal with respect to the previous one.

> 
> > I don't have cycles to do that now as I want to make progresses on other fronts.
> 
> We can't progress in other fronts by dropping working code. We need to move cycles to fix upstream, then we come back to continue our work.
> 
> > Agree but we need to set priorities. For me now the priority is to make progresses on the pack transformations and how it propagates.
> 
> Not at the cost of dropping functionality. The priority is to give only positive steps, and sometimes that means fixing upstream code that isn't ours.

I am not dropping functionalities, the passes we rely on upstream changed their behavior and "break" our test.

> 
> > The problem are the missing fusion call-back and the extra copies generated by bufferization.
> 
> Can we cope with it by running a local cleanup pass to merge the copies? If this is as bad as I think it is, such a cleanup would probably be relevant to other people upstream as well... At least until the original pass stops creating useless copies on its own.
> 
> > Regarding LLVM we need to move forward otherwise our propagation will not work (and pack/unpack propagation is what I am working on now). There was a bug in MLIR which is now fixed: [llvm/llvm-project@9bcac22](https://github.com/llvm/llvm-project/commit/9bcac22be559842cf488db469964fce37c2035a3). Without this commit our propagation pass will assert. This bump is not even enough to get there, so expect a follow-up bump.
> 
> This does not look to me like a stable move.

I do get "This does not look to me like a stable move."

chelini:
> I'm still not comfortable with the XFAIL. We need to make forward progress by fixing this upstream.

I can fix the test based on the upstream changes, and avoid XFAIL.

rengolin:
> There are already tests for this, see for example: `transform-pack.mlir`. Here the test is exercising all the patterns together.

Ack.

> Ok let me clarify, maybe I explain myself wrongly. The commits did not break anything it just introduce performance regression in the MLP example we have. We have now copies that should not be there and we cannot fuse as before by relying on LICM. The IR is correct it is just sub-optimal with respect to the previous one.

Oh, ok, then perhaps the test needs changing on what it checks for.

The performance regression is a problem, and we need to create an issue for that (with a reproducer IR and command lines), but it should not stop this merge.

> I am not dropping functionalities, the passes we rely on upstream changed their behavior and "break" our test.

Marking a test as XFAIL is never the right response to a change in behaviour or an upstream change breaking our stuff.

If the test breaks because the output has changed, we change the output checks. If it's really broken, we fix upstream before merging another LLVM. If there's a performance regression, we create an issue with a clear reproducer so that anyone can work on it. But never marking tests as XFAIL.

> I can fix the test based on the upstream changes, and avoid XFAIL.

That'd be preferable, yes, with the extra issue and a reproducer for the performance problem.

I did notice the CHECK lines in that test are super specific. Maybe we need to check for less specific shapes, just a sequence of instructions, and let the performance problems be caught by proper (future) benchmarks.

rengolin (approved):

PR #93, opened 2022-10-31
Refactoring xsmm dispatches to be weakly typed

KavithaTipturMadhu:
Updating just the dispatch calls in this patch for easier reviews and merging.


chelini (approved):
Nice thanks!

PR #99, opened 2022-10-31
Improve diagnostic and documentation

chelini:
- better diagnostic when mapping to BRGEMM

- better doc for createMapToBatchReduceGEMMPass

### DIFF
--- a/lib/TPP/MapToBatchReduceGEMM.cpp
+++ b/lib/TPP/MapToBatchReduceGEMM.cpp
@@ -98,19 +98,6 @@ static LogicalResult checkBody(linalg::LinalgOp linalgOp) {
   return success();
 }
 
-static LogicalResult MapToBRGEMMOpPreconditions(linalg::LinalgOp linalgOp) {
-  if (!isa<linalg::GenericOp>(linalgOp))
-    return failure();
-
-  if (linalgOp.hasDynamicShape())
-    return failure();
-
-  if (failed(checkStructure(linalgOp)) ||
-      failed(checkAccessPatterns(linalgOp)) || failed(checkBody(linalgOp)))
-    return failure();
-  return success();
-}
-
 static FailureOr<SmallVector<Value>>
 getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
                   linalg::LinalgOp linalgOp, ValueRange valuesToUse) {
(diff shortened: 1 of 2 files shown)


chelini:
Note: `// TODO: this should not be required.` is addressed in a follow-up.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -163,7 +166,6 @@ mlir::linalgx::mapToBRGEMMOp(RewriterBase &rewriter,
     FailureOr<SmallVector<Value>> maybeSlicedOperands =
         getSlicedOperands(builder, loc, localIvs, linalgOp, operandValuesToUse);
     if (failed(maybeSlicedOperands)) {
-      // TODO: is safe to just return{} ?
       assert(0 && "failed to generate loops");
```
rengolin, line 169: why not just `return maybeSlicedOperands` instead of this going around?

chelini: Sorry what do you mean?

chelini: If we fail to extract the slices when cannot build the loop nest and we assert.

chelini: https://mlir.llvm.org/doxygen/structmlir_1_1linalg_1_1GenerateLoopNest.html

chelini: maybe we can avoid failing. In fact, if we return failure something went really really wrong, what do you think?

rengolin: That's what I'm thinking... asserting and failing will both stop execution, but an assert is harder than a failure, and the latter may prove more user-friendly...

rengolin: Then either `return failure()` of some version with an error message would be more appropriate than `assert() && return {}`.

chelini: I cannot propagate the failure, that my main problem. So basically the only option is to asset, as far as I can see reading the code in https://mlir.llvm.org/doxygen/structmlir_1_1linalg_1_1GenerateLoopNest.html

rengolin: ah! damn, ignore me, being thick is what's it. This code is in the lambda..

rengolin (approved):

PR #101, opened 2022-10-31
Apply transposition when propagating through an elementwise operation

chelini:
Unpack has `transpose` semantics with `outer_dims_perm` when propagating through an operation we need to apply transposition if `outer_dims_perm` is available.

### DIFF
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -779,8 +779,8 @@ struct PropagateThroughElementWiseOp
         unpackOutputs.push_back(unpackOp.getOutput());
     }
 
-    // XXX: innerDimsPos and outerDimsPerm must have been unsigned in the first
-    // place. `AffineMap::getPermutationMap` requires unsigned, convert.
+    // FIXME: innerDimsPos and outerDimsPerm must have been unsigned in the
+    // first place. `AffineMap::getPermutationMap` requires unsigned, convert.
     AffineMap permutationMap;
     if (!tileLoopPerms.empty()) {
       SmallVector<unsigned> interchangeVectorUInt;
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -13,6 +13,7 @@
 #include "mlir/Dialect/Bufferization/IR/Bufferization.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Traits.h"
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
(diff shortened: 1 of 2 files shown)


lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -587,17 +601,38 @@ struct PropagateThroughElementWiseOp
     : public OpRewritePattern<linalg::GenericOp> {
   using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
 
-  // Check sinking preconditions: a) all loops must
-  // be parallel.
-  LogicalResult checkPreconditions(linalg::GenericOp linalgOp) const {
-    if (linalgOp.getNumLoops() != linalgOp.getNumParallelLoops())
+  LogicalResult isElementwise(linalg::GenericOp linalgOp) const {
+    if (!linalg::isElementwise(linalgOp))
```
rengolin, line 605: This looks like a nop function, why not just use `linalg::isElementwise` directly?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -587,17 +601,38 @@ struct PropagateThroughElementWiseOp
...
+                        [](AffineMap map) { return map.isMinorIdentity(); });
+  }
+
+  // Require operands to come from a single `unpack` operation.
+  LogicalResult hasOnlyOnePackedOperand(linalg::GenericOp linalgOp) const {
+    bool oneOf = false;
+    for (OpOperand &operand : linalgOp->getOpOperands()) {
+      linalgx::UnPackOp unpackOp =
+          operand.get().getDefiningOp<linalgx::UnPackOp>();
+      if (unpackOp) {
```
rengolin, line 622: I'm guessing there aren't many operands, and this logic while meaningful, is a bit convoluted.

Probably be simpler to just count and `return count != 1`?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -703,6 +789,19 @@ struct PropagateThroughElementWiseOp
         unpackOutputs.push_back(unpackOp.getOutput());
     }
 
+    // XXX: innerDimsPos and outerDimsPerm must have been unsigned in the first
```
rengolin, line 792: What does XXX mean?

chelini: Something bad that I would like to fix later. But essentially the problem is that we should have designed `pack` and `unpack` with unsigned for `inner_dims` and `outer_dims_perms`.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: Done.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: Done.

lib/TPP/ToBlockLayoutAndBack.cpp
rengolin: Then FIXME? or TODO?

chelini: Done.

chelini:
Any more comments here?

rengolin (approved):

PR #102, opened 2022-11-01
Remove unused passes

chelini:
- Bufferization was not used as was a way too long WIP. We can simply use one-shot bufferization
- TppCompilerPipeline was an attempt to have a pipeline but was still a WIP and if we use transform we don't need it
- ConvertTppToVector would be good to have but not enough cycles at the moment


adam-smnk (approved):
seems valid

PR #103, opened 2022-11-01
Remove padding pass

chelini:
There was only one test exercising it, and it was leaking. Remove the pass as we plan to do padding using `pack` or `unpack`.


lib/TPP/TppCompiler.cpp
```diff
@@ -45,8 +45,8 @@ void TppCompilerPipeline::runOnOperation() {
   // map-linalg-to-tpp
   pm.addNestedPass<func::FuncOp>(createMapLinalgToTppPass());
   // enforce-tpp-preconditions
-  if (enablePreconditions)
-    pm.addNestedPass<func::FuncOp>(createPasSIMDDimensionPass());
+  // if (enablePreconditions)
```
rengolin, line 48: Don't comment out, remove the lines

rengolin (commented):
What was the point of that pass anyway?

lib/TPP/TppCompiler.cpp
chelini: The entire pass was removed, thus here I commented out. There is not more this file in main. I will rebase it.

hfp (approved):
Not much to review from my side. This is mostly about removing code.

chelini:
> What was the point of that pass anyway?

We want to have the K dimension in GEMM to be multiple of the SIMD lane. We can do that while packing I think we don't need an explicit padding.

hfp:
> > What was the point of that pass anyway?
> 
> We want to have the K dimension in GEMM to be multiple of the SIMD lane. We can do that while packing I think we don't need an explicit padding.

But then your packing pass would do the padding? Just wonder how the remainder is processed.

lib/TPP/TppCompiler.cpp
rengolin: My point was just about the commenting out vs removing the lines. Leaving code commented out all over the file is a mess and confuses people.

chelini:
> > > What was the point of that pass anyway?
> > 
> > 
> > We want to have the K dimension in GEMM to be multiple of the SIMD lane. We can do that while packing I think we don't need an explicit padding.
> 
> But then your packing pass would do the padding? Just wonder how the remainder is processed.

Yes, packing as an option to pad, so my intuition would be to use `pack` and `unpack` instead. We pad to get full tiles and we don't have remainder. For now we assume full tiles, we will get to that later.

PR #105, opened 2022-11-01
Introduce `MapLinalgToTppOp`

chelini:
The op runs `populateMapLinalgToTppPatterns` to annotate linalg.generic operations with attributes that represent mappable tpp operations.


adam-smnk (approved):
pretty clever solution, not sure if 100% compliant with library_call attr but worst case it'll lower to loops

PR #112, opened 2022-11-02
Xsmm Runtime weakly typed invoke calls

KavithaTipturMadhu:


KavithaTipturMadhu:
Fixes https://github.com/plaidml/tpp-sandbox/issues/108

tpp-rt/XsmmRunnerUtils.cpp
```diff
@@ -344,62 +287,44 @@ extern "C" void _mlir_ciface_matrix_copy_NC_to_NCNC(
   matrix_copy_NC_to_NCNC(addr_input, addr_output, 1, C, N, c, n);
 }
 
-extern "C" void _mlir_ciface_xsmm_brgemm_invoke_f32(
-    int64_t addr, UnrankedMemRefType<float> *A, UnrankedMemRefType<float> *B,
-    UnrankedMemRefType<float> *C, int64_t numBatches) {
+extern "C" void _mlir_ciface_xsmm_brgemm_invoke(const libxsmm_datatype dType,
+                                                int64_t addr,
+                                                UnrankedMemRefType<char> *A,
```
chelini, line 292: Not an expert here but would void work instead of char?

chelini (approved):
Nice, thanks! Just a minor comment.

tpp-rt/XsmmRunnerUtils.cpp
KavithaTipturMadhu: problem is dynamicmemreftype does not support void type, which we need to get the data pointer and the offset.

Issue #95, opened 2022-10-31
More copies in `mlp-packed.mlir`


chelini:
Upstream commits: https://github.com/llvm/llvm-project/commit/1b99f3a224afcab4a08f3b5b32ece63aa22cae0c made bufferization more conservative. The copies are not needed as we access different chunks for each iteration in the outermost loop. 
```
// Minimal reprodcuer: tpp-opt -empty-tensor-to-alloc-tensor -one-shot-bufferize="allow-return-allocs bufferize-function-boundaries"
 #map0 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d3)>
 #map1 = affine_map<(d0, d1, d2, d3) -> (d0, d3, d2)>
 #map2 = affine_map<(d0, d1, d2, d3) -> (d1, d2)>
 module @predict_function {
  func.func @main(%arg0: tensor<128x256xf32>, %arg1: tensor<256x512xf32>, %arg2: tensor<512xf32>, %arg3: tensor<512x1024xf32>, %arg4: tensor<1024xf32>, %arg5: tensor<1024x2048xf32>, %arg6: tensor<2048xf32>, %arg7: tensor<2048x1024xf32>, %arg8: tensor<1024xf32>, %arg9: tensor<128x1024xf32>, %arg10: tensor<128x2048xf32>, %arg11: tensor<128x1024xf32>, %arg12: tensor<128x512xf32>) -> tensor<4x32x32x32xf32> {
    %cst = arith.constant 1.000000e+00 : f32
    %c1 = arith.constant 1 : index
    %c64 = arith.constant 64 : index
    %c32 = arith.constant 32 : index
    %c0 = arith.constant 0 : index
    %c4 = arith.constant 4 : index
    %0 = tensor.empty() : tensor<4x64x32x32xf32>
    %inserted = tensor.insert %cst into %0[%c0, %c0, %c0, %c0] : tensor<4x64x32x32xf32>
    %1 = tensor.empty() : tensor<4x32x32x32xf32>
    %2 = scf.for %arg13 = %c0 to %c4 step %c1 iter_args(%arg14 = %1) -> (tensor<4x32x32x32xf32>) {
      %3 = tensor.empty() : tensor<32x32x32xf32>
      %extracted_slice = tensor.extract_slice %inserted[%arg13, 0, 0, 0] [1, 64, 32, 32] [1, 1, 1, 1] : tensor<4x64x32x32xf32> to tensor<64x32x32xf32>
[...]

PR #114, opened 2022-11-02
Use transform to optimize MLP

chelini:
Using transform we can have a better control over fusion and we can generate only 5 loops. 1 loop for each layer (4 layers) + an outermost
one. Bufferization still remains a problem (see issue: #95).

### DIFF
--- a/include/TPP/Dialect/LinalgX/TransformOps/LinalgXTransformOps.td
+++ b/include/TPP/Dialect/LinalgX/TransformOps/LinalgXTransformOps.td
@@ -207,4 +207,39 @@ def MapLinalgToTppOp : Op<Transform_Dialect, "structured.map_linalg_to_tpp", [
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// FoldUnitExtentDimsOp
+//===----------------------------------------------------------------------===//
+
+def FoldUnitExtentDimsOp : Op<Transform_Dialect, "structured.fold_unit_extent_dims", [
+    FunctionalStyleTransformOpTrait,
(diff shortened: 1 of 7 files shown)


lib/TPP/TransformDialectInterpreter.cpp
```diff
@@ -38,9 +38,27 @@ struct TransformDialectInterpreter
   }
 };
 
+struct TransformDropSchedulePass
```
rengolin, line 41: I'm surprised this isn't done automatically by the pass, or that there isn't this functionality in MLIR proper...

include/TPP/Dialect/LinalgX/TransformOps/LinalgXTransformOps.td
```diff
@@ -207,4 +207,39 @@ def MapLinalgToTppOp : Op<Transform_Dialect, "structured.map_linalg_to_tpp", [
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// FoldUnitExtentDimsOp
```
rengolin, line 211: I don't understand why this is in `LinalgX` and not on its own dialect. While this operates on `linalg` ops, it's not an extension to the `linalg` dialect.

rengolin (approved):
Looks good to me. My comments are for future work only.

adam-smnk (approved):

lib/TPP/TransformDialectInterpreter.cpp
chelini: It is not exposed, as Transform is still in test-mode. Something similar is done in IREE too.

Issue #87, opened 2022-10-24
Remove the usage of `sparse-compiler` in tests


rengolin:
The following tests are using the option `-sparse-compiler` for `tpp-opt`. Identify why we're using it and remove it no longer relevant.

Here are the tests using it:
```
test/BF16/matmul-pbf16.mlir
test/BF16/mlp-all-bf16.mlir
test/BF16/mlp-single-layer-blocked-bf16.mlir
test/BF16/mlp-single-layer-bf16.mlir
test/Integration/relayout-gemm.mlir
test/Integration/conv-on-loops.mlir
test/Integration/conv-on-tensor.mlir
test/Integration/subview-on-tensor.mlir
test/Integration/relayout-more-interesting.mlir
test/Integration/conv-mapping-to-gemm.mlir
```

rengolin:
I'm getting a non-terminating loop on `relayout-more-interesting.mlir` when I change the command line opt from `-sparse-compiler` to the same options in `brgemm-tpp.mlir`.

rengolin:
The common pattern is `GreedyPatternRewriteDriver::simplify` -> `ReshapeGenericOpForTpp::matchAndRewrite` -> `reshape2D` in `lib/TPP/ConvertLinalgToTpp.cpp:102`, which calls `mlir::linalg::tileLinalgOp`.

@chelini

rengolin:
Printing the `linalgOp` a few times between `continues` at `tileLinalgOp`:

```
(gdb) call linalgOp->dump()
"linalg.generic"(%18569, %18570) ({
^bb0(%arg2476: f32, %arg2477: f32):
  "linalg.yield"(%arg2476) : (f32) -> ()
}) {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0 * 2 + d2, d1 * 2 + d3)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"], library_call = "to-block-layout", operand_segment_sizes = array<i32: 1, 1>} : (memref<2x2xf32, strided<[16, 1], offset: ?>>, memref<1x1x2x2xf32, strided<[32, 4, 2, 1], offset: ?>>) -> ()
```

```
(gdb) call linalgOp->dump()
"linalg.generic"(%28109, %28110) ({
^bb0(%arg3748: f32, %arg3749: f32):
[...]

PR #115, opened 2022-11-02
Test case to verify tpp_identity library call

KavithaTipturMadhu:
Covers test 1 of https://github.com/plaidml/tpp-sandbox/issues/94

### DIFF
--- a/test/Integration/tpp-identity.mlir
+++ b/test/Integration/tpp-identity.mlir
@@ -1,0 +1,44 @@
+// RUN: tpp-opt %s -canonicalize -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-check-to-func  -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf |\
+// RUN: tpp-run \
+// RUN:  -e entry -entry-point-result=void  \
+// RUN: -shared-libs=%llvmlirdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext
+// 
+
+#map = affine_map<(d0, d1, d2, d3) -> (d3)>
+#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+
+module{
+func.func private @generate_1D_source(%width : index) -> tensor<64xf32> {
+  %init_source = arith.constant dense<0.0> : tensor<64xf32>
+  %source = linalg.generic {
+      indexing_maps = [affine_map<(d0) -> (d0)>],
+      iterator_types = ["parallel"]}
+      outs(%init_source : tensor<64xf32>) {
+    ^bb0(%b0 : f32):
+      %inner = linalg.index 0 : index
+      %inner_val_i32 = arith.index_cast %inner : index to i32
+      %inner_val = arith.sitofp %inner_val_i32 : i32 to f32
+      linalg.yield %inner_val :  f32
+  } -> tensor<64xf32>
+  return %source : tensor<64xf32>
+}
+
+ func.func @entry(){
+   %cst = arith.constant 64:index
+   %input_tensor = call @generate_1D_source(%cst): (index) -> (tensor<64xf32>)
+   %1 = tensor.empty() : tensor<12x56x56x64xf32> 
+   %2 = linalg.generic {indexing_maps=[#map,#map1], iterator_types = ["parallel", "parallel", "parallel", "parallel"], library_call = "tpp.identity"} ins(%input_tensor : tensor<64xf32> ) outs(%1 : tensor<12x56x56x64xf32>) {
+      ^bb0(%in: f32, %out: f32):
+        linalg.yield %in : f32
+  }->tensor<12x56x56x64xf32>
+  %3 = tensor.empty() : tensor<12x56x56x64xf32>
+  %4 = linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%input_tensor : tensor<64xf32> ) outs(%3 : tensor<12x56x56x64xf32>) {
+      ^bb0(%in: f32, %out: f32):
+        linalg.yield %in : f32
+  }->tensor<12x56x56x64xf32>
+ 
+  %threshold = arith.constant 0.0: f32
+  check.expect_almost_eq(%2, %4, %threshold): tensor<12x56x56x64xf32>, tensor<12x56x56x64xf32>, f32
+  return 
+ }
+}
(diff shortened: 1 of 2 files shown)


test/Integration/tpp-identity.mlir
```diff
@@ -0,0 +1,48 @@
+// RUN: tpp-opt %s -map-linalg-to-tpp -pre-bufferization -linalg-bufferize  -func-bufferize  -arith-bufferize -tensor-bufferize    -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-check-to-func -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf |\
```
chelini, line 1: `map-linalg-to-tpp` will mark the second generic with `tpp.identity`.

chelini, line 1: Let's also please drop `pre-bufferization` and let's use `one-shot`.

chelini (changes requested):

test/Integration/tpp-identity.mlir
```diff
@@ -0,0 +1,48 @@
...
+      linalg.yield %inner_val :  f32
+  } -> tensor<?xf32>
+  return %source : tensor<?xf32>
+}
+
+ func.func @entry(){
+   %cst = arith.constant 64:index
+   %input_var = call @generate_1D_source(%cst): (index) -> (tensor<?xf32>)
+   %input_tensor = tensor.cast %input_var: tensor<?xf32> to tensor<64xf32>
+   %input = bufferization.to_memref %input_tensor: memref<64xf32, strided<[?], offset:? >>
```
chelini, line 30: Ok I see why you did not use `one-shot`. You can do like this:
instead of `memref.alloc` use `tensor.empty` then prior to bufferization use: `empty-tensor-to-alloc-tensor`

KavithaTipturMadhu: I am not using memref.alloc here, the bufferization.to_memref is the one causing problems with one-shot-bufferization and I am using it to convert tensor to memref. Can you please suggest a workaround for this?

chelini:
``` #map = affine_map<(d0) -> (d0)>
#map1 = affine_map<(d0, d1, d2, d3) -> (d3)>
#map2 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
module {
  func.func private @generate_1D_source(%arg0: index) -> tensor<?xf32> {
    %0 = tensor.empty(%arg0) : tensor<?xf32>
    %1 = linalg.generic {indexing_maps = [#map], iterator_types = ["parallel"]} outs(%0 : tensor<?xf32>) {
    ^bb0(%out: f32):
      %2 = linalg.index 0 : index
      %3 = arith.index_cast %2 : index to i32
      %4 = arith.sitofp %3 : i32 to f32
      linalg.yield %4 : f32
    } -> tensor<?xf32>
    return %1 : tensor<?xf32>
  }
  func.func @entry() {
    %c64 = arith.constant 64 : index
    %0 = call @generate_1D_source(%c64) : (index) -> tensor<?xf32>
    %cast = tensor.cast %0 : tensor<?xf32> to tensor<64xf32>
    %1 = tensor.empty() : tensor<12x56x56x64xf32>
    %2 = linalg.generic {indexing_maps = [#map1, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel"], library_call = "tpp.identity"} ins(%cast : tensor<64xf32>) outs(%1 : tensor<12x56x56x64xf32>) {
    ^bb0(%in: f32, %out: f32):
      linalg.yield %in : f32
    } -> tensor<12x56x56x64xf32>
    %3 = tensor.empty() : tensor<12x56x56x64xf32>
    %4 = linalg.generic {indexing_maps = [#map1, #map2], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%cast : tensor<64xf32>) outs(%3 : tensor<12x56x56x64xf32>) {
    ^bb0(%in: f32, %out: f32):
      linalg.yield %in : f32
    } -> tensor<12x56x56x64xf32>
    check.expect_almost_eq(%2, %4, %cst) : tensor<12x56x56x64xf32>, tensor<12x56x56x64xf32>, f32
    return
  }
} ```

chelini: `./build/bin/tpp-opt test/Integration/tpp-identity.mlir -canonicalize -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize`

chelini: this are the passes you need to bufferize.

KavithaTipturMadhu: I have updated the test, but looks like the test fails to compile (compilation step hangs at convert-linalg-to-tpp stage).

chelini: Then there is a bug during conversion.

KavithaTipturMadhu:
I have added the second test case, but again its not using one-shot-bufferization and I am not sure how to convert memref.subviews to operate on tensors. Can you please suggest some workaround for this? The test itself hangs at linalg-to-tpp conversion, which is the same snippet as test case 1.

chelini:
> I have added the second test case, but again its not using one-shot-bufferization and I am not sure how to convert memref.subviews to operate on tensors. Can you please suggest some workaround for this? The test itself hangs at linalg-to-tpp conversion, which is the same snippet as test case 1.

We should not mix tensor and memref in this way. The entire test should be either in memref or tensor land.

KavithaTipturMadhu:
@chelini After rebasing on top of your changes, I noticed two problems:
1. map-linalg-to-tpp is converting both linalg.generic ops in tpp_identity test to tpps, which I didn't understand why
2. tpp_relu test is hanging at convert-linalg-to-tpp stage again.

chelini:
> @chelini After rebasing on top of your changes, I noticed two problems:
> 
> 1. map-linalg-to-tpp is converting both linalg.generic ops in tpp_identity test to tpps, which I didn't understand why
> 2. tpp_relu test is hanging at convert-linalg-to-tpp stage again.

`map-linalg-to-tpp` should only mark and not convert. `convert-linalg-to-tpp` will do the conversion. For the relu can you create an isolated test case? I suspect the problem comes from `reshape2D`.

rengolin:
> > @chelini After rebasing on top of your changes, I noticed two problems:
> > 
> > 1. map-linalg-to-tpp is converting both linalg.generic ops in tpp_identity test to tpps, which I didn't understand why
> > 2. tpp_relu test is hanging at convert-linalg-to-tpp stage again.
> 
> `map-linalg-to-tpp` should only mark and not convert. `convert-linalg-to-tpp` will do the conversion. For the relu can you create an isolated test case? I suspect the problem comes from `reshape2D`.

It most likely is, as exposed in #87

KavithaTipturMadhu:
The tests seem to be green after reshape2D fix (https://github.com/plaidml/tpp-sandbox/commit/38529869d79938299a12d2e327b925a85d8e35f1). I have created a separate issue for tensor.cast folding. Can we merge this now? @chelini @rengolin

chelini (approved):
Did you make sure that the second generic is not mapped to tpp?

KavithaTipturMadhu:
> Did you make sure that the second generic is not mapped to tpp?

I had missed removing map-linalg-to-tpp flag in the tpp-identity test file, added it in the last force push. The second generic is not mapped to tpp with this.

chelini:
Nice, thanks. Please squash and merge it

PR #123, opened 2022-11-03
More complex convs example

chelini:
Bias -> Conv [1x1] -> Relu -> Pad -> Conv [3x3] -> Relu

Map the first conv to brgemm while the second to gemm


rengolin (approved):

adam-smnk (approved):
Looks good, maybe it could use some more checks in the future

rengolin:
> Looks good, maybe it could use some more checks in the future

We're trying to reduce the number of checks that were too complex and break when the IR changes (but not the semantics).

Though, perhaps here we could have a `// CHECK-NOT: tpp.relu` if the call is fused?

Issue #116, opened 2022-11-02
Replace `mathx.relu` with `arith.maxf`


rengolin:
We have a `mathx` dialect that has `relu` operation, when we don't really need it, we can just use [arith.maxf](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithmaxf-mlirarithmaxfop) with `0.0`.

It'd also be good to transform other `linalg.generic` patterns, like `cmp(gt, 0.0)+select` into `arith.maxf`, so that we can detect and use in our transforms.

One day, these will all be `tcp.clamp`, but for now, we have to make do with what we have.

After replacing it in our passes and tests, we should delete the `mathx` dialect from our tree.

PR #124, opened 2022-11-03
Retire Mathx (NFC)

chelini:


rengolin:
Fixes #116

rengolin:
> more investigation is needed

Please, create an issue with the necessary information to reproduce and understand the problem

PR #128, opened 2022-11-04
Change behavior when converting to tpp

chelini:
- Generate parallel loops as default.
- Check static shape in `checkOperandForTpp`
- Drop duplicate code
- Revert to original behavior where we have two separate passes: 1 for detection and 1 for mapping (`populateMapLinalgToTppPatterns` is not removed from `populateConvertLinalgToTppPatterns`.

### DIFF
--- a/test/TPP/linalg-to-tpp.mlir
+++ b/test/TPP/linalg-to-tpp.mlir
@@ -51,29 +51,3 @@ func.func @matmul_lowering(%arg0: memref<8x9xf32>,
                 outs(%arg2: memref<8x8xf32>)
   return
 }
-
-// -----
-
-#map5 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
-
-// Make sure we detect a relu even if we don't have a tpp.mark.
-// CHECK-LABEL: func.func @relu(
-func.func @relu(%arg3: memref<64x32x32xf32>) -> memref<64x32x32xf32> {
-  // CHECK-DAG: %[[zero:.*]] = arith.constant 0 : index
-  // CHECK-DAG: %[[one:.*]] = arith.constant 1 : index
-  // CHECK-DAG: %[[sixtyfour:.*]] = arith.constant 64 : index
-  // CHECK: scf.parallel ([[i:.*]]) = (%[[zero]]) to (%[[sixtyfour]]) step (%[[one]]) {
-  // CHECK: %[[slice:.*]] = memref.subview
-  // CHECK: tpp.relu ins(%[[slice]] : memref<32x32xf32, #map>) out(%[[slice]] : memref<32x32xf32, #map>)
-  // CHECK: scf.yield
-  %c0 = arith.constant 0.0 : f32
-  linalg.generic {
-    indexing_maps = [#map5], 
-    iterator_types = ["parallel", "parallel", "parallel"]} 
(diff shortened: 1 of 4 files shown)


KavithaTipturMadhu (approved):

test/TPP/linalg-to-tpp.mlir
```diff
@@ -51,29 +51,3 @@ func.func @matmul_lowering(%arg0: memref<8x9xf32>,
                 outs(%arg2: memref<8x8xf32>)
   return
 }
-
-// -----
-
-#map5 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
```
rengolin, line 57: Why was this test dropped? Do we not recognise this any more as `tpp.relu`?

rengolin: Actually, this looks like a repeat of the test above, is that the reason?

chelini: No simply because we don't run the detection in the conversion pass anymore. We revert to original behavior where we first need to run `-map-linalg-to-tpp` and then `convert-linalg-to-tpp`. It is better to keep these as separate passes.

rengolin: Is there a similar test elsewhere, then?

chelini: Yes, the one above. Was the same only without the `tpp.relu` annotation.

rengolin: I see. Do we have tests that check that the mapping happens for all cases?

chelini: with all cases you mean `maxf(0, x)` and `maxf(0, x)`? If so, no. And the mapping is quite not right at the moment as we only check to have a `maxf`.

rengolin: Sorry, no. `(0, x)` can be normalised to `(x, 0)` and we're good.

This test seems to look at all "recogniseable" linalg ops that have the TPP shape (or annotation). IIUC, that's done by the mapping pass.

Do we have tests for the mapping pass, to make sure that test that you removed will continue to be recognised in the future? 

Here, we only have hand-coded IR that already has the annotation, I'm wondering if we have tests that make sure we keep adding annotation to the right constructs.

rengolin: >  And the mapping is quite not right at the moment as we only check to have a maxf

So, if we have a `maxf(x, 2)` it will map to RELU?

chelini: We don't have much: https://github.com/plaidml/tpp-sandbox/blob/main/test/TPP/simple-match.mlir and yes at the moment `maxf(x, 2)` will get annotated with a tpp.relu

chelini: We also need to make the relu single operand see: https://github.com/plaidml/tpp-sandbox/issues/129

rengolin: > We don't have much: https://github.com/plaidml/tpp-sandbox/blob/main/test/TPP/simple-match.mlir

That should be fine for the time being.

> and yes at the moment `maxf(x, 2)` will get annotated with a tpp.relu

#130

PR #131, opened 2022-11-04
Fix reshape2D

chelini:
Reshape2D tiles a linalg.generic by stripping out all but the last two innermost loops. The tile sizes for all but the two innermost loops are essentially the whole tensor dimension. But tiling by one will always succeed and give back the same tensor, thus the infinite loop.


rengolin (approved):
Don't forget to add "Fixes #87" on the commit message, or the issue won't be closed with this merge. (comments don't work, I tried earlier today).

PR #135, opened 2022-11-07
TPP Relu input buffer to be updated in place

KavithaTipturMadhu:


include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -108,16 +108,15 @@ def Tpp_ReluOp : Tpp_Op<"relu", [SameTypeOperands]> {
...
     ```
   }];
 
-  let arguments = (ins TppOperand:$input, TppOperand:$output);
+  let arguments = (ins TppOperand:$input);
 
   let assemblyFormat = [{
-      `ins` `(` $input `:` type($input) `)` 
-      `out` `(` $output `:` type($output) `)` attr-dict
+      `ins` `(` $input `:` type($input) `)`  attr-dict
```
chelini, line 119: Yes, this is what I had in mind. The direction looks good to me. I am not sure we want to use `ins` as the tensor is both the input *and* the output. Maybe `inout` but I don't like it much

KavithaTipturMadhu: I don't see inout defined in opbase.td, introducing it will need changes to LLVM that needs to be upstreamed, yes?

chelini: `ins` is just a string that will get printed, we can call it as we want. Similar for `input` we can have `(ins TppOperand:$inout)`.

chelini: We don't need any upstream extension.

rengolin (commented):
`outs` is accepted to be in+out especially in in-place ops, even though it's a "bit odd" that this op "doesn't have declared inputs". I don't mind either way.

test/TPP/tpp/tpp-invalid.mlir
```diff
@@ -21,7 +21,7 @@ func.func @tpp_add_invalid(%arg0: f32, %arg1: f32) {
 
 func.func @tpp_relu_invalid(%arg0: memref<1x2xf32>, %arg1: memref<2x1xf32>) -> memref<2x1xf32> {
 
-  // expected-error @below {{'tpp.relu' op requires all operands to have the same type}}
```
chelini, line 24: I think here you slightly change what the test was trying to do: check that input and output have the same operands. Since we have a single argument I think we can remove it.

tpp-rt/XsmmRunnerUtils.cpp
```diff
@@ -204,6 +204,13 @@ _mlir_ciface_xsmm_unary_invoke(const libxsmm_datatype dType, int64_t addr,
   kernel(&param);
 }
 
+extern "C" void
+_mlir_ciface_xsmm_unary_invoke_inline(const libxsmm_datatype dType,
```
chelini, line 208: Why do we need an extra function? Can we not use `unary_invoke` and `unary_dispatch`?

KavithaTipturMadhu: Unary invoke takes 4 arguments, whereas in this case, we only need 3. If we use the same unary_invoke as the 4 argument version, we'll have to somehow identify that the second UnrankedMemref operand is not used. Instead, I used a different function for inlined ops which has just 3 arguments. In case we want to inline some other op going further, we can reuse this function. Same dispatch is being used right now.

chelini: What is the other unary op? tpp.identity. I think we should have only 3 arguments, and avoid duplicating code. I am fine merging as is but please add a TODO and issue there. We can fix it later.

chelini (approved):

tpp-rt/XsmmRunnerUtils.cpp
chelini: Up to you (I already approved this PR).

KavithaTipturMadhu: I'll leave a todo and an issue to refactor inline functions to see if we can simplify this further and merge for now.

PR #137, opened 2022-11-07
Fix memory leak for `tpp-relu.mlir`

chelini:


KavithaTipturMadhu (approved):
sorry this was my bad, I should've checked the test for memory leaks.

Issue #129, opened 2022-11-04
Make tpp.relu a single operand op


KavithaTipturMadhu:
Make tpp.relu operate in place on its operand, instead of two operands (one for input and one for result).

PR #140, opened 2022-11-07
Add tests for benchmarks

rengolin:
Adding the benchmarks as tests using the new tpp-run infrastructure.

This allows us to test them at pre-merge checks and only benchmark them on post-merge checks, after building an infrastructure to run these tests through a harness.

Once these tests can be run as benchmarks, we can remove the old benchmark harness that is too complicated and doesn't scale.

A future commit will add the benchmarking harness.

Two changes to tpp-run:
 * Don't try to allocate a vector for the entire tensor, this leads to stack overflow. For now we loop a 2D tensor with a 1D vector. We should really create code that dynamically determines the rank, etc.
 * Changed to non-constant globals, so we can write to them on the kernel and get the updated values to print afterwards.

Continue working towards #74.

### DIFF
--- a/test/Benchmarks/matmul_kernel_12x6x9.mlir
+++ b/test/Benchmarks/matmul_kernel_12x6x9.mlir
@@ -1,0 +1,27 @@
+// RUN: tpp-opt %s -map-linalg-to-tpp \
+// RUN:            -pre-bufferization \
+// RUN:            -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
+// RUN:            -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize \
+// RUN:            -convert-linalg-to-tpp="enable-tiling" -convert-tpp-to-xsmm -loop-invariant-code-motion -convert-xsmm-to-func | \
+// RUN: tpp-run \
+// RUN:  -e entry -entry-point-result=void  \
+// RUN: -shared-libs=%llvmlirdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext | \
+// RUN: FileCheck %s
+
+#map0 = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+
+func.func @entry(%A: tensor<12x9xf32>, %B: tensor<9x6xf32>,
+                  %C: tensor<12x6xf32>) -> tensor<12x6xf32> {
+  %D = linalg.generic {indexing_maps = [#map0, #map1, #map2],
+                         iterator_types = ["parallel", "parallel", "reduction"]}
+    ins(%A, %B: tensor<12x9xf32>, tensor<9x6xf32>) outs(%C: tensor<12x6xf32>) {
+      ^bb0(%a: f32, %b: f32, %c: f32):
+        %0 = arith.mulf %a, %b : f32
+        %1 = arith.addf %c, %0 : f32
+        linalg.yield %1 : f32
+    } -> tensor<12x6xf32>
+  return %D : tensor<12x6xf32>
+}
+// CHECK-COUNT-12: ( 10, 10, 10, 10, 10, 10 )
--- a/test/Benchmarks/matmul_kernel_12x6x9.mlir
+++ b/test/Benchmarks/matmul_kernel_12x6x9.mlir
@@ -1,5 +1,4 @@
 // RUN: tpp-opt %s -map-linalg-to-tpp \
-// RUN:            -pre-bufferization \
 // RUN:            -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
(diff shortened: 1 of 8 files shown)


rengolin:
Interesting, I don't get that on my local build with sanitizers...

test/Benchmarks/matmul_kernel_12x6x9.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: tpp-opt %s -map-linalg-to-tpp \
+// RUN:            -pre-bufferization \
```
chelini, line 2: Can we drop `-pre-bufferization` in all tests? I need to remove this pass but I am waiting for @KavithaTipturMadhu to finish #129. I basically makes the relu and in-place operation for better bufferization. All the rest is now available upstream.

test/Benchmarks/matmul_kernel_12x6x9.mlir
```diff
@@ -0,0 +1,27 @@
...
+                         iterator_types = ["parallel", "parallel", "reduction"]}
+    ins(%A, %B: tensor<12x9xf32>, tensor<9x6xf32>) outs(%C: tensor<12x6xf32>) {
+      ^bb0(%a: f32, %b: f32, %c: f32):
+        %0 = arith.mulf %a, %b : f32
+        %1 = arith.addf %c, %0 : f32
+        linalg.yield %1 : f32
+    } -> tensor<12x6xf32>
+  return %D : tensor<12x6xf32>
+}
+// CHECK-COUNT-12: ( 10, 10, 10, 10, 10, 10 )
```
chelini, line 27: nit: no new line at the end of the file.

test/Benchmarks/mlp_kernel.mlir
```diff
@@ -0,0 +1,42 @@
+// RUN: tpp-opt %s -map-linalg-to-tpp \
+// RUN:            -main-closure -pre-bufferization -pack-matmul="block-factors=2,2" -loop-invariant-code-motion -canonicalize -undo-main-closure \
```
chelini, line 2: I would also drop `main-closure`, `undo-main-closure` and `loop-invariant-code-motion`.

tpp-run/tpp-run.cpp
```diff
@@ -182,18 +182,15 @@ static LogicalResult prepareMLIRKernel(Operation *op) {
...
     // For some reason, memref global op needs dense tensor type
     // See: lib/Dialect/MemRef/IR/MemRefOps.cpp :: GlobalOp::verify
     auto tensorType = RankedTensorType::get(memrefTy.getShape(), memrefTy.getElementType());
     auto floatInit = mlir::DenseElementsAttr::get(tensorType, floatValue);
-    auto constant = UnitAttr::get(ctx);
     auto alignment = builder.getIntegerAttr(builder.getI64Type(), 128);
 
     // Create the global object in the module's region
-    builder.create<memref::GlobalOp>(loc, nameAttr, privAttr, typeAttr, floatInit, constant, alignment);
+    builder.create<memref::GlobalOp>(loc, StringRef(name), privAttr, memrefTy, floatInit, /*constant=*/false, alignment);
```
chelini, line 193: can you please run `git-clang-format HEAD~1` or `clang-format -i tpp-run.cpp`.

tpp-run/tpp-run.cpp
```diff
@@ -221,15 +218,31 @@ static LogicalResult prepareMLIRKernel(Operation *op) {
   }
 
   // Read into a vector and print output
+  // We don't want to alloc the whole tensor as a vector,
+  // so we pick the inner dimension and iterate through the outer ones.
+  auto outputType = dyn_cast_or_null<MemRefType>(result.getType());
+  assert(outputType && "Unsupported return type");
+  VectorType vecType;
+  auto lastDim = outputType.getRank() - 1;
+  ArrayRef<int64_t> outer_dims(1);
```
chelini, line 227: CamelCase

tpp-run/tpp-run.cpp
```diff
@@ -221,15 +218,31 @@ static LogicalResult prepareMLIRKernel(Operation *op) {
...
+  if (outputType.getRank() > 1) {
+    ArrayRef<int64_t> inner_dims(&outputType.getShape()[lastDim], 1);
```
chelini, line 229: CamelCase

chelini (commented):
Quick comments. I will look tomorrow with fresh mind.

rengolin:
Sorry about the style issues, I'm not caring much about it at the moment because this is very early code and not at all generic. There are a number of fixmes that I still need to go through that are pretty fundamental.

I'll fix the issues you pointed for now, but the code will probably change a lot in the next iteration.

The important bits are:
* how to get the library paths
* how to get and change the cmd line options
* how to print dynamic shapes (or higher rank) in IR

test/Benchmarks/mlp_kernel.mlir
rengolin: This specific test does not work without `-pre-bufferization`, I haven't looked why yet. #142

test/Benchmarks/matmul_kernel_12x6x9.mlir
rengolin: Ack

test/Benchmarks/matmul_kernel_12x6x9.mlir
rengolin: Copy and paste from the benchmarks, which are old. Changing.

PR #141, opened 2022-11-08
Retire DecomposeConvToMatmulOrBrgemm and improve transform testing

chelini:


chelini:
This start to be a big PR. Please review as I would like to merge it.
Summary:
1. retire `DecomposeConvToMatmulOrBrgemm` now available via transform
2. find some bugs along the way: `hasMatmulBody` was too restrictive and allow only float. Relax it to allow also integers. The test `matmul-tpp-with-print.mlir` was not mapping to tpp at all, fix it. (Note that tpp only support float thus the change from i32 to f32).

rengolin (approved):

PR #145, opened 2022-11-09
Check to SCF Loops conversion pass

KavithaTipturMadhu:
Fixes https://github.com/plaidml/tpp-sandbox/issues/88

### DIFF
--- a/lib/TPP/ConvertCheckToLoops.cpp
+++ b/lib/TPP/ConvertCheckToLoops.cpp
@@ -1,0 +1,130 @@
+//===- ConvertCheckToLoops.cpp ----------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Dialect/Check/CheckOps.h"
+#include "TPP/Passes.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
+#include "mlir/Dialect/Math/IR/Math.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+using namespace mlir::check;
+using namespace mlir::cf;
+using namespace mlir::arith;
+using namespace mlir::math;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+// Convert 
+// check.expect_almost_equals(%t1:tensor<2x2xf32>, %t2:tensor<2x2xf32>,
+// %threshold:f32) 
+// to the pattern: 
+// scf.some_loop(%i, %j)
+// %0 = load from %t1 
+// %1 = load from %t2 
+// %diff = math.subf %0, %1 
+// %abs = arith.absf %diff 
+// %compare = arith.cmpf le, %abs, %threshold 
+// cf.assert %compare, "Result mismatch"
+struct ConvertAlmostEqualsOp
+    : public OpRewritePattern<check::ExpectAlmostEqOp> {
+  using OpRewritePattern<check::ExpectAlmostEqOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(check::ExpectAlmostEqOp almostEqOp,
+                                PatternRewriter &rewriter) const override {
+    Location loc = almostEqOp.getLoc();
+    SmallVector<Value> ubs;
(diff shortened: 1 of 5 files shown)


KavithaTipturMadhu:
I haven't added a test case yet, will do so. In the meantime, can you review the PR @chelini?

chelini:
> I haven't added a test case yet, will do so. In the meantime, can you review the PR @chelini?

yes.

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+  LogicalResult matchAndRewrite(check::ExpectAlmostEqOp almostEqOp,
+                                PatternRewriter &rewriter) const override {
+    Location loc = almostEqOp.getLoc();
+    SmallVector<Value> ubs;
+    size_t rank = almostEqOp.getLhs().getType().cast<MemRefType>().getRank();
+    for (size_t idx = 0; idx < rank; idx++) {
+      Value dim = rewriter.create<arith::ConstantIndexOp>(
+          loc,
+          almostEqOp.getLhs().getType().cast<MemRefType>().getShape()[idx]);
```
chelini, line 42: I don't recall but we all the check op have static shape right? If we allow dynamic you may want to have something like: https://github.com/plaidml/tpp-sandbox/commit/3e5f2db8b1e51dbb85d7474f4e53e97caa611dda

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -0,0 +1,115 @@
...
+          }
+          b.create<cf::AssertOp>(loc, compare,
+                                 b.getStringAttr("Result mismatch"));
+        });
+    rewriter.eraseOp(almostEqOp);
+    return success();
+  }
+};
+
+// Converts identity op.
```
chelini, line 76: wrong comment I assume.

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -0,0 +1,115 @@
...
+using namespace mlir::cf;
+using namespace mlir::arith;
+using namespace mlir::math;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+#include <iostream>
+namespace {
+
+struct ConvertAlmostEqualsOp
```
chelini, line 30: please document a bit the pattern

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -0,0 +1,115 @@
+//===- ConvertTppToLoops.cpp ------------------------------------*- C++ -*-===//
```
chelini, line 1: Wrong comment.

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -0,0 +1,133 @@
...
+    SmallVector<Value> ubs;
+    if (!almostEqOp.getLhs().getType().isa<MemRefType>()) {
+      return failure();
+    }
+    size_t rank = almostEqOp.getLhs().getType().cast<MemRefType>().getRank();
+    for (size_t idx = 0; idx < rank; idx++) {
+      // Works for dynamic shapes
+      // If the index is a constant, the dim value is folded into a constant
+      Value dimIndex = rewriter.create<arith::ConstantIndexOp>(loc, idx);
+      Value dim = rewriter.create<memref::DimOp>(loc, almostEqOp.getOperand(0),
```
chelini, line 58: Can this get folded? I was typically using `linalg::createFoldedDimOp`

chelini: https://github.com/plaidml/tpp-sandbox/blob/main/lib/TPP/TransformUtils.cpp#L147

KavithaTipturMadhu: This automatically folds if the dimIndex is a constant, from what I've noticed.

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -0,0 +1,131 @@
+//===- ConvertCheckToLoops.cpp ------------------------------------*- C++
```
chelini, line 1: nit. should be 80 chars, otherwise it will get formatted this way.

test/TPP/check/check-ops-loops.mlir
```diff
@@ -0,0 +1,28 @@
+// RUN: tpp-opt %s -map-linalg-to-tpp -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-check-to-loops | FileCheck %s
```
chelini, line 1: please drop `map-linalg-to-tpp` as not needed here.

test/TPP/check/check-ops-loops.mlir
```diff
@@ -0,0 +1,28 @@
...
+    ]> : tensor<4x4xf32>
+ %c =  arith.constant dense<[
+     [ 1.1, 2.1, 3.1, 4.1 ],
+     [ 1.2, 2.2, 3.2, 4.2 ],
+     [ 1.3, 2.3, 3.3, 4.3 ],
+     [ 1.4, 2.4, 3.4, 4.35 ]
+    ]> : tensor<4x4xf32>
+
+  %threshold = arith.constant 0.1: f32
+  // CHECK: scf.for {{.*}}{
```
chelini, line 18: should we also check loop bounds?

chelini (approved):

PR #146, opened 2022-11-09
Introduce more transform ops

chelini:


test/TPP/transform/transform-mlp.mlir
```diff
@@ -104,9 +104,16 @@ transform.sequence failures(propagate) {
     // Detect relus and identity
     %3 = transform.structured.match ops{["func.func"]} in %arg1
     transform.structured.map_linalg_to_tpp %3
+
+    transform.bufferization.one_shot_bufferize %arg1 {
```
rengolin, line 108: We should do this to all tests, right?

chelini: yes. Now we have all the transforms to go from linalg (at tensor) to tpp (at memref).

rengolin:
What's the plan to re-enable these tests without the transform blocks? We're removing a lot of those tests, but the functionality can't die, so we need that kind of tests anyway.

I think we should keep all the tests, and when we replace the passes with a default transform block, we don't need to add them all over again.

chelini:
> What's the plan to re-enable these tests without the transform blocks? We're removing a lot of those tests, but the functionality can't die, so we need that kind of tests anyway.
> 
> I think we should keep all the tests, and when we replace the passes with a default transform block, we don't need to add them all over again.

We are changing the driver not the functionality, if this makes sense. Let me know if this is not what you had in mind and you prefer something different.

rengolin:
> We still don't have a default pipeline (we actually never had one). The plan is to first migrate all to transform. Once we have all the basic block in transform we can add default pipeline tests. For default test I would expect something like conv or gemm. The functionality is just moving to transform.

I wonder if we need the same tests on both sides. If not, then how can we guarantee that our default transforms can catch all cases? This will be helpful as we develop the transforms later, to make sure we don't regress on the default pipeline by amending the default transforms accordingly as we add new functionality.

chelini:
ping @rengolin @KavithaTipturMadhu @adam-smnk Can I merge?

adam-smnk (commented):
Overall, looks good to me. I'm fine with the general direction and the changes.
But I'll let @rengolin make the final call.

rengolin (approved):

PR #151, opened 2022-11-10
Fix ConvertLinalgToTpp member self initialization

adam-smnk:
Noticed due to this warning:
`../lib/TPP/ConvertLinalgToTpp.cpp:447:26: warning: implicitly-declared ‘mlir::Pass::Option<bool>& mlir::Pass::Option<bool>::operator=(const mlir::Pass::Option<bool>&)’ is deprecated [-Wdeprecated-copy]
  447 |     this->enableTiling = enableTiling;`


chelini (approved):
Thanks

PR #154, opened 2022-11-11
Detect `maxf(x, N)` operation

### LINKED ISSUE #130: Fix match `maxf(x, N)` to relu
As exposed in #128, a `linalg.generic` with `arith.maxf(x, N)` (with N any non-zero number) will map to ReLU, which is wrong.

To map to ReLU, N must be `0.0`.

adam-smnk:
Fixes #130

The current logic might not cover all the possible sources of maxf zero arguments but the logic should be easily extendible when such need comes.

### DIFF
--- a/include/TPP/Dialect/Tpp/TppUtils.h
+++ b/include/TPP/Dialect/Tpp/TppUtils.h
@@ -36,6 +36,9 @@ bool hasMatmulBody(linalg::LinalgOp linalgOp);
 // Return true if the linalg operation has copy semantics.
 bool hasCopySemantics(linalg::LinalgOp linalgOp);
 
+// Return true if linalg generic region contains a maxf(x, 0) operation.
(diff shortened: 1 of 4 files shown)


rengolin (commented):
All of the cases I can think of, and it seems all of the tests, too, operate on floating point values (not tensors, memrefs). Is this just for completion, and if it is, how do we map it to `relu`?

adam-smnk:
All the extra logic is need to follow `maxf` operands outside of `linalg.generic` block.
Also, at least with the current pipeline, I cannot really expect to operate purely on tensors at the time of analysis. 
For example, when bufferization is enabled, it can bring operands into memref and inject extra `memref.copy` operations. The `@reluZeroMemref` test case explicitly shows such scenario.

rengolin (approved):
Looks good, thanks!

PR #156, opened 2022-11-13
(WIP) Retire additional passes

chelini:
- Retire `map-linalg-to-tpp`

- Retire `convert-linalg-to-tpp`


chelini:
Retire `convert-linalg-to-tpp` and `map-linalg-to-tpp` is difficult. Let's keep them for now.

PR #164, opened 2022-11-15
Factoring MLIRBench from main file

rengolin:
Changes how we lower MLIR to a set of helper functions. The API isn't settled yet, because we still need to lower benchmark calls and counters, which can complicate a bit, so it looks a bit of a mix between a driver and a static helper class. Once we finish the functionality, we can re-factor to a nicer API. Since it's small, should be easy.

Working towards #74.

### DIFF
--- a/tpp-run/CMakeLists.txt
+++ b/tpp-run/CMakeLists.txt
@@ -25,7 +25,9 @@ set(LLVM_LINK_COMPONENTS
   native
   )
 
-add_llvm_executable(tpp-run tpp-run.cpp)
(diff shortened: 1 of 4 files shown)


rengolin:
For the record, the functionality is **identical** to the previous version, but "factored out" in a helper class. 

This will allow us to play with loops, checks and prints in IR without making the lowering function overly confusing, as it separates the logic that is adding IR constructs (verbose and messy) from the logic that knows what to emit (that is simple and belongs to the runner's logic).

chelini (approved):

PR #166, opened 2022-11-16
Rework `MapAndConvertLinalgToTpp` (WIP)

chelini:


chelini:
Idem here: Retire convert-linalg-to-tpp and map-linalg-to-tpp is difficult. Let's keep them for now.

PR #170, opened 2022-11-17
Remove unused Unit tests directory

rengolin:
Do we plan to use that? It's been empty for a while...

If we want to do actual "unit" tests, we shouldn't be using LIT anyway.


chelini (approved):
Yes, we can remove. Can we also remove the directory?

rengolin:
Yup, I did, it's gone. :)

Issue #158, opened 2022-11-14
Crash when building vector print in `tpp-run`


rengolin:
As reported in #45:
```
********************
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60..
FAIL: TPP_OPT :: Benchmarks/simple-gemm.mlir (69 of 71)
******************** TEST 'TPP_OPT :: Benchmarks/simple-gemm.mlir' FAILED ********************
Script:
--
: 'RUN: at line 1';   /home/asiemien/tpp-sandbox/build/bin/tpp-opt /home/asiemien/tpp-sandbox/test/Benchmarks/simple-gemm.mlir -map-linalg-to-tpp  -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize  -convert-linalg-to-tpp -convert-tpp-to-xsmm  -convert-xsmm-to-func |  /home/asiemien/tpp-sandbox/build/bin/tpp-run   -e entry -entry-point-result=void   -shared-libs=/home/asiemien/llvm-project/build/./lib/libmlir_c_runner_utils.so,/home/asiemien/tpp-sandbox/build/lib//libtpp_c_runner_utils.so |  /home/asiemien/llvm-project/build/bin/FileCheck /home/asiemien/tpp-sandbox/test/Benchmarks/simple-gemm.mlir
: 'RUN: at line 12';   /home/asiemien/tpp-sandbox/build/bin/tpp-opt /home/asiemien/tpp-sandbox/test/Benchmarks/simple-gemm.mlir -map-linalg-to-tpp  -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize  -convert-linalg-to-tpp | /home/asiemien/llvm-project/build/bin/FileCheck -check-prefix=TPP /home/asiemien/tpp-sandbox/test/Benchmarks/simple-gemm.mlir
--
Exit Code: 2

Command Output (stderr):
--
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
[...]

PR #171, opened 2022-11-18
Disable benchmarks

chelini:
After #158 we can re-enable them.


adam-smnk:
Did benchmarks start actively failing on CI or sth?

chelini:
> Did benchmarks start actively failing on CI or sth?

Not on the CI, I think. If we want to keep checking this on the CI I can have this change locally.

rengolin:
Isn't this a problem only when using GCC?

rengolin (changes requested):
Let's not disable tests just because they fail locally. We've been through this. We need to open issues, reproduce the CI locally and change/add build modes once we fix them. But removing tests like that goes against forward progress and we can't have that.

chelini:
Ok. I will keep the change local.

PR #173, opened 2022-11-18
Fix TPP conversion

chelini:
We should *not* use annotation for converting linalg to tpp. This is a first patch to untangle the mapping and conversion pass.

### DIFF
--- a/include/TPP/Dialect/Tpp/TppUtils.h
+++ b/include/TPP/Dialect/Tpp/TppUtils.h
@@ -42,7 +42,33 @@ bool hasCopySemantics(linalg::LinalgOp linalgOp);
 bool hasMaxfZeroOp(linalg::LinalgOp linalgOp);
 
 // Returns true if the linalg generic is a tpp.matmul.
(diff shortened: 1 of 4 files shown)


adam-smnk (approved):

chelini:
> Are these `isTpp*` functions any different from `canMapToTpp*`?

Yes they are. For example, with `canMapToTppAdd` we can mark an addition with 4 parallel loops while this is not the case with `isTppAdd` that requires to have at most 2 loops. I don't like this approach either, but I don't see any better solution if we keep relying on linalg.generic, we need to bring back information by looking at the op. Suggestions/better flows are more than welcome. I think we should start looking at `linalg.map` and similars.

rengolin (approved):
So much better, thanks!

adam-smnk:
Ah I see, makes sense.
I share your view on generics but for now they are probably still the goto for most frontends. So, it makes sense to focus on them.
Maybe we could invest time into more gradual conversion infrastructure but I don't really see any major payoff atm.

PR #178, opened 2022-11-21
Use `mlir-cpu-runner` for integration

chelini:
`tpp-run` is not a replacement for `mlir-cpu-runner`. Currently, `tpp-run` rejects any MLIR module with an `entry` func.func. Thus use `mlir-cpu-runner` for integration tests and use `tpp-run` only for benchmarking. Most of the integration tests have already an entry func.func.


rengolin (approved):
Thanks! Agreed!

PR #182, opened 2022-11-23
Preserve parallelism when mapping to matmul

chelini:
At buffer level we can preserve parallelism when mapping a convolution to a matmul operation.

### DIFF
--- a/lib/TPP/MapConvToMatmul.cpp
+++ b/lib/TPP/MapConvToMatmul.cpp
@@ -276,9 +276,15 @@ mlir::linalgx::mapConvToMatmul(RewriterBase &rewriter,
   };
 
   Location loc = linalgOp.getLoc();
-  linalg::GenerateLoopNest<scf::ForOp>::doit(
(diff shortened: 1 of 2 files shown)


lib/TPP/MapConvToMatmul.cpp
```diff
@@ -276,9 +276,15 @@ mlir::linalgx::mapConvToMatmul(RewriterBase &rewriter,
   };
 
   Location loc = linalgOp.getLoc();
-  linalg::GenerateLoopNest<scf::ForOp>::doit(
-      rewriter, loc, loopRanges, linalgOp, linalgOp.getIteratorTypesArray(),
-      gemmBuilder);
+  if (linalgOp.hasBufferSemantics()) {
+    linalg::GenerateLoopNest<scf::ParallelOp>::doit(
```
KavithaTipturMadhu, line 280: Does this mean that linalg with memrefs is guaranteed to not have dependences resolved through memory?

KavithaTipturMadhu (approved):

Issue #175, opened 2022-11-21
Replace `benchmarks` with `tpp-run`


rengolin:
After #74 is complete, we can finally replace the programs in `benchmarks` with simple scripts calling `tpp-run`.

@hfp

PR #184, opened 2022-11-23
Benchmark driver for MLIR kernels

rengolin:
This change adds a benchmark harness to drive `tpp-run` through MLIR kernels, auto-detecting arguments so that we can just call them on select files or use it as an actual benchmark driver.

We still need to replace the existing benchmarks with calling the harness.

Working towards #175

### DIFF
--- a/benchmarks/harness/FileCheckParser/__init__.py
+++ b/benchmarks/harness/FileCheckParser/__init__.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+"""
+ Parse IR files for comments indicating metadata that we can actually use.
+
+ Usage:
+  params = FileCheckParser(self.logger).parse('kernel.mlir')
+
+ Params: A dictionary containing the following fields, if found
+    - 'iters': Iterations, from a RUN line
+    - 'entry': Kernel function name, from a RUN line
+    - 'libs': Shared library argument, from a RUN line
+    - 'expected_mean': From a BENCH_EXPECTED_MEAN line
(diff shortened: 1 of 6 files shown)


benchmarks/harness/FileCheckParser/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+    - 'expected_stdev': From a BENCH_EXPECTED_STDEV line
+"""
+
+import re
+
+class FileCheckParser(object):
+    """Parsers IR files for FileCheck lines to extract informatio
+       about the execution of the kernel"""
+
+    def __init__(self, logger=None):
```
adam-smnk, line 22: nit: methods assume presence of a `logger` so default `None` might blow up

benchmarks/harness/Logger/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+import logging
+import coloredlogs
+
+class Logger(object):
+    def __init__(self, name, parser, verbosity):
+        self.logger = logging.getLogger(name)
+
+        # Default level is WARNING (no output other than warnings and errors)
+        start = 30
```
adam-smnk, line 24: nit: `logging` should have enums for these e.g., `logging.WARNING`

adam-smnk (approved):
Really nice infrastructure

benchmarks/harness/FileCheckParser/__init__.py
rengolin: good point, now it's compulsory

PR #187, opened 2022-11-28
Replace old benchmarks with new infrastructure

rengolin:
This commit replaces the old benchmark infrastructure that relied on an implementation of memref in C to a generic MLIR runner (tpp-run) and a reference implementation in C.

The main changes are:
 * We no longer rely on a particular implementation of memref in C
 * We have a reference (plain C) implementation as a baseline
 * We have options to enable an XSMM version (still not implemented)
 * We count performance in GFLOPs on both sides (MLIR and C)
 * The benchmark driver is a simple bash script that cna be easily extended and runs both MLIR and C programs, passing args

Still pending:
 * Actual XSMM implementation on the C side
 * Improve the shell scripts to call all three (plain, xsmm, mlir)
 * Allow the mlir version to skip TPP opts (to calculate delta)
 * Allow the MLP program to define the number of layers (C and mlir)

Once all the pending tasks are done, we can start adding more complex vision and language models (convolution+residual, multi attention head) and start the actual benchmarking.

Working towards #175


benchmarks/run-all.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+source common.sh
+
+echo "SIMPLE COPY:"
+pushd simple_copy > /dev/null
+./run.sh
```
adam-smnk, line 7: nit: could pass arguments given to `run-all.sh` to the subscripts (assuming the arguments are common for all the runs)

benchmarks/common.sh
```diff
@@ -1,83 +1,193 @@
...
+  ARGS="$*"
+  CMD="$HARNESS -q $SRC -n $ITER $ARGS"
+  if ! $CMD; then
+    >&2 echo "ERROR: Harness failed"
+    >&2 echo "   $CMD"
+    exit 1
+  fi
+}
+
+# run_benchmark()
```
adam-smnk, line 149: nit: mix of positional and getopts arguments makes UX a bit confusing

adam-smnk (approved):
Tested it a bit locally, it seems to be running fine.

rengolin:
General comment: I don't care much now about the numbers, just the infrastructure. The numbers quality will improve in future PRs.

benchmarks/common.sh
rengolin: It is, and I intend to fix it, but it's not easy when the arguments that I want to pass down are different. Note that there is no UX here, this is mostly to drive the benchmarks via CI, and help somewhat reproduce the errors.

Users should really be using the harness (for the MLIR part) and the C part should be stable and consistent, so no need to modify or inspect the results.

benchmarks/run-all.sh
rengolin: I think I can make all args to be the same (or similar), it will improve usability not just for users, but also to make the benchmark more flexible, I agree. Let me try something...

benchmarks/common.sh
adam-smnk: Fair point. We might want to tweak the baseline in the future but there’s no substantial gain in making it particularly pretty at the backend.

benchmarks/run-all.sh
rengolin: Honestly, bash's `getopts` is really bad. It assumes that you know all arguments and that it's impossible to do partial parsing (only look at some of the args, not all) and it errors out on any unknown options, even if I chose to handle the error, I can't just continue. So, I'll let it be as is for now and honestly, we should have a more robust driver soon.

rengolin:
After trying to improve this, I found out that it's leading us into a cul-de-sac of bad decisions, so I'll rewrite this in a more appealing way, and make use of configurations for benchmarks, as discussed with @hfp earlier this week. Closing this for now.

PR #189, opened 2022-11-30
VNNI to Tpp lowering via transforms

KavithaTipturMadhu:


KavithaTipturMadhu:
@chelini The PR is ready for review.

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
```diff
@@ -0,0 +1,102 @@
...
+#define GET_OP_LIST
+#include "TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp.inc"
+        >();
+  }
+};
+} // namespace
+
+namespace {
+// A simple pattern rewriter that implements no special logic.
+class SimpleRewriter : public PatternRewriter {
```
chelini, line 51: now in upstream MLIR they have this: https://github.com/llvm/llvm-project/commit/a8850312c106d71f9e35fd902f9dcd3c4ac0a690

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
```diff
@@ -0,0 +1,102 @@
...
+        matmulOp.getMatrixC());
+    return success();
+  }
+};
+
+  
+DiagnosedSilenceableFailure transform::MapVNNIToTppOp::applyToOne(
+    Operation *target, SmallVector<Operation *> &results,
+    TransformState &state) {
+  if (!target->hasTrait<OpTrait::IsIsolatedFromAbove>()) {
```
chelini, line 84: can you please document this in the op description? It also seems that the patter in this case is relatively easy so we can get away with a failureOr<tpp::VNNIMatmul> instead of running the pattern rewriter?

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
rengolin: We haven't got that yet on our tree. It should be a trivial fix when we bump LLVM again, no need to push LLVM in this PR.

test/BF16/vnni-to-tpp.mlir
```diff
@@ -0,0 +1,33 @@
+// RUN: tpp-opt  --one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" --transform-dialect-interpreter %s | FileCheck %s 
+
+module {
+  transform.sequence failures(propagate) {
+  ^bb0(%arg0: !pdl.operation):
+    %0 = transform.structured.match ops{["matmul_static"]} in %arg0
+    transform.structured.map_vnni_to_tpp %0
+  }
+
+  transform.sequence failures(propagate) {
```
rengolin, line 10: Why the two ways? Don't they mean the same thing? Which one applies first?

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
KavithaTipturMadhu: Running the conversion on an entire function requires that I use pattern rewriter, because applyToOne does not get invoked with just one op (vnni.matmul) but with the whole function, right? Also, I am invoking conversion on the whole function because we need to convert brgemm calls as well. Does this seem right to you?

test/BF16/vnni-to-tpp.mlir
KavithaTipturMadhu: my bad, I'll get rid of the redundant one.

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
KavithaTipturMadhu: I have updated the patch so that "map_vnni_to_tpp" transform applies to vnni.matmul ops directly instead of func. With this change, the op needn't be isolated from above and hence, I have removed it.

rengolin: I think for the VNNI to TPP pass it should be fine to be a function pass because it will always be 1:1 and the shapes should have been checked (and matched) by the Linalg to VNNI pass.

rengolin:
@chelini any more comments? Looks good to me otherwise. Thanks!

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
```diff
@@ -0,0 +1,88 @@
...
+public:
+  SimpleRewriter(MLIRContext *context) : PatternRewriter(context) {}
+};
+} // namespace
+
+DiagnosedSilenceableFailure
+transform::MapVNNIToTppOp::applyToOne(Operation *target,
+                                      SmallVector<Operation *> &results,
+                                      TransformState &state) {
+  if (auto matmulOp = dyn_cast<vnni::MatmulOp>(target)) {
```
chelini, line 61: can you please return failure immediately instead of nesting inside an if, thanks

chelini:
Just left a comment. Looking at the PR it seems that we plan to do lowering with transform, just wondering if this is what we want.

rengolin (approved):
If Lorenzo is happy, I'm happy. :)

chelini (approved):

PR #196, opened 2022-12-02
Replace old benchmarks with new infrastructure

### LINKED ISSUE #175: Replace `benchmarks` with `tpp-run`

rengolin:
This commit replaces the old benchmark infrastructure that relied on an implementation of memref in C to a generic MLIR runner (tpp-run) and a reference implementation in C.

The main changes are:
 * We no longer rely on a particular implementation of memref in C
 * We have a reference (plain C) implementation as a baseline
 * We have options to enable an XSMM version (still not implemented)
 * We count performance in GFLOPs on both sides (MLIR and C)
 * The benchmark driver is a simple Python script that can be easily more benchmarks, variations, arguments

Still pending:
 * Actual XSMM implementation on the C side

Once all the pending tasks are done, we can start adding more complex vision and language models (convolution+residual, multi attention head) and start the actual benchmarking.

Fixes #175


rengolin:
Without logs, the output of the driver is:
```
$ ./driver.py 
Benchmark: simple_copy
ref                 :     7.399 +-     0.548 gflops
mlir                :     7.130 +-     0.467 gflops

Benchmark: mlp
ref                 :     1.427 +-     0.017 gflops
mlir                :    56.335 +-     0.225 gflops

Benchmark: matmul
ref_48x64x96        :     3.237 +-     0.218 gflops
mlir_48x64x96       :   131.940 +-     4.236 gflops
ref_64x48x96        :     3.229 +-     0.084 gflops
mlir_64x48x96       :   117.948 +-     3.782 gflops
ref_64x64x64        :     3.820 +-     0.001 gflops
mlir_64x64x64       :   131.704 +-     5.901 gflops
```

benchmarks/driver.py
```diff
@@ -0,0 +1,333 @@
...
+        return True
+
+class CPPRun(BaseRun):
+    """ C++ runs """
+
+    def __init__(self, name, env, json, logger):
+        BaseRun.__init__(self, name, env, json, logger)
+        assert(json["type"] == "C++")
+        source_dir = os.path.dirname(self.source)
+        self.binary = os.path.join(source_dir, name)
```
adam-smnk, line 127: I'd suggest adding some prefix/suffix to the created binaries and adding it to `.gitignore` to avoid accidental commits with binaries when `-k` flag is used.

benchmarks/matmul/matmul.cc
```diff
@@ -0,0 +1,142 @@
...
+
+  // Cmd-line args
+  int opt;
+  while ((opt = getopt(argc, argv, "n:xv")) != -1) {
+    switch (opt) {
+      case 'n':
+        iter = atoi(optarg);
+        break;
+      case 'x':
+        //kernel = matmul_xsmm;
```
adam-smnk, line 69: nit: could use assert like the other benchmarks

adam-smnk (approved):
+1 elegant and easily extensible infrastructure
I haven't looked much into the ref implementations but I assume they will be tweaked separately later if needed.

rengolin:
> I haven't looked much into the ref implementations but I assume they will be tweaked separately later if needed.

Correct. Next we'll add XSMM implementations, etc. Since there's a lot of common code, I also want to factor everything but the kernel, but since the kernels have different arguments (and I don't want to use `vararg`), I'll have to do that in C++ with some templates and lambdas. Then it should be pretty easy to just add a kernel and pass that to a C++ "driver" that will run, measure and print the results.

All of that for later, though.

rengolin:
Thanks Adam. I'm going to merge this and we can iterate in-tree.

Issue #190, opened 2022-11-30
Add a resnet-like test/benchmark


rengolin:
We need to break Resnet down into blocks and add those blocks as individual tests.

Later layers of a Resnet 50 have the "bottleneck", which replaces two `3x3` convolutions with `{ 1x1, 3x3, 1x1 }`. This allows us to test BRGEMMs for the `1x1` and GEMMs for the `3x3`. We can also use the first convolution to test a `7x7` filter.

The tasks in this issue are:
1. Get the first bottleneck into a self-contained MLIR file in `test/Benchmarks`
   1. Add `tpp-opt` checks for the appropriate TPP operations
2. Create fake dense weights as global memrefs and leave only the inputs as parameters
   1. Add `tpp-run` checks for output
3. Add the MLIR file as a benchmark in `benchmarks`

Repeat for the  `7x7` layer. We may want to also add the fully-connected layer at the end

Once benchmarks are working with XSMM calls, implement an XSMM function that reproduces a bottleneck for comparison with the MLIR file.

rengolin:
@nhasabni

nhasabni:
Reference PR #239

rengolin:
Is this complete with the merge of #239?

nhasabni:
> Is this complete with the merge of #239?

I think 7x7 conv block is not complete yet. Otherwise, bottleneck block is complete.

rengolin:
With 1x1 and 3x3 done on both output and execution, I think the 7x7 should be fine with just `tpp-opt` output calling TPP ops after the expected pack and we're done, then.

rengolin:
Also, the C++ implementation in XSMM will be done using the libxsmm-dnn repository and is covered by a separate issue (#197)

nhasabni:
Sounds good then. Feel free to close the issue then.

rengolin:
Ok, let's focus on other benchmarks before increasing the scope of this one.

PR #204, opened 2022-12-05
Reduce diff between upstream pack/unpack and in-tree ones

chelini:
Use `DenseI64ArrayAttr` and reduce diff with upstreamed pack and unpack operations.


lib/TPP/Dialect/LinalgX/LinalgXOps.cpp
```diff
@@ -128,43 +128,43 @@ LogicalResult UnPackOp::canonicalize(UnPackOp unpackOp,
...
+            innerTiles, builder.getDenseI64ArrayAttr(staticInnerTiles),
+            /*padding_value=*/nullptr);
+    else
+      build(builder, result, TypeRange{}, input, output,
+            builder.getDenseI64ArrayAttr(outerDimPerm),
+            builder.getDenseI64ArrayAttr(innerDimPos), innerTiles,
+            builder.getDenseI64ArrayAttr(staticInnerTiles),
+            /*padding_value=*/{});
+
+  } else {
```
rengolin, line 155: assert(tensor)?

lib/TPP/Dialect/LinalgX/LinalgXOps.cpp
```diff
@@ -176,30 +176,30 @@ void UnPackOp::build(OpBuilder &builder, OperationState &result, Value input,
...
-  SmallVector<int64_t> staticInnerTiles;
-  dispatchIndexOpFoldResults(tiles, innerTiles, staticInnerTiles,
-                             ShapedType::kDynamic);
-  if (!outerDimPerm.size())
-    build(builder, result, llvm::None, input, output, /*outerDimPerm=*/{},
-          innerDimsPos, innerTiles, builder.getI64ArrayAttr(staticInnerTiles));
-  else
-    build(builder, result, llvm::None, input, output, outerDimPerm,
-          innerDimsPos, innerTiles, builder.getI64ArrayAttr(staticInnerTiles));
+  } else {
```
rengolin, line 191: ditto

test/TPP/transform/transform-convolutions.mlir
```diff
@@ -285,3 +285,145 @@ transform.sequence failures(propagate) {
     %9 = transform.structured.interchange %8 { iterator_interchange = [0, 1, 4, 2, 3, 5] }
     transform.structured.map_to_brgemm %9
 }
+
```
rengolin, line 288: please, drop this change. This should be attached to #190 and done in one go, not half-way on an uinrelated PR.

rengolin:
The failure is probably because you need to rebase

chelini:
> Another change is the tensor return type / output,. which is the source of the #45 problem. Can you fix that too? Should be quick and if if fixes the bug, win win!



> sor return type / output,. w

What exactly should I fix it?

rengolin:
The type in the assembly format to look like this: https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Tensor/IR/TensorOps.td#L1754

rengolin (approved):
As discussed offline, the change in the return type is due to the op here supporting both tensors and memrefs, and we can't change that now because the upstream ones only support tensors.

This can be a change for another PR. LGTM, thanks!

PR #211, opened 2022-12-08
Fix for VNNI pack to be applied to the second op and not first

KavithaTipturMadhu:


include/TPP/Dialect/VNNI/VNNIInterfaces.td
```diff
@@ -0,0 +1,86 @@
+#ifndef VNNI_IR_VNNIINTERFACES
```
chelini, line 1: this is new? Look to me I already review this code.

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -104,15 +104,15 @@ struct ConvertTppVNNIMatmulOp : public OpRewritePattern<tpp::VNNIMatmulOp> {
...
-      return rewriter.notifyMatchFailure(matmulOp, "Cannot compute lda");
-    lda = lda / (*divLdaDim);
 
     auto ldbDim = getLeadingDim(memrefB);
     if (failed(ldbDim))
       return rewriter.notifyMatchFailure(matmulOp, "Cannot compute ldb");
     int64_t ldb = *ldbDim;
+    auto divLdbDim = getLeadingDim(memrefB, 1);
+    if (failed(divLdbDim))
+      return rewriter.notifyMatchFailure(matmulOp, "Cannot compute ldb");
```
chelini, line 114: can you please adjust error msg?

lib/TPP/Dialect/VNNI/BufferizableOpInterfaceImpl.cpp
```diff
@@ -72,8 +72,8 @@ struct MatmulLayoutInterface
       return failure();
     Value srcBufferB = *maybeSrcBufferB;
 
-    rewriter.create<vnni::MatmulOp>(op->getLoc(), destBuffer.getType(),
-                                    srcBufferA, srcBufferB, destBuffer);
+    rewriter.create<vnni::MatmulOp>(op->getLoc(), TypeRange{}, srcBufferA,
```
chelini, line 75: as a follow-up maybe you can think of adding a custom builder.

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
```diff
@@ -0,0 +1,92 @@
...
+#define GET_OP_LIST
+#include "TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp.inc"
+        >();
+  }
+};
+} // namespace
+
+namespace {
+// A simple pattern rewriter that implements no special logic.
+class SimpleRewriter : public PatternRewriter {
```
chelini, line 51: this should not be needed anymore after my llvm bump. Simply include the file: https://github.com/llvm/llvm-project/commit/a8850312c106d71f9e35fd902f9dcd3c4ac0a690. Bonus point if you do that also for other transform we have.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -404,16 +404,16 @@ mlir::linalgx::packVNNIMatmulOp(RewriterBase &rewriter,
...
 
   OpFoldResult tileOnI = tiles[0];
-  SmallVector<OpFoldResult, 1> tilesOnA = {tileOnI};
+  SmallVector<OpFoldResult, 1> tilesOnB = {tileOnI};
 
   Location loc = matmulOp.getLoc();
   // reshape input A.
-  Value packedMatrixA =
-      toPackLayout_VNNI(loc, matmulOp.getInputs()[0], tilesOnA, rewriter);
+  Value packedMatrixB =
```
chelini, line 411: pls adjust comment.

chelini (approved):

include/TPP/Dialect/VNNI/VNNIInterfaces.td
KavithaTipturMadhu: The patch is rebased on top of the vnni to tpp code, you'll probably see the same changes here. Once we merge the previous patch, we can review this better.

lib/TPP/Dialect/VNNI/TransformOps/VNNITransformOps.cpp
KavithaTipturMadhu: I have fixed this transform in  the other patch https://github.com/plaidml/tpp-mlir/pull/189. I will update this patch with trivialpatternrewriter for the other transform.

PR #214, opened 2022-12-09
Remove irrelevant Resnet file

rengolin:
This file was generated by PlaidML on an MLIR from a year ago and is no longer relevant. We can't use it for testing or benchmarking anyway.


chelini (approved):

rengolin:
#190 will take care to add a test/benchmark for Resnet like modules.

PR #215, opened 2022-12-09
Improve propagation of unpack

chelini:
Originally we required a single unpack producer for the linalg generic (element-wise); we now relax this restriction and enable multiple consumers as long as the information provided is non-conflicting.

### DIFF
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -482,6 +482,98 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
 // PropagateThroughElementWiseOp
 //===----------------------------------------------------------------------===//
 
+struct PackInfo {
+  llvm::DenseMap<int64_t, OpFoldResult> dimAndTileMapping;
+  SmallVector<int64_t> tileLoopPerms;
+  SmallVector<int64_t> interchangeVector;
+
+  LogicalResult extractPackInfoFromProducers(linalg::GenericOp genericOp);
+  bool isConsistent(int64_t dim, OpFoldResult tile);
+};
+
+// Return true if the tile stored in `dimAndTileMapping` is consistent
+// with `tile`.
+bool PackInfo::isConsistent(int64_t dim, OpFoldResult tile) {
+  if (!dimAndTileMapping.count(dim))
+    return true;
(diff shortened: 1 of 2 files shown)


lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -482,6 +482,98 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
 // PropagateThroughElementWiseOp
 //===----------------------------------------------------------------------===//
 
+struct PackInfo {
```
rengolin, line 485: This is very different from https://reviews.llvm.org/D139680. We already have a big delta between upstream pack and unpack, can we keep it low by following what is being done upstream here too?

chelini: the struct or the entire PR?

rengolin: For now just the struct. We need to converge on everything that we can, so that we can pull the upstream changes as soon as possible.

chelini: I can align the name better but we have `SmallVector<int64_t> interchangeVector;` which is something not upstream. Also notice that this PR is for the unpack while upstream is for pack. Is this name-changing fine?

rengolin: Sorry, `DenseMap`

rengolin: > Could you please be more specific on what you are referring to?
> 
> `packedTileSize` -> `dimAndTileMapping` [same container] `outerDimsOnDomainPerm`->`tileLoopPerms` [same container]
> 
> `interchangeVector` is not in the upstream as it is not used.

Ah, you're right! Perhaps just a renaming should do it then.

chelini:
@rengolin can I merge it?

rengolin:
This has the similar functionality as https://reviews.llvm.org/D139680, so if we want to reduce the cost of rebasing later, we should do the implementation of whatever they have there in the same way, most strikingly `getPackingInfoFromConsumer` and `getOrCreatePackedViewOfOperand`.

It also seems the upstream discussion isn't concluded, so we're not sure that what is implemented there is really going to be final. We need to make sure the upstream version is solid and then follow suite locally.

Issue #100, opened 2022-10-31
Perf dialect


rengolin:
With the new runner, we need to time the execution of the kernel only, but because the runner calls the JIT compiler, we can't separate the compilation from the execution, and the preparation of tensors from the kernel. But if we had a simple dialect that allows us to start/stop a timer and collect some statistics, we could have that in the input IR (or even add it as a pass), and then we wouldn't need any callbacks, etc.

We'll need a set of ops like:
* `%t = perf.start_timer: perf.timer_t` for some definition of `timer_t` (could be a unique integer)
* `%delta = perf.stop_timer(%t): f64` for the delta from a specific timer, closes the timer (ie. can't take another stop, `%t` is dead)
* `perf.push(%list: memref<1000xf64>, %delta: f64): void` for adding the current delta to a list of results
* `%mean = perf.mean(%list: memref<1000xf64>): f64` for the mean of those 1000 values
* `%dev = perf.stdev(%list: memref<1000xf64>): f64` for the standard deviation of those 1000 values

And a simple runtime implementation for those, calling `std::chrono` or something.

If we get this right, this could be a simple dialect that could easily be upstreamed to MLIR directly without much effort.

chelini:
Indeed would be nice to have.

adam-smnk:
Summary:
- perf dialect added
- perf lowering passes added:
-- `perf.bench` to loops
-- other perf ops to function calls - some have native MLIR implementation, timing related logic requires runtime support
- our benchmarks now use perf dialect for the main benchmarking loop

PR #218, opened 2022-12-12
Perf dialect (#100)

adam-smnk:
Adds a new infrastructure dialect that facilitates code performance benchmarking directly at the IR level.

The perf dialect introduces a set of operations that allows to collect and process basic code runtime statistics.
The dialect positions itself as a high-level generic utility which relies on support of a runtime that supplies platform dependent functionality such as timers.

Work towards #100 

This PR introduces just core operations and basic utilities.
Lowering to loops and func will follow separately.

### DIFF
--- a/include/TPP/Dialect/Perf/PerfDialect.h
+++ b/include/TPP/Dialect/Perf/PerfDialect.h
@@ -1,0 +1,17 @@
+//===- TppDialect.h - Tpp dialect -------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TPP_PERF_DIALECT_H
+#define TPP_PERF_DIALECT_H
+
+// clang-format off
+#include "mlir/IR/Dialect.h"
+#include "TPP/Dialect/Perf/PerfOpsDialect.h.inc"
+// clang-format on
+
+#endif // TPP_PERF_DIALECT_H
--- a/include/TPP/Dialect/Perf/PerfDialect.h
+++ b/include/TPP/Dialect/Perf/PerfDialect.h
@@ -1,4 +1,4 @@
-//===- TppDialect.h - Tpp dialect -------------------------------*- C++ -*-===//
+//===- PerfDialect.h - Perf dialect -----------------------------*- C++ -*-===//
 //
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
--- a/include/TPP/Dialect/Perf/PerfOps.h
+++ b/include/TPP/Dialect/Perf/PerfOps.h
@@ -9,6 +9,7 @@
 #ifndef TPP_DIALECT_PERF_PERFOPS_H
 #define TPP_DIALECT_PERF_PERFOPS_H
 
+#include "TPP/Dialect/Perf/PerfTypes.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Dialect.h"
 #include "mlir/IR/OpDefinition.h"
--- a/include/TPP/Dialect/Perf/PerfOps.h
+++ b/include/TPP/Dialect/Perf/PerfOps.h
@@ -1,0 +1,23 @@
+//===- TppOps.h - Tpp dialect ops -------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TPP_DIALECT_PERF_PERFOPS_H
+#define TPP_DIALECT_PERF_PERFOPS_H
+
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Dialect.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/OpImplementation.h"
+#include "mlir/Interfaces/ControlFlowInterfaces.h"
+#include "mlir/Interfaces/InferTypeOpInterface.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+
+#define GET_OP_CLASSES
+#include "TPP/Dialect/Perf/PerfOps.h.inc"
+
+#endif // TPP_DIALECT_PERF_PERFOPS_H
--- a/include/TPP/Dialect/Perf/PerfOps.h
+++ b/include/TPP/Dialect/Perf/PerfOps.h
@@ -1,4 +1,4 @@
-//===- TppOps.h - Tpp dialect ops -------------------------------*- C++ -*-===//
+//===- PerfOps.h - Perf dialect ops -----------------------------*- C++ -*-===//
 //
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
--- a/include/TPP/Dialect/Perf/PerfOps.td
+++ b/include/TPP/Dialect/Perf/PerfOps.td
@@ -1,0 +1,261 @@
+//===- TppOps.td - Tpp dialect ops -------------------------*- tablegen -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TPP_PERF_OPS
+#define TPP_PERF_OPS
+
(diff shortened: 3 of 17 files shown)


include/TPP/Dialect/Perf/PerfDialect.h
```diff
@@ -0,0 +1,17 @@
+//===- TppDialect.h - Tpp dialect -------------------------------*- C++ -*-===//
```
chelini, line 1: fix please.

include/TPP/Dialect/Perf/PerfOps.h
```diff
@@ -0,0 +1,23 @@
+//===- TppOps.h - Tpp dialect ops -------------------------------*- C++ -*-===//
```
chelini, line 1: fix please.

include/TPP/Dialect/Perf/PerfOps.td
```diff
@@ -0,0 +1,261 @@
+//===- TppOps.td - Tpp dialect ops -------------------------*- tablegen -*-===//
```
chelini, line 1: same.

include/TPP/Dialect/Perf/PerfOps.td
```diff
@@ -0,0 +1,261 @@
...
+
+    ```mlir
+
+    %deltas = perf.bench (%n) {
+      ... // ops under measurement
+    } -> memref<?xf64>
+
+    ```
+
+    Optionally, a yield op can be added:
```
chelini, line 44: This comment is confusing what you want to say is mentioned in the yield op: If the parent operation defines no values, then the `perf.yield`...

lib/TPP/Dialect/Perf/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+using namespace mlir;
+using namespace mlir::bufferization;
+using namespace mlir::perf;
+
+namespace mlir {
+namespace perf {
+namespace {
+
+// TODO: bufferization interface for check ops
```
chelini, line 23: wrong comment?

lib/TPP/Dialect/Perf/PerfOps.cpp
```diff
@@ -0,0 +1,49 @@
...
+      ++numStopTimers;
+  }
+  if (numStopTimers != 1)
+    return emitOpError("timer stopped multiple times");
+
+  return success();
+}
+
+LogicalResult YieldOp::verify() {
+  // Get the parent operation to check its return values
```
chelini, line 35: please finish sentence with `.`

lib/TPP/Dialect/Perf/PerfOps.cpp
```diff
@@ -0,0 +1,49 @@
...
+  auto benchOp = (*this)->getParentOfType<BenchOp>();
```
chelini, line 36: I believe this is already taken into account in `HasParent<"BenchOp">`

lib/TPP/Dialect/Perf/PerfOps.cpp
```diff
@@ -0,0 +1,49 @@
...
+  if (!benchOp)
+    return emitOpError("invalid parent operation");
+
+  // Check that body results match the yield operands
+  auto types =
```
chelini, line 41: Probably also this logic can be move in the .td file. See for example https://github.com/llvm/llvm-project/blob/78c033b541f00d12b7e1ba2a676b02e022c9beb1/mlir/include/mlir/Dialect/Tensor/IR/TensorOps.td#L1669

include/TPP/Dialect/Perf/PerfOps.td
```diff
@@ -0,0 +1,261 @@
...
+  let builders = [OpBuilder<(ins), [{ /* nothing to do */ }]>];
+
+  let assemblyFormat =
+    [{  attr-dict ($results^ `:` type($results))? }];
+
+  let hasVerifier = 1;
+}
+
+//===----------------------------------------------------------------------===//
+// StartTimerOp
```
rengolin, line 97: Perhaps begin with these ops first, then say `perf.bench` is like: `loop(n) { start; body; stop; yield delta }`

include/TPP/Dialect/Perf/PerfOps.td
```diff
@@ -0,0 +1,261 @@
...
+    The `perf.start_timer` operation creates a new timer
+    which begins time measurement.
+
+    See `perf.stop_timer` for timer termination.
+
+    Example:
+
+    ```mlir
+
+    %timer = perf.start_timer : i64
```
rengolin, line 112: What is the semantics of this `i64`? Unique? auto-increment? tied to some implementation detail?

include/TPP/Dialect/Perf/PerfOps.td
```diff
@@ -0,0 +1,261 @@
...
+    `(` $input `:` type($input) `,` $mean `:` type($mean) `)` attr-dict
+    `:` type($stdev)
+  }];
+}
+
+//===----------------------------------------------------------------------===//
+// DoNotOptOp
+//===----------------------------------------------------------------------===//
+
+def Perf_DoNotOptOp : Perf_Op<"do_not_opt", [ConditionallySpeculatable]> {
```
rengolin, line 229: Nit: I'd call this `perf.sink`?

test/TPP/perf/perf-ops.mlir
```diff
@@ -0,0 +1,107 @@
...
+
+// CHECK-LABEL: @perf_matmul_bench
+func.func @perf_matmul_bench(%A: tensor<4x8xf32>,
+          %B: tensor<8x4xf32>, %C: tensor<4x4xf32>, %n: i64) {
+  // CHECK: perf.bench
+  %deltas = perf.bench (%n) {
+    // CHECK: linalg.matmul
+    %D = linalg.matmul ins(%A, %B: tensor<4x8xf32>, tensor<8x4xf32>) outs(%C: tensor<4x4xf32>) -> tensor<4x4xf32>
+    // CHECK: perf.do_not_opt
+    perf.do_not_opt(%D) : tensor<4x4xf32>
```
rengolin, line 11: I thought the `perf.yield` was supposed to be inserted when there's no other yield op?

test/TPP/perf/perf-ops.mlir
```diff
@@ -0,0 +1,107 @@
...
+  // CHECK: perf.stop_timer
+  perf.stop_timer(%t : i64) : f64
+
+  return %c : i32
+}
+
+// -----
+
+// CHECK-LABEL: @perf_timer
+func.func @perf_timer(%a: i32, %b: i32, %n: i64) -> i32 {
```
rengolin, line 50: duplicated test?

test/TPP/perf/perf-ops.mlir
```diff
@@ -0,0 +1,107 @@
...
+// CHECK-LABEL: @perf_stdev
+func.func @perf_stdev(%arg0: memref<?xf64>, %mean: f64) -> f64 {
+  // CHECK: perf.stdev
+  %stdev = perf.stdev(%arg0 : memref<?xf64>, %mean : f64) : f64
+  return %stdev : f64
+}
+
+// -----
+
+// CHECK-LABEL: @perf_matmul_loops
```
rengolin, line 81: also please add a full test with `perf.bench` alternative

adam-smnk: That would be the currently present `perf_matmul_bench` test, if I get you correctly.
I'll reorder the test cases so the two are together.

lib/TPP/Dialect/Perf/PerfOps.cpp
adam-smnk: Yup that worked, thanks.
I moved the result check to `perf.bench` directly and, with the other fix, I removed the `perf.yield` verifier.

test/TPP/perf/perf-ops.mlir
adam-smnk: After simple parsing, the yield op stays implicit. I believe it is in line with `scf.for` behavior.
But good point anyway as I have not tested the `BenchOp::build` builder. A custom builder might be needed to cover the yield insertion.

test/TPP/perf/perf-ops.mlir
rengolin: Right, I mean full bench + mean/stdev, so that we have "documented" both usages with the most common one represented.

What would be helpful (at this stage) is to write the above test as returning two floats, mean/stdev, then calling `bench` + `yield` + `mean` + `stdev` and returning those. Then the one below, with a comment like "this is how it's supposed to lower to", and a literal loop + start/stop, etc.

So that when you add the lowering, we can CHECK against the structure below.

It doesn't change the implementation or the design, but it documents what's expected and what are the next steps, so make it easier for you to follow up on the work and people reviewing the patches to know what to expect and spend less time on the next review.

rengolin (approved):
This is looking great, thanks!

test/TPP/perf/perf-ops.mlir
adam-smnk: Yield insertion was missing, solved with a custom `BenchOp::build`.

include/TPP/Dialect/Perf/PerfOps.td
adam-smnk: Right now it models unique timer instances in line with the current benchmark timer design.
As discussed, I moved to a custom `!perf.timer` type. For lowering to function calls, it will probably still default to `i64` that will represent a unique timer ID.

PR #220, opened 2022-12-12
Improve mapping to tpp

chelini:
Make sure all the maps are projected permutation and the rank is less than or equal to 2.

### DIFF
--- a/lib/TPP/Dialect/Tpp/TppUtils.cpp
+++ b/lib/TPP/Dialect/Tpp/TppUtils.cpp
@@ -308,15 +308,24 @@ bool canMapToTppAdd(linalg::GenericOp linalgOp) {
   return hasOnlyScalarElementwiseOp<arith::AddFOp>(linalgOp.getRegion());
 }
 
-// TODO: check access pattern using affine map.
+static bool allIndexingsAreProjectedPermutation(linalg::GenericOp genericOp) {
+  return llvm::all_of(genericOp.getIndexingMapsArray(), [](AffineMap m) {
+    return m.isProjectedPermutation(/*allowZeroInResults=*/true);
+  });
+}
+
+// Return true if the linalg.generic can be mapped to a tpp.add.
 bool isTppAdd(linalg::GenericOp linalgOp) {
   if (linalgOp.getNumLoops() != linalgOp.getNumParallelLoops())
     return false;
-  if ((linalgOp.getNumDpsInputs() != 1) || (linalgOp.getNumDpsInits() != 1))
+  if (linalgOp.getNumDpsInits() != 1)
     return false;
   if (linalgOp.hasTensorSemantics())
     return false;
-  return hasOnlyScalarElementwiseOp<arith::AddFOp>(linalgOp.getRegion());
+  if (!hasStaticShape(linalgOp))
+    return false;
+  return allIndexingsAreProjectedPermutation(linalgOp) &&
+         hasOnlyScalarElementwiseOp<arith::AddFOp>(linalgOp.getRegion());
 }
 
 bool canMapToTppRelu(linalg::GenericOp linalgOp) {
@@ -327,7 +336,7 @@ bool canMapToTppRelu(linalg::GenericOp linalgOp) {
   return hasOnlyScalarElementwiseOp<arith::MaxFOp>(linalgOp.getRegion());
 }
 
-// TODO: check access pattern using affine map.
+// Return true if the linalg.generic can be mapped to a tpp.relu.
 bool isTppRelu(linalg::GenericOp linalgOp) {
   if (linalgOp.getNumLoops() != linalgOp.getNumParallelLoops())
     return false;
@@ -335,7 +344,10 @@ bool isTppRelu(linalg::GenericOp linalgOp) {
     return false;
   if (linalgOp.hasTensorSemantics())
     return false;
-  return hasOnlyScalarElementwiseOp<arith::MaxFOp>(linalgOp.getRegion());
+  if (!hasStaticShape(linalgOp))
+    return false;
+  return allIndexingsAreProjectedPermutation(linalgOp) &&
+         hasOnlyScalarElementwiseOp<arith::MaxFOp>(linalgOp.getRegion());
 }
 
 bool canMapToTppIdentity(linalg::GenericOp linalgOp) {
@@ -344,14 +356,15 @@ bool canMapToTppIdentity(linalg::GenericOp linalgOp) {
   return (hasStaticShape(linalgOp) && hasCopySemantics(linalgOp));
 }
 
-// TODO: check access pattern using affine map. We need to take into account
-// broadcasting too.
+// Return true if the linalg.generic can be mapped to a tpp.identity.
 bool isTppIdentity(linalg::GenericOp linalgOp) {
   if (linalgOp.getNumLoops() != linalgOp.getNumParallelLoops())
     return false;
   if ((linalgOp.getNumDpsInputs() != 1) || (linalgOp.getNumDpsInits() != 1))
     return false;
   if (linalgOp.hasTensorSemantics())
+    return false;
+  if (!hasStaticShape(linalgOp))
     return false;
   return hasCopySemantics(linalgOp);
 }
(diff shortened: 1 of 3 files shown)


lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -327,15 +361,25 @@ bool canMapToTppRelu(linalg::GenericOp linalgOp) {
   return hasOnlyScalarElementwiseOp<arith::MaxFOp>(linalgOp.getRegion());
 }
 
-// TODO: check access pattern using affine map.
+// Return true if the linalg.generic can be mapped to a tpp.relu
+// 1. All loops are parallel
+// 2. Single output
+// 3. Buffer semantics
+// 4. Has only a scalar operation (arith.addf)
```
adam-smnk, line 368: `arith.maxf`?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -344,15 +388,23 @@ bool canMapToTppIdentity(linalg::GenericOp linalgOp) {
...
-// TODO: check access pattern using affine map. We need to take into account
-// broadcasting too.
+// Return true if the linalg.generic can be mapped to a tpp.identity
+// 1. All loops are parallel
+// 2. Single output
+// 3. Buffer semantics
+// 4. Has only a scalar operation (arith.addf)
+// 5. All the maps are identity
+// 6. All memref have rank <= 2
+// 7. Static shape
```
adam-smnk, line 398: Needs to be adjusted

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -308,15 +308,49 @@ bool canMapToTppAdd(linalg::GenericOp linalgOp) {
   return hasOnlyScalarElementwiseOp<arith::AddFOp>(linalgOp.getRegion());
 }
 
-// TODO: check access pattern using affine map.
+// Return true if the linalgOp has only identity maps and its operands are
+// shapedType with rank <= 2.
+static bool checkOperandsAndMaps(linalg::GenericOp linalgOp) {
```
rengolin, line 313: Bad name for this function, doesn't reflect what it does. Comment is not visible at the call site.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -308,15 +308,49 @@ bool canMapToTppAdd(linalg::GenericOp linalgOp) {
...
+    ShapedType operandType = operand->get().getType().cast<ShapedType>();
+    // Reject 0-rank memref or memref with rank > 2.
+    if ((operandType.getRank() == 0))
+      return false;
+    ArrayRef<int64_t> shape = operandType.getShape();
+    size_t unitDims = 0;
+    for (int64_t dim : shape)
+      if (dim == 1)
+        unitDims++;
+    if (operandType.getRank() - unitDims > 2)
```
rengolin, line 329: Brackets around comparison to avoid cognitive load of having to remember operator precedence.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -308,15 +308,49 @@ bool canMapToTppAdd(linalg::GenericOp linalgOp) {
...
+      return false;
+  }
+  return true;
+}
+
+// Return true if the linalg.generic can be mapped to a tpp.add
+// 1. All loops are parallel
```
rengolin, line 336: It's best if these comments are inline in the function, before their implementation. Same below. Helps make sure that the comment is always in sync with the code and accurate.

test/TPP/linalg-to-tpp.mlir
```diff
@@ -99,3 +99,29 @@ func.func @main() -> memref<8x32x32x32xf32> {
...
+      linalg.yield %0 : f32
+  }
+  return 
+}
+
+// -----
+
+#map = affine_map<() -> ()>
+func.func @add_mapping(%arg0: memref<f32>, %arg1: memref<f32>) {
+  // CHECK-NOT: tpp.add
```
rengolin, line 120: Make sure to add a comment as to why this should fail. someone looking at the test won't necessarily see the commit. Also, add tests for > 2 and all corner cases for the dim=1 escape hatch.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -308,15 +308,45 @@ bool canMapToTppAdd(linalg::GenericOp linalgOp) {
   return hasOnlyScalarElementwiseOp<arith::AddFOp>(linalgOp.getRegion());
 }
 
-// TODO: check access pattern using affine map.
```
rengolin, line 311: Is this TODO not relevant any more? Doesn't seem to be what you're implementing in this patch. Same for the one below that you also removed.

test/TPP/linalg-to-tpp.mlir
```diff
@@ -99,3 +99,60 @@ func.func @main() -> memref<8x32x32x32xf32> {
...
+
+// -----
+
+#map = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+func.func @add_mapping(%arg0: memref<10x10x10xf32>, %arg1: memref<10x10x10xf32>) {
+  // CHECK-DAG: %[[C0:.+]] = arith.constant 0 : index
+  // CHECK-DAG: %[[C10:.+]] = arith.constant 10 : index
+  // CHECK-DAG: %[[C1:.+]] = arith.constant 1 : index
+  // CHECK: scf.parallel ({{.*}}) to (%[[C10]]) step (%[[C1]])
+  // CHECK: tpp.add
```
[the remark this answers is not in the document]
chelini: We can peel iterations and map to tpp (reshape2D does that).

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: I wanted to check the affine maps to be projected permutations, see: `allIndexingsAreProjectedPermutation`

test/TPP/linalg-to-tpp.mlir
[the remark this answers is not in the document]
rengolin: Right, so this is not testing the exit condition in the code. Can you add another test that does that? (and explain in a comment why this test doesn't meet that criteria?)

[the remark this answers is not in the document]
chelini: the exit condition is already tested here:

```
// Scalar operands we don't expect any mapping to tpp.
#map = affine_map<() -> ()>
func.func @add_mapping(%arg0: memref<f32>, %arg1: memref<f32>) {
  // CHECK-NOT: tpp.add
  linalg.generic {indexing_maps = [#map, #map], iterator_types = []} ins(%arg0: memref<f32>) outs(%arg1: memref<f32>) {
    ^bb0(%in: f32, %out: f32):
      %0 = arith.addf %in, %out : f32
      linalg.yield %0 : f32
  }
  return
}```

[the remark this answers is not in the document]
chelini: Do you want a test that checks the affine maps?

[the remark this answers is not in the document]
rengolin: Only one condition is checked there, the one where the rank is zero. The other condition, where the rank (minus the 1 dms) is greater than two, isn't.

I'm expecting to see an error when you try to apply the transform on a tensor that is `10x10x10`. If you do other transforms on the existing test, then you're not testing that condition. 

To e clear, we need to have a test, with `CHECK-NOT: tpp.add` because the tensor rank doesn't match the conditions you are creating.

It would also be nice to have tests beyond `1x10x10`, like `10x1x10` and `10x10x1`, as well as `1x1` and `1x1x1`, to make sure there are no corner cases that you missed.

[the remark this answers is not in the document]
chelini: This function is part of a conversion pass and we can map tensors like 10x10x10 by pulling out the outermost dimension. Looking at the conversion again there is not need to check rank in `isTppAdd` but rank is ensured to be 2d by `rankReducingSubviewDroppingUnitDims` in `ConvertLinalgToTpp.cpp`. Here we can check only the affine maps. I added more tests as suggested.

chelini:
ping here @rengolin and @adam-smnk

rengolin (approved):
Awesome, thanks!

We just need a few additional shapes to some tpp.add CHECK lines (see inline comment), but otherwise looks good to me.

Feel free to merge once you have addressed those.

adam-smnk (approved):

PR #221, opened 2022-12-13
Lower BRGEMM from linalg to VNNI to TPP

KavithaTipturMadhu:


lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -85,6 +85,17 @@ static Value handleLayout_VNNI(OpBuilder &builder, Location loc, Value input,
                             /*outerDimsPerm=*/{});
 }
 
+static Value handleBRGemmLayout_VNNI(OpBuilder &builder, Location loc,
+                                     Value input, Value output,
+                                     ArrayRef<OpFoldResult> tiles) {
+  assert(tiles.size() == 1 && "expect 1 block for VNNI");
+  SmallVector<int64_t> innerDimPos = {1};
+  if (!output)
```
rengolin, line 93: IIRC, this distinction is about tensor vs memref semantics right? If the requirement below (need to be on tensors) is right, than we really don't need this distinction.

rengolin (commented):
This looks reasonable to me. Please add a negative test on the known problems, for example multiple tile sizes, dynamic shapes, buffer semantics. (I know other ops have the same issues and don't have those tests, but we need to fix this).

test/BF16/brgemm-vnni.mlir
```diff
@@ -0,0 +1,16 @@
+// RUN: tpp-opt -transform-dialect-interpreter %s | FileCheck %s
```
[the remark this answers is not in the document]
KavithaTipturMadhu: I have added a new test which goes from linalg all the way to tpp. I want to keep this test as is because this lowers only upto vnni and we want that covered separately.

lib/TPP/ToBlockLayoutAndBack.cpp
KavithaTipturMadhu: I have removed the unpack call from here, it doesn't make sense indeed.

KavithaTipturMadhu:
> This looks reasonable to me. Please add a negative test on the known problems, for example multiple tile sizes, dynamic shapes, buffer semantics. (I know other ops have the same issues and don't have those tests, but we need to fix this).

I have added tests for unbufferized case and dynamic shape. When you say multiple tile sizes, do you mean mismatching tile sizes? That's covered in tpp tests already.

test/BF16/vnni-to-tpp-fail.mlir
```diff
@@ -0,0 +1,39 @@
...
+  vnni.matmul ins(%arg0 : memref<?xbf16>, %0 : memref<256x1024x2xbf16>) out(%arg2 : memref<?xbf16>)
+  return %arg2 : memref<?xbf16>
+}
+
+// -----
+
+func.func @brgemm_static(%arg0: tensor<4x256x512xbf16>, %arg1: tensor<4x512x1024xbf16>, %arg2: tensor<256x1024xbf16>) -> tensor<256x1024xbf16> {
+  %0 = tensor.empty() : tensor<4x256x1024x2xbf16>
+  %1 = linalgx.pack %arg1 inner_dims_pos = [1] inner_tiles = [2] into %0 : (tensor<4x512x1024xbf16> tensor<4x256x1024x2xbf16>) -> tensor<4x256x1024x2xbf16>
+  // CHECK-NOT: tpp.vnni_brgemm
```
rengolin, line 26: Why isn't this transformed?

rengolin (commented):
Just one comment on the brgemm test, but otherwise looks good to me, thanks for the extra work with the passes!

test/BF16/vnni-to-tpp-fail.mlir
KavithaTipturMadhu: Vnni needs to be bufferized before lowering to tpp. We have an explicit bufferized check when lowering.

rengolin: Ah, right! Could just add a comment here, as i expected it to pass when looking at the test.

rengolin (approved):
Otherwise, looks good, thanks!

PR #223, opened 2022-12-15
Perf to loops

adam-smnk:
Adds a conversion pass from perf dialect to loops.

A perf.bench operation is materialized as an scf.for loop with extra benchmarking code injected into the loop's body.
The optional bench args are passed around as loop iter_args.
Additionally, perf.yield arguments are renamed for more consistent API.

Work towards #100

### DIFF
--- a/lib/TPP/ConvertPerfToLoops.cpp
+++ b/lib/TPP/ConvertPerfToLoops.cpp
@@ -1,0 +1,107 @@
+//===- ConvertPerfToLoops.cpp ------------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Dialect/Perf/PerfOps.h"
+#include "TPP/Passes.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/RegionUtils.h"
+
+using namespace mlir;
+using namespace mlir::perf;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+struct ConvertBenchToLoops : public OpRewritePattern<perf::BenchOp> {
+  using OpRewritePattern<perf::BenchOp>::OpRewritePattern;
+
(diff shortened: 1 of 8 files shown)


lib/TPP/ConvertPerfToLoops.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+namespace {
+
+struct ConvertBenchToLoops : public OpRewritePattern<perf::BenchOp> {
+  using OpRewritePattern<perf::BenchOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(perf::BenchOp benchOp,
+                                PatternRewriter &rewriter) const override {
+    auto loc = benchOp.getLoc();
+    auto benchYield = benchOp.getRegion().front().getTerminator();
```
rengolin, line 31: `assert(op.isa<YieldOp>)`

lib/TPP/ConvertPerfToLoops.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Wrap the benchmark kernel in timer calls.
+    OpBuilder::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPointToStart(loop.getBody());
+    auto timer = rewriter.create<perf::StartTimerOp>(
+        loc, TimerType::get(rewriter.getContext()));
+    rewriter.setInsertionPointToEnd(loop.getBody());
+    auto delta = rewriter.create<perf::StopTimerOp>(loc, rewriter.getF64Type(),
+                                                    timer.getTimer());
+
+    // Move all perf.sink ops after the timer to prevent influencing
```
rengolin, line 58: Isn't `perf.sink` just a nop?

lib/TPP/ConvertPerfToLoops.cpp
```diff
@@ -0,0 +1,107 @@
...
+        op.moveAfter(delta.getOperation());
+    }
+
+    // Store measured time delta on each iteration.
+    rewriter.create<memref::StoreOp>(loc, delta.getDelta(), benchOp.getDeltas(),
+                                     loop.getInductionVar());
+
+    // Replace uses of bench args within the benchmark body with their
+    // equivalent loop-carried variables.
+    for (auto pair : llvm::zip(benchOp.getArgs(), loop.getRegionIterArgs()))
```
rengolin, line 71: I think you can use `for (auto [bop, larg] = llvm::zip...)` to avoid the `get<0>` below.

turns out it's UB if the two `zip` args don't have the same size. Perhaps `assert(size() == size())` here?

rengolin (commented):
Small comments but looks great, thanks!

lib/TPP/ConvertPerfToLoops.cpp
adam-smnk: Ideally, it would be a nop. However, my current approach is to lower all `perf.sink`s to function calls to prevent any further compilation stages (like LLVM backend) from removing results/instructions that we want to preserve.

rengolin (approved):

PR #224, opened 2022-12-15
Perf to func

adam-smnk:
Converts perf operations into function calls to a runtime
environment that provides required functionality.
    
The perf dialect is materialized through an external runtime as
the core operations such as access to timers are commonly platform
dependent and can vary greatly between different targets.

Work towards #100

First lowering pass to a "fat" runtime. It will allow for initial testing with the benchmark infrastructure.
Later, we can reduce the number of dependencies on custom runtime functions.

### DIFF
--- a/lib/TPP/ConvertPerfToFunc.cpp
+++ b/lib/TPP/ConvertPerfToFunc.cpp
@@ -87,10 +87,15 @@ static SmallVector<Value> getNormalizedOperands(OpBuilder &b, Location loc,
 }
 
 // Create a perf function prototype.
-// Assumes that the current insertion point is in the correct place.
 static func::FuncOp createPerfFuncPrototype(Location loc, std::string funcName,
                                             Operation *op,
                                             PatternRewriter &rewriter) {
+  // Insert before module terminator.
+  ModuleOp module = op->getParentOfType<ModuleOp>();
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(module.getBody(),
+                             std::prev(module.getBody()->end()));
+
   FlatSymbolRefAttr fnName = SymbolRefAttr::get(op->getContext(), funcName);
   auto libFnType = rewriter.getFunctionType(
       extractNormalizedTypes(rewriter, op->getOperands()),
@@ -107,11 +112,11 @@ static func::FuncOp createPerfFuncPrototype(Location loc, std::string funcName,
 static LogicalResult buildPerfMeanFunc(Location loc, std::string funcName,
                                        Operation *op,
                                        PatternRewriter &rewriter) {
-  OpBuilder::InsertionGuard guard(rewriter);
   auto funcOp = createPerfFuncPrototype(loc, funcName, op, rewriter);
 
   // Create function body
   Block *block = funcOp.addEntryBlock();
+  OpBuilder::InsertionGuard guard(rewriter);
   rewriter.setInsertionPointToEnd(block);
 
   // Check assumptions on function arguments.
@@ -168,11 +173,11 @@ static LogicalResult buildPerfMeanFunc(Location loc, std::string funcName,
 static LogicalResult buildPerfStdevFunc(Location loc, std::string funcName,
                                         Operation *op,
                                         PatternRewriter &rewriter) {
-  OpBuilder::InsertionGuard guard(rewriter);
   auto funcOp = createPerfFuncPrototype(loc, funcName, op, rewriter);
 
   // Create function body
   Block *block = funcOp.addEntryBlock();
+  OpBuilder::InsertionGuard guard(rewriter);
   rewriter.setInsertionPointToEnd(block);
 
   // Check assumptions on function arguments.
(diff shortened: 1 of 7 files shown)


rengolin:
Going to look at this after #223 is merged to make it easier to see the difference.

adam-smnk:
Update: `perf.mean` and `perf.stdev` implementations are now generated directly as MLIR IR.

I also investigated some potential options for directly generating timestamp calls like `std::chrono` directly instead of going through a timer runtime. However, there are some issues with this approach:
- std::chrono is heavily templated and relies on custom types which is not compatible with default `extern "C"` name mangling
- MLIR code generation heavily favors interoperability with and generation of C instead of C++ which further increases integration and adoption cost of this solution
- this solution could be potentially forced by generating calls to pre-mangled functions, however, there are portability issues (for example, looking at you MSVC)

Overall, I would stick with runtime support which just requires user provided: `perf_start_timer`, `perf_stop_timer`, and (optionally) `perf_sink_*` implementations. We could add some default implementation for the `perf_*_timer` functions to, for example, MLIR `RunnerUtils.cpp`.

lib/TPP/ConvertPerfToFunc.cpp
```diff
@@ -0,0 +1,361 @@
...
+  if (!module.lookupSymbol(fnName.getAttr())) {
+    OpBuilder::InsertionGuard guard(rewriter);
+    // Insert before module terminator.
+    rewriter.setInsertionPoint(module.getBody(),
+                               std::prev(module.getBody()->end()));
+
+    auto libFnType = rewriter.getFunctionType(
+        extractNormalizedTypes(rewriter, op->getOperands()),
+        extractNormalizedTypes(rewriter, op->getResults()));
+    func::FuncOp funcOp =
```
rengolin, line 244: This disconnects the function from the builder and creates coupling between the creation outside of `buildPerfMeanFunc` and adding the entry block inside.

This kind of code generation is best if the functions are declared/defined outline and the function symbol is kept on a map to be used by code generation.

`Sink` is the only one that needs to be "on demand" because of the argument type, but that's an empty function, so it's the trivial case.

lib/TPP/ConvertPerfToFunc.cpp
```diff
@@ -0,0 +1,361 @@
...
+                              PatternRewriter &rewriter) {
+  OpBuilder::InsertionGuard guard(rewriter);
+
+  // Create function body
+  Block *block = func.addEntryBlock();
+  rewriter.setInsertionPointToEnd(block);
+
+  // Check assumptions on function arguments.
+  auto argTypes = func.getFunctionType().getInputs();
+  assert((argTypes.size() == 1) && "expect only 1 function argument");
```
rengolin, line 118: These asserts all disappear when you build the function in one go offline and use it later on the rewriter for the perf op.

lib/TPP/ConvertPerfToFunc.cpp
```diff
@@ -0,0 +1,361 @@
...
+  auto result = rewriter.create<arith::ConstantFloatOp>(
+      loc, APFloat::getZero(floatType.getFloatSemantics()), floatType);
+  auto mean = block->getArguments()[1];
+
+  auto loopNest = scf::buildLoopNest(
+      rewriter, loc, /*lbs=*/ValueRange{zero}, /*ubs=*/ValueRange{len},
+      /*steps=*/ValueRange{one}, ValueRange{result},
+      [&](OpBuilder &b, Location loc, ValueRange localIvs,
+          ValueRange iterArgs) -> scf::ValueVector {
+        auto meas = rewriter.create<memref::LoadOp>(loc, deltas, localIvs);
```
rengolin, line 201: `meas`?

lib/TPP/ConvertPerfToFunc.cpp
```diff
@@ -0,0 +1,361 @@
...
+      rewriter, loc, /*lbs=*/ValueRange{zero}, /*ubs=*/ValueRange{len},
+      /*steps=*/ValueRange{one}, ValueRange{result},
+      [&](OpBuilder &b, Location loc, ValueRange localIvs,
+          ValueRange iterArgs) -> scf::ValueVector {
+        auto delta = rewriter.create<memref::LoadOp>(loc, deltas, localIvs);
+        auto sum = rewriter.create<arith::AddFOp>(loc, delta, iterArgs[0]);
+
+        return scf::ValueVector({sum});
+      });
+  assert((loopNest.results.size() == 1) && "expect only 1 loop result");
```
rengolin, line 148: Why would it be more than one?

adam-smnk: Testing leftover, shouldn't be needed.

lib/TPP/ConvertPerfToFunc.cpp
adam-smnk: Good point, moved to proper op errors.

lib/TPP/ConvertPerfToFunc.cpp
adam-smnk: The builders should be more self contained now.

I don't think cross-module `func.call`s are really supported so I gave up on the LUT approach as you could have perf ops present in different modules.

lib/TPP/ConvertPerfToFunc.cpp
```diff
@@ -0,0 +1,386 @@
...
+          res.push_back(cast);
+        })
+        .Default([&](Type t) { res.push_back(op); });
+  }
+
+  return res;
+}
+
+// Create a perf function prototype.
+// Assumes that the current insertion point is in the correct place.
```
rengolin, line 90: This assumption is dangerous because functions can only be inserted into modules.

You also have two levels of `InsertionGuard` (main loop and `buildPerf...`) which are redundant.

Technically, you only need one, and change the insertion point to the module at the `buildPerf...` level, after creating the `InsertionGuard`, and it should all work, without assumptions.

rengolin (approved):
Small comment inline. With that change, looks good to me.

lib/TPP/ConvertPerfToFunc.cpp
adam-smnk: True, these functions should be more independent.
Double guards introduce a bit of redundancy but it should be safer to have them whenever the insertion point is changed. Then the functional call inserter can just blindly call the builders.

Issue #96, opened 2022-10-31
Fusion in `mlp-packed.mlir`


chelini:
Avoid using LICM to fix-up fusion, prefer instead call-back to have more control on what get fused. Callback implementation: https://github.com/chelini/llvm-project/commit/f2a09adbe37de17306b50dea3b1892ce6867f782.

The callback allows to fuse together the BRGEMM and the relu.

chelini:
We don't use LICM anymore.

rengolin:
@adam-smnk @chelini is this still a problem in the default pipeline?

chelini:
There is not more need for fusion callback. Now we can control upstream tile and fuse because they exposed what we need.

PR #225, opened 2022-12-16
Add passes back

chelini:


adam-smnk (approved):
If I'm not mistaken, this OR just restores some of the older logic. So, it should be good to go.

adam-smnk:
@chelini I guess we'll have to revisit issues #95 and #96 now that we're bringing passes, right? Any other immediate problems that come to your mind?

chelini:
> If I'm not mistaken, this OR just restores some of the older logic. So, it should be good to go.

Yes, we used to match rely this way. I follow-up with a fix and more tests.

chelini:
> @chelini I guess we'll have to revisit issues #95 and #96 now that we're bringing back passes, right? Any other immediate problems that come to your mind?

Yes, #95 is the more problematic one. #96 can be closed.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -337,10 +337,12 @@ bool canMapToTppRelu(linalg::GenericOp linalgOp) {
 }
 
 // Return true if the linalg.generic can be mapped to a tpp.relu.
+// TODO: check the output bbarg to not be used.
 bool isTppRelu(linalg::GenericOp linalgOp) {
   if (linalgOp.getNumLoops() != linalgOp.getNumParallelLoops())
     return false;
-  if ((linalgOp.getNumDpsInputs() != 0) || (linalgOp.getNumDpsInits() != 1))
+  if (/*(linalgOp.getNumDpsInputs() != 1) || */ (linalgOp.getNumDpsInits() !=
```
rengolin, line 344: Please avoid commented out code. Remove the code and add a comment if needed.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -217,3 +217,46 @@ mlir::linalgx::mapToBRGEMMOp(RewriterBase &rewriter,
...
   return outermostLoop ? outermostLoop->getResults() : tensorResults;
 }
+
+//===----------------------------------------------------------------------===//
+// Passes
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+struct DoItOnGeneric : public OpRewritePattern<linalg::GenericOp> {
```
rengolin, line 227: Herbie Hancock style!

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -0,0 +1,216 @@
...
+  return static_cast<bool>(strideAsInt);
+}
+
+static int64_t getSizeRange(Range loopRange) {
+  assert(isStaticRange(loopRange));
+  return *getConstantIntValue(loopRange.size) -
+         *getConstantIntValue(loopRange.offset);
+}
+
+// Check if all the loops have static ranges.
```
rengolin, line 46: Same thing here, we can pull those later if needed.

rengolin (commented):
Can some tests be moved from transform already with this?

chelini:
> Can some tests be moved from transform already with this?

Yes, the entire MLP. Talking yesterday with Alex this is what we want to do. Convolutions can wait.

lib/TPP/TileConsumerAndFuseProducers.cpp
chelini: done.

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: done.

rengolin:
> Yes, the entire MLP. Talking yesterday with Alex this is what we want to do. Convolutions can wait. Also the way we used to map to convs was bad by passing attributes, so maybe it is just better to write them again when we needed.

what about all the other tests that don't involve convs? The ones we converted to transforms a while ago, I imagine they can all be moved to passes and removing their transform blocks, like we discussed, right?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -776,9 +776,169 @@ struct PropagateThroughElementWiseOp
...
+    if (failed(genericOp))
+      return failure();
+    return success();
+  }
+
+private:
+  SmallVector<int64_t> blockingFactors;
+};
+
+struct PackConv2DNchwFchw : public PackConv2DNchwFchwBase<PackConv2DNchwFchw> {
```
rengolin, line 867: What about the tests for these transforms? Are they all using transforms now? If so, why are we bringing this code back if it's not going to be tested?

test/TPP/passes/mlp-fusion.mlir
```diff
@@ -0,0 +1,31 @@
...
+      linalg.yield %arg9 : f32
+  } -> tensor<128x512xf32>
+  // CHECK-NEXT: tpp.matmul
+  %2 = linalg.generic {indexing_maps = [#map2, #map3, #map4], iterator_types = ["parallel", "parallel", "reduction"]} ins(%arg0, %arg1 : tensor<128x256xf32>, tensor<256x512xf32>) outs(%1 : tensor<128x512xf32>) attrs =  {iterator_ranges = [128, 512, 256]} {
+    ^bb0(%arg9: f32, %arg10: f32, %arg11: f32):
+      %16 = arith.mulf %arg9, %arg10 : f32
+      %17 = arith.addf %arg11, %16 : f32
+      linalg.yield %17 : f32
+  } -> tensor<128x512xf32>
+  // CHECK-NEXT: tpp.relu
```
rengolin, line 23: IIUC, this is just fusing `id + matmul + relu` inside the loop, not fusing TPP ops, right?

We should have a second RUN line that goes all the way down fusing TPP ops too, later on.

test/TPP/passes/mlp-fusion.mlir
```diff
@@ -0,0 +1,31 @@
...
+#map4 = affine_map<(d0, d1, d2) -> (d0, d1)>
+  
+func.func @main(%arg0: tensor<128x256xf32>, %arg1: tensor<256x512xf32>,
+  %arg2: tensor<512xf32>,  %output: tensor<128x512xf32>) -> tensor<128x512xf32> {
+  // CHECK: tpp.identity
+  %1 = linalg.generic {indexing_maps = [#map0, #map1], iterator_types = ["parallel", "parallel"]} ins(%arg2 : tensor<512xf32>) outs(%output : tensor<128x512xf32>) {
+    ^bb0(%arg9: f32, %arg10: f32):
+      linalg.yield %arg9 : f32
+  } -> tensor<128x512xf32>
+  // CHECK-NEXT: tpp.matmul
```
rengolin, line 16: CHECK shapes, please. These shapes (non-packed) are bound to be different than the one in `mlp-packed`, and we need to make sure we get them right.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -337,10 +337,13 @@ bool canMapToTppRelu(linalg::GenericOp linalgOp) {
...
 
 // Return true if the linalg.generic can be mapped to a tpp.relu.
+// TODO: check input arguments:
+// 1. Single input, single output as long as the output bbarg is not used
+// 2. Single output.
 bool isTppRelu(linalg::GenericOp linalgOp) {
   if (linalgOp.getNumLoops() != linalgOp.getNumParallelLoops())
     return false;
-  if ((linalgOp.getNumDpsInputs() != 0) || (linalgOp.getNumDpsInits() != 1))
+  if (linalgOp.getNumDpsInits() != 1)
```
rengolin, line 346: I'm surprised this wasn't being tested by any test we have. I was expecting such a change to break something...

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -776,9 +776,169 @@ struct PropagateThroughElementWiseOp
...
+  ArrayRef<int64_t> blockingFactors;
+};
+
+// From linalg.generic to linalg.matmul.
+struct DeGeneralizeMatmul : public OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (!linalgOp.hasTensorSemantics())
```
rengolin, line 810: We need to add an issue to move this pack to tensors, so we can bring the upstream ops.

After the pass work, this is one of the most important things we need to finish.

test/TPP/passes/mlp-packed.mlir
```diff
@@ -0,0 +1,65 @@
+// RUN: tpp-opt %s -pack-matmul="block-factors=32,32,32" -map-linalg-to-tpp -transform-dialect-interpreter -canonicalize -tile-consumer-and-fuse-producers="tile-sizes=1,0,0,0" -canonicalize -tile-consumer-and-fuse-producers="tile-sizes=1,0,0" -canonicalize -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize -map-linalg-to-tpp -convert-linalg-to-tpp="use-parallel-loops=false" -map-to-brgemm -convert-linalg-to-tpp | FileCheck %s
```
rengolin, line 1: Duplicated use of `tile-consumer-and-fuse-producers` in this RUN line

include/TPP/Passes.td
```diff
@@ -115,4 +115,78 @@ def TransformDropSchedulePass : Pass<"transform-drop-schedule", "ModuleOp"> {
...
+    the Matmul has a relu operation as its consumer block also the relu operation.
+  }];
+  let options = [
+    ListOption<"blockingFactors", "block-factors", "int64_t", 
+               "Blocking factor for relayout">
+  ];
+  let constructor = "mlir::tpp::createPackMatmulPass()";
+}
+
+def PackConv2DNchwFchw : Pass<"pack-conv2DNchwFchw", "func::FuncOp"> {
```
rengolin, line 131: Untested code

include/TPP/Passes.td
```diff
@@ -115,4 +115,78 @@ def TransformDropSchedulePass : Pass<"transform-drop-schedule", "ModuleOp"> {
...
+    Pack the output's channel K with a block factor BK.
+  }];
+  let options = [
+    ListOption<"blockingFactors", "block-factors", "int64_t",
+               "Blocking factor for relayout">
+  ];
+  let constructor = "mlir::tpp::createPackConv2DNchwFchwPass()";
+}
+
+def PackConv2DNhwcHwcf : Pass<"pack-conv2DNhwcHwcf", "func::FuncOp"> {
```
rengolin, line 147: Untested code

chelini: the passes and the transform uses the same code. I would like to avoid duplicating tests.

test/TPP/passes/mlp-packed.mlir
chelini: It is not duplicated. We have two rounds of fusion

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: Sorry did not get here. Line 808 is just a de-linearization pattern.

test/TPP/passes/mlp-fusion.mlir
chelini: yes, we don't have fusion on tpps call. We just fuse at linalg level and maps to tpp.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: Do you want to use transform for packing?

chelini:
> > Yes, the entire MLP. Talking yesterday with Alex this is what we want to do. Convolutions can wait. Also the way we used to map to convs was bad by passing attributes, so maybe it is just better to write them again when we needed.
> 
> what about all the other tests that don't involve convs? The ones we converted to transforms a while ago, I imagine they can all be moved to passes and removing their transform blocks, like we discussed, right?

Can you please be more specific on the tests you would like to see moved to pass? Yesterday the objective was to have MLP, and I executed in this direction.

include/TPP/Passes.td
rengolin: The passes are not being tested, even if the code that the pass uses is being tested. Since we're moving to passes, and will eventually delete the transforms from `main`, we need those tests to be in passes.

rengolin:
> Can you please be more specific on the tests you would like to see moved to pass? Yesterday the objective was to have MLP, and I executed in this direction.

One objective is to have MLP, but the objective of this PR is to move back to passes, independent of MLP. The more commits we have for this, the more other people will have to rebase their current work, which is exactly what we're trying to reduce with this move.

test/TPP/passes/mlp-packed.mlir
rengolin: I see. Weird, but fine, we can worry about that later if needed.

rengolin (changes requested):
To be clear, we need all tests as pass, and the old passes back, being tested, as passes.

We don't need the transforms and their tests to be removed in this PR, so a copy & replace will be fine.

But we shouldn't work on the transform pipeline in `main` to avoid the churn that it caused to others.

chelini:
> To be clear, we need all tests as pass, and the old passes back, being tested, as passes.
> 
> We don't need the transforms and their tests to be removed in this PR, so a copy & replace will be fine.
> 
> But we shouldn't work on the transform pipeline in `main` to avoid the churn that it caused to others.

I will strongly oppose to remove transform form main. As we don't have a good reason on why passes are a better way forward.

rengolin:
> > To be clear, we need all tests as pass, and the old passes back, being tested, as passes.
> > We don't need the transforms and their tests to be removed in this PR, so a copy & replace will be fine.
> > But we shouldn't work on the transform pipeline in `main` to avoid the churn that it caused to others.
> 
> I will strongly oppose to remove transform form main. As we don't have a good reason on why passes are a better way forward.

#226 goes through some of those reasons. For the things that we care now (BRGEMM for matmuls/1x1convs) we don't need anything special, and we had those things as passes before, so it has been proven to work in this repository.

We may need transforms later for more complex passes, but they'll need to be auto-generated to be used by the end-to-end pipeline.

test/TPP/passes/map-linalg-to-tpp.mlir
```diff
@@ -0,0 +1,135 @@
...
+
+// -----
+
+#map0 = affine_map<(d0, d1) -> (d0, d1)>
+
+
+// CHECK-LABEL: func.func @identity
+func.func @identity(%arg1: memref<1x512xf32>) {
+  %0 = memref.alloc() : memref<1x512xf32>
+  // CHECK: tpp.identity
```
rengolin, line 129:
```suggestion
  // CHECK: library_call = "tpp.identity"
```

test/TPP/passes/map-to-brgemm.mlir
```diff
@@ -0,0 +1,41 @@
+// RUN: tpp-opt %s -map-to-brgemm | FileCheck %s
+
+#map3 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map4 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
+#map5 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
+
+transform.sequence failures(propagate) {
```
rengolin, line 7: Left over transform

rengolin (commented):
Great! This should be most of the tests that needed replicating, thanks!

I'm looking at tests in `Integration/transforms` and it seems we already have something similar elsewhere?

Also, there's a test in `Integration/conv-mapping-to-gemm.mlir` which has a transform block. Probably should be moved to the `transform` directory.

chelini:
> 

Yes, they will be auto-generated by a pass. What I want to say is that I don't want to remove what we have.

chelini:
> 

Move the test in transform. Maybe we should have something like /trasform/Integration to avoid mixing tests but we can follow-up.

lib/TPP/DecomposeConvsToMatmulOrBrgemm.cpp
```diff
@@ -0,0 +1,630 @@
...
+
+// Interchange iterator in a linalg.generic marked as
+// 'tpp.BlockedConv2DNchwFchwOp' to expose a GEMM operation.
+struct InterchangeIteratorsConv2DNchwFchw
+    : OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    if (!tpp::utils::isMarkedWithTpp(genericOp, "tpp.BlockedConv2DNchwFchwOp"))
```
chelini, line 251: FYI: @adam-smnk @rengolin and @KavithaTipturMadhu  this is what I dislike the most. Each pattern picks up when the other left.

chelini: You run a canonicalization in between and you are dead.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: @KavithaTipturMadhu I think this was something you added, can you check if the tests you had are still ok?

lib/TPP/DecomposeConvsToMatmulOrBrgemm.cpp
rengolin: Worst case we can run the same pattern matcher instead of annotating, but for now we can just make sure we don't canonicalize in between.

lib/TPP/Dialect/Tpp/TppUtils.cpp
KavithaTipturMadhu: I didn't add this. I'm not sure i understand what's going on in this change, are dpsinputs never 0?

chelini: I will fix it on Monday. Just made sure the tests are working. We used to assume to have 1 inputs for relu.

rengolin: Ah, right, I thought this was some VNNI alternative shape, but relu doesn't care, my mistake.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -352,6 +352,9 @@ bool isUnaryOp(linalg::GenericOp linalgOp) {
 }
 
 // Return true if the linalg.generic can be mapped to a tpp.relu.
+// TODO: check input arguments:
+// 1. Single input, single output as long as the output bbarg is not used
+// 2. Single output.
```
rengolin, line 357: Is this comment still relevant? I have pushed a similar PR here, no?

rengolin (approved):
Small question on a comment, but LGTM. Thanks!

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: Will remove.

adam-smnk:
> ... 2) build a different fusion driver in the pass, still unclear how to do this.

It seems that we need more fine-grained control to prevent fusion pass from being applied globally as we need to fuse different parts of IR at different times. So, for the approach 2), the idea is to apply the current transform approach i.e., match and fuse locally, at the pass level. 
To achieve that, we could either try generating a small transform kernel and try applying it immediately, or call underlying transform logic from a pass. Either sounds feasible but we'll see how it goes in practice.

rengolin:
> It seems that we need more fine-grained control to prevent fusion pass from being applied globally as we need to fuse different parts of IR at different times. So, for the approach 2), the idea is to apply the current transform approach i.e., match and fuse locally, at the pass level. To achieve that, we could either try generating a small transform kernel and try applying it immediately, or call underlying transform logic from a pass. Either sounds feasible but we'll see how it goes in practice.

Let's avoid creating small transform kernels for now. This will increase impedance mismatch with the rest of the passes.

If the greedy algorithm is finding broken solutions, there's something wrong with the algorithm or our usage. If it's just not finding optimal solutions, let's merge and keep the issue open. We can think about cleanups or other solutions later when we tackle those issues, but we really need to merge this soon.

chelini:
> > It seems that we need more fine-grained control to prevent fusion pass from being applied globally as we need to fuse different parts of IR at different times. So, for the approach 2), the idea is to apply the current transform approach i.e., match and fuse locally, at the pass level. To achieve that, we could either try generating a small transform kernel and try applying it immediately, or call underlying transform logic from a pass. Either sounds feasible but we'll see how it goes in practice.
> 
> Let's avoid creating small transform kernels for now. This will increase impedance mismatch with the rest of the passes.
> 
> If the greedy algorithm is finding broken solutions, there's something wrong with the algorithm or our usage. If it's just not finding optimal solutions, let's merge and keep the issue open. We can think about cleanups or other solutions later when we tackle those issues, but we really need to merge this soon.

Ok. Let me address the comments and I will merge.

PR #236, opened 2022-12-21
Perf dialect runtime

adam-smnk:
Adds runtime bindings required to execute perf dialect operations.
The existing manual MLIR benchmarking loop is now moved to rely on the perf dialect ops.
Also, necessary lowering passes are added to the `tpp-run`.

Work toward #100

### DIFF
--- a/include/Utils/Perf.h
+++ b/include/Utils/Perf.h
@@ -43,7 +43,7 @@ class PerfResults {
   }
 
   /// Zero a time_point
-  void zero(std::chrono::high_resolution_clock::time_point& point) {
+  void zero(std::chrono::high_resolution_clock::time_point &point) {
     point = std::chrono::high_resolution_clock::time_point();
   }
 
@@ -56,7 +56,7 @@ public:
   }
 
   /// Stops the timer, accumulates, clears state
-  void stopTimer() {
+  double stopTimer() {
     assert(!locked && "Stop called after stats produced");
     assert(!isZero(start) && "Stop called before start");
     assert(isZero(stop) && "Stop called twice");
@@ -67,6 +67,8 @@ public:
     timings.push_back(val);
     zero(start);
     zero(stop);
+
+    return val;
   }
 
   /// Get mean of timings. Locks the timer, only calculate stats once.
--- a/include/Utils/Perf.h
+++ b/include/Utils/Perf.h
@@ -56,7 +56,7 @@ public:
   }
 
   /// Stops the timer, accumulates, clears state
-  double stopTimer() {
+  void stopTimer() {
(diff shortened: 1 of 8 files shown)


tpp-rt/PerfRunnerUtils.cpp
```diff
@@ -29,25 +33,73 @@ int64_t _mlir_ciface_timer_alloc() {
...
+// A generic sink function.
+// Its aim is to ensure that the passed data and its producers cannot be
+// optimized away such that the time measured by a benchmark loop correctly
+// represents the full workload.
+static void __attribute__((optnone)) perf_sink(void *data) { (void)data; }
+
+/*
+  Perf dialect runtime bindings for common perf.sink op argument types.
+*/
+void _mlir_ciface_perf_sink_memref_i8(UnrankedMemRefType<int8_t> *val) {
```
rengolin, line 86: I don't think we need runtime support for this. Declaring the functions at the IR level and marking as opt-none / no-inline may be enough to get all compile time types needed.

rengolin (commented):
Looks much cleaner, thanks!

Some comments inline.

tpp-rt/PerfRunnerUtils.cpp
adam-smnk: I wasn't sure how to make it work directly in MLIR before but I finally figured it out. It's much cleaner now.

adam-smnk:
`perf.sink` functionality was verified with a small manual diff: 
[perf_sink_present.patch](https://github.com/plaidml/tpp-mlir/files/10279603/perf_sink_present.patch)
[perf_sink_absent.patch](https://github.com/plaidml/tpp-mlir/files/10279602/perf_sink_absent.patch)

Results:
- present:
```Benchmark: mlp
ref                 :     1.596 +-     0.094 gflops
mlir                :   132.403 +-    20.538 gflops
```
- absent:
```Benchmark: mlp
ref                 :     1.656 +-     0.132 gflops
mlir                : 1415357.311 +- 1256694.567 gflops
```

include/Utils/Perf.h
```diff
@@ -43,7 +43,7 @@ class PerfResults {
   }
 
   /// Zero a time_point
-  void zero(std::chrono::high_resolution_clock::time_point& point) {
+  void zero(std::chrono::high_resolution_clock::time_point &point) {
```
rengolin, line 46: Unnecessary change for this PR

rengolin (approved):
Awesome! Just one nit on the spurious change and LGTM, thanks!

This closes the issue, right?

adam-smnk:
> This closes the issue, right?

Yup, I think that's it.

Issue #201, opened 2022-12-05
Move our use of `pack`/`unpack` to upstream `tensor`


rengolin:
Our `Linalgx` implementation is very different from the upstream and we need to move to it as soon as possible. If that means we add more extensions to serialize the ops locally, so be it. At least we're progressing with the upstream implementation and not having extensions for things that already exist.

Looking at #45, the syntax of our pack/unpack is really off, and I wouldn't be surprised if there are some corner cases in the parser that don't expect the things we do. During the upstream review, there were a lot of mentions to "non-standard" way of doing things and the upstream implementation looks a lot better.

@chelini, what are the missing bits here?

rengolin:
This depends on the work in #202 as well as tiling interface and surrounding code that Han Chung is working from IREE and upstreaming, so we'll have to wait for him to finish.

rengolin:
We cannot continue to move in a direction that is against the upstream current. We need to join the upstream efforts (code review, patches, testing) and have a parallel branch with the changes needed to bring our work towards upstream.

Once we can bridge the gap, we flip to using upstream and continue to work to reduce the gap until upstream is as good as it's going to get.

Examples of what I mean, we need to:
* pull the upstream lowering and tiling work
* use upstream `pack`/`unpack` on tensors only, see what breaks on memrefs
* change our implementation at memref level to match, take notes, feed back into upstream review
* rinse, repeat

rengolin:
Assigning to @chelini as he's working on it (#238)

PR #238, opened 2022-12-21
Partially retire linalgx pack and unpack

chelini:
Partially retire linalgx.pack and linalgx.unpack, and switched to upstreamed versions. Disabled test: `perf-to-loops.mlir`: rebasing to the new LLVM version surface an issue tracked in #277 

All other tests have been updated to use the upstream version. For some transform tests, we check the IR at the tensor level instead of memref as currently we don't have a way in transform to generalize tensor.pack and tensor.unpack, I can follow up if necessary with some small additions. The old operations linalgx.pack and linalgx.unpack are left as some tests depend on them, but the ops are now restricted to memref only and not used in the code-gen flow.


rengolin:
This implements #201. If you plan to completely remove the LinalgX in this PR, then please mark it as a fix for #201.

chelini:
@rengolin, @KavithaTipturMadhu and @adam-smnk switch to new tensor.pack and tensor.unpack op. The PR is ready for review. I disabled three tests: two BF16 related and one on the perf dialect, see summary message.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -105,8 +101,6 @@ struct DefaultTppPasses : public DefaultTppPassesBase<DefaultTppPasses> {
 
     // Lower all Check ops.
     pm.addPass(createConvertCheckToLoopsPass());
-    // Lower all LinalgX ops.
-    pm.addPass(createLinalgXToLoopsPass());
```
rengolin, line 109: Don't we need the generalize pack pass here, too?

lib/TPP/Dialect/Transform/LinalgXTransformOps.cpp
```diff
@@ -7,10 +7,8 @@
 //===----------------------------------------------------------------------===//
 
 #include "TPP/Dialect/Transform/LinalgXTransformOps.h"
```
rengolin, line 9: Probably best to rename this file to PackOps or something.

test/BF16/brgemm-tpp.mlir
```diff
@@ -1,17 +1,7 @@
...
-  %0 = transform.structured.match ops{["linalg.batch_reduce_matmul"]} in %arg1
-    transform.structured.pack %0 { use_vnni=true, blocking_factors = [2] }
-}
+// RUN: tpp-opt -pack-vnni="block-factors=2" -generalize-tensor-pack-unpack -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" --convert-vnni-to-tpp %s | FileCheck %s
 
 func.func @myfunc(%arg0: tensor<32x4x4xbf16>, %arg1: tensor<32x4x4xbf16>, %arg2: tensor<4x4xbf16>) -> tensor<4x4xbf16>{
-// CHECK: %[[PACK:.+]] = memref.alloc() {alignment = 128 : i64} : memref<32x2x4x2xbf16>
-// CHECK: linalgx.pack %arg1 inner_dims_pos = [1] inner_tiles = [2] into %[[PACK]] : (memref<32x4x4xbf16> memref<32x2x4x2xbf16>) 
-// CHECK: tpp.vnni_brgemm ins(%arg0 : memref<32x4x4xbf16>, %[[PACK]] : memref<32x2x4x2xbf16>) out(%arg2 : memref<4x4xbf16>)
+  // CHECK: tpp.vnni_brgemm ins(%{{.*}} : memref<32x4x4xbf16>, %{{.*}} : memref<32x2x4x2xbf16>) out(%{{.*}} : memref<4x4xbf16>)
```
rengolin, line 4: are there no `tensor.pack`s here?

test/BF16/matmul-pbf16.mlir
```diff
@@ -15,26 +15,28 @@ func.func @matmultpp(%A: memref<4x8xbf16>,
...
   %c0 = arith.constant 0 : index
   %f0 = arith.constant 1.0 : bf16
-  %da = memref.alloc() :memref<4x8xbf16>
+  
+  %da = memref.alloc() : memref<4x8xbf16>
   linalg.fill ins(%f0 : bf16) outs(%da : memref<4x8xbf16>)
-  %db = memref.alloc() :memref<8x4xbf16>
-  linalg.fill ins(%f0:bf16) outs (%db:memref<8x4xbf16>)
+  
+  %db = memref.alloc() : memref<4x4x2xbf16>
```
rengolin, line 22: This test was testing the packing, now it isn't any more. We really shouldn't remove that functionality from these tests.

rengolin (commented):
Some initial comments. 

Overall, what's the strategy for re-enabling the XFAILed tests?

lib/TPP/DefaultTppPasses.cpp
chelini: I can add it. There was no packing pass in the default pipeline so no-need at the moment, but should end up there at some point.

> Some initial comments.
> 
> Overall, what's the strategy for re-enabling the XFAILed tests?

The test on perf should be fixed in #277 and then re-enables. For the bf16 we can try to move them at tensor level, what do you think @KavithaTipturMadhu ?

lib/TPP/Dialect/Transform/LinalgXTransformOps.cpp
chelini: You mean rename the file?

test/BF16/brgemm-tpp.mlir
chelini: It is at memref level there is no pack at that abstraction.

test/BF16/matmul-pbf16.mlir
chelini: Looking at the passes in the test I am not sure the intent was to test the packing pass, there is no pack. It is written as IR and check e2e. Integration for pack is added in another test.

lib/TPP/DefaultTppPasses.cpp
rengolin: Don't the gemm/conv passes add the packs themselves?

rengolin: > For the bf16 we can try to move them at tensor level, what do you think @KavithaTipturMadhu ?

Are there any BF16 packing tests left? The idea was to leave `linalgx.pack` for BF16 until @KavithaTipturMadhu had time to fix, but if we're going to remove, then we must add something in its place.

test/BF16/brgemm-tpp.mlir
rengolin: Then why do we need `-pack-vnni="block-factors=2" -generalize-tensor-pack-unpack`?

Do we have other VNNI packing tests? We can't simply remove all of them and not replace with anything.

test/BF16/brgemm-tpp.mlir
```diff
@@ -1,17 +1,7 @@
-// RUN: tpp-opt -transform-dialect-interpreter -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" --convert-vnni-to-tpp %s | FileCheck %s
-
-transform.sequence failures(propagate) {
- ^bb0(%arg1: !pdl.operation):
-  %0 = transform.structured.match ops{["linalg.batch_reduce_matmul"]} in %arg1
-    transform.structured.pack %0 { use_vnni=true, blocking_factors = [2] }
-}
+// RUN: tpp-opt -pack-vnni="block-factors=2" -generalize-tensor-pack-unpack -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" --convert-vnni-to-tpp %s | FileCheck %s
```
rengolin, line 1: We really should stop using `-empty-tensor-to-alloc-tensor` in all our tests. We need to start building the tests correctly, or fixing them to be correct. This generic allocation passes create more problems than they solve.

test/BF16/matmul-pbf16.mlir
rengolin: It had an explicit `pack` before a `matmul`, which was testing the VNNI packing.

Sure we can replace that with an already "packed" tensor, but then this wouldn't be the same test.

test/BF16/mlp-all-bf16-tpprun.mlir
```diff
@@ -1,7 +1,8 @@
-// RUN: tpp-opt %s -convert-check-to-loops -linalg-ext-to-loops -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -arith-expand -convert-math-to-llvm  -convert-vector-to-scf -convert-scf-to-cf -lower-affine  -reconcile-unrealized-casts |\
+// RUN: tpp-opt %s -convert-check-to-loops  -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -arith-expand -convert-math-to-llvm  -convert-vector-to-scf -convert-scf-to-cf -lower-affine  -reconcile-unrealized-casts |\
 // RUN: tpp-run \
 // RUN:  -e entry -entry-point-result=void
 //
+// XFAIL:*
```
rengolin, line 5: There must be some solution to this, we can't just mark as XFAIL.

test/BF16/mlp-all.mlir
```diff
@@ -1,17 +1,19 @@
...
+// RUN: mlir-cpu-runner \
+// RUN:  -e entry -entry-point-result=void \
+// RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext
 //
-// RUN: tpp-opt %s --pack-vnni="block-factors=2"  --one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize  --linalg-ext-to-loops --convert-linalg-to-tpp --convert-vnni-to-tpp | FileCheck %s -check-prefix=TPP
+// RUN: tpp-opt %s -pack-vnni="block-factors=2" -generalize-tensor-pack-unpack -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-tpp -convert-vnni-to-tpp | FileCheck %s -check-prefix=TPP
```
rengolin, line 6: If the new generalize pass needs an alloc tensor pass right afterwards, than we may be doing something wrong. We should not use those passes unless there's no other way.

If this is just about the one `empty` tensor in main, than it could very well be a `fill`.

test/BF16/mlp-single-layer-bf16.mlir
```diff
@@ -1,28 +1,31 @@
...
-  linalg.fill ins(%c0:bf16) outs(%arg3:memref<128x512xbf16>)
+  
+  %arg3 = memref.alloc() : memref<128x512xbf16>
+  linalg.fill ins(%c0 : bf16) outs(%arg3 : memref<128x512xbf16>)
+  
   tpp.identity ins(%arg2 : memref<512xbf16>) out(%arg3 : memref<128x512xbf16>)
-  %wt = memref.alloc():memref<128x512x2xbf16>
-  linalgx.pack %arg1 inner_dims_pos=[0] inner_tiles=[2] into %wt: (memref<256x512xbf16> memref<128x512x2xbf16>)
-  tpp.vnni_matmul ins(%arg0 : memref<128x256xbf16>, %wt : memref<128x512x2xbf16>) out(%arg3 : memref<128x512xbf16>)
+  tpp.vnni_matmul ins(%arg0 : memref<128x256xbf16>, %arg1 : memref<128x512x2xbf16>) out(%arg3 : memref<128x512xbf16>)
```
rengolin, line 22: Also removing the pack here...

test/BF16/mlp-single-layer-blocked-bf16.mlir
```diff
@@ -1,9 +1,10 @@
-// RUN:tpp-opt %s -linalg-ext-to-loops -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts |\
+// RUN:tpp-opt %s -convert-linalg-to-loops -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts |\
 // RUN: mlir-cpu-runner \
 // RUN:  -e entry -entry-point-result=void  \
 // RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext | \
 // RUN: FileCheck %s
 //
+// XFAIL:*
```
rengolin, line 7: Same XFAIL comment as above.

test/Integration/conv-to-matmul.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts | \
+// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts | \
```
rengolin, line 1: Don't use `-empty-tensor-to-alloc-tensor`

test/Integration/pack-unpack-conversion.mlir
```diff
@@ -0,0 +1,276 @@
+// RUN: tpp-opt %s -generalize-tensor-pack-unpack -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts | \
```
rengolin, line 1: Don't use `-empty-tensor-to-alloc-tensor`

test/BF16/matmul-pbf16.mlir
KavithaTipturMadhu: This test just checks VNNI matmul results for packed layout, it just so happens that I am using pack operation to transform the layout. So, its fine to use the buffer in packed shape directly.

test/BF16/mlp-single-layer-bf16.mlir
KavithaTipturMadhu: This is not a pack test, pack tests usually have -pack-vnni="block-factors..." as a rule of thumb. I think its fine to shape the weight buffer into VNNI format directly. I am anyway introducing an e2e test in the patch for matmul to vnni lowering, so that test covers the pack/unpack cases.

test/BF16/mlp-all.mlir
chelini: generalize-tensor-pack-unpack does not emit any `tensor.empty`. The op is there because we build pack/unpack with a tensor.empty as destination. We can always allocate for tensor.pack or tensor.unpack but we need more thinking especially if you start fusion or vectorizing this ops.

test/Integration/conv-to-matmul.mlir
chelini: done.

test/BF16/brgemm-tpp.mlir
chelini: I can follow-up on this.

test/BF16/brgemm-tpp.mlir
chelini: @KavithaTipturMadhu do we have?

test/BF16/mlp-all-bf16-tpprun.mlir
chelini: @KavithaTipturMadhu what do you think we should do here? If the goal of the test is mapping to VNNI ops I think we can remove it?

test/BF16/brgemm-tpp.mlir
KavithaTipturMadhu: This one's a unit test that checks if linalg.batch_reduce_matmul is lowered correctly. This one needs a pack that must be checked using filecheck.

KavithaTipturMadhu: yes, as long as there's an explicit check for pack, its fine.

chelini: Done in: e158ef01c51ed5036fb40932f1b79d65ca6cb4e1

KavithaTipturMadhu (approved):

PR #239, opened 2022-12-22
Adding ResNet50 bottleneck block as a benchmark

nhasabni:
Converts `linalg.conv2d` into `linalg.matmul` via `transform` dialect. `linalg.matmul` then gets mapped into `xsmm` MatMul call.


alheinecke (commented):
Thanks! We should add a bottleneck example where the BN is folded into the weights. For inference there is no explicit BN.

As this is not a full topology, I suggest removing BN and just keeping Bias and ReLU

rengolin (commented):
This isn't a benchmark yet (we need to C version to compare), so I created a new directory in tests (`Models`) for this kind of testing. We can copy to benchmarks later when we have a C implementation.

rengolin:
The test fails with parsing error. It's possible that you tested this with an old version of the repository bound to an old version of LLVM.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,653 @@
...
+            ^bb0(%in: f32, %in_34: f32, %out: f32):
+                %sum = arith.addf %in, %in_34 : f32
+                linalg.yield %sum : f32
+    } -> !second_conv1x1_output_tensor_t
+
+    return %2 : !second_conv1x1_output_tensor_t
+}
+
+// Output
+// CHECK: ( TBD )
```
rengolin, line 645: We have to have checks to be sure the test is doing what it should. Just executing isn't enough to know much.

We need either:
* `FileCheck` looking at the IR output from `tpp-opt`
* `FileCheck` looking at the execution output from `tpp-run`
* Check dialect, comparing the result with a "golden" tensor, and asserting on difference.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,653 @@
...
+
+transform.sequence failures(propagate) {
```
rengolin, line 647: No transforms in generic tests. This needs to use the conv passes that Lorenzo added back recently.

rengolin (commented):
Thanks! It's a huge test, perhaps could be broken into smaller pieces? One for each function, for example?

We also need to split into CHECK IR (first) and CHECK Output (later) tests.

Some comments inline.

test/Benchmarks/resnet50_bottleneck.mlir
chelini: DecomposeConvToMatmul should do the trick.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,653 @@
+// RUN: tpp-opt %s -transform-dialect-interpreter -map-linalg-to-tpp -empty-tensor-to-alloc-tensor \
+// RUN: -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
+// RUN: -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize \
+// RUN: -map-linalg-to-tpp -convert-linalg-to-tpp="use-parallel-loops=false" \
+// RUN: -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func \
+// RUN: -expand-strided-metadata -convert-memref-to-llvm -lower-affine \
```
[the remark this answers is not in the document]
nhasabni: Addressed.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,653 @@
...
+// RUN: -convert-arith-to-llvm -convert-scf-to-cf -convert-cf-to-llvm -convert-arith-to-llvm -convert-math-to-llvm -cse | \
+// RUN: head -n -9 | awk '{print $0} END{print "}"}' | \
```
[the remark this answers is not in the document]
nhasabni: It was needed because of `transform` dialect appearing in the output of `tpp-opt`. Removed it now.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,653 @@
...
+// RUN: tpp-run -n 10 -print=false \
```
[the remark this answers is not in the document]
nhasabni: Added `CHECK` calls to check the lowering to XSMM.

test/Benchmarks/resnet50_bottleneck.mlir
nhasabni: Using `FileCheck` to check the IR output of `tpp-opt` now.

test/Benchmarks/resnet50_bottleneck.mlir
nhasabni: Done. Thanks @chelini for pointer.

rengolin:
Works towards #190

Issue #235, opened 2022-12-21
Add `expect-sane` operation to `check` dialect


rengolin:
We could add a `check-sane` (or some better name) to the `check` dialect, that would return true if the tensor/memref doesn't have NaNs or Infs. We could even add a `delta` to make sure the normalization happened correctly (`-1 < x < 1`).

@KavithaTipturMadhu

KavithaTipturMadhu:
I am not sure I understand the normalization part, why would we need a delta to check normalization of floating point?

rengolin:
The point was to check if the values are all between `-1 ` and `1` after normalization, as a "sanity" check, but just realised it can be done with `check_almost_eq(tensor, 0.0, 1.0)`. I think #241 fixes this issue completely.

PR #241, opened 2022-12-22
Expect Sane Check Op

KavithaTipturMadhu:
Fixes https://github.com/plaidml/tpp-mlir/issues/235. Delta argument for normalization pending.


lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -107,8 +106,70 @@ struct ConvertExpectTrueOp : public OpRewritePattern<ExpectTrueOp> {
...
+// %2 = arith.cmpf one, %0, %inf : f32
+// %compare = arith.andi %1, %2 : i1
+// cf.assert %compare, "Buffer can't contain NaNs or Infinite values"
+struct ConvertExpectSaneOp : public OpRewritePattern<ExpectSaneOp> {
+  using OpRewritePattern<ExpectSaneOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ExpectSaneOp expectSaneOp,
+                                PatternRewriter &rewriter) const override {
+    Location loc = expectSaneOp.getLoc();
+    auto elementType = expectSaneOp.getOperand().getType().getElementType();
```
rengolin, line 124: Nit: cache operand, type and element type to use in the code below, makes for clearer code.

test/Dialect/Check/check-ops-fail.mlir
```diff
@@ -1,18 +1,24 @@
...
-// RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext
+// DEFINE: %{option} = entry
+// DEFINE: %{command} = tpp-opt %s -map-linalg-to-tpp -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map"  -canonicalize -drop-equivalent-buffer-results -finalizing-bufferize -convert-check-to-loops -convert-linalg-to-loops -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func -convert-vector-to-scf -convert-scf-to-cf -convert-vector-to-llvm -convert-func-to-llvm -convert-memref-to-llvm -convert-math-to-llvm -canonicalize -reconcile-unrealized-casts |\
+// DEFINE: mlir-cpu-runner \
+// DEFINE:  -e %{option} -entry-point-result=void  \
+// DEFINE: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext
 //
 
-// XFAIL: *
+// RUN: %{command}
```
rengolin, line 8: Cant you use split-file?

test/Dialect/Check/check-ops-fail.mlir
```diff
@@ -30,3 +36,25 @@ func.func @myfunc2() {
   check.expect_almost_eq(%a, %b, %threshold): tensor<4x4xf32>, tensor<4x4xf32>, f32
   return
 }
+
+// -----
+// REDEFINE: %{option} = entry3
+// RUN: %{command}
+
+func.func @entry3(%arg0:tensor<4x4xf32>){
+  check.expect_sane(%arg0): tensor<4x4xf32>
```
rengolin, line 45: this may have unexpected results if the memory happens to have "reasonable" float values.

lib/TPP/ConvertCheckToLoops.cpp
```diff
@@ -107,8 +106,70 @@ struct ConvertExpectTrueOp : public OpRewritePattern<ExpectTrueOp> {
...
+    SmallVector<Value> steps(rank, one);
+    (void)scf::buildLoopNest(
+        rewriter, loc, lbs, ubs, steps,
+        [&](OpBuilder &b, Location loc, ValueRange localIvs) {
+          Value scalarOperand = b.create<memref::LoadOp>(
+              loc, expectSaneOp.getOperand(), localIvs);
+          Value zeroVal = b.create<arith::ConstantOp>(
+              loc, elementType, rewriter.getZeroAttr(elementType));
+
+          Value inf = rewriter.create<arith::ConstantOp>(
```
rengolin, line 149: what about `-Inf`?

rengolin (commented):
Looks good, thanks! Some comments inline.

We also need a bit more tests in there. For example:
* Signaling NaNs, and NaNs with mantissa
* Negative infinity (`log(0)`), arithmetic (`1/0`) infinity
* Positive cases (where it should pass)

test/Dialect/Check/check-ops-fail.mlir
KavithaTipturMadhu: Split file doesn't work on mlir-cpu-runner.

test/Dialect/Check/check-ops-fail.mlir
KavithaTipturMadhu: Fair enough, removing this test.

lib/TPP/ConvertCheckToLoops.cpp
KavithaTipturMadhu: Modifying the check to compare against absolute value of the operand.

KavithaTipturMadhu:
@rengolin The current implementation of check.expect_sane doesn't return anything, and just asserts for non NaN/Inf values. Should we return a value from expect_sane and then assert on the return value explicitly?

rengolin:
> @rengolin The current implementation of check.expect_sane doesn't return anything, and just asserts for non NaN/Inf values. Should we return a value from expect_sane and then assert on the return value explicitly?

I thought about that, too. I think both `check_almost_eq` and `check_sane` should have the same behaviour, whatever it is.

The earlier discussion raised this question too, and I think the result was that there isn't much we can do is the tensors are not sane/equals other than assert, so there's not much point in needing to call `check_true` on the result of `check_sane` every time.

The one thing we could do is to use that on a conditional branch and print some error message, so that the remaining tests can continue to run, for example. But that's not a trivial thing to do in MLIR testing with FileCheck.

So I think we can continue with the assert for now and when we create the RFC upstream, we describe the choices and ask for opinions.

rengolin (approved):

Issue #234, opened 2022-12-21
Create a pre-pass in `tpp-run` with the default pipeline


rengolin:
`tpp-run` needs MLIR to be in a certain state (post-TPP), adds the main function, then lowers the rest to LLVM Dialect (as a post-pass) and executes.

We need to change it to accept any MLIR file from "linalg on tensors" to "scf with xsmm calls on memrefs" (not LLVM Dialect), so that we can run any file through it.

The first step is to create a pre-pass, just like the post-pass, that lowers the IR to the required format. This will, in turn, be our default pipeline (which can be reproduced by `tpp-opt` by hand).

The main goals of this issue are:
* Change all execution tests to run on `tpp-run` instead of `mlir-cpu-runner`
* Add new tests to pass increasingly lower IR forms to make sure the pipeline works
* Be able to add the main file (perhaps optional) to IR from any level (up to XSMM, not LLVM).

adam-smnk:
Note to future self.
Added basic collection of default passes (`-default-tpp-passes`) to lower TPP to core low-level dialects.

Next steps:
- create test cases with progressive lowering and mixed levels of IR
- move from mlir-cpu-runner to tpp-run
- try adding more specialized passes e.g., packing and tiling with some default values, brgemm stuff etc.

rengolin:
We need to get the convs too. We should add at least one benchmark with a 1x1 conv to test that.

PR #245, opened 2022-12-23
`tpp-opt` default passes

adam-smnk:
Creates a new pass `-default-tpp-passes` which acts as a default `tpp-opt` pre-pass i.e., a collection of different conversions, lowerings, and other preprocessing steps, that ensures none of the custom TPP ops exist after the pass runs.

Work towards #234 

The existing test benchmarks were moved to use the new pass as they already utilize `tpp-run`,

Next steps:
- create test cases with progressive lowering and mixed levels of IR
- move from `mlir-cpu-runner` to `tpp-run`
- try adding more specialized passes e.g., packing and tiling with some default values, brgemm stuff etc.

### DIFF
--- a/include/TPP/Passes.h
+++ b/include/TPP/Passes.h
@@ -23,6 +23,24 @@ class FuncDialect;
 } // namespace mlir
 
 namespace mlir {
+namespace bufferization {
+class BufferizationDialect;
+} // namespace bufferization
+} // namespace mlir
+
+namespace mlir {
+namespace math {
+class MathDialect;
+} // namespace math
+} // namespace mlir
+
+namespace mlir {
+namespace arith {
+class ArithDialect;
+} // namespace arith
+} // namespace mlir
+
+namespace mlir {
 namespace vector {
 class VectorDialect;
 } // namespace vector
@@ -41,9 +59,21 @@ class SCFDialect;
 } // namespace mlir
 
 namespace mlir {
+namespace tensor {
+class TensorDialect;
+} // namespace tensor
+} // namespace mlir
+
+namespace mlir {
 namespace memref {
 class MemRefDialect;
 } // namespace memref
+} // namespace mlir
+
+namespace mlir {
+namespace tpp {
+class TppDialect;
+} // namespace tpp
 } // namespace mlir
 
 namespace mlir {
@@ -90,7 +120,7 @@ createTileConsumerAndFuseProducersPass();
 std::unique_ptr<OperationPass<func::FuncOp>>
 createDecomposeConvToMatmulOrBrgemmPass();
 
-ModulePass *createDefaultTppPass();
+std::unique_ptr<OperationPass<ModuleOp>> createDefaultTppPass();
 } // namespace tpp
 } // namespace mlir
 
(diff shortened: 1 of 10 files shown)


lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -455,8 +455,9 @@ struct ConvertTppAddOp : public OpRewritePattern<tpp::AddOp> {
       return failure();
 
     MemRefType outputMemRef = outputType.cast<MemRefType>();
-    int64_t m = outputMemRef.getShape()[0];
-    int64_t n = outputMemRef.getShape()[1];
+    auto memShape = outputMemRef.getShape();
+    int64_t m = memShape[0];
+    int64_t n = memShape.size() == 1 ? 1 : memShape[1];
```
adam-smnk, line 460: Temporaryt test change - should be dropped

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "TPP/Dialect/Transform/LinalgXTransformOps.h"
+#include "TPP/Dialect/VNNI/BufferizableOpInterfaceImpl.h"
+#include "TPP/Dialect/VNNI/VNNIDialect.h"
+#include "TPP/Dialect/Xsmm/XsmmDialect.h"
+
+using namespace mlir;
+using namespace mlir::tpp;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
```
adam-smnk, line 49: leftover

include/TPP/Passes.h
```diff
@@ -22,6 +22,24 @@ class FuncDialect;
 } // namespace func
 } // namespace mlir
 
+namespace mlir {
```
rengolin, line 25: nit: this is getting silly. We should only have one `namespace mlir` and all others inside. :)

include/TPP/Passes.td
```diff
@@ -237,4 +237,18 @@ def DecomposeConvToMatmulOrBrgemm : Pass<"decompose-conv-to-matmul-or-brgemm",
   let constructor = "mlir::tpp::createDecomposeConvToMatmulOrBrgemmPass()";
 }
 
+def DefaultTppPasses : Pass<"default-tpp-passes", "ModuleOp"> {
+  let summary = "Convert perf to loops";
```
rengolin, line 241: wrong comment

lib/TPP/DefaultTppPasses.cpp
adam-smnk: oh nvm that's fine

adam-smnk:
In our current TPP pipeline there's currently one fork: lowering TPP to either XSMM or loops.
I opted to go XSMM route by default, as it seems more useful for us, with an optional flag to redirect TPP to loops.

Also, currently the default passes only cover major common lowering and bufferization.
I'll look into adding other more specialized passes like `-decompose-conv-to-matmul-or-brgemm` once I have better collection of test cases to figure out proper pipeline order.

rengolin (approved):
Small nits but looks good to me, thanks!

We can split the "default passes" pass later when we start running them on other compilers, to more easily interpolate them with their own passes.

PR #249, opened 2022-12-30
Adding output CHECKs for ResNet50 bottleneck block with tpp-run

nhasabni:
Depends on fix in #248

### DIFF
--- a/test/Benchmarks/resnet50_bottleneck.mlir
+++ b/test/Benchmarks/resnet50_bottleneck.mlir
@@ -1,8 +1,9 @@
 // RUN: tpp-opt %s -decompose-conv-to-matmul-or-brgemm -map-linalg-to-tpp -empty-tensor-to-alloc-tensor \
 // RUN: -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
-// RUN: -drop-equivalent-buffer-results -finalizing-bufferize -buffer-results-to-out-params -buffer-deallocation -canonicalize \
+// RUN: -drop-equivalent-buffer-results -finalizing-bufferize \
+// RUN: -buffer-results-to-out-params -buffer-deallocation -canonicalize \
 // RUN: -map-linalg-to-tpp -convert-linalg-to-tpp="use-parallel-loops=false" \
-// RUN: -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func \
+// RUN: -convert-tpp-to-xsmm -convert-xsmm-to-func \
 // RUN: -expand-strided-metadata -lower-affine | \
 // RUN: tpp-run -n 10 -print \
 // RUN: -e resnet50_bottleneck_block -entry-point-result=void \
--- a/test/Benchmarks/resnet50_bottleneck.mlir
+++ b/test/Benchmarks/resnet50_bottleneck.mlir
@@ -1,0 +1,241 @@
+// RUN: tpp-opt %s -decompose-conv-to-matmul-or-brgemm -map-linalg-to-tpp -empty-tensor-to-alloc-tensor \
+// RUN: -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
+// RUN: -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize \
+// RUN: -map-linalg-to-tpp -convert-linalg-to-tpp="use-parallel-loops=false" \
+// RUN: -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func \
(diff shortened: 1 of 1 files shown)


rengolin (approved):

rengolin:
The leaks detected are likely on the MLIR file. You need to make sure you deallocate all tensors/memrefs that you allocate. It's hard to detect exactly there, because the leak is detected on the JIT execution and there is no more correlation between IR and code at that point.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,241 @@
...
+        } 
+        ins(%2 : !first_conv1x1_output_tensor_t) 
+        outs(%4 : !first_conv1x1_output_tensor_t) {
+            ^bb0(%in: f32, %out: f32):
+            %sum = arith.addf %in, %out : f32
+            linalg.yield %sum : f32
+    } -> !first_conv1x1_output_tensor_t
+    
+    // ReLU
+    %6 = tensor.empty() : !first_conv1x1_output_tensor_t
```
rengolin, line 78: Why don't you use `%5` here, instead of a new `empty`?

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,241 @@
...
+        } 
+        ins(%2 : !conv3x3_output_tensor_t) 
+        outs(%4 : !conv3x3_output_tensor_t) {
+            ^bb0(%in: f32, %out: f32):
+            %sum = arith.addf %in, %out : f32
+            linalg.yield %sum : f32
+    } -> !conv3x3_output_tensor_t
+
+    // ReLU
+    %6 = tensor.empty() : !conv3x3_output_tensor_t
```
rengolin, line 133: Here, too. Is this how TF lowers? If so, we may have to canonicalize to avoid these extra allocations.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,241 @@
...
+        } 
+        ins(%2 : !second_conv1x1_output_tensor_t) 
+        outs(%4 : !second_conv1x1_output_tensor_t) {
+            ^bb0(%in: f32, %out: f32):
+            %sum = arith.addf %in, %out : f32
+            linalg.yield %sum : f32
+    } -> !second_conv1x1_output_tensor_t
+
+    // ReLU
+    %6 = tensor.empty() : !second_conv1x1_output_tensor_t
```
rengolin, line 188: Same here.

rengolin (changes requested):
I think the leak problem is that you create tensors in each function and don't return nor free. We need to make sure that we don't have those `empty` all over the place. We should only have one per op and do everything we can in-place.

It's possible that even this won't fix all the leaks, because `tpp-run` doesn't deallocate the result tensor after the print. If that's the case, then we also need to add a `dealloc` op on the tensor IFF it's a return value (not the last operand).

test/Benchmarks/resnet50_bottleneck.mlir
nhasabni: Yes, the code came from IREE's TF importer.

rengolin: Ok, try to dealloc any non-returned tensor and see if the leak stops. Try building with sanitizers locally to avoid the round trip to CI.

nhasabni: Yes, that's what I am doing now. :)

nhasabni: Fixed the leaks. Please check now.

test/Benchmarks/resnet50_bottleneck.mlir
```diff
@@ -0,0 +1,241 @@
+// RUN: tpp-opt %s -decompose-conv-to-matmul-or-brgemm -map-linalg-to-tpp -empty-tensor-to-alloc-tensor \
+// RUN: -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
+// RUN: -drop-equivalent-buffer-results -finalizing-bufferize -buffer-results-to-out-params -buffer-deallocation -canonicalize \
+// RUN: -map-linalg-to-tpp -convert-linalg-to-tpp="use-parallel-loops=false" \
+// RUN: -convert-linalg-to-tpp -convert-tpp-to-xsmm -convert-xsmm-to-func \
```
chelini, line 5: nit: please drop this second `convert-linalg-to-tpp`.

rengolin (approved):
Thanks! It's good to have this as an example, since it comes from a real framework. If we want to deal with the extra allocations differently, we can do so later, as a cleanup pass (there are other sources of pointless allocations, as discussed elsewhere).

rengolin:
Fixes #190

PR #258, opened 2023-01-06
Re-work tpp semantics and operations

chelini:
Ready for RV

### DIFF
--- a/include/TPP/Dialect/Tpp/TppOps.td
+++ b/include/TPP/Dialect/Tpp/TppOps.td
@@ -65,15 +65,26 @@ class Tpp_UnaryOp<string mnemonic, list<Trait> traits = []> :
 def Tpp_AddOp : Tpp_BinaryOp<"add"> {
   let summary = "Element-wise addition.";
   let description = [{
-    The `tpp.add` operation performs element-wise addition
-    on two-dimensional memref C = A + B.
+    The `tpp.add` operation performs element-wise addition on two-dimensional
+    memrefs, writing the result on the output memref. No checks or assumption are
+    made on the input/output arguments so the same memref can be passed both as
+    input and output. 
 
     Example:
 
     ```mlir
 
+    // A = A + A
+    tpp.add ins(%1 : memref<2x2xf32>, %1: memref<2x2xf32>)
+            out(%1 : memref<2x2xf32>)
+
+    // B = A + B
     tpp.add ins(%1 : memref<2x2xf32>, %2: memref<2x2xf32>) 
             out(%2: memref<2x2xf32>)
+
+    // C = A + B
+    tpp.add ins(%1 : memref<2x2xf32>, %2: memref<2x2xf32>)
+            out(%3 : memref<2x2xf32>)
 
     ```
   }];
@@ -88,7 +99,13 @@ def Tpp_AddOp : Tpp_BinaryOp<"add"> {
 def Tpp_IdentityOp : Tpp_UnaryOp<"identity"> {
   let summary = "Copies input to output.";
   let description = [{
-    The `tpp.identity` copies input memref to output memref.
+    The `tpp.identity` copies input memref to output memref. It supports
+    broadcast comparing the memref shape element-wise. Starting with the trailing
+    dimensions and working from right to left. Two dimensions are broadcast
+    compatible if:
+    - They are equal
+    - One if them is 1 For more details, see: 
+    https://numpy.org/doc/stable/user/basics.broadcasting.html#general-broadcasting-rules
 
     Example:
 
--- a/include/TPP/Dialect/Tpp/TppOps.td
+++ b/include/TPP/Dialect/Tpp/TppOps.td
@@ -125,7 +125,9 @@ def Tpp_ReluOp : Tpp_UnaryOp<"relu", [SameTypeOperands]> {
     tpp.relu ins(%0: memref<2x2xf32>) out(%0: memref<2x2xf32>)
 
     ```
-  }]; 
+  }];
+
+  let hasVerifier = 1; 
 }
 
 //===----------------------------------------------------------------------===//
--- a/include/TPP/Dialect/Tpp/TppOps.td
+++ b/include/TPP/Dialect/Tpp/TppOps.td
@@ -29,47 +29,63 @@ def TppBRGEMMVNNIMemrefInput : StaticMemRefRankOf<[AnyFloat], [4]>;
 
 // Tpp operands is a scalar float or a static memref with rank 1 or 2.
 def TppOperand : AnyTypeOf<[TppMemRef, AnyFloat]>;
+
 // Tpp operands for VNNI layout.
 def TppVNNIOperand : AnyTypeOf<[TppVNNIMemrefInput, AnyFloat]>;
+
+
+// Class for tpp binary operations.
+class Tpp_BinaryOp<string mnemonic, list<Trait> traits = []> :
+  Tpp_Op<mnemonic, !listconcat(traits, [SameTypeOperands])> {
+
+  let arguments = (ins TppOperand:$lhs, TppOperand:$rhs, TppOperand:$out);
+
+  let assemblyFormat = [{
+        `ins` `(` $lhs `:` type($lhs) `,` $rhs `:` type($rhs) `)`
+        `out` `(` $out `:` type($out) `)` attr-dict
+  }];
+}
+
+// Class for tpp.unary operations.
+class Tpp_UnaryOp<string mnemonic, list<Trait> traits = []> :
+  Tpp_Op<mnemonic, !listconcat(traits, [])> {
+  
+  let arguments = (ins TppOperand:$input, TppOperand:$output);
+  
+  let assemblyFormat = [{
+    `ins` `(` $input `:` type($input) `)` `out` `(` $output `:` type($output) `)` 
+    attr-dict
+  }];
+}
 
 //===----------------------------------------------------------------------===//
 // AddOp
 //===----------------------------------------------------------------------===//
 
-def Tpp_AddOp : Tpp_Op<"add", [SameTypeOperands]> {
-    let summary = "Element-wise addition.";
-    let description = [{
-        The `tpp.add` operation performs element-wise addition
-        on two-dimensional memref.
+def Tpp_AddOp : Tpp_BinaryOp<"add"> {
+  let summary = "Element-wise addition.";
+  let description = [{
+    The `tpp.add` operation performs element-wise addition
+    on two-dimensional memref C = A + B.
 
-        Example:
+    Example:
 
-        ```mlir
+    ```mlir
 
-        tpp.add ins(%1 : memref<2x2xf32>) out(%2: memref<2x2xf32>)
+    tpp.add ins(%1 : memref<2x2xf32>, %2: memref<2x2xf32>) 
+            out(%2: memref<2x2xf32>)
 
-        ```
-    }];
+    ```
+  }];
 
-    let arguments = (ins TppOperand:$lhs, TppOperand:$rhs);
-
-    let assemblyFormat = [{
-        `ins` `(` $lhs `:` type($lhs) `)`
-        `out` `(` $rhs `:` type($rhs) `)` attr-dict
-    }];
-
-    let extraClassDeclaration = [{
-      Value getOutput() { return getRhs(); };
-    }];
-
-    let hasVerifier = 1;
+  let hasVerifier = 1;
 }
 
 //===----------------------------------------------------------------------===//
 // IdentityOp
 //===----------------------------------------------------------------------===//
 
-def Tpp_IdentityOp : Tpp_Op<"identity"> {
+def Tpp_IdentityOp : Tpp_UnaryOp<"identity"> {
   let summary = "Copies input to output.";
   let description = [{
     The `tpp.identity` copies input memref to output memref.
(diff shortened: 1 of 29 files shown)


lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -288,8 +286,11 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
       return success();
     }
     if (tpp::utils::isTppAdd(linalgOp)) {
+      // A = A + B
```
rengolin, line 289: IIRC from the discussion yesterday, `A = A + B` requires `A.shape == B.shape`, but for `C = A + B`, we could have `A.rank < B.rank`, with automatic broadcast (if the lower rank dims match).

The definition above uses `SameTypeOperands`, so the latter case would be invalid, which is ok as a first approximation, but if/when we have that case, we'll probably need to have an explicit broadcast before the TPP operations, and match both ops to the auto-broadcast TPP op.

tpp-rt/XsmmRunnerUtils.cpp
```diff
@@ -259,47 +262,6 @@ _mlir_ciface_xsmm_unary_scalar_invoke(const libxsmm_datatype dType,
   kernel(&param);
 }
 
-LIBXSMM_INLINE void matrix_copy_NC_to_NCNC(float *src, float *dst, int T, int N,
```
rengolin, line 262: This looks like an unrelated change...

chelini: Yes, house cleaning. I can do it separately if needed.

rengolin: Easy NFC commit outside of PR.

include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -84,21 +100,14 @@ def Tpp_IdentityOp : Tpp_Op<"identity"> {
...
   let hasVerifier = 1;
 }
 
 //===----------------------------------------------------------------------===//
 // ReluOp
 //===----------------------------------------------------------------------===//
 
-def Tpp_ReluOp : Tpp_Op<"relu"> {
+def Tpp_ReluOp : Tpp_UnaryOp<"relu", [SameTypeOperands]> {
   let summary = "Applies a Rectified Linear Unit function in place.";
```
adam-smnk, line 111: I think some verifier should check that input and output are the same destination for unary ops.
As an example, currently `tpp.relu` happily accepts two different arguments and further lowering just implicitly ignores the input.

adam-smnk: Slightly modified case from `tpp-to-loops` that will pass:
```// CHECK: func.func @relu_to_loops(
func.func @relu_to_loops(%arg0: memref<3x3xf32>, %arg1: memref<3x3xf32>) {
  // CHECK-DAG: %[[ub:.*]] = arith.constant 3 : index
  // CHECK-DAG: %[[lb:.*]] = arith.constant 0 : index
  // CHECK-DAG: %[[step:.*]] = arith.constant 1 : index
  // CHECK-DAG: %[[relu:.*]] = arith.constant 0.000000e+00 : f32
  // CHECK: scf.for %[[i:.*]] = %[[lb]] to %[[ub]] step %[[step]] {
  // CHECK:   scf.for %[[j:.*]] = %[[lb]] to %[[ub]] step %[[step]] {
  // CHECK:     %[[load:.*]] = memref.load %[[ARG0]][%[[i]], %[[j]]] : memref<3x3xf32>
  // CHECK:     %[[max:.*]] = arith.maxf %[[load]], %[[relu]] : f32
  // CHECK:     memref.store %[[max]], %[[ARG0]][%[[i]], %[[j]]] : memref<3x3xf32>
  tpp.relu ins(%arg1: memref<3x3xf32>) out(%arg0: memref<3x3xf32>)
  return
}
```

chelini: ops that's wrong the load should be from %arg1 and the store to %arg0. thx

adam-smnk: So, it is no longer in place only op, right? If that's the case, I think the TableGen docs need slight updates too.

chelini: yes, is not longer in-place but can be if you pass the same subview/memref as input and output.

chelini: will update the .td doc.

chelini: The point is that you can still have in-place beh. if you pass ins and out the same memref. But the new spec is 1 input and 1 output.

adam-smnk (approved):
lgtm

chelini:
@rengolin this is ready for review when you have time

rengolin (commented):
Looking great, thanks! Some comments inline, but nothing serious.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -314,15 +314,53 @@ static bool allIndexingsAreProjectedPermutation(linalg::GenericOp genericOp) {
...
-    return false;
   if (linalgOp.hasTensorSemantics())
     return false;
-  if (!hasStaticShape(linalgOp))
+  return hasStaticShape(linalgOp);
+}
+
+// Return true if the operation is binary.
+bool isBinaryOp(linalg::GenericOp linalgOp) {
+  if ((linalgOp.getNumDpsInputs() == 1) && (linalgOp.getNumDpsInits() == 1)) {
```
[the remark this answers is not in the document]
chelini: The uses are the one in the region of the generic. Just want to make sure we are using the input/output instead of other values, because the lianlg.generic is non-isolated from above and can capture outside IR.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -314,15 +314,53 @@ static bool allIndexingsAreProjectedPermutation(linalg::GenericOp genericOp) {
...
+    Value outArg = linalgOp.getRegionOutputArgs()[0];
+    SmallVector<Operation *> usersOut(outArg.getUsers().begin(),
```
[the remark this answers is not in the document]
chelini: used std::distance.

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -449,14 +457,15 @@ struct ConvertTppAddOp : public OpRewritePattern<tpp::AddOp> {
...
-
+    Type outputType = addOp.getOut().getType();
+    assert(outputType.isa<MemRefType>() && "expect a memref type");
     MemRefType outputMemRef = outputType.cast<MemRefType>();
-    int64_t m = outputMemRef.getShape()[0];
-    int64_t n = outputMemRef.getShape()[1];
+    assert((outputMemRef.getRank() == 1 || outputMemRef.getRank() == 2) &&
+           "expect memref with rank 1 or 2");
+    
+    int64_t m = (outputMemRef.getRank() == 2) ? outputMemRef.getShape()[0] : 1;
```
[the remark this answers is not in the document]
chelini: I am planning to change this code once more when we increase tpp ops. If you don't have string feeling on this I can come back to it later.

include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -29,47 +29,63 @@ def TppBRGEMMVNNIMemrefInput : StaticMemRefRankOf<[AnyFloat], [4]>;
...
+  let hasVerifier = 1;
 }
 
 //===----------------------------------------------------------------------===//
 // IdentityOp
 //===----------------------------------------------------------------------===//
 
-def Tpp_IdentityOp : Tpp_Op<"identity"> {
+def Tpp_IdentityOp : Tpp_UnaryOp<"identity"> {
   let summary = "Copies input to output.";
```
nhasabni, line 89: I think `tpp.identity` also supports broadcasting. If so, can you please update the description to mention about it?

include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -29,47 +29,63 @@ def TppBRGEMMVNNIMemrefInput : StaticMemRefRankOf<[AnyFloat], [4]>;
...
-    }];
+    Example:
 
-    let extraClassDeclaration = [{
-      Value getOutput() { return getRhs(); };
-    }];
+    ```mlir
+
+    tpp.add ins(%1 : memref<2x2xf32>, %2: memref<2x2xf32>) 
+            out(%2: memref<2x2xf32>)
```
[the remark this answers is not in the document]
nhasabni: I see the code for mapping `linalg.generic` into `tpp.add`. Looks like we are supporting both in-place and out-of-place add. If so, can you please clarify it in the description and also add an example of out-of-place add?

lib/TPP/Dialect/Tpp/TppUtils.cpp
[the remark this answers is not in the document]
rengolin: Makes sense.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -314,15 +314,55 @@ static bool allIndexingsAreProjectedPermutation(linalg::GenericOp genericOp) {
...
+}
+
+static bool hasZeroUser(Value val) {
+  return std::distance(val.getUsers().begin(), val.getUsers().end()) == 0;
+}
+
+// Return true if the operation is binary.
+static bool isBinaryOp(linalg::GenericOp linalgOp) {
+  if ((linalgOp.getNumDpsInputs() == 1) && (linalgOp.getNumDpsInits() == 1)) {
+    Value outArg = linalgOp.getRegionOutputArgs()[0];
```
rengolin, line 337: If the region (of a non-binary linalg region) has no output arguments, this will crash. You only check for DPS inputs in the if expression.

Perhaps it could be easier to just put a simple check outside of both IFs in the first line:
```
  if (linalgOp.getOutputs().size() < 1)
    return false;
  Value outArg = linalgOp.getRegionOutputArgs()[0];
  ...
```

test/Conversion/TppToLoops/tpp-to-loops.mlir
```diff
@@ -1,149 +1,202 @@
...
-  // CHECK:   }
-  // CHECK: }
+  // CHECK:     %[[load1:.*]] = memref.load %[[ARG0]][%[[i]], %[[j]]] : memref<3x3xf32>
+  // CHECK:     %[[load2:.*]] = memref.load %[[ARG1]][%[[i]], %[[j]]] : memref<3x3xf32>
+  // CHECK:     %[[add:.*]] = arith.addf %[[load1]], %[[load2]] : f32
+  // CHECK:     memref.store %[[add]], %[[ARG1]][%[[i]], %[[j]]] : memref<3x3xf32>
+  tpp.add ins(%arg0: memref<3x3xf32>, %arg1: memref<3x3xf32>) out(%arg1: memref<3x3xf32>)
+  return
+}
+
```
nhasabni, line 72: Can you also add a test for `add_to_loops` with out-of-place `add`? Looks like it is missing.

nhasabni:
Some comments, otherwise LGTM.

include/TPP/Dialect/Tpp/TppOps.td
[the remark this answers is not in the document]
chelini: Update doc.

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: I think, at least in the generic, getNumDpsInits is the number of output: https://github.com/llvm/llvm-project/blob/4c6d6fc176a115d196194c6d5091d3279d27fffa/mlir/include/mlir/Dialect/Linalg/IR/LinalgStructuredOps.td#L221

chelini: https://github.com/llvm/llvm-project/blob/4c6d6fc176a115d196194c6d5091d3279d27fffa/mlir/include/mlir/Interfaces/DestinationStyleOpInterface.td#L78

adam-smnk: Yeah, it looks like inits map to output - I found this example a bit clearer: https://github.com/llvm/llvm-project/blob/4c6d6fc176a115d196194c6d5091d3279d27fffa/mlir/lib/Dialect/Linalg/Transforms/DropUnitDims.cpp#L558

Although, I wouldn't mind some more obvious API for that.

chelini: Welcome to MLIR

lib/TPP/ConvertTppToXsmm.cpp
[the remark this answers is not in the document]
rengolin: Makes sense.

include/TPP/Dialect/Tpp/TppOps.td
rengolin: I'm not seeing this change in the current state.

chelini: Added doc here: https://github.com/plaidml/tpp-mlir/pull/258/commits/0e9e2107a4c6c6d0d4014bdd304c870668388ea9

rengolin: Oh, I must have been looking at the wrong time. Github isn't always smart on email links.

rengolin (approved):
LGTM, thanks!

PR #262, opened 2023-01-10
VNNI packing to linalg generic without blocking factors specified

KavithaTipturMadhu:


lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -32,6 +33,8 @@ static LogicalResult checkStructure(linalg::LinalgOp linalgOp) {
   if (iteratorTypes.size() < 4)
     return failure();
   size_t size = iteratorTypes.size() - 1;
+  if (size == 6)
```
chelini, line 36: Would it make sense to split the logic out of this function? `checkStructure` was specific for linalg.brgemm

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -80,11 +89,17 @@ static LogicalResult checkAccessPatterns(linalg::LinalgOp linalgOp) {
...
-  // Expected access patterns of BRGEMM
-  SmallVector<AffineMap> expectedMaps =
-      infer({{r1, p3, r2}, {r1, r2, p4}, {p3, p4}});
+  AffineExpr r1, p3, p4, r2, r3;
+  SmallVector<AffineMap> expectedMaps;
+  if (hasVNNIMap) {
+    bindDims(linalgOp.getContext(), r1, p3, p4, r2, r3);
+    // Expected access patterns of BRGEMM
+    expectedMaps =
+        infer({{r1, p3, r2}, {r1, r2.floorDiv(2), p4, r3}, {p3, p4}});
```
chelini, line 98: We are going to always use 2?

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -175,19 +199,25 @@ mlir::linalgx::mapToBRGEMMOp(RewriterBase &rewriter,
...
-                  ValueRange{slicedOperands[0], slicedOperands[1]},
-                  slicedOperands[2])
-            : builder.create<linalg::BatchReduceMatmulOp>(
-                  loc, ValueRange{slicedOperands[0], slicedOperands[1]},
-                  slicedOperands[2]);
-
-    tensorResults = insertSlicesBack(builder, loc, linalgOp, slicedOperands,
-                                     brgemm->getResults());
-
+    if (slicedOperands[1].getType().cast<ShapedType>().getRank() == 3) {
```
chelini, line 202: yes, I would definitely split this to another function.

chelini (commented):
Mind adding a test on how the IR looks like? It easier to review.

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: right now, we're only considering bf16 packing by a factor of 2. Therefore, I have a hardcoded 2 here. I can change this such that a variable is used, which is assigned 2 by default.

lib/TPP/MapToBatchReduceGEMM.cpp
rengolin: Also, we shouldn't rely on just the size for this check. This needs to make sure the extra dimension is really a VNNI one (dim/type = 2/bf16, 4/bf8 and coming from a pack etc).

lib/TPP/MapToBatchReduceGEMM.cpp
rengolin: Here there could be a `static always_inline getVNNIPackingRatio(Type)` or something that returns 2 for BF16 and 4 for BF8 (or something like that).

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -102,17 +117,23 @@ static LogicalResult checkBody(linalg::LinalgOp linalgOp) {
...
-    FailureOr<Value> slicedOperand = utils::getSliceOperand(
-        builder, operand, linalgOp, localIvs, valuesToUse, 3);
-    if (failed(slicedOperand))
-      return failure();
+    FailureOr<Value> slicedOperand;
+    AffineMap map = linalgOp.getMatchingIndexingMap(operand);
+    if (map.getNumResults() == 5) {
+      slicedOperand = utils::getSliceOperand(builder, operand, linalgOp,
+                                             localIvs, valuesToUse, 4);
+    } else {
```
rengolin, line 131: No error checking here?

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: Added getVNNIBlockingSize utility

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: what kind of error checking do you have in mind? I am extracting a map from the affine map of each opoperand and have a check for opoperand 1 to have a map with5 results. Do you think this map ought to be checked for vnni blocking?

include/TPP/TransformUtils.h
```diff
@@ -53,7 +53,7 @@ Value getSliceOperand(OpBuilder &builder, linalg::LinalgOp linalgOp,
 FailureOr<SmallVector<Range>> getLoopsToMaterialize(RewriterBase &rewriter,
                                                     linalg::LinalgOp linalgOp,
                                                     unsigned upTo);
-
+int getVNNIBlockingFactor(Type type);
```
chelini, line 56: Please add some doc here.

lib/TPP/Dialect/Transform/LinalgXTransformOps.cpp
```diff
@@ -49,22 +49,15 @@ transform::PackOpExt::applyToOne(linalg::LinalgOp target,
         packedOp = mlir::linalgx::packConv2DNhwcHwcfOp(rewriter, convOp,
                                                        blockingFactors);
       })
+      .Case([&](linalg::GenericOp matmulOp) {
```
chelini, line 52: Why this changes on the transform?

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -18,13 +19,38 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "llvm/Support/Debug.h"
 
+#include <iostream>
```
chelini: drop please.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -18,13 +19,38 @@
...
+    return failure();
+  size_t size = iteratorTypes.size() - 1;
+  bool match = linalg::isReductionIterator(iteratorTypes[size]) &&
+               linalg::isReductionIterator(iteratorTypes[size - 1]) &&
+               linalg::isParallelIterator(iteratorTypes[size - 2]) &&
+               linalg::isParallelIterator(iteratorTypes[size - 3]) &&
+               linalg::isReductionIterator(iteratorTypes[size - 4]);
+  if (!match)
+    return failure();
+  size = size - /*BRGEMM loops*/ 4;
```
chelini, line 43: BRGEMM has 3 loops in linalg can we maybe specify VNNI BRGEMM or something similar?

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -58,6 +83,45 @@ static bool isOutputOperand(linalg::LinalgOp linalgOp, OpOperand &operand) {
...
+      assert(isOutputOperand(linalgOp, operand));
+      if (map.getNumResults() < 2)
+        return failure();
+      maps.push_back(map.getMinorSubMap(2));
+    }
+  }
+  SmallVector<AffineMap> compressedDimMaps = compressUnusedDims(maps);
+  using MapList = ArrayRef<ArrayRef<AffineExpr>>;
+  auto infer = [](MapList m) { return AffineMap::inferFromExprList(m); };
+  AffineExpr r1, p3, p4, r2, r3;
```
chelini, line 109: Not sure I understand this part. Are you looking for two parallel loops and 3 reductions? Why is called p3 and p4?

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -102,17 +167,23 @@ static LogicalResult checkBody(linalg::LinalgOp linalgOp) {
...
 
   SmallVector<Value> slicedOperands;
   for (OpOperand *operand : linalgOp.getDpsInputOperands()) {
-    FailureOr<Value> slicedOperand = utils::getSliceOperand(
-        builder, operand, linalgOp, localIvs, valuesToUse, 3);
-    if (failed(slicedOperand))
-      return failure();
+    FailureOr<Value> slicedOperand;
+    AffineMap map = linalgOp.getMatchingIndexingMap(operand);
+    if (map.getNumResults() == 5) {
```
chelini, line 178: This if switches on VNNI brgemm and traditional BRGEMM? If this is the case I don't think is enough.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -125,6 +196,17 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
   return slicedOperands;
 }
 
+static bool isLinalgVNNILoop(linalg::LinalgOp linalgOp) {
```
chelini, line 199: avoid the if pls. Why is called `isLinalgVNNILoop`?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -8,6 +8,7 @@
 #include "TPP/Dialect/Tpp/TppUtils.h"
 #include "TPP/Dialect/VNNI/VNNIOps.h"
 #include "TPP/Passes.h"
+#include "TPP/TransformUtils.h"
```
chelini, line 11: Maybe we should have a `TransformUtilsVNNI.h`?

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -138,19 +220,25 @@ mlir::linalgx::mapToBRGEMMOp(RewriterBase &rewriter,
   if (!isa<linalg::GenericOp>(linalgOp))
     return rewriter.notifyMatchFailure(linalgOp, "expects a linalg.generic");
 
-  if (failed(checkStructure(linalgOp)))
+  bool isVNNILoop = isLinalgVNNILoop(linalgOp);
```
chelini, line 223: I am a bit concerned of having all this checks here. Can we have a MapToBatchReduceVNNIGemm.cpp? I feel is already complicated the logic and adding VNNI on top makes it even harder to debug. What do you think?

test/BF16/mlp-all.mlir
```diff
@@ -13,92 +13,66 @@
...
+  // TPP: scf.parallel
+  // TPP: tpp.vnni_brgemm
   %10 = linalg.matmul ins(%7, %arg5 : tensor<128x1024xbf16>, tensor<1024x2048xbf16>) outs(%9 : tensor<128x2048xbf16>) -> tensor<128x2048xbf16>
   %11 = linalg.generic {indexing_maps = [#map1], iterator_types = ["parallel", "parallel"]} outs(%10 : tensor<128x2048xbf16>) {
   ^bb0(%arg9: bf16):
     %16 = arith.maxf %arg9, %c0 : bf16
     linalg.yield %16 : bf16
   } -> tensor<128x2048xbf16>
 
-  // TPP: tpp.identity ins(%[[ARG8]] : memref<1000xbf16>) out(%[[OUTPUT]] : memref<128x1000xbf16>)
```
chelini, line 83: why we don't look for tpp.identity anymore?

lib/TPP/Dialect/Transform/LinalgXTransformOps.cpp
KavithaTipturMadhu: Earlier, there was a structured pack transform that used to operate on gemm, now the plan is to get it to work on linalg.generic op to spit out vnni.brgemm instead of matmul. Do you think this needs to be removed entirely?

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: There is an extra trailing reduction dimension in linalg.generic in case of VNNI matmul, so I am checking for it in the maps. I thought the name of expressions doesn't matter, does it? Should I use p4, p5?

lib/TPP/MapToBatchReduceGEMM.cpp
rengolin: It's not always possible to differentiate a packed `<NxMxnxTy>` from a standard `<MxNxKxTy>` where both `n` and `K` are 2.

the main tell-tales are:
* inner dim is 2 and type is BF16 || inner dim is 4 and type is bf8
* the _other_ shape in the operation doesn't have the same dimension (is this always true?)
* the producer is a `pack` operation with VNNI shape at some point in the use-def chain (this is hard to check)
* none of the ops in the use-def chain undo the VNNI pack (like matmul) (also hard to check)

In the end, it's not a trivial check, so perhaps we could keep a set somewhere of all packed shapes, adding them when we issue a `pack` and removing when we issue an `unpack` or a `matmul` that we know undoes the pack.

A quick check on that set would tell us that the shape is packed, and looking at the inner dimension + type should tell us which way.

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: I could refactor to have a separate MapToBatchReduceVNNIGemm.cpp, only problem is lots of code is likely to end up duplicated.

test/BF16/mlp-all.mlir
KavithaTipturMadhu: The test seems to be failing for argument checks even. I have stripped it down to a minimum set of filechecks and intend to get the pack ops' checks back in another PR. Filechecks exist for single layer in another test so we can rest assured that this is covered elsewhere.

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: I'll rename this, I called it isLinalgVNNILoop because I check for loop's dimensions to be 7 in the method.

lib/TPP/MapToBatchReduceGEMM.cpp
rengolin: I just meant:
```
  if (failure(slicedOperand))
    return failure;
```

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: This if checks if the op's map has result == 5 implying that the operand is packed for VNNI. I have added a flag to cover the linalg op check for VNNI. I can't think of anything more to check, can you point out what other check might have to be covered?

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: I've updated the names of affine expressions to p4, p5 instead of p3, p4.

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: I've also updated this check to see if the second operand's source is a VNNI pack operation, this works in tensor semantics only.

include/TPP/VNNIUtils.h
```diff
@@ -0,0 +1,23 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Types.h"
+
+namespace mlir {
+namespace vnni {
+namespace utils {
+//Returns the VNNI blocking factor: 2 for BF16/4 for BF8
```
chelini, line 17: nit add space and finish sentence with '.'

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -59,6 +85,45 @@ static bool isOutputOperand(linalg::LinalgOp linalgOp, OpOperand &operand) {
   return !isInputOperand(linalgOp, operand);
 }
 
+static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
+  SmallVector<AffineMap> maps;
+  for (OpOperand &operand : linalgOp->getOpOperands()) {
```
chelini, line 90: this nesting still looks a bit entangled can you first scan the input and then the output?

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -59,6 +85,45 @@ static bool isOutputOperand(linalg::LinalgOp linalgOp, OpOperand &operand) {
...
+    AffineMap map = linalgOp.getMatchingIndexingMap(&operand);
+    if (isInputOperand(linalgOp, operand)) {
+      if (map.getNumResults() < 3)
+        return failure();
+      if (operand.getOperandNumber() == 1) {
+        assert(map.getNumResults() == 5);
```
chelini, line 96: not sure if asserting is the right think as we are checking the access pattern, bailing out for me would be better.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -59,6 +85,45 @@ static bool isOutputOperand(linalg::LinalgOp linalgOp, OpOperand &operand) {
...
+      maps.push_back(map.getMinorSubMap(2));
+    }
+  }
+  SmallVector<AffineMap> compressedDimMaps = compressUnusedDims(maps);
+  using MapList = ArrayRef<ArrayRef<AffineExpr>>;
+  auto infer = [](MapList m) { return AffineMap::inferFromExprList(m); };
+  AffineExpr r1, p4, p5, r2, r3;
+  SmallVector<AffineMap> expectedMaps;
+  bindDims(linalgOp.getContext(), r1, p4, p5, r2, r3);
+  // Expected access patterns of BRGEMM
```
chelini, line 114: nit: end sentence with '.'

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -126,6 +200,53 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
   return slicedOperands;
 }
 
+// Returns true if linalg op is using its second operand in VNNI format
+static bool isLinalgOpVNNI(linalg::LinalgOp linalgOp) {
```
chelini, line 204: Let's split this method a bit. It is difficult to follow.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -126,6 +200,53 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
...
+              .getType()
+              .cast<ShapedType>()
+              .getShape()[linalgOp->getOperands()[1]
+                              .getType()
+                              .cast<ShapedType>()
+                              .getRank() -
+                          1] ==
+          vnni::utils::getVNNIBlockingFactor(
+              linalgOp->getOperands()[1].getType())) {
+    if (!linalgOp.hasBufferSemantics()) {
```
chelini, line 221: this check is too restrictive. Looking for a pack may not be right, if you run this after propagation. The pack cannot be there anymore.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -383,47 +385,108 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
   return replacementOp;
 }
 
-FailureOr<vnni::MatmulOp>
+bool isVNNIPacked(linalg::GenericOp matmulOp) {
+  // TODO add VNNI packing checks here
```
chelini, line 389: should you check the body too? `tpp::utils::hasMatmulBody`?

test/BF16/matmul-vnni.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -pack-vnni="block-factors=2"  -canonicalize | FileCheck %s
+// RUN: tpp-opt %s --pack-matmul="block-factors=32,32,32"  --pack-vnni -map-to-brgemm  -canonicalize | FileCheck %s
```
chelini, line 1: are you also retiring pack-vnni and use only pack-matmul for both bf16 and f32?

KavithaTipturMadhu: I am not retiring -pack-vnni, just getting rid of the blocking factor specification as an argument to pack-vnni.

lib/TPP/ToBlockLayoutAndBack.cpp
KavithaTipturMadhu: I check that where I call isVNNIPacked method.

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: I'll get rid of this.

lib/TPP/MapToBatchReduceGEMM.cpp
KavithaTipturMadhu: If I get rid of the pack op check, would you recommend refactoring still?

lib/TPP/VNNIUtils.cpp
```diff
@@ -0,0 +1,22 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+
+#include "TPP/VNNIUtils.h"
+
+namespace mlir {
+namespace vnni {
+namespace utils {
+int getVNNIBlockingFactor(Type type) {
```
chelini, line 15: Why this is not a simple constant? Why do you need to pass the type?

lib/TPP/MapToBatchReduceGEMM.cpp
chelini: line 211 till 215 are still difficult to follow for me. Maybe moving the operand into a variable would help. Something like:

```
auto firstOperand = linalgOp->getOperands()[1].getType().cast<ShapedType>();
if (linalgOp.getNumLoops() == 7 && linalgOp->getOperands().size() >= 3 && firstOperand.isBF16() && firstOperand.getShape()[firstOperand.getRank() - 1] == 2) {

}
```
 Also this if seems to check two different things: 1 structural property of the ops (i.e., number of loops) and 2 a specific operand of the op. We could split the if by checking first the structural properties and then the operands.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -383,47 +385,108 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
...
 }
 
 FailureOr<vnni::BRGemmOp>
 mlir::linalgx::packVNNIBRGemmOp(RewriterBase &rewriter,
-                                linalg::BatchReduceMatmulOp brgemmOp,
-                                ArrayRef<OpFoldResult> tiles) {
-  if (tiles.size() != 1)
-    return rewriter.notifyMatchFailure(brgemmOp, "require 1 blocking factor");
+                                linalg::BatchReduceMatmulOp brgemmOp) {
+  if (!brgemmOp.getInputs()[0]
```
chelini, line 470: this seems a recurring pattern. Introduce it as helper function in VNNIUtils.h

lib/TPP/VNNIUtils.cpp
KavithaTipturMadhu: We want to check the element type to be bf16/bf8. The method is a wrapper on the type check code.

chelini: I see, here you check only for BF16. Also more than a check is an assert. Maybe return something like: 
```
Optional<int64_t> getVNNIBlockingFactor(Type t) {
if (!type.isBF16())
  return std::nullopt
return 2
}
```

KavithaTipturMadhu:
@chelini can you please take a look?

include/TPP/VNNIUtils.h
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Types.h"
+
+namespace mlir {
+namespace vnni {
+namespace utils {
+// Returns the VNNI blocking factor: 2 for BF16/4 for BF8.
```
chelini, line 17: Returns the VNNI blocking factor: 2 for BF16/4 for BF8 -> Returns the VNNI blocking factor: 2 for BF16 and 4 for BF8

lib/TPP/VNNIUtils.cpp
```diff
@@ -0,0 +1,26 @@
...
+namespace vnni {
+namespace utils {
+
+Optional<int64_t> getVNNIBlockingFactor(Type type) {
+  if (!type.cast<ShapedType>().getElementType().isBF16())
+    return std::nullopt;
+  return 2;
+}
+bool isBF16Type(Type type) {
+  return type.cast<ShapedType>().getElementType().isBF16();
```
chelini, line 22: this crashes if the type is not ShapedType.
``` 
if (!isa<ShapedType>(type))
return false
```

lib/TPP/VNNIUtils.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+
+#include "TPP/VNNIUtils.h"
+
+namespace mlir {
+namespace vnni {
+namespace utils {
+
+Optional<int64_t> getVNNIBlockingFactor(Type type) {
+  if (!type.cast<ShapedType>().getElementType().isBF16())
```
chelini, line 17: this crashes if the type is not ShapedType.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -383,47 +385,101 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
...
+  AffineMap mapB =
+      AffineMap::get(/*dims=*/7, /*symbols=*/0,
+                     {p2, r1, r2.floorDiv(*blockingFactor), p4, r3}, ctx);
+  AffineMap mapC =
+      AffineMap::get(/*dims=*/7, /*symbols=*/0, {p1, p2, p3, p4}, ctx);
+  Value matrixC = matmulOp.getOutputs()[0];
+  linalg::GenericOp replacementOp = rewriter.create<linalg::GenericOp>(
+      loc, matrixC.getType(), packedInputs, ValueRange{matrixC},
+      ArrayRef<AffineMap>{mapA, mapB, mapC},
+      ArrayRef<mlir::utils::IteratorType>{mlir::utils::IteratorType::parallel,
```
chelini, line 453: please drop `mlir::`

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -383,47 +385,101 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
...
 
-  OpFoldResult tileOnI = tiles[0];
-  SmallVector<OpFoldResult, 1> tilesOnB = {tileOnI};
+  if (!isMatmulOp(matmulOp))
+    return rewriter.notifyMatchFailure(matmulOp, "require matmul semantics");
+
+  if (isVNNIPacked(matmulOp))
+    return rewriter.notifyMatchFailure(matmulOp, "already packed to VNNI");
 
+  assert(matmulOp.getInputs()[0]
```
chelini, line 421: can you use the util method here?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -383,47 +385,101 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
...
+}
+
+FailureOr<linalg::GenericOp>
 mlir::linalgx::packVNNIMatmulOp(RewriterBase &rewriter,
-                                linalg::MatmulOp matmulOp,
-                                ArrayRef<OpFoldResult> tiles) {
-  if (tiles.size() != 1)
-    return rewriter.notifyMatchFailure(matmulOp, "require 1 blocking factor");
+                                linalg::GenericOp matmulOp) {
+  if (matmulOp.getInputs().size() > 0 && !matmulOp.getInputs()[0]
```
chelini, line 402: can you use the util method here?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -366,10 +366,12 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
   linalg::GenericOp replacementOp = rewriter.create<linalg::GenericOp>(
       loc, packMatrixC.getType(), packedInputs, ValueRange{packMatrixC},
       ArrayRef<AffineMap>{mapA, mapB, mapC},
-      ArrayRef<utils::IteratorType>{
-          utils::IteratorType::parallel, utils::IteratorType::parallel,
-          utils::IteratorType::reduction, utils::IteratorType::parallel,
-          utils::IteratorType::parallel, utils::IteratorType::reduction},
+      ArrayRef<mlir::utils::IteratorType>{mlir::utils::IteratorType::parallel,
```
chelini, line 369: please drop mlir::

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -139,19 +241,26 @@ mlir::linalgx::mapToBRGEMMOp(RewriterBase &rewriter,
...
         linalgOp, "failed to match BRGEMM access patterns");
 
-  if (failed(checkBody(linalgOp)))
-    return rewriter.notifyMatchFailure(linalgOp, "expects a GEMM-like body");
-
-  // Materialize outer loops.
+  // Materialize outer loops
   unsigned upTo = linalgOp.getNumLoops() - /*BRGEMM loops=*/4;
+  // VNNI loop
+  if (isVNNILoop) {
```
chelini, line 261: please drop braces. Single stmt should not have braces.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -126,6 +206,28 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
   return slicedOperands;
 }
 
+// Returns true if linalg op is using its second operand in VNNI format.
+static bool isLinalgOpVNNI(linalg::LinalgOp linalgOp) {
+  // Linalg loop must have 7 dimensions.
+  if (linalgOp.getNumLoops() != 7)
+    return false;
+  if (linalgOp->getOperands().size() < 3)
```
chelini, line 214: linalgOp->getOperands().size() -> linalgOp->getNumOperands()

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -126,6 +206,28 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
...
+    return false;
+  auto firstOperand = linalgOp->getOperands()[1];
+  auto firstOperandType = firstOperand.getType();
+  auto blockingFactor = vnni::utils::getVNNIBlockingFactor(firstOperandType);
+  if (!blockingFactor)
+    // Unsupported blocking factor for type.
+    return false;
+  if (vnni::utils::isBF16Type(firstOperandType) &&
```
chelini, line 222: avoid return after if. Just return vnni::utils::isBF16Type(firstOperandType) ...

include/TPP/VNNIUtils.h
```diff
@@ -0,0 +1,27 @@
...
+Optional<int64_t> getVNNIBlockingFactor(Type type);
+
+// Checks if the type is BF16.
```
chelini, line 20: // Checks if the type is BF16. -> // Return true if the type is BF16.

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -59,6 +85,51 @@ static bool isOutputOperand(linalg::LinalgOp linalgOp, OpOperand &operand) {
   return !isInputOperand(linalgOp, operand);
 }
 
+static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
+  SmallVector<AffineMap> maps;
+  for (OpOperand &operand : linalgOp->getOpOperands()) {
```
chelini, line 90: Suggest to replace with:
```
for (OpOperand *operand : linalgOp.getInputOperands()) {
   AffineMap map = linalgOp.getMatchingIndexingMap(operand);
   if (map.getNumResults() < 3) return failure();
   if (operand.getOperandNumber() == 1) {
        if (map.getNumResults() != 5)
          return failure();
        maps.push_back(map.getMinorSubMap(4));
      } else {
        maps.push_back(map.getMinorSubMap(3));
      }
}
```

lib/TPP/MapToBatchReduceGEMM.cpp
```diff
@@ -59,6 +85,51 @@ static bool isOutputOperand(linalg::LinalgOp linalgOp, OpOperand &operand) {
...
+      if (operand.getOperandNumber() == 1) {
+        if (map.getNumResults() != 5)
+          return failure();
+        maps.push_back(map.getMinorSubMap(4));
+      } else {
+        maps.push_back(map.getMinorSubMap(3));
+      }
+    }
+  }
+  for (OpOperand &operand : linalgOp->getOpOperands()) {
```
chelini, line 104: You can do the same here using `linalgOp.getDpsInitOperands`. This basically avoid you checking if an operand is the input or the output.

KavithaTipturMadhu:
@chelini another round of review?

include/TPP/VNNIUtils.h
```diff
@@ -0,0 +1,28 @@
+#ifndef TPP_VNNIUTILS_H
+#define TPP_VNNIUTILS_H
+
+//===- VNNIUtils.cpp ---------------------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
chelini, line 10: nit new line before include.

lib/TPP/VNNIUtils.cpp
```diff
@@ -0,0 +1,28 @@
...
+namespace vnni {
+namespace utils {
+
+Optional<int64_t> getVNNIBlockingFactor(Type type) {
+  if (!isa<ShapedType>(type) ||
+      !type.cast<ShapedType>().getElementType().isBF16())
+    return std::nullopt;
+  return 2;
+}
+bool isBF16Type(Type type) {
```
chelini, line 21: new line before `bool isBF16Type(Type type) {`

chelini (approved):
Thank you! Looks good to me

KavithaTipturMadhu:
@rengolin do you have any comments on this?

PR #264, opened 2023-01-11
Retire map-linalg-to-tpp pass

chelini:
Originally conversion to tpp was happening as a two-stages approach: mapping and conversion based on annotate ops. Time proved this approach fragile, and now convert-linalg-to-tpp can do both tasks.


adam-smnk (approved):
lgtm - from my experience, `convert-linalg-to-tpp` covers everything

rengolin (approved):

PR #265, opened 2023-01-11
MobileNet v2 model with and without BN

### LINKED ISSUE #254: Add MobileNet end-to-end model as a model
Import TensorFlow MobileNet model into Linalg MLIR and ensure that the MLIR code can be optimized using `tpp-opt` and can run using `tpp-run`. Add appropriate CHECKs to verify TPP lowerings, XSMM calls, and also correctness of output.

nhasabni:
Does not have CHECKs in place yet .. Starting a PR to get feedback/comments. We do not seem to lower 3x3 Conv2D with stride 2 and depthwise Conv2D into TPP and XSMM calls yet.

Fixes #254


test/Models/mobilenet-without-batchnorm.mlir
```diff
@@ -0,0 +1,900 @@
+// RUN: tpp-opt %s -decompose-conv-to-matmul-or-brgemm -empty-tensor-to-alloc-tensor \
```
rengolin, line 1: Try to use the default pipeline option. We should be moving to that for all integration tests soon enough.

rengolin (commented):
We don't yet support strides in convolution, I just created a new issue to work on that (#266). I've also added an issue to work on depth-wise convolution (#267).

Since we're using batch = 1, we don't really need batch norm, so we could have a single one, the one without bath norm.

Thanks again for such detailed output, it'll help a lot in understanding what all that means.

Just to be clear, there is nothing at all in this model that gets converted into XSMM calls?

test/Models/mobilenet-without-batchnorm.mlir
adam-smnk: FYI, at the moment `-expand-strided-metadata -lower-affine` have to be added separately after the default pipeline.
I'll add these two passes to `tpp-run` soon.

rengolin:
@nhasabni #269 should make a big difference here.

nhasabni:
> @nhasabni #269 should make a big difference here.

@rengolin @chelini Confirming that #269 lowers strided Conv2D into xsmm matmul calls. We are only left with lowering depthwise Conv2D in this model now.

rengolin:
> @rengolin @chelini Confirming that #269 lowers strided Conv2D into xsmm matmul calls. We are only left with lowering depthwise Conv2D in this model now.

Great, thanks! Just need to remove the batchnorm version and add CHECK lines for the strided convolutions, and this is good to go. Thanks!

rengolin (approved):
Thanks!

PR #268, opened 2023-01-12
Use the default pipeline for Resnet50

adam-smnk:
Adds an extra test run with the default pipeline to the ResNet50 model test.
The original manually built lowering pipeline is left for comparison and ease of future experimentation with different passes.
ResNet50 integration test is moved fully to the default pipeline.

Small changes to the `CHECK`s are needed due to: LICM in the default pipeline which wasn't present in the original manual pipeline and use of parallel loops for linalg to tpp conversion is enabled.

### DIFF
--- a/test/Benchmarks/resnet50_bottleneck.mlir
+++ b/test/Benchmarks/resnet50_bottleneck.mlir
@@ -1,15 +1,3 @@
-// RUN: tpp-opt %s -decompose-conv-to-matmul-or-brgemm -empty-tensor-to-alloc-tensor \
-// RUN: -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" \
-// RUN: -drop-equivalent-buffer-results -finalizing-bufferize \
-// RUN: -buffer-results-to-out-params -buffer-deallocation -canonicalize \
-// RUN: -convert-linalg-to-tpp="use-parallel-loops=false" \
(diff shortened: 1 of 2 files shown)


rengolin:
About the leak, comparing the default pass, I don't see `-buffer-results-to-out-params`.

adam-smnk:
CI detected memory leaks after switching benchmark to the default pipeline. To be investigated.

adam-smnk:
> About the leak, comparing the default pass, I don't see `-buffer-results-to-out-params`.

Good point, it'd make sense. I'll add a small test for it.

adam-smnk:
> > About the leak, comparing the default pass, I don't see `-buffer-results-to-out-params`.
> 
> Good point, it'd make sense. I'll add a small test for it.

That worked but I quickly realized changing func signatures by default will be annoying for normal Model testing.
It might be better to add `-buffer-results-to-out-params` to the `tpp-run` backend.

adam-smnk:
@rengolin I reverted `tpp-run` changes as they didn't work too well by default. I haven't investigated it much deeper after that. I'll have to look into the default backend pipeline separately.
Plus as discussed, it is still unclear whether these out-params are desired by default.

For now, out-params bufferization will remain as a explicit opt-in that has to be added to benchmarks on per-need basis.
I'll create a follow-up ticket to investigate where to include these passes in the default lowering or if it is even needed at all.

PR #269, opened 2023-01-12
Add tests for strided convs

### LINKED ISSUE #266: Support convolution with stride != 1 to GEMM
For MobileNet we need convolution with stride = 2, and currently that's not even getting mapped to a GEMM, let alone a BRGEMM.

Since convolutions are complicated things for BRGEMM, we can try first with GEMM, to at least get some TPP (and perhaps fuse some stuff, tile, etc). Later we can look at BRGEMM, together with filters greater than 1x1, but not in this issue.

chelini:
The logic was there but no tests were available.

Fixes #266


rengolin (approved):
Awesome! This should make the Mobilenet case work, right?

Don't forget to update the commit message to fix the right issue.

chelini:
> Awesome! This should make the Mobilenet case work, right?
> 
> Don't forget to update the commit message to fix the right issue.

yes, strided convs can now be converted to matmul. The only missing piece is to detect also packed convolutions and feed them to the same flow. I expect it to work just need to write the matchers in `decomposeConvToMatmulOrBrgemm`. Will add the fix line.

PR #270, opened 2023-01-12
Benchmark using CMake, C++ classes

### LINKED ISSUE #199: Factor benchmark C++ code boilerplate from kernels
Right now we have a `utils.h` header that implements some basic functionality, but there's still a lot of boilerplate in the tests themselves because of the different arguments for the kernel functions.

We should common that up and pass the kernel as a lambda or functor and use templates to simplify the calling, so that we don't lose time on indirect dispatches.

### LINKED ISSUE #251: Move benchmark compilation to CMake
The benchmark driver is complex and needs information from CMake (so it looks at the build and test directory). This could be simplified by moving it to CMake, and having a target `ninja benchmark` to build and run it.

This would not only simplify the building process, and move it to the right place, but also allow for the compilation command line to be included in `compile_commands.txt` and help editors to parse the files correctly.

The reason why it hasn't been done so is because the same file is compiled multiple times, because some options are determined at compile-time, so to make this work, we probably need to finish #199 first, and make the matrix sizes run-time options? Or make it easier to have very small `main` wrappers for a parametrized kernel call, so we can have one C file per benchmark?

rengolin:
This refactors the benchmarks to simplify adding new code, commoning up most logic in simple classes like Bench, Tensor and Config.

It also simplifies the build and execution of the benchmarks by building the C++ benchmarks via CMake and only calling the binaries via the driver.

There's a ninja benchmark target that calls the script in the right way, depending on the three benchmarks, so you can just call `ninja benchmark` and it will rebuild the C++ benchmarks and the MLIR tools (if necessary) and run the benchmark correctly.

Fixes #199
Fixes #251


rengolin:
Once this is merged, I'll change the CI to just call `ninja benchmark`. Right now we're calling the script directly, which works in the same way, so no funny state in between.

benchmarks/README.md
```diff
@@ -5,24 +5,35 @@ This allows us to measure how much off are we when compared with ninja-written c
...
+
+The C++ code represent what the compiler _should_ be doing if it can get the transforms right.
+The MLIR code should be within 95% of the performance from C++ runs.
+
+## How to Build
+
+The C++ benchmakrs are build by CMake and the binaries will be available in the build directory.
+The sources are in `benchmarks/CPPHarness`, each directory under `src` is a separate benchmark.
+
+The MLIR benchmarks use the `MLIRHarness`, which in turn uses `tpp-run` to JIT-compile the MLIR files in `test/Benchmarks` using the default pass pipeline.
```
adam-smnk, line 36: Are these two `*Harness` paths outdated or maybe just not added to git?

benchmarks/README.md
```diff
@@ -5,24 +5,35 @@ This allows us to measure how much off are we when compared with ninja-written c
...
+These are the ninja optimized code, either by hand in C++ calling libxsmm directly, or by the Tensor Compiler, generating the calls automatically.
 The blocking/tiling/fusing parameters are in optimal configuration and should be the fastest runs.
-They also represent what the compiler _should_ be doing if it can get the transforms right.
+
+The C++ code represent what the compiler _should_ be doing if it can get the transforms right.
+The MLIR code should be within 95% of the performance from C++ runs.
+
+## How to Build
+
+The C++ benchmakrs are build by CMake and the binaries will be available in the build directory.
```
adam-smnk, line 33: The C++ benchmakrs are build by CMake and the binaries will be available in the build directory.
```suggestion
The C++ benchmarks are build by CMake and the binaries will be available in the build directory.
```

benchmarks/README.md
```diff
@@ -36,47 +47,46 @@ This is what the CI does.
 This will run both C++ and MLIR versions and will print out the results in order.
 The output is semi-formatted, human readable and machine parseable, and you can use that to track timings over time.
 
+You can run it from the `build` directory via `ninja benchmarks`, or you can use the Python script directly for more controll.
```
adam-smnk, line 50:
```suggestion
You can run it from the `build` directory via `ninja benchmarks`, or you can use the Python script directly for more control.
```

benchmarks/README.md
```diff
@@ -36,47 +47,46 @@ This is what the CI does.
...
-The compilation is assuming `clang` is in the path and picking the other flags from the repository.
-You can get those options from a `./driver.py -vv` run.
-
-Once compiled, the binary is generated on the same directory as the source and you can pass certain arguments to it, for example:
+The binaries generated on the build directory accept certain arguments to it, for example:
 * `-x`: Runs the XSMM version (if available), not the reference one.
-* `-n`: Changes the number of iterations to run
+* `-n N`: Changes the number of iterations to run.
+* `-m MxNxK`: Sets the required shape for the tensors for simpler benchmarks.
+* `-r`: Sets the input to be random. If ommited, inputs are constant `all_ones`.
```
adam-smnk, line 66:
```suggestion
* `-r`: Sets the input to be random. If omitted, inputs are constant `all_ones`.
```

benchmarks/benchmarks.json
```diff
@@ -3,69 +3,59 @@
...
-      "flags": ""
+      "benchmark": "bench_simple_copy",
+      "flags": []
     },
     "mlir": {
       "type": "MLIR",
-      "source": "../test/Benchmarks/simple_copy.mlir",
-      "iters": "1000",
-      "flags": ""
+      "benchmark": "simple_copy.mlir",
```
adam-smnk, line 11: Should these include file extension `.mlir`?

benchmarks/README.md
rengolin: Outdated, I'll change that.

benchmarks/include/Config.h
```diff
@@ -0,0 +1,106 @@
...
+      llvm::cl::desc("Emit results in GFLOP/s (default is ms)"),
+      llvm::cl::init(false)};
+};
+} // namespace
+
+/// Argument parser and configuration for benchmarks. Sets the variables to the
+/// defaults, overridden from the command line and potential internal logic
+/// later on.
+///
+/// Defauls overriden:
```
chelini, line 41: nit: Defaults

benchmarks/include/Config.h
```diff
@@ -0,0 +1,106 @@
+#pragma once
```
chelini, line 1: can we follow LLVM style here?

benchmarks/include/Config.h
```diff
@@ -0,0 +1,106 @@
...
+      dims.insert(dims.begin(), m, m+found);
+
+    // Override iter, if zero
+    iter = options.iter;
+    if (!iter) {
+      // For an "average size" matrix multiply, this should be enough
+      iter = defaultIter;
+      if (dims.size() != 0) {
+        // Divide the "average size" by the actual size to get the ratio
+        float size = std::accumulate(dims.begin(), dims.end(), 1.0, std::multiplies<float>());
```
chelini, line 94: is it clang-formatted?

benchmarks/benchmarks.json
rengolin: I think so. This is to show (to developers adding benchmarks) that the benchmark in question is an MLIR file. Given that I have removed the path from the config, it's a bit more clear this way.

chelini (commented):
some small comments.

benchmarks/matmul/matmul.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  if (config.verbose) {
+    std::cerr << "Kernel version: " << ((config.xsmm) ? "xsmm" : "ref") << std::endl;
+    std::cerr << "[ " << m << ", " << n << " ] = "
+              << "[ " << m << ", " << k << " ] * "
+              << "[ " << k << ", " << n << " ] X " << config.iter << std::endl;
+  }
+
+  // Init benchmark (TODO: support BF16)
+  double gflops = (double)(2*n*m*k) / 1e9;
```
adam-smnk, line 96:
```suggestion
  double gflops = static_cast<double>(2*n*m*k) / 1e9;
```

benchmarks/mlp/mlp.cpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+  if (config.verbose) {
+    std::cerr << "Kernel version: " << ((config.xsmm) ? "xsmm" : "ref") << std::endl;
+    std::cerr << "[ " << m << ", " << n << " ] = "
+              << "[ " << m << ", " << k << " ] * "
+              << "[ " << k << ", " << n << " ] X " << config.iter << std::endl;
+  }
+
+  // Init benchmark (TODO: support BF16)
+  double gflops = (double)((n*m) + (2*n*m*k) + (n*m)) / 1e9;
```
adam-smnk, line 83:
```suggestion
  double gflops = static_cast<double>((n*m) + (2*n*m*k) + (n*m)) / 1e9;
```

benchmarks/simple_copy/simple_copy.cpp
```diff
@@ -0,0 +1,72 @@
...
+    m = config.dims[0];
+    n = config.dims[1];
+  }
+  if (config.verbose) {
+    std::cerr << "Kernel version: " << ((config.xsmm) ? "xsmm" : "ref") << std::endl;
+    std::cerr << "[ " << m << ", " << n << " ] X " << config.iter << std::endl;
+  }
+
+  // Init benchmark (TODO: support BF16)
+  double gflops = (double)(n*m)/1e9;
```
adam-smnk, line 44:
```suggestion
  double gflops = static_cast<double>(n*m)/1e9;
```

benchmarks/include/Config.h
rengolin: I don't think we need to, because this will never touch LLVM code, we'll never want to upstream this code, and `#pragma once` is superior.

With the `ifndef` trick, it's all too common to copy and paste and forget to change, and the compiler won't complain until you include both headers on the same file.

adam-smnk (approved):
Nice +1 - the new kernel interface approach makes things pretty clear.

rengolin:
> Nice +1 - the new kernel interface approach makes things pretty clear.

I wanted to common up the running of benchmarks, too, as it's basically the same thing, but because there are different arguments and checks, and because that'd be yet another layer of indirection, I decided to leave it for later.

My objective was to have a class that implements the kernel interface and a main that creates a benchmark with some arguments and has one method: run(). Once we have more functionality there we'll be able to see what's really common and simplify it further.

PR #272, opened 2023-01-12
Benchmarks memory leak fix

adam-smnk:
Adds missing memory deallocation in the benchmarks Tensor helper copy assignment operator.

For completeness, all benchmarks now perform conversion of memref function results to out-params and deallocation cleanup to prevent potential memory leaks.

Work towards #234


adam-smnk:
Still leaky, I'll have to reproduce it offline.

rengolin (approved):
D'oh! I tried with vectors first, and pointers were easier, but were an afterthought. Thanks!

Issue #256, opened 2023-01-06
Deprecate MLIR benchmark harness in favour of `tpp-run` + passes + discovery


rengolin:
The [harness](https://github.com/plaidml/tpp-mlir/tree/main/benchmarks/harness) was a stop-gap to remove the previous benchmark programs, but it adds complexity, where we don't need one. The main issue is that we have C++ benchmarks that use C++ argument handling, output, etc and `tpp-run` is in C++, so we could use the same infrastructure for both without needing Python at all.

This is related to #199, #251 and #234.

The main issues we need to resolve here are:
* Common argument parsing, so that both C++ and MLIR benchmarks (via `tpp-run`) have the same args
* Encode the XSMM and MLIR library paths into `tpp-run` (via `rpath`) from CMake directly
* Common timing-to-flops calculation, with number of FLOPs coming from different places (C++ variable, MLIR annotation)

The harness' execution is basically just a call to `tpp-run` and parsing the return value, which should be done at the `driver.py` level, if we make sure the outputs are the same from either C++ or MLIR benchmarks.

@alheinecke FYI

rengolin:
We basically have solved all of those issues, we can now do this work. However, the one thing we will miss is the ability to run `tpp-opt` before `tpp-run` in case we have some special magic that `tpp-run` can't handle.

Long term we should not have such cases, but this may not be true today.

@adam-smnk any opinion on this?

adam-smnk:
Our benchmarking needs have evolved since this ticket was created. Currently, we need to run various binaries to both establish baseline performance (`xsmm_dnn_mlp`) as well as create various kernels (`mlp-gen`) to evaluate our compiler. With the current setup, we could even do sth like `python --flags tensorflow_model.py` and pipe resulting MLIR IR through tpp-mlir all within our config+harness stack.
I do not see need to change anything in the benchmarking infrastructure right now.

Perhaps some aspects could be simplified or the logic could be moved from python scripts to other places. However, current setup works fine, provides decent generic platform to execute whatever we need, and it is not too hard to extend it.
Imo, `tpp-run` is better off as a dedicated MLIR runner and we should have external tooling to handle other stuff like IR generation, running ninja code implementation etc.

Overall, I would leave things as they are unless you have something different in mind or there is burning need to move to a different benchmarking stack.

rengolin:
Agreed.

PR #274, opened 2023-01-12
Add libraries to tpp-run to avoid `--shared-libs` arg

rengolin:
This facilitates the execution of our tests with tpp-run, that we can't do with mlir-cpu-runner, by making it simpler to execute MLIR files without havin to pass a build-time option that has the location of the MLIR and TPP libraries.

This reduces the need for extra complexity in the harness.

Works towards #256


adam-smnk:
FYI, the solution works for me locally on WSL Ubuntu
There must be some hidden assumption on the build system or some config that breaks for CI (and potentially other users), no clue where though

rengolin:
> FYI, the solution works for me locally on WSL Ubuntu There must be some hidden assumption on the build system or some config that breaks for CI (and potentially other users), no clue where though

I think I know what it is... The build dir isn't accessible from all places. I'll check.

rengolin:
@adam-smnk fixed the path issue, thanks for testing!

adam-smnk (approved):

Issue #275, opened 2023-01-13
F32 matmul results invalid


KavithaTipturMadhu:
```  func.func @entry()->tensor<128x512xf32>{
  %c1 = arith.constant 1.0 : f32
  %c0 = arith.constant 0.0 : f32
  %arg0Val = tensor.empty():tensor<128x256xf32>
  %arg0 = linalg.fill ins(%c1:f32) outs(%arg0Val:tensor<128x256xf32>)-> tensor<128x256xf32>
  %arg1Val = tensor.empty():tensor<256x512xf32>
  %arg1 = linalg.fill ins(%c1:f32) outs(%arg1Val:tensor<256x512xf32>)-> tensor<256x512xf32>
  %1 = tensor.empty(): tensor<128x512xf32>
  %2 = linalg.matmul ins(%arg0, %arg1: tensor<128x256xf32>, tensor<256x512xf32>) outs(%1: tensor<128x512xf32>) -> tensor<128x512xf32>
  %constant = arith.constant 256.0:f32
  %buf = tensor.empty():tensor<128x512xf32>
  %constantBuf = linalg.fill ins(%constant:f32) outs(%buf:tensor<128x512xf32>)->tensor<128x512xf32>
  //check.expect_almost_eq(%2, %constantBuf, %constant):tensor<128x512xf32>, tensor<128x512xf32>, f32
  return %2 : tensor<128x512xf32>
}
```
I compiled the above mlir using the flags  ``` -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts``` When run using tpp-run, the printed results show nans and invalid results instead of 256 everywhere as expected.

chelini:
Can you please create a test case?

KavithaTipturMadhu:
Test case created in https://github.com/plaidml/tpp-mlir/pull/276

KavithaTipturMadhu:
[...]

PR #276, opened 2023-01-13
Gemm end to end test f32 datatype

KavithaTipturMadhu:
Test case for issue https://github.com/plaidml/tpp-mlir/issues/275

### DIFF
--- a/test/Integration/gemm.mlir
+++ b/test/Integration/gemm.mlir
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts |\
+// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize -buffer-results-to-out-params -buffer-deallocation  --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts |\
 // RUN: tpp-run \
 // RUN:  -e entry -entry-point-result=void  \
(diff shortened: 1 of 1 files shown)


KavithaTipturMadhu:
Looks like there's a memory leak in the test which is why the CI is failing. I checked locally where the test was failing because the results mismatched with the expected result.

test/Integration/gemm.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts |\
```
adam-smnk, line 1:
```suggestion
// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize -buffer-results-to-out-params -buffer-deallocation --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts |\
```

adam-smnk (commented):
These two extra flags should address the leak.

test/Integration/gemm.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts |\
+// RUN: tpp-opt %s -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs function-boundary-type-conversion=identity-layout-map" -drop-equivalent-buffer-results -finalizing-bufferize -buffer-results-to-out-params -buffer-deallocation  --linalg-ext-to-loops --convert-check-to-loops --convert-vnni-to-tpp --convert-linalg-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine  -arith-expand -reconcile-unrealized-casts |\
```
rengolin, line 1: How does that compare with the default pipeline? @adam-smnk

adam-smnk: It looks pretty much the same.
`-default-tpp-passes -buffer-results-to-out-params -buffer-deallocation -lower-affine -arith-expand` should also work.

rengolin: We probably should move all integration tests to the default pipeline soon

PR #283, opened 2023-01-18
Introduce a pass to rewrite scf.for to scf.parallel

chelini:
`scf.parallel` does not work at the tensor level. To maintain parallelism at the tensor level, we have the following choices: `scf.foreach_theread` or looking at gml dialect in mlir-hlo. Preliminary discussion seems that we don't like either option: `scf.foreach_theread` does not have a lowering, and it is too 'thread' like, and it is unclear how to lower it to OMP, while gml requires adding mlir-hlo as a submodule. If we exclude these two options, we should either extend upstream `scf.parallel` on tensor or raise back and `scf.for` to `scf.parallel` after bufferization. This PR implements the latter option, which is wrong, but it is the path with less friction.

We only raise loop marked as 'parallel'; other passes like tile-consumer-and-fuse-producers are responsible for injecting the attribute when a loop with parallel semantics is materialized.

### DIFF
--- a/include/TPP/Passes.h
+++ b/include/TPP/Passes.h
@@ -88,6 +88,7 @@ createDecomposeConvToMatmulOrBrgemmPass();
 std::unique_ptr<OperationPass<ModuleOp>> createDefaultTppPass();
 std::unique_ptr<OperationPass<func::FuncOp>>
 createGeneralizeTensorPackAndUnPackPass();
+std::unique_ptr<OperationPass<func::FuncOp>> createRaiseToParallelLoopPass();
 
 } // namespace tpp
 } // namespace mlir
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -246,4 +246,13 @@ def GeneralizeTensorPackAndUnPack : Pass<"generalize-tensor-pack-unpack",
   let constructor = "mlir::tpp::createGeneralizeTensorPackAndUnPackPass()";
 }
 
+def RaiseToParallelLoop : Pass<"raise-to-parallel-loop",
+                               "func::FuncOp"> {
+  let summary = "Rewrite scf.for to scf.parallel.";
+  let description = [{
+
+  }]; 
+  let constructor = "mlir::tpp::createRaiseToParallelLoopPass()";
+}
+
 #endif // TPP_DIALECT_TPP_PASSES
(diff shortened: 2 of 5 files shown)


rengolin:
Shouldn't we also merge multiple loops marked as parallel?

chelini:
(never tried but should work).

rengolin:
> We can use: https://github.com/llvm/llvm-project/blob/f81151edbfa3d65acb9f84d6ef214da60495556a/mlir/include/mlir/Dialect/SCF/Transforms/Passes.td#L48

I thought there was one already, thanks for the pointer. @adam-smnk, this sounds like a nice default option for the default pipeline at some cleanup points.

adam-smnk:
> > We can use: https://github.com/llvm/llvm-project/blob/f81151edbfa3d65acb9f84d6ef214da60495556a/mlir/include/mlir/Dialect/SCF/Transforms/Passes.td#L48
> 
> I thought there was one already, thanks for the pointer. @adam-smnk, this sounds like a nice default option for the default pipeline at some cleanup points.

Definitely, I'll add both passes to the default pipeline with extra tests.

rengolin:
I've done some experimentation and it doesn't seem to work straight off the bat.

The pass only fuses parallel loops if they have no side effects (which `store` does), but if I change it to have arguments and use `yield` to return the sum, the pass added in this PR doesn't convert the `for` loops into `parallel` any more.

I think an issue is in order to get there, otherwise we may not get the result we want (and the one @KavithaTipturMadhu got on her tests) automatically, and this change loses a lot of the utility.

rengolin:
Look at my [parallel branch](https://github.com/rengolin/tpp-mlir/tree/parallel) for my experiments.

chelini:
`-scf-parallel-loop-collapse` does not exist in my LLVM/MLIR. I have `scf-parallel-loop-collapsing` but it crashes :)

chelini:
I think it is also misleading. It collapsed indices not loop.

rengolin:
Right, I was kinda shooting in the dark. Don't take that example too seriously.

chelini:
But what would be the expectation to squash the two loops together? and why is needed?

rengolin:
I'm hoping that we can do things [like this](https://github.com/plaidml/tpp-mlir/blob/main/test/Conversion/LinalgToTpp/linalg-to-tpp.mlir#L79) automatically from two `scf.for` loops, and that we're able to use OMP [collapse](https://www.openmp.org/spec-html/5.2/openmpsu30.html) if the destination accumulation tensor is parallelisable at that level.

The main difference of the existing pass is that we **can** collapse two `scf.parallel` even having side effects, if we know that the side effects affect non-overlapping areas of memory, which we can obtain by looking at the iteration space of the accumulation (output) matrix.

chelini:
I think scf.parallel are always squash-able.

PR #285, opened 2023-01-19
Allow test kernels to return results

adam-smnk:
Adds automatic memory deallocation to `tpp-run` to allow test kernels to pass test outputs through function return without memory leaks.
This change then allows us to stop using `-buffer-results-to-out-params` for more predictable kernel APIs and removes need for non-default passes. Extra manual memory management is added when needed.

Work towards #234

### DIFF
--- a/test/Integration/result-out-arg.mlir
+++ b/test/Integration/result-out-arg.mlir
@@ -1,0 +1,49 @@
+// RUN: tpp-opt %s -default-tpp-passes | \
+// RUN: tpp-run -print -n 10 \
+// RUN:  -e entry -entry-point-result=void | \
+// RUN: FileCheck %s
+
+#map0 = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+
+func.func @matmultpp(%A: memref<2x2xf32>,
+          %B: memref<2x2xf32>, %C: memref<2x2xf32>) attributes {llvm.emit_c_interface} {
+  linalg.generic {indexing_maps = [#map0, #map1, #map2],
+                         iterator_types = ["parallel", "parallel", "reduction"]}
+    ins(%A, %B: memref<2x2xf32>, memref<2x2xf32>) outs(%C: memref<2x2xf32>) {
+        ^bb0(%a: f32, %b: f32, %c: f32):
+          %0 = arith.mulf %a, %b : f32
+          %1 = arith.addf %c, %0 : f32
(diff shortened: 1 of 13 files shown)


test/Integration/result-out-arg.mlir
```diff
@@ -0,0 +1,51 @@
+// RUN: tpp-opt %s -default-tpp-passes | \
+// RUN: tpp-run -print -n 10 \
+// RUN:  -e entry -entry-point-result=void  \
+// RUN:  -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext | \
```
rengolin, line 4: We don't need the shared-libs option anymore, right?

tpp-run/MLIRBench.cpp
```diff
@@ -222,6 +226,13 @@ void MLIRBench::printVector(Value vector) {
 }
 
 LogicalResult MLIRBench::printMemRef(mlir::Value memRef) {
+  OpBuilder::InsertionGuard guard(builder);
+
+  // Build print logic directly after the kernel return value
+  // before the output gets potentially deallocated.
+  if (isa<func::CallOp>(memRef.getDefiningOp()))
```
rengolin, line 233: While this works it's a little surprising after reading the function above and expecting the deallocation to happen just after the call, unless there's a print, etc. It can cause confusion on developers, for example, assuming the deallocation immediately after could optimize things, or end up exposing undefined behaviour.

I'd keep the `printMemref` as it is and move the deallocation to some new `cleanup` method that is called at the end, either by hand on the main file or from the lowering call (that also adds a return).

rengolin (commented):
Some comments. Also, the `setInsertionPoint` business is fragile as is, side effect of moving from a monolithic function. We really need to start using `InsertionGuard`s, but that's for another PR.

tpp-run/MLIRBench.cpp
adam-smnk: With the previous approach of passing results through an output argument it was possible to have a single reusable buffer and just clean it up afterwards.
But now if we use function return for kernel output then each kernel call allocates memory internally. Which means that we have plenty of allocation in the main benchmarking loop so we need to clean it up immediately anyway.

I probably could pick the insertion point for `printMemRef` in a bit nicer way though.

adam-smnk: I rearranged the printing logic a bit to make it cleaner and decouple it from the explicit assumption that there is any dealloc.

rengolin (approved):
It's still confusing but that's not this PR's fault, the original code needs some re-thinking now that we have more information on what we want to do. For another PR, this one looks good, thanks!

PR #288, opened 2023-01-20
Fix xsmm tests

### LINKED ISSUE #286: Invalid XSMM results compared to loops
Three integration tests `relayout-gemm.mlir`, `relayout-more-interesting.mlir`, and `smoke.mlir` report different test kernel results depending whether TPP is lowered to XSMM or loops.
In all three cases, lowering to loops (`-default-tpp-passes="tpp-to-loops"`) produces valid expected results, while lowering to XSMM (`-default-tpp-passes`) gives quite different and incorrect results.

Quick comparison of IR produced by the default pipeline is the two configs shows that the general structure of these kernels i.e., allocs, outer loops and their IVs, affine maps, general order of operations etc., appears to be the same. The main difference is presence of either inner loops or XSMM calls.

The issue can be reproduced by running:
* correct:
```
tpp-opt -default-tpp-passes="tpp-to-loops" -expand-strided-metadata -lower-affine <test>.mlir | tpp-run -e entry -entry-point-result=void
```
* invalid:
```
tpp-opt -default-tpp-passes -expand-strided-metadata -lower-affine <test>.mlir | tpp-run -e entry -entry-point-result=void
```

adam-smnk:
Fixes #286 

Fixes:
- converts `smoke.mlir` to `f32` as our runtime doesn't currently handle `f64` dispatch
- restricts XSMM ops from `AnyFloat` to `f32` and `bf16` to avoid future surprises like with the `smoke.mlir` test
  - this should be reverted once we decide to have full LIBXSMM type support
- fixes ldi computation for `tpp.identity` to address `relayout-*.mlir` tests issues

### DIFF
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -13,9 +13,9 @@ include "XsmmDialect.td"
 include "XsmmAttr.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
 
-def XsmmMemRef : AnyTypeOf<[MemRefRankOf<[AnyFloat], [1, 2, 3, 4]>, AnyFloat, I64]>;
-def Xsmm2DMemRef : AnyTypeOf<[MemRefRankOf<[AnyFloat], [2]>]>;
-def Xsmm4DMemRef : AnyTypeOf<[MemRefRankOf<[AnyFloat], [4]>]>;
+def XsmmMemRef : AnyTypeOf<[MemRefRankOf<[F32, BF16], [1, 2, 3, 4]>, F32, BF16, I64]>;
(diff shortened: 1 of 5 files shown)


rengolin:
Did the `smoke.mlir` test worked with f64 after the fix? If so, then we may not need to restrict to f32/16 only.

As Alex said, libxsmm does support f64 for most ops.

adam-smnk:
> Did the `smoke.mlir` test worked with f64 after the fix? If so, then we may not need to restrict to f32/16 only.
> 
> As Alex said, libxsmm does support f64 for most ops.

Nope, the conversion fix addresses only the `relayout` test errors.
Our lack of `f64` runtime is related to some other logic but, as far as I'm aware, it was a deliberate choice to not actively support it at the moment.

rengolin (approved):
Ah, yes, the runtime. I forgot about that. Fair enough. LGTM, thanks!

PR #293, opened 2023-01-23
Decouple convolutions rewrite pass from packing

chelini:
We already have specific passes for packing. The rewrite pass should not pack; instead, it should pick up a packed convolution. Add relevant tests.


lib/TPP/RewriteConvsToMatmulOrBrgemm.cpp
```diff
@@ -240,15 +240,14 @@ struct BlockConv2DNchwFchw : OpRewritePattern<linalg::Conv2DNchwFchwOp> {
...
-// 'tpp.BlockedConv2DNchwFchwOp' to expose a GEMM operation.
+// Interchange a blocked convolutions to expose a linalg.matmul.
 struct InterchangeIteratorsConv2DNchwFchw
     : OpRewritePattern<linalg::GenericOp> {
   using OpRewritePattern::OpRewritePattern;
 
   LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
                                 PatternRewriter &rewriter) const override {
-    if (!tpp::utils::isMarkedWithTpp(genericOp, "tpp.BlockedConv2DNchwFchwOp"))
+    if (!linalgx::utils::isBlockedConvolution(genericOp))
```
rengolin, line 250: Why is this in `linalgx` namespace? We want to remove that, not add more functionality there.

You could just add that to the `linalg` namespace, and "extend" it that way. No need for an `x` at the end.

But only if we really want to upstream that, or if there's no TPP logic in it. Otherwise, it should be local and in some tpp namespace (as it was before).

lib/TPP/TransformUtils.cpp
```diff
@@ -184,6 +185,26 @@ FailureOr<SmallVector<Range>> getLoopsToMaterialize(RewriterBase &rewriter,
   return loopRanges;
 }
 
+bool isBlockedConvolution(linalg::LinalgOp linalgOp) {
+  auto iteratorTypes = linalgOp.getIteratorTypesArray();
+  if (iteratorTypes.size() != 9)
+    return false;
+  bool match = linalg::isParallelIterator(iteratorTypes[0]) &&
```
adam-smnk, line 192: nit: you could probably replace this check with `std::all_of` for better readability

test/Passes/pass-conv-to-matmul.mlir
```diff
@@ -1,4 +1,8 @@
-// RUN: tpp-opt %s -rewrite-conv-to-matmul-or-brgemm -canonicalize -split-input-file | FileCheck %s
+// RUN: tpp-opt %s -rewrite-conv-to-matmul-or-brgemm -canonicalize -split-input-file | FileCheck -check-prefix=UNBLCKHWCF %s
```
rengolin, line 1: I'm guessing each RUN line matches only one of the tests, so it would be good to check that the other two are not matched and still have the original operation.

lib/TPP/RewriteConvsToMatmulOrBrgemm.cpp
chelini: I feel linalgx is a better namespace than tpp because these methods works at linalg. I prefer not to pollute linalg namespace.

lib/TPP/TransformUtils.cpp
chelini: I check for parallel and reduction. why std::all_of?

lib/TPP/RewriteConvsToMatmulOrBrgemm.cpp
chelini: We map both blocked and non-blocked convs. So the first run also map to linalg even if it does not block the conv.

test/Passes/pass-conv-to-matmul.mlir
chelini: We map both blocked and non-blocked convs. So the first run also map to linalg even if it does not block the conv.

lib/TPP/TransformUtils.cpp
adam-smnk: Just personal opinion 2 lines vs 9 - if you think it doesn't matter, feel free to ignore this.

adam-smnk (approved):
Looks fine to me.

chelini:
@rengolin and @adam-smnk any further comments?

lib/TPP/RewriteConvsToMatmulOrBrgemm.cpp
rengolin: I don't like the namespace but it doesn't matter much. If we want this to be linalg-only, then perhaps we should upstream those helpers?

test/Passes/pass-conv-to-matmul.mlir
rengolin: Would be nice to have a CHECK line for each run on each example, to make explicit what we're expecting here.

chelini: Made some changes. I just want to check that we are able to map a blocked conv to matmul. I updated the test this way.

lib/TPP/RewriteConvsToMatmulOrBrgemm.cpp
chelini: Let's keep them in-house a little longer. Don't think are relevant for upstream at the moment.

test/Passes/pass-conv-to-matmul.mlir
rengolin: As it is, this are three separate tests. Even though the other RUN lines pass on the other (non-CHECKed) tests, we have no way of knowing if their execution is correct, as there are no CHECK lines. Keeping them like this gives the wrong impression that we know it works.

My point is simple: either add all CHECKs to all splits, or split this into three separate tests.

chelini: I understand your point. There are not more three lines of RUN, just a single one.

rengolin: I'm sorry, Github showed me the old file still. What a crap interface!

rengolin (approved):

PR #296, opened 2023-01-27
tpp-run default pipeline

### LINKED ISSUE #234: Create a pre-pass in `tpp-run` with the default pipeline

adam-smnk:
Integrates the default tpp passes and affine lowering directly into tpp-run.

Benchmarks now use only tpp-run without any extra preprocessing.
The integration tests are expanded with additional test runs that validate default pipeline using only tpp-run. tpp-opt is still used but only to accommodate test-specific preprocessing cases when necessary.
These changes allows tpp-run to operate as a standalone binary which can accept IR at any level, automatically lowers it to LLVM, and finally executes it.
Additionally, benchmarking infrastructure is updated to handle input kernels operating on tensors.

Resolves #234


test/Integration/conv-to-matmul.mlir
```diff
@@ -15,6 +15,15 @@
 // RUN: FileCheck %s -check-prefix=TRANSFORM
 //
 
+// Validate default pipeline
+// RUN: tpp-run %s -print \
+// RUN:  -e entry -entry-point-result=void | \
+// RUN: FileCheck %s -check-prefix=LINALG
+
+// RUN: tpp-run %s -print \
```
rengolin, line 23: Not clear to me what's the difference here. The original seems to be between with and without `-rewrite-conv-to-matmul-or-brgemm`, but the default pipeline will not skip that.

I don't know if the original intention can be mapped to lower-to-loops {true, false}.

adam-smnk:
@rengolin I left `tpp-opt` handling in the benchmarks harness as it still might be useful in the future. For example, in case we need to add some test-specific, non-default preprocessing or just for ease of debugging.

test/Integration/conv-to-matmul.mlir
```diff
@@ -195,5 +204,10 @@ func.func @entry() {
   // TRANSFORM-SAME:   ( 0, 28, 56, 84, 112, 140, 168, 196 ) ) ) )
   //
   vector.print %v2 : vector<1x2x2x8xf32>
+
+  bufferization.dealloc_tensor %img_seed : tensor<?xf32>
```
chelini, line 208: is this not inserted by bufferization?

chelini (approved):

test/Integration/conv-to-matmul.mlir
adam-smnk: Nope, there are memory leaks without this explicit deallocation.

rengolin (commented):
Looking good. 

Initial comments on the tests...
1. They're adding `tpp-run`, not replacing the existing RUN lines. This is ok as a transient state, but ultimately, we don't want any `tpp-opt` in `Integration` tests. Those should be smaller, IR tests.
2. There seems to be a `LINALG` vs `TRANSFORM` check on some tests that only check for the BRGEMM pass, so I'm guessing this can be mapped to the new option "tpp-to-loops", which results need to be the same. Actual transform tests seem correct, though.
3. Some Integration tests don't check output, so we have no idea if they're actually passing (correct output) or just successfully running.

Will look at the code later.

test/Integration/conv-to-matmul.mlir
rengolin: Yeah, the IR itself allocates them, so it should also deallocate.

chelini: I did not see any leaks here for example: https://github.com/plaidml/tpp-mlir/commit/464116e4517d06db4d1a6bae0c83f664c4ebaa3d

adam-smnk: Although, now that I double-checked it, the leaks are present only when the default pipeline is used. The older `RUN` statements execute correctly. I'll need to have a look where this difference is coming from.

test/Integration/conv-to-matmul.mlir
adam-smnk: Good catch, in this case it doesn't matter for `tpp-run` as this conv rewrite is always enabled in the default pipeline.
One of these runs can be removed.

Regarding your comment 2) - I'll have to double check them, in some cases there are difference with an extra `tpp-opt` preprocessing but there might be a few that could be cleaned up.
About comment 3) - the tests that don't print/return outputs rely only on `check` dialect validation. But to be sure, I'll check again if that's truly the case in all of them.

test/Integration/conv-to-matmul.mlir
adam-smnk: The generated LLVM IR (manual pipeline vs tpp-run) looks to be mostly the same - identical in the number of allocs and frees. I think that our `mlir-cpu-runner` just doesn't include any sanitizers (I tried building LLVM with sanitizers but ran into some errors) whereas they are enabled for `tpp-run` which then reports the memory leaks.

@chelini one-shot or `BufferDeallocationPass` cannot handle cases when an explicitly allocated buffer is returned from a function. In the example you linked it is fine because the alloc is within the main function.

rengolin: You can build `mlir-cpu-runner` with santizers (may need some pushing, may have temporary upstream bugs on the commit we choose, etc), but you have to build the entire LLVM+MLIR. 

It's doable, but using `tpp-run` directly is much easier and more controllable.

rengolin (approved):

adam-smnk:
Cleanup some tests - mostly removed duplicated custom `CHECK` cases and added a note related to #298.
Also, I checked that all integration tests that don't use `-print` contain `check` dialect test cases.

PR #297, opened 2023-01-30
Relax `SameTypeOperands` constraint for `tpp.add`

chelini:
Requiring all operands to have the same type is too restrictive, and we may lose mapping opportunities. For example, one operand can be a strided memref like `memref<56x32xf32, strided<[32, 1], offset: ?>>` while another, a non-strided one `memref<56x32xf32>`. The types of the memref are different, but we still want to allow both in `tpp.add`. Relax the operand check and now require each operand to have the same *shape* as we do not support broadcast yet.

### DIFF
--- a/lib/TPP/Dialect/Tpp/TppUtils.cpp
+++ b/lib/TPP/Dialect/Tpp/TppUtils.cpp
@@ -305,27 +305,43 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
 // strided<[32, 1], offset: ?>> B: memref<56x32xf32> allOperandsHaveSameShape(A,
 // B) return true. C: memref<1x32xf32> allOperandsHaveSameShape(A, C) return
 // false.
-static bool allOperandsHaveSameShape(linalg::GenericOp linalgOp) {
-  if (!hasStaticShape(linalgOp))
+static bool allOperandsHaveSameShapeAndStrides(linalg::GenericOp linalgOp) {
+  auto operandType = linalgOp->getOperand(0).getType();
+  if (!operandType.isa<MemRefType>())
     return false;
-  Type operandType = linalgOp->getOperand(0).getType();
-  if (!operandType.isa<ShapedType>())
-    return false;
-  // Validate rank.
-  int64_t rankOperand = operandType.cast<ShapedType>().getRank();
+  auto firstOperandType = operandType.cast<MemRefType>();
+
+  // Step1. Validate rank.
+  int64_t rankOperand = firstOperandType.getRank();
   for (Value operand : linalgOp->getOperands()) {
     Type currentType = operand.getType();
-    if (!currentType.isa<ShapedType>())
+    if (!currentType.isa<MemRefType>())
       return false;
-    if (currentType.cast<ShapedType>().getRank() != rankOperand)
+    if (currentType.cast<MemRefType>().getRank() != rankOperand)
       return false;
   }
-  // Validate shape.
-  ArrayRef<int64_t> shapeOperand = operandType.cast<ShapedType>().getShape();
+
+  // Step2. Validate shape, and strides.
+  // Get stride and offset for the first operand.
+  int64_t offsetFirstOperand = 0;
+  SmallVector<int64_t> stridesFirstOperand;
+  if (failed(getStridesAndOffset(firstOperandType, stridesFirstOperand,
+                                 offsetFirstOperand)))
+    return false;
+  ArrayRef<int64_t> shapeOperand = firstOperandType.getShape();
   for (Value operand : linalgOp->getOperands()) {
-    ArrayRef<int64_t> shapeCurrentOperand =
-        operand.getType().cast<ShapedType>().getShape();
+    // Compare the shape.
+    auto currentOperandType = operand.getType().cast<MemRefType>();
+    ArrayRef<int64_t> shapeCurrentOperand = currentOperandType.getShape();
     if (shapeCurrentOperand != shapeOperand)
+      return false;
+    int64_t offsetCurrentOperand = 0;
+    SmallVector<int64_t> stridesCurrentOperand;
+    // Compare the strides.
+    if (failed(getStridesAndOffset(currentOperandType, stridesCurrentOperand,
+                                   offsetCurrentOperand)))
+      return false;
+    if (stridesFirstOperand != stridesCurrentOperand)
       return false;
   }
   return true;
@@ -337,7 +353,7 @@ bool isTppAdd(linalg::GenericOp linalgOp) {
     return false;
   if (!isBinaryOp(linalgOp))
     return false;
-  if (!allOperandsHaveSameShape(linalgOp))
+  if (!allOperandsHaveSameShapeAndStrides(linalgOp))
     return false;
   return allIndexingsAreProjectedPermutation(linalgOp) &&
          hasOnlyOp<arith::AddFOp>(linalgOp.getRegion());
--- a/lib/TPP/Dialect/Tpp/TppUtils.cpp
+++ b/lib/TPP/Dialect/Tpp/TppUtils.cpp
@@ -300,21 +300,15 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
   return false;
 }
 
-// Returns true if all the operands of the current linalgOp are shape type and
-// have the same statically-known shape. For example: A: memref<56x32xf32,
-// strided<[32, 1], offset: ?>> B: memref<56x32xf32> allOperandsHaveSameShape(A,
-// B) return true. C: memref<1x32xf32> allOperandsHaveSameShape(A, C) return
-// false.
-static bool allOperandsHaveSameShapeAndStrides(linalg::GenericOp linalgOp) {
-  auto operandType = linalgOp->getOperand(0).getType();
-  if (!operandType.isa<MemRefType>())
+bool allOperandsHaveSameShapeAndStrides(TypeRange types) {
+  assert(types.size() > 1 && "expect one or more types");
+  if (!types[0].isa<MemRefType>())
     return false;
-  auto firstOperandType = operandType.cast<MemRefType>();
+  auto firstOperandType = types[0].cast<MemRefType>();
 
   // Step1. Validate rank.
   int64_t rankOperand = firstOperandType.getRank();
-  for (Value operand : linalgOp->getOperands()) {
-    Type currentType = operand.getType();
+  for (Type currentType : types) {
     if (!currentType.isa<MemRefType>())
       return false;
     if (currentType.cast<MemRefType>().getRank() != rankOperand)
@@ -328,23 +322,31 @@ static bool allOperandsHaveSameShapeAndStrides(linalg::GenericOp linalgOp) {
   if (failed(getStridesAndOffset(firstOperandType, stridesFirstOperand,
                                  offsetFirstOperand)))
     return false;
-  ArrayRef<int64_t> shapeOperand = firstOperandType.getShape();
-  for (Value operand : linalgOp->getOperands()) {
+  ArrayRef<int64_t> shapeFirstOperand = firstOperandType.getShape();
+  for (Type currentOperandType : types) {
     // Compare the shape.
-    auto currentOperandType = operand.getType().cast<MemRefType>();
-    ArrayRef<int64_t> shapeCurrentOperand = currentOperandType.getShape();
-    if (shapeCurrentOperand != shapeOperand)
+    ArrayRef<int64_t> shapeCurrentOperand =
+        currentOperandType.cast<MemRefType>().getShape();
+    if (shapeCurrentOperand != shapeFirstOperand)
       return false;
     int64_t offsetCurrentOperand = 0;
     SmallVector<int64_t> stridesCurrentOperand;
     // Compare the strides.
-    if (failed(getStridesAndOffset(currentOperandType, stridesCurrentOperand,
+    if (failed(getStridesAndOffset(currentOperandType.cast<MemRefType>(),
+                                   stridesCurrentOperand,
                                    offsetCurrentOperand)))
       return false;
     if (stridesFirstOperand != stridesCurrentOperand)
       return false;
   }
   return true;
+}
+
+static bool allOperandsHaveSameShapeAndStrides(linalg::GenericOp linalgOp) {
+  SmallVector<Type> types;
+  for (Value operand : linalgOp->getOperands())
+    types.push_back(operand.getType());
+  return allOperandsHaveSameShapeAndStrides(types);
 }
 
 // Return true if the linalg.generic can be mapped to a tpp.add.
--- a/lib/TPP/Dialect/Tpp/TppUtils.cpp
+++ b/lib/TPP/Dialect/Tpp/TppUtils.cpp
@@ -300,12 +300,32 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
   return false;
 }
 
-// Return true if the operands are all the same.
-static bool allOperandsHaveSameType(linalg::GenericOp linalgOp) {
+// Returns true if all the operands of the current linalgOp are shape type and
+// have the same statically-known shape. For example: A: memref<56x32xf32,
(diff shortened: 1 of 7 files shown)


lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -349,12 +349,32 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
...
 }
 
-// Return true if the operands are all the same.
-static bool allOperandsHaveSameType(linalg::GenericOp linalgOp) {
+// Returns true if all the operands of the current linalgOp are shape type and
+// have the same statically-known shape. For example: A: memref<56x32xf32,
+// strided<[32, 1], offset: ?>> B: memref<56x32xf32> allOperandsHaveSameShape(A,
+// B) return true. C: memref<1x32xf32> allOperandsHaveSameShape(A, C) return
+// false.
+static bool allOperandsHaveSameShape(linalg::GenericOp linalgOp) {
```
adam-smnk, line 357: Is it still valid when strides are dynamic?

rengolin: Or different/dynamic offsets?

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,87 @@
+// RUN: tpp-opt %s -convert-linalg-to-tpp -convert-linalg-to-loops -convert-tpp-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -canonicalize -reconcile-unrealized-casts | \
+// RUN: mlir-cpu-runner \
```
rengolin, line 2: Can you try to use `tpp-run` instead? We have more checks than `mlir-cpu-runner`.

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,87 @@
...
+      ins(%arg0, %arg1 : memref<56x32xf32>, memref<56x32xf32>)
+      outs(%subview : memref<56x32xf32, strided<[32, 1], offset: ?>>) {
+        ^bb0(%in: f32, %in_0: f32, %out: f32):
+          %0 = arith.addf %in, %in_0 : f32
+          linalg.yield %0 : f32
+      }
+    }
+  }
+  %to_print = memref.cast %arg2 : memref<1x2x56x56x32xf32> to memref<*xf32>
+  call @printMemrefF32(%to_print) : (memref<*xf32>) -> ()
```
rengolin, line 83: No CHECK lines for output?

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,87 @@
...
+// RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%llvmlibdir/libmlir_runner_utils%shlibext &> %s.dump.linalg.loops.mlir
+//
+
+// RUN: tpp-opt %s -convert-linalg-to-loops -convert-tpp-to-loops -convert-tpp-to-xsmm -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -convert-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -canonicalize -reconcile-unrealized-casts | \
+// RUN: mlir-cpu-runner \
+// RUN:  -e entry -entry-point-result=void  \
+// RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%llvmlibdir/libmlir_runner_utils%shlibext &> %s.dump.tpp.xsmm.mlir
+//
+
+// RUN: diff --strip-trailing-cr -I 'Unranked Memref base' %s.dump.tpp.loops.mlir %s.dump.linalg.loops.mlir
```
rengolin, line 19: This isn't a good idea. If the test crashes, it leaves output all over the place.

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: The offset in the example is dynamic.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -334,12 +334,32 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
...
+      return false;
+    if (currentType.cast<ShapedType>().getRank() != rankOperand)
+      return false;
+  }
+  // Validate shape.
+  ArrayRef<int64_t> shapeOperand = operandType.cast<ShapedType>().getShape();
+  for (Value operand : linalgOp->getOperands()) {
+    ArrayRef<int64_t> shapeCurrentOperand =
+        operand.getType().cast<ShapedType>().getShape();
+    if (shapeCurrentOperand != shapeOperand)
```
rengolin, line 362: Looking at #331, this might not be enough, if the shape is the same but the stride is incompatible.

chelini: why? I cannot think of an example. what do you have in mind?

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,81 @@
...
+    scf.for %arg4 = %c0 to %c56 step %c1 {
+      %subview = memref.subview %arg2 [0, %arg3, %arg4, 0, 0] [1, 1, 1, 56, 32] [1, 1, 1, 1, 1] 
+        : memref<1x2x56x56x32xf32> to memref<56x32xf32, strided<[32, 1], offset: ?>>
+      // CHECK: tpp.add
+      linalg.generic {
+      indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>, 
+                       affine_map<(d0, d1) -> (d0, d1)>], 
+      iterator_types = ["parallel", "parallel"]}
+      ins(%arg0, %arg1 : memref<56x32xf32>, memref<56x32xf32>)
+      outs(%subview : memref<56x32xf32, strided<[32, 1], offset: ?>>) {
```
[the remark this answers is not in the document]
chelini: Same here. Memref are dense.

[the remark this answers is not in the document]
rengolin: Sorry, by "dense" I mean "the stride reflects the actual shape" (as described in #331).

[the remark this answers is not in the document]
chelini: Yeah, this is a good idea. We can check strides along each dimensions and make sure they are equal, instead of checking the shape, and rank. Will fix tomorrow

test/Integration/tpp-add.mlir
chelini: Getting: Main function has to be 'void', even if the kernel return's a value, because that's the type of the wrapper we create here

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -334,24 +334,67 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
   return false;
 }
 
-// Return true if the operands are all the same.
-static bool allOperandsHaveSameType(linalg::GenericOp linalgOp) {
-  Type operandType = linalgOp->getOperand(0).getType();
-  for (Value operand : linalgOp->getOperands()) {
-    Type currentType = operand.getType();
-    if (currentType != operandType)
+// Returns true if all the operands of the current linalgOp are shape type and
```
rengolin, line 337: Nit: this function really isn't specific to `linalgOp`. The one below is.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -334,24 +334,67 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
...
+  }
+
+  // Step2. Validate shape, and strides.
+  // Get stride and offset for the first operand.
+  int64_t offsetFirstOperand = 0;
+  SmallVector<int64_t> stridesFirstOperand;
+  if (failed(getStridesAndOffset(firstOperandType, stridesFirstOperand,
+                                 offsetFirstOperand)))
+    return false;
+  ArrayRef<int64_t> shapeOperand = firstOperandType.getShape();
```
rengolin, line 364: Nit: `shapeFirstOperand`?

Also, perhaps end with `Op` instead, to reduce line length?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -334,24 +334,67 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
...
+    if (shapeCurrentOperand != shapeOperand)
+      return false;
+    int64_t offsetCurrentOperand = 0;
+    SmallVector<int64_t> stridesCurrentOperand;
+    // Compare the strides.
+    if (failed(getStridesAndOffset(currentOperandType.cast<MemRefType>(),
+                                   stridesCurrentOperand,
+                                   offsetCurrentOperand)))
+      return false;
+    if (stridesFirstOperand != stridesCurrentOperand)
```
rengolin, line 378: Shouldn't we also care about offsets?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -334,24 +334,67 @@ static bool isBinaryOp(linalg::GenericOp linalgOp) {
...
       return false;
   }
   return true;
 }
 
+static bool allOperandsHaveSameShapeAndStrides(linalg::GenericOp linalgOp) {
+  SmallVector<Type> types;
+  for (Value operand : linalgOp->getOperands())
+    types.push_back(operand.getType());
+  return allOperandsHaveSameShapeAndStrides(types);
```
rengolin, line 388: Can't you just pass `getOperands` to the function and capture as a reference?

You probably wouldn't need the two functions, then.

test/Integration/tpp-add.mlir
rengolin: `tpp-run` can cope with `void main`. Also, please replace the long list of passes with the default pipeline. This could probably just be a single line.

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: I am passing a typeRange, getOperands return a ValueRange.

rengolin: Perhaps a helper from `ValueRange` to `TypeRange` then?

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: But shouldn't all the offsets start at the same place? Otherwise, the TPP call could segfault.

chelini: Not necessary, we take into account the offset when computing the starting pointer for LIBXSMM. It is perfectly fine to iterate at different position in the tensor.

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: What exactly do you want to achieve here? Remove `allOperandsHaveSameShapeAndStrides(linalg::GenericOp linalgOp)`?

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: This is guaranteeing that the shapes and strides are the same, if the offsets are different, there's something wrong.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: For starters, yes. Any such special wrappers pollute the file with similar code that gets hard to maintain later.

This is true for the rest of the code, too. So, as you go through the code, if you find the same pattern of super-specialized functions, try to generalize them and move up to some common place, and as a last resort, TppUtils (we should also move that file to a more generic place, it's not specific to the tpp dialect any more).

rengolin: I think I've seen a helper function that extracts type ranges from value ranges in MLIR upstream...

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: You suggest to compare also the offset?

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: remember the name?

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: Yes, if possible. 

If they're dynamic, I'm not sure what to do because I don't know where this is coming from. If we are creating those, then it gets tricky, but if not, perhaps can we just fail on dynamic offsets?

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: Yes, I updated my previous comment. Turns out is just `ValueRange.getTypes()`.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: Discussed offline, the offset is introduced by the subviews we create, so should be "safe". Once we start adding TPP ops from non-linalg regions we can re-think the approach and start checking the offsets.

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,78 @@
+// RUN: tpp-opt %s -default-tpp-passes="tpp-to-loops" | \
```
rengolin, line 1: You can use `tpp-run -tpp-to-loops` without needing `tpp-opt` for this.

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,78 @@
...
+// RUN: tpp-run -print \
+// RUN:  -e entry -entry-point-result=void | \
+// RUN: FileCheck %s -check-prefix=EXE
+
+// RUN: tpp-opt %s -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -finalize-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -canonicalize -reconcile-unrealized-casts | \
+// RUN: tpp-run -print \
+// RUN:  -e entry -entry-point-result=void | \
+// RUN: FileCheck %s -check-prefix=EXE
+
+// RUN: tpp-opt %s -default-tpp-passes | \
```
rengolin, line 11: You can use `tpp-run` without needing `tpp-opt` for this.

test/Integration/tpp-add.mlir
```diff
@@ -0,0 +1,78 @@
+// RUN: tpp-opt %s -default-tpp-passes="tpp-to-loops" | \
+// RUN: tpp-run -print \
+// RUN:  -e entry -entry-point-result=void | \
+// RUN: FileCheck %s -check-prefix=EXE
+
+// RUN: tpp-opt %s -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -finalize-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -canonicalize -reconcile-unrealized-casts | \
```
rengolin, line 6: Why are you testing this alternative pipeline? Passing LLVM dialect to `tpp-run` doesn't work, and this doesn't test anything additional than the other two tests.

chelini: I want to check with Linalg.

rengolin: Ah, right, then remove the llvm passes and it should work.

Issue #229, opened 2022-12-20
Random Initialization in Benchmarks/Tests


rengolin:
For now, `tpp-run` initializes the input tensors as `dense(1.0)`, which makes it easier to check results at the end but it may not catch numeric issues with the algorithms. We need random initialization, but in order to check outputs, we need to make them predictable and comparable.

The main idea is to have two flags: `-random` to force a random initialization and `-seed N` to force a particular seed for the PRNG.

There are two main ways we can do that:
1. **Compare to golden files**

If we run the kernel unoptimized and save the results to a file (text, protobuf, etc), we can compare the results of the optimized run against it. But then, it won't be truly random. 

We can "randomize" frequently, and save the golden files in the repo, but that pollutes the commit history. We can have that infra in the CI machines without being in Git, but then we lose control of the process, and it's not reproducible.

2. **Each run computes both vanilla and optimized**

If we clone the kernel and only optimize one, we can run the vanilla one for its golden output and then compare that with the optimized version. We only need to run the optimized version in the benchmark loop.

This is more complex to do and it can take a considerable amount of time (optimizations we do usually extract 50x-100x performance), but we can use the same input tensors for both, so no need to keep seeds or golden files.

Due to speed, it would need to be a special flag, so that we only build that on nights or weekends, for example, in the CI.

@KavithaTipturMadhu like we discussed this week.
@hfp Could this be an interesting framework for libxsmm too?

rengolin:
IIUC, [this new attribute](https://github.com/llvm/llvm-project/commit/995ab92964d667123efd90d1f8016602c4a9df01)  allows us to create a constant from a `dense_resource` and mmap a file to that resource ([as expanded here](https://github.com/llvm/llvm-project/commit/18546ff8dd45a81e72c0a2ed0561b5aec8c15ca3#diff-ab2649ffae320add26eff630dd372468f410cc7e4a581da40bacdd0747de64b2)).

But it's not clear what's the completion state of this (doesn't seem to support MMAP yet, at least not writable resources for outputs). We may have to upstream that ourselves if we want to use it.

rengolin:
@chelini pointed me in the direction of [linalg.fill_rng_2d](https://mlir.llvm.org/docs/Dialects/Linalg/#linalgfill_rng_2d-mlirlinalgfillrng2dop), which has operands `seed`, `min` and `max`, which is pretty much what we want.

[...]

PR #302, opened 2023-02-01
Remove linalgx.pack from tpp bf16 tests

KavithaTipturMadhu:
I am removing linalgx.pack from BF16 tests by changing the layout of buffers to packed layouts directly as tpp operates at memref level whereas tensor.pack is at tensor level.  I will introduce new tests to cover packing with vnni operations.

### DIFF
--- a/test/BF16/matmul-pbf16.mlir
+++ b/test/BF16/matmul-pbf16.mlir
@@ -17,11 +17,9 @@ func.func @entry() {
   %f0 = arith.constant 1.0 : bf16
   %da = memref.alloc() :memref<4x8xbf16>
   linalg.fill ins(%f0 : bf16) outs(%da : memref<4x8xbf16>)
-  %db = memref.alloc() :memref<8x4xbf16>
-  linalg.fill ins(%f0:bf16) outs (%db:memref<8x4xbf16>)
   // Call kernel.
   %0 = memref.alloc() : memref<4x4x2xbf16>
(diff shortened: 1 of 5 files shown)


chelini (approved):
Sounds good to me. Please also remove linagx dir in follow-up commits.

rengolin:
It should be fine to have packed memrefs as input for IR checks, but not runtime checks, because half of the job in VNNI is packing correctly.

The tests that we need are:
* Already packed, BRGEMM tests (the ones you have now)
* Only VNNI packed (of smaller memrefs) into correct packed dimensions
* Matmul/Conv pack + VNNI pack into correct packed dimensions

The IR tests should be in `test/Passes` for the VNNI packing passes, using `tpp-opt`. The input initialization is irrelevant.

The execution tests should be in `test/Integration`, using `tpp-run`, and the input initialization should be variable enough to show packing works and CHECKing the output.

One trick I'm using with #229 is to have a smaller pattern which doesn't divide the tensor size (small odd number should work for most cases), and init them with constant values between 0 and 1. For example: `{ 0.3, 0.6, 0.9 }` or `{0.2, 0.4, 0.5, 0.6, 0.8 }`, etc.

This PR should be fine to merge as is if you create those tests soon after.

rengolin:
> Please also remove linagx dir in follow-up commits.

We decided to do that on a separate commit so we can one day "revert" it (even for testing purposes) without also bringing unrelated history.

PR #304, opened 2023-02-01
Default pipeline raise-to-parallel-loop pass

adam-smnk:
Adds `raise-to-parallel-loop` to the default pipeline and a corresponding pass test.
Additionally, adds default pipeline test cases using `tpp-run` to more tests that already use `mlir-cpu-runner`.

### DIFF
--- a/test/BF16/mlp-single-layer-bf16.mlir
+++ b/test/BF16/mlp-single-layer-bf16.mlir
@@ -4,6 +4,10 @@
 // RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext
 //
 
+// Validate default pipeline
+// RUN: tpp-run %s \
+// RUN:  -e entry -entry-point-result=void
+
 func.func @entry(){
   %c0 = arith.constant 1.0:bf16
   %arg0 = memref.alloc():memref<128x256xbf16>
--- a/test/Passes/DefaultPipeline/mixed.mlir
+++ b/test/Passes/DefaultPipeline/mixed.mlir
@@ -40,7 +40,6 @@ module @predict_function  {
     %2 = xsmm.unary.dispatch relu [128, 512, 512, 512](broadcast none dataType f32)
     xsmm.unary relu(dataType f32, %2, %arg3, %arg3) : (i64, memref<128x512xf32>, memref<128x512xf32>) -> ()
 
-    // CHECK: return
     return
   }
 }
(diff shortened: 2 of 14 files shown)


rengolin (commented):
Hopefully we can reuse the same CHECK lines, otherwise, LGTM.

test/Passes/DefaultPipeline/mixed.mlir
```diff
@@ -96,3 +96,22 @@ func.func @buffer_no_dealloc(%A: memref<4x8xf32>,
...
+          %src: memref<32xf32>) {
+  // CHECK-NOT: scf.for
+  // CHECK: scf.parallel
+  scf.for %i = %lb to %ub step %step {
+    %load = memref.load %src[%i] : memref<32xf32>
+    %add = arith.addf %load, %load : f32
+    memref.store %add, %src[%i] : memref<32xf32>
+  } {parallel}
+
+  // CHECK: return
```
chelini, line 115: Why are we checking return?

adam-smnk: Just a copy-paste template. I should probably clean-up these empty returns.

test/BF16/mlp-single-layer-bf16.mlir
```diff
@@ -4,6 +4,10 @@
 // RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext,%tpplibdir/libtpp_c_runner_utils%shlibext
 //
 
+// Validate default pipeline
+// RUN: tpp-run %s -print \
+// RUN:  -e entry -entry-point-result=void
```
rengolin, line 9: These really should have CHECK lines on both, too. I know "it wasn't there before" but it would be easy to add and make sure that the test is really correct. We've been bitten by that recently...

adam-smnk: I reviewed this and other integration tests that don't use print once more. And I still don't see any point in adding CHECKs to these tests. Printing is not practical due to either buffer sizes e.g., ` memref<128x512xbf16>` in this test, or because they would just repeat existing validation using `check` ops.

Maybe I'm misinterpreting your goal here. But if there are any concerns about validation using `check` dialect, I'd rather add extra tests for `check` ops and keep using them whenever possible.

test/Passes/DefaultPipeline/mixed.mlir
adam-smnk: Cleaned-up these checks.

test/BF16/mlp-single-layer-bf16.mlir
rengolin: Argh, sorry, check dialect checks. You're right, it's fine.

rengolin (approved):

PR #306, opened 2023-02-02
Remove Linalgx dialect

### LINKED ISSUE #201: Move our use of `pack`/`unpack` to upstream `tensor`

KavithaTipturMadhu:
Fixes #201


chelini (approved):
Please mention https://github.com/plaidml/tpp-mlir/issues/201 in the commit msg. For me is good to go.

PR #307, opened 2023-02-02
Misc 

chelini:
This is now in MLIR: 6bb0ab0de039ddbcc70eafc549b50da7867fb617

Some CHECK line changes (transform-convolutions.mlir) are due to bubbling up tensor.pack which we did not have before, as we were not relying on upstream.


chelini:
Open a pull request. It is not yet green as I am waiting for this to land (https://reviews.llvm.org/D143087).

rengolin (commented):
Can you push the indentation and whitespace changes in an NFC commit, rebase this PR and just leave the actual changes here? 

It's hard to know what's a real change and what isn't.

chelini:
> Can you push the indentation and whitespace changes in an NFC commit, rebase this PR and just leave the actual changes here?
> 
> It's hard to know what's a real change and what isn't.

Do you want to split Retire PropagateThroughElementWiseOp in multiple PR?

rengolin:
> Do you want to split Retire PropagateThroughElementWiseOp in multiple PR? Or just push rename files and methods as NFC?

There a a few whitespace and indentation changes in the Perf/Check dialect files, mixed with some actual changes. Github isn't good enough to show what actually change in the line, so we have to parse-by-eye and that's prone to errors. There may be others that I didn't get to. cleaning those up means we're only looking at what was intended to change, as part of the PR.

Anything that actually changes something can remain in the PR...

chelini:
> 

Should be better now. Splitted the bump and the retire. Not sure why git still shows commits landed in main. But now each commit should be easy to rev.

rengolin:
> Not sure why git still shows commits landed in main

I don't see them myself... could be web-page refresh issues? Try CTRL+F5.

I'm having some Github formatting problems today, like newlines where none are needed on the file listings... Maybe they're mucking up something that will soon resolve?

rengolin:
Yeah, now that I looked all the changes, it is quite convoluted, a mix bag of many things... it's not easy to review a "Misc" PR.

The warnings fix (`kEmitRawAttributesFolder`) and unnecessary namespace can be NFC direct pushes, NFC renaming probably a separate PR to make sure we all agree.

The LLVM bump should be alongside the one that actually needs, not just some random cleanups, so that it's clear why we're moving. If there's no reason, it's just a move, then it probably should be in a PR of its own, to make the intention clear.

The rest I see the different issues:
* Retire PropagateThroughElementWiseOp
* Move propagation into a pass
* Add pass to fold pack into splat tensor
* Avoid packing linalg.fill

It's not clear to me which ones are related and what is the aim of each.

When creating a PR, please make sure to give it a clear name and enough description to let reviewers know what's the intent, not just the code.

rengolin:
If you separate the NFCs from the rest, and give it a proper description, it should be easier to review as one PR.

PRs depending on others are a mess... :/

chelini:
Closing. I splitted this into multiple smaller PRs

PR #312, opened 2023-02-06
Add pass to fold pack into splat tensor

chelini:
Packing a constant-tensor can be avoided by rewriting the constant. Currently support only splat tensor.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -253,4 +253,12 @@ def PropagatePackUnPack : Pass<"propagate-pack-and-unpack", "func::FuncOp"> {
   let constructor = "mlir::tpp::createPropagatePackUnPackPass()"; 
 }
 
+def ConstantFoldPack : Pass<"constant-fold-pack", "ModuleOp"> {
+  let summary = "Constant fold tensor.pack";
+  let description = [{
+    Reduce pack overhead by folding tensor.pack into constants.
+  }];
+  let constructor = "mlir::tpp::createConstantFoldPackPass()";
+}
+
 #endif // TPP_DIALECT_TPP_PASSES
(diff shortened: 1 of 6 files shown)


include/TPP/Passes.td
```diff
@@ -253,4 +253,12 @@ def PropagatePackUnPack : Pass<"propagate-pack-and-unpack", "func::FuncOp"> {
   let constructor = "mlir::tpp::createPropagatePackUnPackPass()"; 
 }
 
+def ConstantFold : Pass<"constant-fold", "ModuleOp"> {
```
rengolin, line 256: Too generic name, make sure it's clear it's pack specific

lib/TPP/ConstantFold.cpp
```diff
@@ -0,0 +1,51 @@
...
+      if (!cst.isa<DenseElementsAttr>())
+        return;
+      auto oldDense = cast<DenseElementsAttr>(cst);
+      if (!oldDense.isSplat())
+        return;
+      auto newDense = oldDense.reshape(packOp.getDestType());
+      rewriter.setInsertionPoint(cstOp);
+      auto newCstOp =
+          rewriter.create<arith::ConstantOp>(cstOp.getLoc(), newDense);
+      rewriter.replaceOp(packOp, newCstOp.getResult());
```
rengolin, line 42: Does this leave the previous `empty` dangling? Do we need a cleanup to avoid silly allocations?

rengolin (commented):
I'm imagining you're also planning on replacing other types of dense throughout?

lib/TPP/ConstantFold.cpp
chelini: they are dead, so they will be removed.

include/TPP/Passes.td
chelini: Done.

adam-smnk (approved):
Nice - simple but saves a lot of runtime work.
We can keep extending the pass with further patterns later when we find new use cases.

PR #313, opened 2023-02-06
Introduce an interchange pass to expose matmul in blocked convolutions

chelini:
Introduce a specific pass to interchange loops to expose a matmul as the innermost operation. Introducing this pass is *ugly* as we cannot have such a specific pass in our pipeline. We should reevaluate the transform dialect.

### DIFF
--- a/lib/TPP/InterchangeBlockConvToExposeMatmul.cpp
+++ b/lib/TPP/InterchangeBlockConvToExposeMatmul.cpp
@@ -1,0 +1,84 @@
+//===- InterchangeBlockConvToExposeMatmul.cpp --------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Passes.h"
+#include "TPP/TransformUtils.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+// Interchange a blocked convolutions to expose a linalg.matmul.
+struct InterchangeBlockConvToExposeMatmulImpl
+    : OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    if (!linalgx::utils::isBlockedConvolution(genericOp))
+      return failure();
+
+    // clang-format off
+    // N                [parallel]
+    //  K               [parallel - blocked]
+    //    P             [parallel]
+    //      Q           [parallel]
+    //        k         [parallel - block of K]
+    //          C       [reduction - blocked]
+    //            R     [reduction]
+    //              S   [reduction]
+    //                c [reduction - block of C]
+
+    // expose matmul by interchange
+
+    // N                [parallel]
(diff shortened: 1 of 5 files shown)


adam-smnk:
Maybe instead of this being a standalone interchange rewriting pass, it could be an analysis pass that another rewriting/lowering pass uses?
Of course, we can always treat this one as a utility/helper pass and just run it as a part of a different pass, even though, it's definitely questionable design.

lib/TPP/InterchangeBlockConvToExposeMatmul.cpp
```diff
@@ -0,0 +1,84 @@
...
+struct InterchangeBlockConvToExposeMatmulImpl
+    : OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    if (!linalgx::utils::isBlockedConvolution(genericOp))
+      return failure();
+
+    // clang-format off
```
rengolin, line 33: Is this the only form of blocked convolution that `isBlockedConvolution` detects?

lib/TPP/InterchangeBlockConvToExposeMatmul.cpp
```diff
@@ -0,0 +1,84 @@
...
+    //        R         [reduction]
+    //          S       [reduction]
+    //            Q     [parallel]
+    //              k   [parallel - block of K]
+    //                c [reduction - block of C]
+    //
+    // Matmul: m = %Q, n = %k and k = %c
+    // clang-format on
+
+    SmallVector<unsigned> interchangeVector = {0, 1, 2, 5, 6, 7, 3, 4, 8};
```
rengolin, line 59: What if there is a VNNI pack after this pack, does it need to be changed as well?

rengolin (commented):
What is the effect of this pass on fusion of other ops?

Once we add this pass to the default pipeline it may break all sorts of pattern matches if we're not careful.

We should at least have a pass test shows that we can still fuse an element-wise after the repacking.

rengolin:
> Maybe instead of this being a standalone interchange rewriting pass, it could be an analysis pass that another rewriting/lowering pass uses?

Yeah, that's a good point. The points I made about VNNI and fusion are pertinent here. If other packing use these rewrites as stepping stones into their own transforms, we don't need the intermediate state.

chelini:
> Maybe instead of this being a standalone interchange rewriting pass, it could be an analysis pass that another rewriting/lowering pass uses? Of course, we can always treat this one as a utility/helper pass and just run it as a part of a different pass, even though, it's definitely questionable design.

It is an enabler pass, without this rewrite-to-brgemm does not work. I am also not happy about this pass. Where do you suggest to plug it?

chelini:
> What is the effect of this pass on fusion of other ops?
> 
> Once we add this pass to the default pipeline it may break all sorts of pattern matches if we're not careful.
> 
> We should at least have a pass test shows that we can still fuse an element-wise after the repacking.

Where do you see the connection with fusion? This looks for a generic relayout-convolution and interchange to expose matmul operations.

lib/TPP/InterchangeBlockConvToExposeMatmul.cpp
chelini: yes, all the convolutions are blocked the same way.

lib/TPP/InterchangeBlockConvToExposeMatmul.cpp
chelini: I would assume propagation to happen first. But it does not affect the packing, just the interation space.

adam-smnk:
> > Maybe instead of this being a standalone interchange rewriting pass, it could be an analysis pass that another rewriting/lowering pass uses? Of course, we can always treat this one as a utility/helper pass and just run it as a part of a different pass, even though, it's definitely questionable design.
> 
> It is an enabler pass, without this rewrite-to-brgemm does not work. I am also not happy about this pass. Where do you suggest to plug it?

I think in that case this could be part of brgemm rewriter. Unless this interchange has some more value on its own, enables more passes etc.

rengolin:
> Where do you see the connection with fusion? This looks for a generic relayout-convolution and interchange to expose matmul operations.

If the previous/next ops have been packed in the same way to be fused with this convolution, they'd also need to change the iteration space to fuse?

chelini:
Not relevant anymore.

PR #314, opened 2023-02-06
Random fill of inputs in `tpp-run`

rengolin:
This change allows passing a `-seed` to `tpp-run` command line that will randomly initialize the input tensors to allow for better testing. We're using the `linalg.fill_rng_2d` op with values between 0 and 1 to help with floating point accuracy.

Also changed:
 * Added a new option `-tpp-to-loops` that lowers TPP ops to loops. Used to compare results from the same `-seed` between loops and XSMM.
 * Common up constant and dense generation to simplify generic code.

Works towards #229.

This is just the initial part because:
 * Lowering to loops accumulates in a different order and there are small differences (10^-4) between XSMM and loops, which makes it hard to just verify with CHECKs.
 * We can't select loops or xsmm for individual functions, so to use `check.check_almost_equals` we'd have to run tests with constant seeds and save the expected result as a constant dense of the output shape with a close result.

This change at least helps us run the tool with some random input for manual testing, and later find a better strategy for the testing side.

### DIFF
--- a/tpp-run/tpp-run.cpp
+++ b/tpp-run/tpp-run.cpp
@@ -53,14 +53,23 @@ llvm::cl::opt<unsigned>
 // Print result
 llvm::cl::opt<bool> printKernelResult("print",
                                       llvm::cl::desc("Print kernel result"),
-                                      llvm::cl::value_desc("true/false"),
                                       llvm::cl::init(false));
+
+// Lower to loops (for validation purposes)
+llvm::cl::opt<bool> tppToLoops("tpp-to-loops",
+                                      llvm::cl::desc("Lower TPP to loops"),
+                                      llvm::cl::init(false));
+
+// Random seed, if zero, don't emit randominputs
+llvm::cl::opt<int>
+    seed("seed", llvm::cl::desc("Random seed, defualt 0 (no random)"),
+                  llvm::cl::value_desc("int"), llvm::cl::init(0));
 
 // This function will be called by the pass manager after parsing,
 // so we can modify the IR with the needed wrappers
 static LogicalResult prepareMLIRKernel(Operation *op,
                                        JitRunnerOptions &options) {
-  MLIRBench bench(op);
+  MLIRBench bench(op, seed, tppToLoops);
 
   // Basic checks
   if (options.mainFuncType != "void")
@@ -88,7 +97,7 @@ static LogicalResult prepareMLIRKernel(Operation *op,
     return bench.emitError("Cannot create kernel inputs");
 
   // Call kernel once, to bootstrap (JIT compile, warm up caches)
-  auto call = bench.callKernel();
+  auto *call = bench.callKernel();
   if (!call)
     return bench.emitError("Cannot generate a call to the kernel");
 
--- a/tpp-run/tpp-run.cpp
+++ b/tpp-run/tpp-run.cpp
@@ -62,7 +62,7 @@ llvm::cl::opt<bool> tppToLoops("tpp-to-loops",
 
 // Random seed, if zero, don't emit randominputs
 llvm::cl::opt<int>
-    seed("seed", llvm::cl::desc("Random seed, defualt 0 (no random)"),
+    seed("seed", llvm::cl::desc("Random seed, default 0 (no random)"),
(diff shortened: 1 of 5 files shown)


tpp-run/tpp-run.cpp
```diff
@@ -53,14 +53,23 @@ llvm::cl::opt<unsigned>
...
                                       llvm::cl::init(false));
 
+// Lower to loops (for validation purposes)
+llvm::cl::opt<bool> tppToLoops("tpp-to-loops",
+                                      llvm::cl::desc("Lower TPP to loops"),
+                                      llvm::cl::init(false));
+
+// Random seed, if zero, don't emit randominputs
+llvm::cl::opt<int>
+    seed("seed", llvm::cl::desc("Random seed, defualt 0 (no random)"),
```
adam-smnk, line 65:
```suggestion
    seed("seed", llvm::cl::desc("Random seed, default 0 (no random)"),
```

adam-smnk (commented):
I still want to test a bit by hand first but looks good overall.

rengolin:
> I still want to test a bit by hand first but looks good overall.

Good idea. Let me know how it goes.

adam-smnk:
> > I still want to test a bit by hand first but looks good overall.
> 
> Good idea. Let me know how it goes.

A bit randomly but I ran into an issue with running `mlp-single-layer-bf16.mlir`:
- Works:
  - tpp-run default: `./bin/tpp-run -print -e entry -entry-point-result=void ../test/BF16/mlp-single-layer-bf16.mlir`
  - tpp-opt default + tpp-run: `./bin/tpp-opt -default-tpp-passes ../test/BF16/mlp-single-layer-bf16.mlir | ./bin/tpp-run -print -e entry -entry-point-result=void`
  - tpp-opt loops + tpp-run: `./bin/tpp-opt -default-tpp-passes="tpp-to-loops" ../test/BF16/mlp-single-layer-bf16.mlir | ./bin/tpp-run -print -e entry -entry-point-result=void`
  - tpp-opt default + tpp-run loops: `./bin/tpp-opt -default-tpp-passes ../test/BF16/mlp-single-layer-bf16.mlir | ./bin/tpp-run -tpp-to-loops -print -e entry -entry-point-result=void`
- Fails:
  - tpp-run loops: `./bin/tpp-run -tpp-to-loops -print -e entry -entry-point-result=void ../test/BF16/mlp-single-layer-bf16.mlir`
  - tpp-opt loops + tpp-run loops: `./bin/tpp-opt -default-tpp-passes="tpp-to-loops" ../test/BF16/mlp-single-layer-bf16.mlir | ./bin/tpp-run -tpp-to-loops -print -e entry -entry-point-result=void`

The broken cases fail with:
```
loc("<stdin>":10:14): error: failed to legalize operation 'builtin.unrealized_conversion_cast' that was explicitly marked illegal
ERROR: Failed to lower Module to LLVM dialect
```

But I have a feeling that this issue might be unrelated to this PR and there might sth wrong with the pipeline.
I'll try to debug it a bit more.

rengolin:
> A bit randomly but I ran into an issue with running `mlp-single-layer-bf16.mlir`:

Yup, I had the same problems. That was the first one I went to change and basically had the exact same tests and problems you did.

I went further and tried to change `test/Benchmarks/mlp.mlir` and ended up with saturated results for the output. I then tried to allocate the weights by hand, but then I get memory leaks because the tensors get allocated but never deallocated (as we know).

I don't think either problem is directly related to this PR, and it's more about building a coherent and "fail-proof" default pipeline as well as writing effective and relevant tests (it's normally OR, not AND).

My intention with this PR isn't to have that as a feature we rely on for validation, but as something we'll use to probe the current tests and benchmarks and improve the default pipeline, and then we'll start using those in tests. Without this in `main`, we can't really test and it'd be us rebasing branches on each other for the functionality.

adam-smnk (approved):
Looks good and in general works fine too
Pipeline will have to be debugged separately.

adam-smnk:
@rengolin Yeah, I agree.
I also looked a bit more into the IR that gets produced but I still no clue where the code generation flow diverges and leaves some leftover casts. But it should be easier to debug with smaller pipeline stages so I'll return to this issue soon after modularization.

PR #315, opened 2023-02-07
Decouple bf16 from VNNI layout in xsmm

KavithaTipturMadhu:


adam-smnk:
Do we really need an explicit `isVNNI` flag?
It's not necessarily wrong but I'm curious about motivation behind this approach.

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -72,10 +72,20 @@ struct ConvertTppMatmulOp : public OpRewritePattern<tpp::MatmulOp> {
         rewriter.getContext(), ArrayRef<int64_t>{m, n, k, lda, ldb, ldc});
     xsmm::TernaryKindAttr attr = xsmm::TernaryKindAttr::get(
         matmulOp.getContext(), xsmm::TernaryKind::MATMUL);
-    xsmm::DataTypeAttr dtype =
-        xsmm::DataTypeAttr::get(matmulOp.getContext(), xsmm::DataType::F32);
+    xsmm::DataTypeAttr dtype;
```
chelini, line 75: FYI: I have a change coming-in on this. It is just too much code. I pulled into an helper function. Let's land this as is and I will rebase and clean-up.

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -117,11 +117,11 @@ def Xsmm_TernaryDispatchOp : Xsmm_Op<"ternary.dispatch",[Pure]> {
   }];
 
   let arguments = (ins Xsmm_TernaryKind:$kind, DenseI64ArrayAttr:$inputs,
-                       Xsmm_DataType:$dataType);
+                       Xsmm_DataType:$dataType, BoolAttr:$isVNNI);
   let results = (outs I64:$results);
 
   let assemblyFormat = [{
-    $kind $inputs `(` `dataType` $dataType `)` attr-dict
+    $kind $inputs `(` `dataType` $dataType `isVNNI` $isVNNI `)` attr-dict
```
chelini, line 124: I would add a comma in between, like: `(` `dataType` $dataType `,` `isVNNI` $isVNNI `)`. Do we also expect to have this flag for other dispatches?

KavithaTipturMadhu:
> Do we really need an explicit `isVNNI` flag? It's not necessarily wrong but I'm curious about motivation behind this approach.

BF16 doesn't necessarily mean vnni packed. We need the datatype in xsmm runtime to cast the buffer pointers to the right type/invoke libxsmm with the right datatype. On the other hand, we want vnni flag to invoke VNNI operations or normal operations depending on the flag.

include/TPP/Dialect/Xsmm/XsmmOps.td
KavithaTipturMadhu: No, this flag is only for ternary dispatches.

chelini: Ok perfect, I also assume this flag is only valid for BF16? Do we check this?

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -117,11 +117,11 @@ def Xsmm_TernaryDispatchOp : Xsmm_Op<"ternary.dispatch",[Pure]> {
   }];
 
   let arguments = (ins Xsmm_TernaryKind:$kind, DenseI64ArrayAttr:$inputs,
-                       Xsmm_DataType:$dataType);
+                       Xsmm_DataType:$dataType, BoolAttr:$isVNNI);
```
adam-smnk, line 120: Docs probably could use an update with the flag description.

include/TPP/Dialect/Xsmm/XsmmOps.td
KavithaTipturMadhu: we don't, I'll add an assert in the dispatch method for this.

chelini (approved):
I don't have other comments. Thanks. @adam-smnk or @rengolin ?

adam-smnk (approved):
+1 Looks fine

PR #318, opened 2023-02-10
Introduce a pattern to `bubbleup` a fill operation

chelini:
Packing a fill op can be avoided by rewriting the `linalg.fill` to operate on the packed domain.

### DIFF
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -558,6 +558,62 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
 };
 
 //===----------------------------------------------------------------------===//
+// BubbleUpThroughFillOp
+//===----------------------------------------------------------------------===//
+
+// Attempt to avoid packing a fill op. Instead create a 'packed' fill.
+// %0 = tensor.empty
+// %packed = tensor.empty
+// %1 = linalg.fill ins(%cst) outs(%0)
+// %2 = tensor.pack %1 into %packed
+// %3 = some_packed_op %2
+// %4 = tensor.unpack %3 into %1
+// --->
+// %0 = tensor.empty
+// %1 = linalg.fill ins(%cst) outs (%packed)
+// %2 = some_packed_op %1
+// %3 = tensor.unpack %2 into %0
+//
+struct BubbleUpThroughFillOp : public OpRewritePattern<tensor::PackOp> {
+  using OpRewritePattern<tensor::PackOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(tensor::PackOp packOp,
+                                PatternRewriter &rewriter) const override {
+    Value source = packOp.getSource();
+    auto fillOp = source.getDefiningOp<linalg::FillOp>();
+    if (!fillOp)
+      return failure();
+
+    Value fillRes = fillOp.getResult(0);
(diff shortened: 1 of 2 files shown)


lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -557,6 +557,62 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
...
+    Value source = packOp.getSource();
+    auto fillOp = source.getDefiningOp<linalg::FillOp>();
+    if (!fillOp)
+      return failure();
+
+    Value fillRes = fillOp.getResult(0);
+    auto users = fillRes.getUsers();
+
+    // Only two users: a pack and an unpack.
+    if (std::distance(users.begin(), users.end()) != 2)
```
adam-smnk, line 591: I don't think presence of `unpack` (or another user) should be strictly required.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -557,6 +557,62 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
...
+      return failure();
+
+    bool foundPack, foundUnPack = false;
+    for (Operation *op : users) {
+      if (isa_and_nonnull<tensor::UnPackOp>(op))
+        foundUnPack = true;
+      if (isa_and_nonnull<tensor::PackOp>(op))
```
rengolin, line 598: Why would the user of the Op's result be a `pack`? Or rather, if it's actually a `pack`, isn't that the next relayout operation and not this one?

rengolin (commented):
This looks like it could be an upstream patch...

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: Yes, it does. The fillOp must be dead otherwise you introduce re-computation.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: The users of the fill must be a pack or unpack for this patter to apply.

lib/TPP/ToBlockLayoutAndBack.cpp
adam-smnk: Ah I see, I initially thought the idea was to simply fold a `pack` into producing `fill` if there are no other users. But it is a bit more specific pattern.

adam-smnk (approved):

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: yeah, fairly specific.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: Does this answer your question?

rengolin: I understood the code and its intention clearly, I just don't get why it would be a `pack`.

rengolin: Or better still, in which case it would be an `unpack` and why don't you have a test for it?

rengolin: Discussed offline, this looks good!

rengolin (approved):

PR #319, opened 2023-02-10
Enable fusion for convolution-like ops followed by eltwise

chelini:
- Avoid pushing in the worklist already fused ops


chelini:
@KavithaTipturMadhu this is just the start, to get what we discussed there is a bit more. But there are some fixes that should go in before.

KavithaTipturMadhu (approved):

rengolin:
I'd like us to not need to use `main` as a communication device. If you two are working on something, please create a branch that will hold that something while you share incomplete patches on the branch, and then once it's ready for review, rebase+squash+submit a PR.

PR #321, opened 2023-02-10
Improve fusion pass

chelini:


include/TPP/Passes.td
```diff
@@ -197,7 +197,9 @@ def TileConsumerAndFuseProducers : Pass<"tile-consumer-and-fuse-producers",
   }];
   let constructor = "mlir::tpp::createTileConsumerAndFuseProducersPass()";
   let options = [
-    ListOption<"tileSizes", "tile-sizes", "int64_t", "Tile sizes">
+    ListOption<"tileSizes", "tile-sizes", "int64_t", "Tile sizes">,
+    Option<"maxDepth", "max-depth", "int64_t", "5",
```
rengolin, line 201: Why do you need a max-depth and why is this a compiler flag? How does that help with the generic case where the pass goes through the whole IR?

@adam-smnk how does this look with the default pipeline?

chelini: I use the flag to control how many producers I pull in into the fusion. For the pipeline, we can have a greedy strategy.

adam-smnk: In the default pipeline, we could configure it explicitly to have no limit or it could be better to set it to greedy as the default pass option value.

rengolin: Makes sense, thanks!

rengolin:
@adam-smnk @KavithaTipturMadhu please review

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -436,16 +433,30 @@ fuseWithEltwise(RewriterBase &rewriter, TilingInterface consumer,
...
 }
 
 static SmallVector<int64_t> getDefaultTileSizes() { return {32, 32}; }
 
+static FailureOr<Operation *>
+getImmediateElementWiseConsumer(linalg::LinalgOp linalgOp) {
+  assert(linalgOp->getNumResults() == 1);
+  Value res = linalgOp->getResult(0);
+  // If we allow use, we may end up doing recomputation. Unclear if it is
+  // profitablem thus disallow for now.
```
KavithaTipturMadhu, line 451: nit:typo

test/Passes/tile-and-fuse-mlp.mlir
```diff
@@ -0,0 +1,50 @@
...
+// CHECK-DAG: %[[C128:.+]] = arith.constant 128 : index
+// CHECK-DAG: %[[C1:.+]] = arith.constant 1 : index
+// CHECK: %{{.+}} = scf.for %[[I:.+]] = %[[C0]] to %[[C32]] step %[[C1]]
+// CHECK: %{{.+}} = scf.for %[[J:.+]] = %[[C0]] to %[[C128]] step %[[C1]]
+// CHECK: %{{.+}} = linalg.generic
+// CHECK-SAME:  indexing_maps = [#[[MAP]], #[[MAP1]], #[[MAP2]]]
+// CHECK-SAME:  iterator_types = ["reduction", "parallel", "parallel", "reduction"]
+// CHECK: linalg.generic
+// CHECK-SAME:  indexing_maps = [#[[MAP3]], #[[MAP4]], #[[MAP3]]]
+// CHECK-SAME:  iterator_types = ["parallel", "parallel"]
```
KavithaTipturMadhu, line 50: This fuses addf and maxf too, yes? Can we add CHECKs to check the body as well?

KavithaTipturMadhu (approved):

test/Passes/tile-and-fuse-mlp.mlir
chelini: yes, will do.

PR #323, opened 2023-02-10
Forward output of `linalg.fill` to `tensor.unpack`

chelini:
Let's consider the following IR:

```
%e = tensor.empty() : tensor<2x2xf32>
%0 = linalg.fill ins(%cst : f32) outs(%e : tensor<2x2xf32>)
%1 = linalg.matmul ins( ... ) outs(%0 : tensor<2x2xf32>)
```

When blocking the matmul a `tensor.unpack` is inserted right after the blocked matmul. The `tensor.unpack` will unpack the output of the matmul to the orginal un-blocked output (%0) in this case. This is fine becuase %0 will alias %e after bufferization but the IR looks better and some canonicalization patterns are simplified if we unpack directly in %e.

### DIFF
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -117,6 +117,8 @@ static Value toUnPackLayoutImpl(OpBuilder &builder, Location loc, Value input,
                                 Value output, ArrayRef<OpFoldResult> tiles,
                                 ArrayRef<int64_t> innerDimPos,
                                 ArrayRef<int64_t> outerDimsPerm) {
+  if (auto fillOp = output.getDefiningOp<linalg::FillOp>())
+    output = fillOp.getOutputs()[0];
   return builder.create<tensor::UnPackOp>(loc, input, output, innerDimPos,
                                           tiles, outerDimsPerm);
 }
@@ -650,8 +652,9 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
 // %1 = linalg.fill ins(%cst) outs(%0)
 // %2 = tensor.pack %1 into %packed
 // %3 = some_packed_op %2
-// %4 = tensor.unpack %3 into %1
+//
 // --->
+//
 // %0 = tensor.empty
 // %1 = linalg.fill ins(%cst) outs (%packed)
 // %2 = some_packed_op %1
@@ -668,21 +671,7 @@ struct BubbleUpThroughFillOp : public OpRewritePattern<tensor::PackOp> {
       return failure();
 
     Value fillRes = fillOp.getResult(0);
-    auto users = fillRes.getUsers();
-
-    // Only two users: a pack and an unpack.
-    if (std::distance(users.begin(), users.end()) != 2)
-      return failure();
-
-    bool foundPack, foundUnPack = false;
-    for (Operation *op : users) {
-      if (isa_and_nonnull<tensor::UnPackOp>(op))
-        foundUnPack = true;
-      if (isa_and_nonnull<tensor::PackOp>(op))
-        foundPack = true;
-    }
-
-    if (!foundUnPack || !foundPack)
+    if (!fillRes.hasOneUse())
       return failure();
 
     // Replace result with output.
--- a/test/Passes/pack-unpack-propagation.mlir
+++ b/test/Passes/pack-unpack-propagation.mlir
@@ -381,7 +381,7 @@ func.func @fill(%arg0: f32, %arg1: tensor<1x56x56x64xf32>, %arg2: tensor<1x1x64x
       %7 = arith.addf %out, %6 : f32
       linalg.yield %7 : f32
   } -> tensor<1x2x56x56x32xf32>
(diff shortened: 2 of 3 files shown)


test/Passes/pack-unpack-propagation.mlir
```diff
@@ -381,7 +381,7 @@ func.func @fill(%arg0: f32, %arg1: tensor<1x56x56x64xf32>, %arg2: tensor<1x1x64x
       %7 = arith.addf %out, %6 : f32
       linalg.yield %7 : f32
   } -> tensor<1x2x56x56x32xf32>
-  %unpack = tensor.unpack %5 outer_dims_perm = [0, 3, 1, 2] inner_dims_pos = [3] inner_tiles = [32] into %1 : tensor<1x2x56x56x32xf32> -> tensor<1x56x56x64xf32>
+  %unpack = tensor.unpack %5 outer_dims_perm = [0, 3, 1, 2] inner_dims_pos = [3] inner_tiles = [32] into %0 : tensor<1x2x56x56x32xf32> -> tensor<1x56x56x64xf32>
```
rengolin, line 384: Wasn't this IR taken from an actual model?

Does it mean we have to change `%1` to `%0` before this pass?

test/Passes/pass-matmul-blocking.mlir
```diff
@@ -40,10 +40,44 @@ func.func @block_linalg_matmul(
...
+// CHECK: %[[EMPTY_ARG0:.+]] = tensor.empty() : tensor<4x4x32x32xf32>
+// CHECK: %[[PACK_ARG0:.+]] = tensor.pack %[[ARG0]] 
+// CHECK-SAME:  inner_dims_pos = [0, 1] inner_tiles = [32, 32] 
+// CHECK-SAME:  into %[[EMPTY_ARG0]] : tensor<128x128xf32> -> tensor<4x4x32x32xf32>
+// CHECK: %[[EMPTY_ARG1:.+]] = tensor.empty() : tensor<4x4x32x32xf32>
+// CHECK: %[[PACK_ARG1:.+]] = tensor.pack %[[ARG1]] 
+// CHECK-SAME:  outer_dims_perm = [1, 0] inner_dims_pos = [0, 1] inner_tiles = [32, 32] 
+// CHECK-SAME:  into %[[EMPTY_ARG1]] : tensor<128x128xf32> -> tensor<4x4x32x32xf32>
+// CHECK: %[[EMPTY_FILL:.+]] = tensor.empty() : tensor<4x4x32x32xf32>
+// CHECK: %[[PACK_FILL:.+]] = tensor.pack %[[FILL]] inner_dims_pos = [0, 1] inner_tiles = [32, 32]
```
rengolin, line 79: Wasn't the previous change to remove the packing of fills? Why do we have them back now?

chelini: This test checks only the packing not canonicalization patterns, that happen later in the pipeline. Here I want to make sure that the output of the fill is forwarded to the unpack so that the unpack write to the destination of the fill not on the fill itself.

test/Passes/pack-unpack-propagation.mlir
chelini: Yes, it is but we can control how we pack.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -650,8 +652,9 @@ struct PropagateThroughPadOp : public OpRewritePattern<tensor::PadOp> {
 // %1 = linalg.fill ins(%cst) outs(%0)
 // %2 = tensor.pack %1 into %packed
 // %3 = some_packed_op %2
-// %4 = tensor.unpack %3 into %1
+//
```
rengolin, line 655: In the test, the tensor is still unpacking into the empty `%0`, so the comment should probably reflect that

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -668,21 +671,7 @@ struct BubbleUpThroughFillOp : public OpRewritePattern<tensor::PackOp> {
...
-    bool foundPack, foundUnPack = false;
-    for (Operation *op : users) {
-      if (isa_and_nonnull<tensor::UnPackOp>(op))
-        foundUnPack = true;
-      if (isa_and_nonnull<tensor::PackOp>(op))
-        foundPack = true;
-    }
-
-    if (!foundUnPack || !foundPack)
+    if (!fillRes.hasOneUse())
```
rengolin, line 674: I'm not sure that's equivalent to the check that was there before.

Sure, the fill has "more than one use" was encoded before, but also was that those need to be both pack and unpack (at least). Now, IIUC, any two users (even non-pack ones) will trigger this replacement?

chelini: It is not equivalent. Now the fill result must have a single user (tensor.pack).

rengolin: No, I mean, before it was checking for a specific op (pack/unpack), now "any" op will do. What if `fill` has a user that is not `pack`?

chelini: No, it can have only the pack. You get the fillOp by getting the pack operand, thus the result of the fillOp will be the pack.

rengolin: Of course! I need to click to expand the code more often, as Github hides too much of the context.

rengolin (approved):

PR #325, opened 2023-02-13
Reduction dimension moved to correct place

KavithaTipturMadhu:


lib/TPP/RewriteToBatchReduceGEMM.cpp
```diff
@@ -106,20 +107,23 @@ static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
     maps.push_back(map.getMinorSubMap(2));
   }
 
+  llvm::errs() << "compressed dims maps\n";
   SmallVector<AffineMap> compressedDimMaps = compressUnusedDims(maps);
   using MapList = ArrayRef<ArrayRef<AffineExpr>>;
+  for (auto dimMap : compressedDimMaps)
```
chelini, line 113: same here.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -507,27 +510,50 @@ mlir::linalgx::packVNNIMatmulOp(RewriterBase &rewriter,
...
-      ArrayRef<mlir::utils::IteratorType>{mlir::utils::IteratorType::parallel,
-                                          mlir::utils::IteratorType::parallel,
-                                          mlir::utils::IteratorType::reduction,
-                                          mlir::utils::IteratorType::parallel,
-                                          mlir::utils::IteratorType::parallel,
-                                          mlir::utils::IteratorType::reduction,
-                                          mlir::utils::IteratorType::reduction},
-      /*doc=*/"", /*libraryCall=*/"");
+  linalg::GenericOp replacementOp;
+  if (matmulOp.getInputs()[1].getType().cast<ShapedType>().getRank() == 4) {
```
chelini, line 518: why do we have to differentiate here?

chelini (commented):
please also add a test.

lib/TPP/RewriteToBatchReduceGEMM.cpp
```diff
@@ -106,20 +107,23 @@ static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
     maps.push_back(map.getMinorSubMap(2));
   }
 
+  llvm::errs() << "compressed dims maps\n";
```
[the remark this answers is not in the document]
rengolin: you can use the LLVM_DBG macro thing, if this is something that will help development in the future, not just a quick local print

lib/TPP/ToBlockLayoutAndBack.cpp
KavithaTipturMadhu: The maps and iterators to linalg.generic differ for operands of different shapes (i.e., when shape is 4d, generic is 7 dimensional and maps are as per the code, as compared to 3d shape obtained when using tiling factor of 1,1).

chelini: Now I am worried. If we start adding a case for each combination, this does not scale.

chelini: If tiling is a 1,1 it tiles along parallel dimensions and indeed this seems the case looking at the affine map, I don't think we need to differentiate, we should look only at the last [r, r, p, p, r] dimensions.

KavithaTipturMadhu: Added a todo for now.

KavithaTipturMadhu:
Updated PR with a test case for untiled version of matmul @chelini. Review?

chelini (approved):
Thanks, I am fine merging as is to allow making progress. Please add an issue and feel free to assign it to me.

PR #327, opened 2023-02-14
Add `conv-init-simplify` pass

chelini:
Instead of initializing the output of a conv with zeros and then add the bias. Add the bias upfront to the conv output.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -270,4 +270,14 @@ def ElementWiseFusion : Pass<"element-wise-fusion", "func::FuncOp"> {
   let constructor = "mlir::tpp::createElementWiseFusionPass()";
 }
 
+def ConvInitSimplify : Pass<"conv-init-simplify", "func::FuncOp"> {
+  let summary = "Simplify initialization for convolution";
+  let description = [{
+    Perform a graph-rewrite to simplify initialization for a Conv2DNhwcHwcfOp
+    operation. Specifically, instead of initializing the output of a convolution
+    with zero and then adding the bias, initialize the output with the bias.  
+  }];
+  let constructor = "mlir::tpp::createConvInitSimplifyPass()";
+}
+
 #endif // TPP_DIALECT_TPP_PASSES
--- a/lib/TPP/ConvInitSimplify.cpp
+++ b/lib/TPP/ConvInitSimplify.cpp
@@ -1,0 +1,109 @@
+//===- ConvInitSimplify.cpp --------------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Dialect/Tpp/TppUtils.h"
+#include "TPP/Passes.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+static bool isBroadCastOp(linalg::GenericOp linalgOp) {
+  if (linalgOp->getNumOperands() != 2 || linalgOp->getNumResults() != 1)
+    return false;
+  if (!tpp::utils::hasOnlyYieldOp(linalgOp.getRegion()))
+    return false;
+  SmallVector<unsigned> perm;
+  return linalgOp.getMatchingIndexingMap(linalgOp.getDpsInputOperand(0))
+      .isPermutationOfMinorIdentityWithBroadcasting(perm);
+}
+
+static std::optional<linalg::GenericOp> getBroadCastProdcuer(OpOperand *rhs) {
+  linalg::GenericOp broadcastOp = rhs->get().getDefiningOp<linalg::GenericOp>();
+  if (!broadcastOp || !isBroadCastOp(broadcastOp))
+    return std::nullopt;
+  return broadcastOp;
+}
+
+static bool isFillWithZeros(linalg::Conv2DNhwcHwcfOp convOp) {
+  auto fillOp =
+      convOp.getDpsInitOperand(0)->get().getDefiningOp<linalg::FillOp>();
+  if (!fillOp)
+    return false;
+  return tpp::utils::isValConstZero(fillOp.getInputs()[0]);
+}
+
+// Instead of initializing the output of a convolution with zero and then add a
+// bias, initialize the output of the convolution with the bias.
+struct EliminateZeroInitAndAddBiasToInit
+    : public OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (!linalgOp.hasTensorSemantics())
+      return failure();
+    if (linalgOp->getNumOperands() != 3 || linalgOp->getNumResults() != 1)
+      return failure();
+
+    if (!linalg::isElementwise(linalgOp) ||
+        !tpp::utils::hasOnlyScalarElementwiseOp<arith::AddFOp>(
+            linalgOp.getRegion()))
+      return failure();
+
+    OpOperand *lhs = linalgOp.getDpsInputOperand(0);
+    OpOperand *rhs = linalgOp.getDpsInputOperand(1);
+
+    auto convProducer = lhs->get().getDefiningOp<linalg::Conv2DNhwcHwcfOp>();
+    auto broadCastProducer = getBroadCastProdcuer(rhs);
+    if (!convProducer || !broadCastProducer || !isFillWithZeros(convProducer) ||
+        !convProducer.getTiedOpResult(convProducer.getDpsInitOperand(0)))
+      return failure();
+
+    SmallVector<Value> convInputs;
+    for (OpOperand *operand : convProducer.getDpsInputOperands())
+      convInputs.push_back(operand->get());
+    Value broadCastOutput = broadCastProducer->getTiedOpResult(
+        broadCastProducer->getDpsInitOperand(0));
+
+    auto replOp = rewriter.create<linalg::Conv2DNhwcHwcfOp>(
+        linalgOp.getLoc(), broadCastOutput.getType(), convInputs,
+        broadCastOutput, convProducer.getStrides(),
+        convProducer.getDilations());
+    if (auto metadata = convProducer->getAttr("metadata"))
+      replOp->setAttr("metadata", metadata);
+
+    rewriter.replaceOp(linalgOp, replOp.getResults());
+    return success();
+  }
+};
+
+struct ConvInitSimplify : public ConvInitSimplifyBase<ConvInitSimplify> {
+  void runOnOperation() override {
+    RewritePatternSet patterns(getOperation().getContext());
+    patterns.add<EliminateZeroInitAndAddBiasToInit>(patterns.getContext());
+    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
+  }
+};
+
+} // namespace
+
+std::unique_ptr<OperationPass<func::FuncOp>>
+mlir::tpp::createConvInitSimplifyPass() {
+  return std::make_unique<ConvInitSimplify>();
+}
--- a/lib/TPP/ConvInitSimplify.cpp
+++ b/lib/TPP/ConvInitSimplify.cpp
@@ -26,7 +26,7 @@ namespace {
 static bool isBroadCastOp(linalg::GenericOp linalgOp) {
   if (linalgOp->getNumOperands() != 2 || linalgOp->getNumResults() != 1)
     return false;
-  if (!tpp::utils::hasOnlyYieldOp(linalgOp.getRegion()))
+  if (!tpp::utils::hasOnlyOp<linalg::YieldOp>(linalgOp.getRegion()))
     return false;
   SmallVector<unsigned> perm;
   return linalgOp.getMatchingIndexingMap(linalgOp.getDpsInputOperand(0))
@@ -62,8 +62,7 @@ struct EliminateZeroInitAndAddBiasToInit
       return failure();
 
     if (!linalg::isElementwise(linalgOp) ||
-        !tpp::utils::hasOnlyScalarElementwiseOp<arith::AddFOp>(
-            linalgOp.getRegion()))
+        !tpp::utils::hasOnlyOp<arith::AddFOp>(linalgOp.getRegion()))
       return failure();
 
     OpOperand *lhs = linalgOp.getDpsInputOperand(0);
--- a/lib/TPP/ConvInitSimplify.cpp
+++ b/lib/TPP/ConvInitSimplify.cpp
@@ -40,14 +40,6 @@ static std::optional<linalg::GenericOp> getBroadCastProdcuer(OpOperand *rhs) {
   return broadcastOp;
 }
 
-static bool isFillWithZeros(linalg::Conv2DNhwcHwcfOp convOp) {
-  auto fillOp =
-      convOp.getDpsInitOperand(0)->get().getDefiningOp<linalg::FillOp>();
-  if (!fillOp)
-    return false;
-  return tpp::utils::isValConstZero(fillOp.getInputs()[0]);
-}
-
 // Instead of initializing the output of a convolution with zero and then add a
 // bias, initialize the output of the convolution with the bias.
 struct EliminateZeroInitAndAddBiasToInit
@@ -70,7 +62,8 @@ struct EliminateZeroInitAndAddBiasToInit
 
     auto convProducer = lhs->get().getDefiningOp<linalg::Conv2DNhwcHwcfOp>();
     auto broadCastProducer = getBroadCastProdcuer(rhs);
-    if (!convProducer || !broadCastProducer || !isFillWithZeros(convProducer) ||
+    if (!convProducer || !broadCastProducer ||
+        !tpp::utils::isZeroTensor(convProducer.getDpsInitOperand(0)->get()) ||
         !convProducer.getTiedOpResult(convProducer.getDpsInitOperand(0)))
       return failure();
 
--- a/lib/TPP/Dialect/Tpp/TppUtils.cpp
+++ b/lib/TPP/Dialect/Tpp/TppUtils.cpp
@@ -131,12 +131,13 @@ bool isMarkedWithTpp(linalg::LinalgOp linalgOp, const std::string &target) {
   return libraryCall.compare(target) == 0;
 }
 
-// Returns true if the region of the linalgOp has only a single operation
-// (linalg.yieldOp).
-static bool hasOnlyYieldOp(Region &region) {
+bool hasOnlyYieldOp(Region &region) {
   if (!region.hasOneBlock())
     return false;
-  return std::distance(region.front().begin(), region.front().end()) == 1;
+  if (std::distance(region.front().begin(), region.front().end()) != 1)
+    return false;
+  Operation &op = region.front().front();
+  return isa<linalg::YieldOp>(op);
 }
 
 bool hasCopySemantics(linalg::LinalgOp linalgOp) {
@@ -172,8 +173,7 @@ static Operation *getPrevUser(Operation *op, Operation *currentUser) {
   return nullptr;
 }
 
-// Returns true if the value is a constant float or integer.
-static bool isValConstZero(Value val) {
+bool isValConstZero(Value val) {
   return matchPattern(val, m_AnyZeroFloat()) || matchPattern(val, m_Zero());
 }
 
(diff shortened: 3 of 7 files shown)


chelini:
@KavithaTipturMadhu maybe this could be of interest to you.

alheinecke:
I thought because of fusion and pack/unpack, we decide to do it afterwards and solve the "bias" add on TPP level?

chelini:
> I thought because of fusion and pack/unpack, we decide to do it afterwards and solve the "bias" add on TPP level?

It is fairly ok to solve this problem at linalg. I see this pattern in resnet, but I don't have a strong opinion. If you manage to solve it a tpp level we can drop this.

rengolin (commented):
IIUC, this is just detecting a single pattern in Resnet and getting rid of the useless zero-init.

The pattern is from:
```
  %0 = empty + fill(0)
  %1 = conv(%arg1, %arg2) -> %0
  %2 = add(%1, %arg3)
```
to:
```
  %1 = conv(%arg1, %arg2) -> %arg3
```

It doesn't work with `fill(non-zero)`, so this technique can't solve the Resnet residual problem.

To solve that, you'd have to move the `add` above the `conv` and convert:
```
  conv + relu
  conv
  identity
  add
  relu
```
into:
```
  conv + relu
  identity + add
  conv + relu
```
so that you can later fuse `conv + relu` and `identity + add`, which can be done at TPP level.

I agree with @alheinecke, the fusion part is best done at TPP level (less information to carry along through all other transforms), but this graph transform should be ok at the linalg level.

We just need to make sure that this does not affect the following passes in the pipeline. @adam-smnk, please check where in the pipeline this would be added and what would be the effect to the integration tests.

I'd like every new pass PR to also add itself to the default pipeline and add one generic integration test, so that we're clear where it goes and what it does to the rest of the tests, too.

PS: I'm not a big fan of the name `fast-math`.

chelini:
> IIUC, this is just detecting a single pattern in Resnet and getting rid of the useless zero-init.
> 
> The pattern is from:
> 
> ```
>   %0 = empty + fill(0)
>   %1 = conv(%arg1, %arg2) -> %0
>   %2 = add(%1, %arg3)
> ```
> 
> to:
> 
> ```
>   %1 = conv(%arg1, %arg2) -> %arg3
> ```
> 
> It doesn't work with `fill(non-zero)`, so this technique can't solve the Resnet residual problem.
> 
> To solve that, you'd have to move the `add` above the `conv` and convert:
> 
> ```
>   conv + relu
>   conv
>   identity
>   add
>   relu
> ```
> 
> into:
> 
> ```
>   conv + relu
>   identity + add
>   conv + relu
> ```
> 
> so that you can later fuse `conv + relu` and `identity + add`, which can be done at TPP level.
> 
> I agree with @alheinecke, the fusion part is best done at TPP level (less information to carry along through all other transforms), but this graph transform should be ok at the linalg level.
> 
> We just need to make sure that this does not affect the following passes in the pipeline. @adam-smnk, please check where in the pipeline this would be added and what would be the effect to the integration tests.
> 
> I'd like every new pass PR to also add itself to the default pipeline and add one generic integration test, so that we're clear where it goes and what it does to the rest of the tests, too.
> 
> PS: I'm not a big fan of the name `fast-math`.

Agree on the name. I did not know how to call it. And indeed I just try to avoid the fill with zero and add.

lib/TPP/FastMathOpts.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+template <typename OP> static bool hasOnlyScalarElementwiseOp(Region &region) {
```
rengolin, line 25: Isn't this on `TppUtils` too?

lib/TPP/FastMathOpts.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+static std::optional<linalg::GenericOp> getBroadCastProdcuer(OpOperand *rhs) {
+  linalg::GenericOp broadcastOp = rhs->get().getDefiningOp<linalg::GenericOp>();
+  if (!broadcastOp || !isBroadCastOp(broadcastOp))
+    return std::nullopt;
+  return broadcastOp;
+}
+
+static bool isConstantZero(linalg::FillOp fillOp) {
```
rengolin, line 66: All of those are repeat of tpp utils and we shouldn't duplicate.

lib/TPP/FastMathOpts.cpp
```diff
@@ -0,0 +1,136 @@
...
+      return failure();
+
+    if (!linalg::isElementwise(linalgOp) ||
+        !hasOnlyScalarElementwiseOp<arith::AddFOp>(linalgOp.getRegion()))
+      return failure();
+
+    OpOperand *lhs = linalgOp.getDpsInputOperand(0);
+    OpOperand *rhs = linalgOp.getDpsInputOperand(1);
+
+    auto convProducer = lhs->get().getDefiningOp<linalg::Conv2DNhwcHwcfOp>();
```
rengolin, line 98: Is this pattern specific to convolutions? Even, to this particular type of convolution?

lib/TPP/FastMathOpts.cpp
```diff
@@ -0,0 +1,136 @@
...
+  auto fillOp =
+      convOp.getDpsInitOperand(0)->get().getDefiningOp<linalg::FillOp>();
+  if (!fillOp)
+    return false;
+  return isConstantZero(fillOp);
+}
+
+// Instead of initializing the output of a convolution with zero and then add a
+// bias, initialize the output of the convolution with the bias.
+struct InitializeConv : public OpRewritePattern<linalg::GenericOp> {
```
rengolin, line 81: This isn't really "initializing a convolution", so `InitializeConv` isn't a good name.

How about `ConvInitSimplify`?

lib/TPP/FastMathOpts.cpp
chelini: yes, resnet.

lib/TPP/FastMathOpts.cpp
chelini: this was not in utils.

lib/TPP/FastMathOpts.cpp
chelini: yes, fixed

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,172 @@
...
+// CHECK-SAME:  outs(%[[EMPTY]] : tensor<1x56x56x64xf32>) -> tensor<1x56x56x64xf32>
+// CHECK: %[[CONV:.+]] = linalg.conv_2d_nhwc_hwcf 
+// CHECK-SAME:  {dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} 
+// CHECK-SAME:  ins(%[[ARG0]], %[[ARG2]] : tensor<1x56x56x64xf32>, tensor<1x1x64x64xf32>) 
+// CHECK-SAME:  outs(%[[FILL]] : tensor<1x56x56x64xf32>) -> tensor<1x56x56x64xf32>
+// CHECK: %[[BROAD:.+]] = linalg.generic 
+// CHECK-SAME:  indexing_maps = [#[[MAP]], #[[MAP1]]] 
+// CHECK-SAME:  iterator_types = ["parallel", "parallel", "parallel", "parallel"]} 
+// CHECK-SAME:  ins(%[[ARG3]] : tensor<64xf32>) outs(%[[EMPTY]] : tensor<1x56x56x64xf32>)
+// CHECK: {{.+}} = linalg.generic
```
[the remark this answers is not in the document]
chelini: added

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,172 @@
...
+// CHECK-SAME:  iterator_types = ["parallel", "parallel", "parallel", "parallel"]} 
+// CHECK-SAME:  ins(%[[CONV]], %[[BROAD]] : tensor<1x56x56x64xf32>, tensor<1x56x56x64xf32>) 
+// CHECK-SAME:  outs(%[[EMPTY]] : tensor<1x56x56x64xf32>)
+
+// -----
+
+#map = affine_map<(d0, d1, d2, d3) -> (d3)>
+#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+
+func.func @main(%arg0: tensor<1x56x56x64xf32>, %arg2: tensor<1x1x64x64xf32>, %arg3: tensor<64xf32>) -> tensor<1x56x56x64xf32> {
```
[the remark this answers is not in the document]
chelini: done

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,172 @@
...
+  } -> tensor<1x56x56x64xf32>
+  %4 = linalg.generic {indexing_maps = [#map1, #map1, #map1], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%2, %3 : tensor<1x56x56x64xf32>, tensor<1x56x56x64xf32>) outs(%0 : tensor<1x56x56x64xf32>) {
+    ^bb0(%in: f32, %in_0: f32, %out: f32):
+      %5 = arith.addf %in, %in_0 : f32
+      linalg.yield %5 : f32
+  } -> tensor<1x56x56x64xf32>
+  return %4 : tensor<1x56x56x64xf32>
+}
+
+// CHECK: #[[MAP:.+]] = affine_map<(d0, d1, d2, d3) -> (d3)>
```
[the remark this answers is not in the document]
chelini: just want to make sure we don't optimize as the fill may be non constant

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,172 @@
...
+
+#map = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+
+func.func @main(%arg0: tensor<1x56x56x64xf32>, %arg2: tensor<1x1x64x64xf32>, %arg3: tensor<1x56x56x64xf32>) -> tensor<1x56x56x64xf32> {
+  %cst = arith.constant 0.0 : f32
+  %0 = tensor.empty() : tensor<1x56x56x64xf32>
+  %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<1x56x56x64xf32>) -> tensor<1x56x56x64xf32>
+  %2 = linalg.conv_2d_nhwc_hwcf {dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>} ins(%arg0, %arg2 : tensor<1x56x56x64xf32>, tensor<1x1x64x64xf32>) outs(%1 : tensor<1x56x56x64xf32>) -> tensor<1x56x56x64xf32>
+  %3 = linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%arg3 : tensor<1x56x56x64xf32>) outs(%0 : tensor<1x56x56x64xf32>) {
```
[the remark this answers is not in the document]
chelini: not sure about that. If the test is wrong I can remove it.

include/TPP/Passes.td
```diff
@@ -270,4 +270,12 @@ def ElementWiseFusion : Pass<"element-wise-fusion", "func::FuncOp"> {
   let constructor = "mlir::tpp::createElementWiseFusionPass()";
 }
 
+def ConvInitSimplify : Pass<"conv-init-simplify", "func::FuncOp"> {
+  let summary = "Simplify initialization for convolution";
+  let description = [{
+    Rewrite the IR by doing some mathematical tricks.
```
rengolin, line 276: This really isn't a good description. The point here is to describe what to expect.

I'd say something like: "Simplifies convolution arguments and users by moving operations around to eliminate redundant computation and to promote instruction-level fusion at the TPP level".

lib/TPP/ConvInitSimplify.cpp
```diff
@@ -0,0 +1,113 @@
+//===- FastMathOpts.cpp.cpp --------------------------------------*- C++-*-===//
```
rengolin, line 1: Renamed

lib/TPP/ConvInitSimplify.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+static std::optional<linalg::GenericOp> getBroadCastProdcuer(OpOperand *rhs) {
+  linalg::GenericOp broadcastOp = rhs->get().getDefiningOp<linalg::GenericOp>();
+  if (!broadcastOp || !isBroadCastOp(broadcastOp))
+    return std::nullopt;
+  return broadcastOp;
+}
+
+static bool isConstantZero(linalg::FillOp fillOp) {
```
rengolin, line 43: All of these should really be in `TppUtils.cpp`

lib/TPP/ConvInitSimplify.cpp
```diff
@@ -0,0 +1,113 @@
...
+  Value cst = fillOp.getInputs()[0];
+  return matchPattern(cst, m_AnyZeroFloat()) || matchPattern(cst, m_Zero());
+}
+
+static bool isFillWithZeros(linalg::Conv2DNhwcHwcfOp convOp) {
```
rengolin, line 48: Avoid specialized functions. Reuse existing `isZero`-type functions and get the value/op before calling them. 

This proliferation of similar functions lead to high maintenance cost and unexpected bugs when we expect the general function to be called and it isn't.

lib/TPP/ConvInitSimplify.cpp
```diff
@@ -0,0 +1,113 @@
...
+  auto fillOp =
+      convOp.getDpsInitOperand(0)->get().getDefiningOp<linalg::FillOp>();
+  if (!fillOp)
+    return false;
+  return isConstantZero(fillOp);
+}
+
+// Instead of initializing the output of a convolution with zero and then add a
+// bias, initialize the output of the convolution with the bias.
+struct ConvInitSimplifyImpl : public OpRewritePattern<linalg::GenericOp> {
```
rengolin, line 58: What is your plan on moving the `addOp` above the `ConvOp`?

If you're going to create a new rewrite pattern (I recommend this), then this one needs to be called by what it does: "eliminate zero init and an add to bias init".

If you'll bundle the "add after relu" here, then it can continue with the name it is. But then the rewrite code will be a bit messy once we start adding more transforms...

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -131,12 +131,13 @@ bool isMarkedWithTpp(linalg::LinalgOp linalgOp, const std::string &target) {
   return libraryCall.compare(target) == 0;
 }
 
-// Returns true if the region of the linalgOp has only a single operation
-// (linalg.yieldOp).
-static bool hasOnlyYieldOp(Region &region) {
+bool hasOnlyYieldOp(Region &region) {
```
rengolin, line 134: All of these `hasOnlyOneXXXOp` functions are the same. This should be easily replaceable by a template function on the header.

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,179 @@
...
+// CHECK-DAG: #[[MAP1:.+]] = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+// CHECK: func.func @expect_opt(
+// CHECK-SAME:  %[[ARG0:.+]]: tensor<1x56x56x64xf32>, 
+// CHECK-SAME:  %[[ARG1:.+]]: tensor<1x1x64x64xf32>, 
+// CHECK-SAME:  %[[ARG2:.+]]: tensor<64xf32>)
+// CHECK: %[[EMPTY:.+]] = tensor.empty() : tensor<1x56x56x64xf32>
+// CHECK: %[[BIAS:.+]] = linalg.generic 
+// CHECK-SAME:  indexing_maps = [#[[MAP]], #[[MAP1]]]
+// CHECK-SAME:  iterator_types = ["parallel", "parallel", "parallel", "parallel"] 
+// CHECK-SAME:  ins(%[[ARG2]] : tensor<64xf32>) outs(%[[EMPTY]] : tensor<1x56x56x64xf32>)
```
rengolin, line 83: Also add the op inside the generic here and any other in this test.

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// CHECK: func.func @simple_copy(
+// CHECK-SAME:  %[[ARG0:.+]]: tensor<1x56x56x64xf32>, 
+// CHECK-SAME:  %[[ARG1:.+]]: tensor<1x1x64x64xf32>, 
+// CHECK-SAME:  %[[ARG2:.+]]: tensor<1x56x56x64xf32>)
+// CHECK: %{{.+}} = linalg.conv_2d_nhwc_hwcf 
+// CHECK-SAME:  dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>
+// CHECK-SAME:  ins(%[[ARG0]], %[[ARG1]] : tensor<1x56x56x64xf32>, tensor<1x1x64x64xf32>) 
+// CHECK-SAME:  outs(%[[ARG2]] : tensor<1x56x56x64xf32>) -> tensor<1x56x56x64xf32>
```
rengolin, line 118: Also add `CHECK-NOT: linalg.generic`

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: I don't see how I can unify `hasOnlyScalarElementwiseOp` and `hasOnlyYieldOp`. The latter only check for a yield operation while the former for a yield plus an additional operation.

rengolin: Well, if the op is empty, you only check for the yield?

lib/TPP/ConvInitSimplify.cpp
chelini: Which function are you referring to? `isZeroTensor`?

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: I think it will be confusion. `hasOnlyScalarElementwiseOp` requires a region with 2 ops, while `hasOnlyYield` a single op in the region. I am also struggling to find a good name, if we merge the two.

lib/TPP/ConvInitSimplify.cpp
chelini: I don't have a plan. But I suppose I can add more logic here when I have a test case, or re-evaluate naming later on.

lib/TPP/ConvInitSimplify.cpp
rengolin: To begin with, your `isConstantZero` above already exists in `TppUtils` (`isValConstZero`).

But here, the `isZeroTensor` already checks to see if a tensor is filled with zeroes. It might not be *exactly* what you're looking for here, but duplicating logic in little static helpers all over the place won't help.

Please start commoning up those methods instead of proliferating them.

lib/TPP/ConvInitSimplify.cpp
rengolin: I'd strongly recommend you create two different patterns, definitely reusing code to detect linalg+add etc, but with different names. So for now, just rename this to what it actually does.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: No it won't. The logic is simple: yield = one op, others = two ops.

I'd name it `hasOnly` with the type being an indication of what "only" means. Yield is assumed to be present always, so it's not confusing.

Something like:
```
template <class OpTy> bool hasOnly(Region &region);
```

called as:
```
// Here is clear
bool hasOnlyYield = hasOnly<YieldOp>(region);

// Here yield is implied, so also clear
bool hasOnlyAdd = hasOnly<AddOp>(region);
```

test/Passes/math-opts.mlir
```diff
@@ -0,0 +1,182 @@
...
+// CHECK-DAG: #[[MAP1:.+]] = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+// CHECK-LABEL: func.func @resnet50v1(
+// CHECK-SAME:  %[[ARG0:.+]]: tensor<1x224x224x3xf64>)
+// CHECK: %[[CST:.+]] = arith.constant 0.000000e+00 : f64
+// CHECK-DAG:  %[[CST_0:.+]] = arith.constant dense<0.000000e+00> : tensor<1x112x112x64xf64>
+// CHECK-DAG:  %[[CST_1:.+]] = arith.constant dense<1.000000e+00> : tensor<7x7x3x64xf64>
+// CHECK-DAG:  %[[CST_2:.+]] = arith.constant dense<5.000000e-01> : tensor<64xf64>
+// CHECK: %[[PAD:.+]] = tensor.pad %[[ARG0]] low[0, 3, 3, 0] high[0, 3, 3, 0]
+// CHECK: %[[EMPTY:.+]] = tensor.empty() : tensor<1x112x112x64xf64>
+// CHECK: %[[BCAST:.+]] = linalg.generic
```
rengolin, line 167: Forgot the op inside this one

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: Got it. But let's do in another commit. It is used in different place.

lib/TPP/ConvInitSimplify.cpp
chelini: `isZeroTensor` does not exist in `ConvInitSimplify.cpp`. There is `isFillWithZeros`. Is this the method you are referring to?

chelini: Ok but you don't make me understand. In the specific case here, what is that you don't want?

rengolin: I want you to reuse the existing functions instead of creating new ones. To refactor existing functions to adapt to all previous and new uses, if they're not quite right.

In this particular case, find a way to adapt `isFilledWithZeros` to `isZeroTensor` (or whatever is more appropriate in utils).

chelini:
@rengolin more comment?

rengolin (approved):

PR #328, opened 2023-02-14
Restore tpp.relu detection test

adam-smnk:
Restores and updates tpp.relu detection test cases that were removed together with the previous Linalg to TPP mapping approach.
Current TPP computation detection is a bit simpler, thus, the zero filled buffer acting as relu weights is no longer accepted.

### DIFF
--- a/test/Conversion/LinalgToTpp/detect-relu.mlir
+++ b/test/Conversion/LinalgToTpp/detect-relu.mlir
@@ -1,0 +1,119 @@
+// RUN: tpp-opt %s -split-input-file -convert-linalg-to-tpp | FileCheck %s
+
+#map0 = affine_map<(d0, d1) -> (d0, d1)>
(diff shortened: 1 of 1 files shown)


rengolin (approved):
All of those are unary ops, while the ones in Resnet are binary with a zero-filled tensor. I don't mind these tests being merged, won't stress the CI but they look like they all go through the constant path not the tensor path, which is what Resnet does.

adam-smnk:
> All of those are unary ops, while the ones in Resnet are binary with a zero-filled tensor. I don't mind these tests being merged, won't stress the CI but they look like they all go through the constant path not the tensor path, which is what Resnet does.

That's true, this logic never tried to consider all possible cases and was written with a few assumptions in mind.
Let's add more detection test cases once this logic gets extended to address resnet use-cases.

PR #330, opened 2023-02-15
Retire `PropagateThroughPadOp` and bump LLVM

chelini:
The pattern is upstreamed in MLIR/LLVM:
https://github.com/llvm/llvm-project/commit/30d542f9b2adf10ee2cb7e07877ad7f0cfdbfea2

Remark: remove `createEmptyTensorEliminationPass` from the default pipeline, as it removes `tensor.empty` operations if a source tensor can be traced back to multiple possible tensor.empty ops. This happens when we generalize tensor.pack and tensor.unpack.

```
%0 = tensor.empty() : tensor<10x10xf32>
%1 = linalg.transpose ins(%a : tensor<10x10xf32>) outs(%0: tensor<10x10xf32>
%2 = tensor.insert_slice %1 into %b
```


adam-smnk:
Pass removal looks good to me.
`createEmptyTensorEliminationPass` was originally introduced because it seemed quite harmless and should've been a free optimization. But it appears that it's either too greedy in elimination or just not compatible with the current state of tensor abstraction i.e., using `tensor.empty` might not be ideal but it remains the default go-to for creating new tensors.

chelini:
> %0 = tensor.empty() : tensor<10x10xf32>
> %1 = linalg.transpose ins(%a : tensor<10x10xf32>) outs(%0: tensor<10x10xf32>
> %2 = tensor.insert_slice %1 into %b

Anyway, I raised the issue on the tpp channel on Discord.

rengolin:
> Anyway, I raised the issue on the tpp channel on Discord.

Now that pack/unpack are upstream, you should start raising these issues on the MLIR channel of the LLVM server instead. We need more than just IREE opinions now.

rengolin:
> Pass removal looks good to me. `createEmptyTensorEliminationPass` was originally introduced because it seemed quite harmless and should've been a free optimization. But it appears that it's either too greedy in elimination or just not compatible with the current state of tensor abstraction i.e., using `tensor.empty` might not be ideal but it remains the default go-to for creating new tensors.

What is the effect of removing these passes on the IR? Does it leave more `empty` tensors around to be allocated by the bufferization dialect, or is it just a noop that gets cleaned up? 

Is the behaviour consistent across all examples, or are we just looking at a particular one?

rengolin:
Just to be sure, this `PropagateThroughPadOp` pass was completely innocuous? Was that a left-over of some time in the past (transforms?) or was it just an experiment that was never used?

This is the second example of code this week that we can just remove and no tests are affected. This is a bit worrying.

chelini:
> Just to be sure, this `PropagateThroughPadOp` pass was completely innocuous? Was that a left-over of some time in the past (transforms?) or was it just an experiment that was never used?
> 
> This is the second example of code this week that we can just remove and no tests are affected. This is a bit worrying.

It is upstreamed: https://github.com/llvm/llvm-project/commit/30d542f9b2adf10ee2cb7e07877ad7f0cfdbfea2

rengolin:
> It is upstreamed: [llvm/llvm-project@30d542f](https://github.com/llvm/llvm-project/commit/30d542f9b2adf10ee2cb7e07877ad7f0cfdbfea2)

Is this upstream pass already present in the default pipeline, by some other dependency?

chelini:
> > It is upstreamed: [llvm/llvm-project@30d542f](https://github.com/llvm/llvm-project/commit/30d542f9b2adf10ee2cb7e07877ad7f0cfdbfea2)
> 
> Is this upstream pass already present in the default pipeline, by some other dependency?

yes, is pulled in: https://github.com/plaidml/tpp-mlir/blob/f88c3f690830eeb6a029f287038a5d7d4e8600be/lib/TPP/ToBlockLayoutAndBack.cpp#L921

adam-smnk:
> > Pass removal looks good to me. `createEmptyTensorEliminationPass` was originally introduced because it seemed quite harmless and should've been a free optimization. But it appears that it's either too greedy in elimination or just not compatible with the current state of tensor abstraction i.e., using `tensor.empty` might not be ideal but it remains the default go-to for creating new tensors.
> 
> What is the effect of removing these passes on the IR? Does it leave more `empty` tensors around to be allocated by the bufferization dialect, or is it just a noop that gets cleaned up?
> 
> Is the behaviour consistent across all examples, or are we just looking at a particular one?

Without this elimination pass, we might end up with more temporary buffer allocations and copies.
It was never explicitly needed or tested in any of our use-cases so, it is hard to judge how it impacts our current code generation or its consistency.
However, as this pass's presence breaks logic in some of our tests, we have no choice to remove it and reevaluate it sometime later in the future. Overall, it probably shouldn't have been added to the default pipeline in the first place.

rengolin:
> Overall, it probably shouldn't have been added to the default pipeline in the first place.

Ack.

> Without this elimination pass, we might end up with more temporary buffer allocations and copies. It was never explicitly needed or tested in any of our use-cases so, it is hard to judge how it impacts our current code generation or its consistency.

We can pick a few relevant tests and run them to file (not FileCheck) with and without the change, and then diff (or even grep for allocas/copies) just to have an idea.

Once we have a working benchmarking infrastructure, that kind of thing gets picked up there, but until it happens, we have to be careful.

> However, as this pass's presence breaks logic in some of our tests, we have no choice to remove it and reevaluate it sometime later in the future. 

Right, we could really only look at the `test/Model`s or `test/Bencmarks` for now, as they're easier to dump to file and diff. This isn't to block this PR, just to make sure we create issues to handle the cases that we're now "introducing".

Otherwise, the PR looks good to me. Thanks!

chelini:
@rengolin improved commit msg.

rengolin:
> @rengolin improved commit msg.

Also good to update the PR description. Github doesn't do that automatically, and if you "squash & merge" it will use the PR's description, not the commit.

rengolin (approved):

PR #331, opened 2023-02-15
Relax `SameTypeOperands` constraint for `tpp.relu`

chelini:


chelini:
depends on #297

test/Conversion/LinalgToTpp/linalg-to-tpp.mlir
```diff
@@ -502,3 +502,25 @@ func.func @entry() -> memref<28x32xf32> {
...
+// CHECK-SAME:  %[[ARG0:.+]]: memref<4x4xf32>,
+// CHECK-SAME:  %[[ARG1:.+]]: memref<4x4xf32, strided<[4, 1], offset: ?>>)
+func.func @relu_mapping_strided(%i: memref<4x4xf32>, %o: memref<4x4xf32, strided<[4, 1], offset: ?>>) {
+  %cst = arith.constant 0.000000e+00 : f32 
+  // CHECK: tpp.relu ins(%[[ARG0]]
+  // CHECK-SAME:  out(%[[ARG1]]
+  linalg.generic {
+    indexing_maps = [#map, #map], 
+    iterator_types = ["parallel", "parallel"]} 
+    ins(%i : memref<4x4xf32>) outs(%o : memref<4x4xf32, strided<[4, 1], offset: ?>>) {
```
rengolin, line 520: Where is the strided requirement coming from?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -378,6 +378,8 @@ bool isTppRelu(linalg::GenericOp linalgOp) {
     return false;
   if (!isUnaryOp(linalgOp))
     return false;
+  if (!allOperandsHaveSameShape(linalgOp))
```
rengolin, line 381: Does this check that the stride is compatible with the iteration order across all operands?

test/Conversion/LinalgToTpp/linalg-to-tpp.mlir
chelini: we need to handle also strided memref, and the naive type checks is too tight. This is an use-case from @KavithaTipturMadhu

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: what do you mean?

chelini: In only checks that the two memref have the same dimensions (i.e., memref<4x4xf32>)

rengolin: > what do you mean?

`memref<4x4xf2>` and `memref<4x4xf32, stride=[4, 1]>` are "compatible".

`memref<4x4xf2>` and `memref<4x4xf32, stride=[4, 2]>` are "not compatible". (jumps one column).

`memref<4x4xf2>` and `memref<4x4xf32, stride=[8, 1]>` are "not compatible". (jumps one line).

rengolin:
This could be a quick fix, now that the dep was merged, no?

PR #332, opened 2023-02-16
Pass to decompose fused linalg ops into separate ones

KavithaTipturMadhu:


KavithaTipturMadhu:
This pass is used to decompose fused linalg ops into separate ones (see test case for example). This is a wrapper on the upstream MLIR pass. I was hoping Adam's pass would land and I wouldnt have to merge mine but he is planning on not merging his anytime soon. This pass is required for MLP example to work. This is not part of the default pipeline because we don't want to keep this around for good, hopefully it will go away after Adam merges his pass.

chelini:
> This pass is used to decompose fused linalg ops into separate ones (see test case for example). This is a wrapper on the upstream MLIR pass. I was hoping Adam's pass would land and I wouldnt have to merge mine but he is planning on not merging his anytime soon. This pass is required for MLP example to work. This is not part of the default pipeline because we don't want to keep this around for good, hopefully it will go away after Adam merges his pass.

I don't think element-wise-fusion is necessary anymore with the latest changes in tile-consumer-and-fuse producers.

KavithaTipturMadhu:
Closing as elementwise fusion isnt required anymore

PR #334, opened 2023-02-21
Wrong call to unary_invoke_inline

rengolin:
Code was generating wrong calls to `xsmm_unary_invoke_inline` by passing three arguments to a two argument function. XSMM tests were using a two-argument version when none was ever generated by TPP-to-XSMM pass.

There's no need to have the `_inline` version since neither TPP-to-XSMM not XSMM-to-LLVM lower less arguments for that version.

I thought this could be the source of the ReLU mismatches, but it isn't. Still, good to clean up incorrect function calls and dangling arguments.


adam-smnk (approved):
lgtm +1

PR #335, opened 2023-02-22
Add support for ReLU detection on dense / global

### LINKED ISSUE #295: ReLU op not recognized by current TPP passes
Running
```
tpp-opt test/Models/resnet50v1.mlir -default-tpp-passes -buffer-results-to-out-params -buffer-deallocation
```

translates following ReLU block from ResNet50v1.mlir
```
// ReLU
%27 = tensor.empty() : tensor<1x112x112x64xf32>
%28 = linalg.generic {
   indexing_maps = [#map1, #map1, #map1], 
   iterator_types = ["parallel", "parallel", "parallel", "parallel"]
   } 
ins(%6, %cst_16 : tensor<1x112x112x64xf32>, tensor<1x112x112x64xf32>) 
outs(%27 : tensor<1x112x112x64xf32>) {
^bb0(%in: f32, %in_34: f32, %out: f32):
  %1591 = arith.maxf %in, %in_34 : f32
  linalg.yield %1591 : f32
} -> tensor<1x112x112x64xf32>
```

into

```
 %alloc_4 = memref.alloc() {alignment = 64 : i64} : memref<1x112x112x64xf32>
 scf.parallel (%arg2) = (%c0) to (%c112) step (%c1) {
   %subview_291 = memref.subview %alloc_3[0, %arg2, 0, 0] [1, 1, 112, 64] [1, 1, 1, 1] : memref<1x112x112x64xf32> to memref<1x1x112x64xf32, strided<[802816, 7168, 64, 1], offset: ?>>
   %subview_292 = memref.subview %63[0, %arg2, 0, 0] [1, 1, 112, 64] [1, 1, 1, 1] : memref<1x112x112x64xf32> to memref<1x1x112x64xf32, strided<[802816, 7168, 64, 1], offset: ?>>
   %subview_293 = memref.subview %alloc_4[0, %arg2, 0, 0] [1, 1, 112, 64] [1, 1, 1, 1] : memref<1x112x112x64xf32> to memref<1x1x112x64xf32, strided<[802816, 7168, 64, 1], offset: ?>>
   linalg.generic {
        indexing_maps = [#map1, #map1, #map1], 
        iterator_types = ["parallel", "parallel"]
      } 
     ins(%subview_291, %subview_292 : memref<1x1x112x64xf32, strided<[802816, 7168, 64, 1], offset: ?>>, memref<1x1x112x64xf32, strided<[802816, 7168, 64, 1], offset: ?>>) 
     outs(%subview_293 : memref<1x1x112x64xf32, strided<[802816, 7168, 64, 1], offset: ?>>) {
     ^bb0(%in: f32, %in_294: f32, %out: f32):
       %296 = arith.maxf %in, %in_294 : f32
       linalg.yield %296 : f32
   }    
   scf.yield
}    
```

We are expecting to see XSMM unary calls for all the ReLU blocks from ResNet50 model.

rengolin:
Resnet model uses a `constant dense<0.0>` at tensor level and `global dense<0.0>` at memref level, so we need to support those two patterns.

Updated the Resnet bottleneck test to match ReLUs, as well as adding more ReLU tests on its own tests and removing the Resnet Benchmark because it was not quite right.

It took me a while to "get it right" on the Model side, so I think we should just keep the models there for both IR and output checks (contrary to my previous recommendation, so mea culpa).

Previous code was iterating through "previous users" which works for some patterns but it's too lenient, especially on recursive patterns.

There are two remaining patterns that we don't catch because of that:
1. empty/alloc + fill(0.0) -> generic
2. same as above + non-writing ops -> generic

In the first case, the "previous user" works, as it detects `fill` as the previous user, checks that it writes "all-zeros" and matches.

The second case is the adversarial generic case that breaks the "previous user" assumption. For example:

```
%c0 = constant 0.0
%c1 = constant 1.0

// Fill + op
%0 = empty
%1 = fill(%c0) into %0
%2 = some_op(%1)

// Using %1 is ok in value but not ok in reference semantics
%2 = whatever... (this could change %1's memory)
%3 = linalg.generic ins(%1)

// Using %2 is only ok if `some_op` isn't changing the underlying value
%2 = copy(%1)
%2 = linalg.generic ins(%2) // OK
// oops
%2 = fill(%c1) into %1 // same as %0 really
%2 = linalg.generic ins(%2) // NOT OK!
```

But there are even more complicated cases that we'd match:
```
%c0 = constant 0.0
%c1 = constant 1.0

// Fill + op + op + generic
%0 = empty
%1 = fill(%c0) into %0
%2 = fill(%c1) into %0
%3 = copy(%1) // in reference semantics, this is already dense<1.0>
%4 = linalg.generic ins(%1)
%5 = some use of %2
```

In this last case, following "previous user" from `%4` would match `copy(%1)` which has def as `fill(%c0)` which matches, but we ignore the second fill, which cannot be a match in reference semantics.

If linalg-on-tensors were pure value semantics, we'd be ok. But it isn't, so we're not ok.

This complication will be left for another patch, since we still don't have a use case for it, but just laying it here so that we know why those tests are still not matches ReLU.

Fixes #295


lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -138,106 +135,105 @@ bool hasCopySemantics(linalg::LinalgOp linalgOp) {
...
+      .Case<IntegerAttr>([](auto attr) {
+          return attr.getInt() == 0;
+      })
+      .Case<DenseElementsAttr>([](auto attr) {
+        if (!attr.getElementType().isIntOrFloat())
+          return false;
+        if (!attr.isSplat())
+          return false;
+        auto value = attr.template getSplatValue<Attribute>();
+        if (auto floatAttr = dyn_cast<FloatAttr>(value)) {
```
adam-smnk, line 191: nit: integer case could also be added for completeness.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -138,106 +135,105 @@ bool hasCopySemantics(linalg::LinalgOp linalgOp) {
...
+  return matchPattern(val, m_AnyZeroFloat()) || matchPattern(val, m_Zero());
+}
+
+// Prototypes
+static bool isZeroOp(Operation *, Operation *);
+
+// Returns true if the value represents a zero filled tensor.
+// Recurse into isZeroOp for defining ops if not imediately obvious
+// Looks past linalg generic's argument (which don't have defining ops)
+static bool isZeroValue(Value val, Operation *userOp) {
```
adam-smnk, line 149: Is the `userOp` needed for anything?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -138,106 +135,105 @@ bool hasCopySemantics(linalg::LinalgOp linalgOp) {
...
+        return isZeroAttr(attr);
+      })
       .Case<linalg::FillOp>([&](auto op) {
-        auto inputs = op.getInputs();
-        return inputs.size() == 1 ? isValConstZero(inputs[0]) : false;
+        if (op.getInputs().size() != 1)
+          return false;
+        return isZeroValue(op.getInputs()[0], op);
       })
       .Case<linalg::CopyOp>([&](auto op) {
```
adam-smnk, line 218: `FillOp` and `CopyOp` cases can be combined.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -138,106 +135,105 @@ bool hasCopySemantics(linalg::LinalgOp linalgOp) {
...
+// Returns true if the value is a constant float or integer.
+bool isValConstZero(Value val) {
+  return matchPattern(val, m_AnyZeroFloat()) || matchPattern(val, m_Zero());
+}
+
+// Prototypes
+static bool isZeroOp(Operation *, Operation *);
+
+// Returns true if the value represents a zero filled tensor.
+// Recurse into isZeroOp for defining ops if not imediately obvious
```
adam-smnk, line 147:
```suggestion
// Recurse into isZeroOp for defining ops if not immediately obvious
```

adam-smnk (approved):
Just some minor comments, overall looks good

adam-smnk:
Maybe also add at least one check (looking for `xsmm_unary_invoke`) to `resnet50v1.mlir` model which shows that ReLUs are now recognized correctly.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: Nope, it was left over from previous attempt at improving the "previous user" chain. Removed.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: Better even, recur back into `isZeroAttr`.

lib/TPP/Dialect/Tpp/TppUtils.cpp
rengolin: Done

rengolin:
> Maybe also add at least one check (looking for `xsmm_unary_invoke`) to `resnet50v1.mlir` model which shows that ReLUs are now detected correctly.

Looking at that test, it has no other checks than for matmul. No bias add, nothing. I think we don't need to worry about that one too much, since it's a "whole model" and not a "known building block" kind of thing.

What we need to worry about in that test is to make sure we count the number of packs, unpacks and make sure the layouts are correct, and let the actual calls to smaller, more tractable tests.

adam-smnk (approved):
Looks good +1

PR #336, opened 2023-02-22
Fold `tensor.pack` in non-splat tensor

chelini:
Enable folding `tensor.pack` into constant tensors. The idea is to work on linearized indexes for both the source (unpacked constant tensor) and the destination (packed constant tensor). Here what happens: Linearize (dest) -> de-linearize [A, B, C, D, E, F] -> apply packing [C, D, [A, E], [B, F]] -> collapse point and tile loops [C, D, X, Y] -> linearize (source)

For resnet, without parallelization

===-------------------------------------------------------------------------===
                         ... Execution time report ...
===-------------------------------------------------------------------------===
  Total Execution Time: 51.8630 seconds

  ----User Time----  ----Wall Time----  ----Name----
    0.5073 (  1.0%)    0.5073 (  1.0%)  Parser
    0.1197 (  0.2%)    0.1191 (  0.2%)  'func.func' Pipeline
    0.0374 (  0.1%)    0.0369 (  0.1%)    PackConv2DNhwcHwcf
    0.0823 (  0.2%)    0.0821 (  0.2%)    PropagatePackUnPack
    0.0320 (  0.1%)    0.0320 (  0.1%)  Canonicalizer
   27.8289 ( 53.5%)   27.8289 ( 53.7%)  ConstantFoldPack
   23.3728 ( 45.0%)   23.3728 ( 45.1%)  Output
    0.0023 (  0.0%)    0.0023 (  0.0%)  Rest
   51.9828 (100.0%)   51.8630 (100.0%)  Total

With parallelization

===-------------------------------------------------------------------------===
                         ... Execution time report ...
===-------------------------------------------------------------------------===
  Total Execution Time: 36.8436 seconds

  ----User Time----  ----Wall Time----  ----Name----
    0.5070 (  1.4%)    0.5070 (  1.4%)  Parser
    0.1175 (  0.3%)    0.1168 (  0.3%)  'func.func' Pipeline
    0.0367 (  0.1%)    0.0362 (  0.1%)    PackConv2DNhwcHwcf
    0.0808 (  0.2%)    0.0806 (  0.2%)    PropagatePackUnPack
    0.0317 (  0.1%)    0.0317 (  0.1%)  Canonicalizer
    9.8325 ( 26.6%)    9.8325 ( 26.7%)  ConstantFoldPack
   26.3527 ( 71.3%)   26.3527 ( 71.5%)  Output
    0.0021 (  0.0%)    0.0021 (  0.0%)  Rest
   36.9611 (100.0%)   36.8436 (100.0%)  Total


test/Passes/fold-pack-into-constant-weight.mlir
```diff
@@ -1,12 +1,78 @@
...
+                               [57.0, 58.0, 59.0, 60.0, 61.0, 62.0, 63.0, 64.0]]> : tensor<8x8xf32>
+  %0 = tensor.empty() : tensor<2x4x2x4xf32>
+  %pack = tensor.pack %cst inner_dims_pos = [1, 0] inner_tiles = [2, 4] 
+    into %0 : tensor<8x8xf32> -> tensor<2x4x2x4xf32>
+  return %pack : tensor<2x4x2x4xf32>
+}
+
+// CHECK-LABEL: func.func @non_splat_with_inner
+// CHECK-NOT: tensor.pack
+// CHECK: %[[CST:.+]] = arith.constant dense<{{.+}}> : tensor<2x4x2x4xf32>
```
adam-smnk, line 77: I think the test should check at least a few values within the new constant to ensure that transposition is handled correctly.

rengolin (commented):
I haven't looked into the code itself, but it has the problem that was discussed upstream: it adds a _separate_ implementation of pack, which needs to be extensively tested (it isn't in this patch), and validated against the run-time version.

The solution that was proposed is to use a compile-time evaluation of code, similar to what `ConstEval` does. If there is infrastructure for us to reuse the existing lowered logic we should use that instead of re-implementing in C++ on some other place.

chelini:
> 

I can constant fold entire Resnet, but did not check-in as it is too large (MB). I also have more integration tests. Anyway, I may miss something here. It is unclear to me how I can use the existing lowering to do compile-time evaluation. The lowering materialize some IR so the only solution I see is JITing. The current approach is in-line with what Linalg does see `ConstantFold.cpp`.

rengolin:
> I can constant fold entire Resnet, but did not check-in as it is too large (MB).

That's not a surprise, but if you're working with our Resnet file, then there's no way to check, since all weights are `dense splats`.

> I also have more integration tests.

Why aren't they in this PR? You add a lot of logic with something that will replace the run-time packing but only have one test without checking any value. Checking the new constant and deleting the pack is way below minimum testing,

> Anyway, I may miss something here. It is unclear to me how I can use the existing lowering to do compile-time evaluation. The lowering materialize some IR so the only solution I see is JITing. The current approach is in-line with what Linalg does see `ConstantFold.cpp`.

The `ConstEval` I pointed to does JITing, yes. If we can use that directly then we carry all of the existing tests on pack/unpack with this. But if we re-write everything like Linalg, then we'll need to write a much larger body of tests to prove that every bit of functionality you added is correct.

test/Passes/fold-pack-into-constant-weight.mlir
rengolin: We're going to need a lot more tests than this...

chelini:
> > I can constant fold entire Resnet, but did not check-in as it is too large (MB).
> 
> That's not a surprise, but if you're working with our Resnet file, then there's no way to check, since all weights are `dense splats`.
> 
> > I also have more integration tests.
> 
> Why aren't they in this PR? You add a lot of logic with something that will replace the run-time packing but only have one test without checking any value. Checking the new constant and deleting the pack is way below minimum testing,
> 
> > Anyway, I may miss something here. It is unclear to me how I can use the existing lowering to do compile-time evaluation. The lowering materialize some IR so the only solution I see is JITing. The current approach is in-line with what Linalg does see `ConstantFold.cpp`.
> 
> The `ConstEval` I pointed to does JITing, yes. If we can use that directly then we carry all of the existing tests on pack/unpack with this. But if we re-write everything like Linalg, then we'll need to write a much larger body of tests to prove that every bit of functionality you added is correct.

The problem with jitting is that is unclear to me how: 1) Jit only a few operations within the module. 2) how to update in-place the constants. Constants are defined within the module and I don't see a way on how to update them during execution.

rengolin:
> The problem with jitting is that is unclear to me how: 1) Jit only a few operations within the module. 2) how to update in-place the constants. Constants are defined within the module and I don't see a way on how to update them during execution.

We can solve that problem later, I was mainly wondering if you had a look and it the infrastructure was there.

You wouldn't eval the whole module, you'd create a new module, insert the ops/constants there, eval that part, take the result and plug it back in. The generic case looks like a nightmare to handle, but a `const+pack` -> `const` would be doable.

> I am not working with our Resnet file, as probably the constants are fake ones. This is the resnet I am looking at: https://tfhub.dev/google/imagenet/resnet_v2_50/classification/5

I see, meno male. But we still need enough tests to show that the packing is correct.

Smaller tests, on smaller tensors, for each kind of pack (inner dim choices, outer dim transpose, pad) with exact values being CHECK-ed in each test.

hfp (approved):
Thanks for the curly braces ;-)

test/Passes/fold-pack-into-constant-weight.mlir
```diff
@@ -1,12 +1,78 @@
...
+                               [49.0, 50.0, 51.0, 52.0, 53.0, 54.0, 55.0, 56.0], 
+                               [57.0, 58.0, 59.0, 60.0, 61.0, 62.0, 63.0, 64.0]]> : tensor<8x8xf32>
+  %0 = tensor.empty() : tensor<2x4x4x2xf32>
+  %pack = tensor.pack %cst inner_dims_pos = [0, 1] inner_tiles = [4, 2] into %0 : tensor<8x8xf32> -> tensor<2x4x4x2xf32>
+  return %pack : tensor<2x4x4x2xf32>
+}
+
+// CHECK-LABEL: func.func @non_splat()
+// CHECK-NOT: tensor.pack
+// CHECK: %[[CST:.+]] = arith.constant dense<{{.+}}> : tensor<2x4x4x2xf32>
```
rengolin, line 32: Can you check for the actual values?

PR #337, opened 2023-02-22
Special lowering to LLVM

rengolin:
This change introduces a custom-lowering to LLVM in addition to the default one from mlir-cpu-runner. After the default passes, we try to find the target triple/cpu (if specified by the user) and run the passes with the chosen optimization targets (speed/size).

Default is: { x86_64, skylake, avx2 }, which is old enough but still relevant to libxsmm.

Also adding options to enable FMA and unsafe math, as we want full speed aways. We have to add unsafe-math attributes to LLVM functions otherwise FMAs don't get lowered by the back-end. This is what clang does but MLIR doesn't.

Just using the defaults we should get FMAs and vectorization, but also allowing options to change and check code generation. Dumping the LLVM IR at the end you'll still see `fmul+fadd` sequences, but piping to `llc -mcpu=skylake -o -` will show `vfmadd132ss`.

### DIFF
--- a/tpp-run/tpp-run.cpp
+++ b/tpp-run/tpp-run.cpp
@@ -9,9 +9,12 @@
 
 #include "MLIRBench.h"
 
+#include "llvm/MC/TargetRegistry.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/InitLLVM.h"
 #include "llvm/Support/TargetSelect.h"
+#include "llvm/Target/TargetMachine.h"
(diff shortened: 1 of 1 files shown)


rengolin:
With this change, the outputs for lowering to loops and libxsmm are identical in the MLP example.

rengolin:
Working towards #229

adam-smnk (approved):
Looks correct - using fast math by default is in line with our rewriting passes

PR #339, opened 2023-02-23
Custom linalg decomposition

### LINKED ISSUE #352: Split linalg generic into multiple ops
Current linalg to TPP mapping is limited to specific `linalg.generic` ops. One of the requirements is that the generic contains only a single scalar operation within its body. This is due to the limitations of the current op analysis (and to some extension conversion) logic. However, TPP-compatible generic ops with multiple body ops might sometimes be present either due to other rewriting passes (like fusion) or generated by a frontend. This leads to missed lowering opportunities.

The first step to increase linalg to TPP conversion coverage is to try decomposing more complex linalg generic ops into separate generics with single operations in their bodies. Then the existing conversion pass can be used.
To achieve decomposition, the upstream MLIR `DecomposeLinalgOp` pass can be used. However, after initial testing, some limitations have been identified:
- the decomposition might introduce temporary buffer zero initialization which breaks our conversion detection
- works only on tensors
  - limits opportunities for "cleanup" in the later (post-bufferization) stages of the lower pipeline
  - cannot be used directly as a part of `convert-linalg-to-tpp` pass to capture more lowering opportunities
  - without CSE, decomposition introduces many extra temporary buffers
  - with CSE, bufferization might still introduces extra temporary buffers
- invalid output buffer propagation between decomposed generics - see example below

To address these issues, it might be beneficial to either expand the upstream pass or create a more specialized custom version.

adam-smnk:
Adds a custom linalg generic decomposition pass which builds on top of the upstream linalg decomposition pattern. The logic is validated for both tensors and memrefs.
For convenience, the upstream decomposition is also exposed as a separate pass.
Additionally, adds, tweaks, and makes public more of the existing TPP utilities.

Notable decomposition changes:
  - generalized to work on both tensors and memrefs
  - reuses the original output buffer whenever possible
  to limit the number of temporary allocations
  - removes the need for temporary buffer initialization
  - correctly propagates output buffer between split ops
  when it is used as an input by any scalar operations

The custom pass limitations:
  - only accepts static shapes

Resolves #352

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -290,4 +290,23 @@ def Bufferize : Pass<"bufferize", "ModuleOp"> {
   let constructor = "mlir::tpp::createBufferizePass()";
 }
 
+def DecomposeLinalgPass : Pass<"decompose-linalg", "ModuleOp"> {
+  let summary = "Custom linalg decomposition pattern";
+  let description = [{
+    Decompose a GenericOp into multiple ops.
+    Supports memrefs and output buffer reuse.
+    Fixes output buffer propagation when used as input.
+    Limited only to static shaped inputs and all inputs having the same shape.
+  }];
+  let constructor = "mlir::tpp::createDecomposeLinalgPass()";
+}
+
+def DecomposeDefaultPass : Pass<"decompose-default", "ModuleOp"> {
+  let summary = "Default linalg decomposition pattern";
+  let description = [{
+    Invoke upstream linalg decomposition pattern.
+  }];
+  let constructor = "mlir::tpp::createDecomposeDefaultPass()";
+}
+
 #endif // TPP_DIALECT_TPP_PASSES
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -290,7 +290,7 @@ def Bufferize : Pass<"bufferize", "ModuleOp"> {
   let constructor = "mlir::tpp::createBufferizePass()";
 }
 
-def DecomposeLinalgPass : Pass<"decompose-linalg", "ModuleOp"> {
+def DecomposeLinalgPass : Pass<"decompose-linalg", "func::FuncOp"> {
   let summary = "Custom linalg decomposition pattern";
   let description = [{
     Decompose a GenericOp into multiple ops.
@@ -301,7 +301,7 @@ def DecomposeLinalgPass : Pass<"decompose-linalg", "ModuleOp"> {
   let constructor = "mlir::tpp::createDecomposeLinalgPass()";
 }
 
-def DecomposeDefaultPass : Pass<"decompose-default", "ModuleOp"> {
+def DecomposeDefaultPass : Pass<"decompose-default", "func::FuncOp"> {
   let summary = "Default linalg decomposition pattern";
   let description = [{
     Invoke upstream linalg decomposition pattern.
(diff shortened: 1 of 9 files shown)


include/TPP/Passes.td
```diff
@@ -290,4 +290,23 @@ def Bufferize : Pass<"bufferize", "ModuleOp"> {
   let constructor = "mlir::tpp::createBufferizePass()";
 }
 
+def DecomposeLinalgPass : Pass<"decompose-linalg", "ModuleOp"> {
```
chelini, line 293: why is a Module pass?

lib/TPP/DecomposeLinalgOps.cpp
```diff
@@ -0,0 +1,481 @@
...
+      residualGenericInOperands, residualGenericOutOperands, indexingMapAttr,
+      genericOp.getIteratorTypes(),
+      /*doc=*/nullptr, /*libraryCall=*/nullptr,
+      [](OpBuilder, Location, ValueRange) {});
+}
+
+LogicalResult
+DecomposeLinalgOp::matchAndRewrite(GenericOp genericOp,
+                                   PatternRewriter &rewriter) const {
+  /// For now only match on operations where the iterator types are all parallel
```
chelini, line 326: use `//` please. `///` end up in doc. Also add `.` at the end of a comment

lib/TPP/DecomposeLinalgOps.cpp
```diff
@@ -0,0 +1,481 @@
...
+  }
+
+  /// If the op has only a single statement (apart from the yield), do nothing.
+  Block *body = genericOp.getBody();
+  if (body->getOperations().size() <= 2) {
+    return rewriter.notifyMatchFailure(genericOp,
+                                       "operation has less than 3 statements");
+  }
+
+  /// Check that the peeled statement has a scalar element type.
```
chelini, line 348: this comment seems off.

lib/TPP/DecomposeLinalgOps.cpp
```diff
@@ -0,0 +1,481 @@
...
+
+  scalarReplacements.reserve(peeledScalarOpNumResults);
+  for (auto num : llvm::seq<unsigned>(0, peeledScalarOpNumResults))
+    scalarReplacements.push_back(residualGenericOpBody->getArgument(
+        num + origNumInputs + origNumOutputs));
+  bool allUsesReplaced = false;
+  rewriter.replaceOpWithinBlock(peeledScalarOperation, scalarReplacements,
+                                residualGenericOpBody, &allUsesReplaced);
+  assert(!allUsesReplaced &&
+         "peeled scalar operation is erased when it wasnt expected to be");
```
chelini, line 442: typo: wasnt

chelini (commented):
First round of comment.

include/TPP/Passes.td
adam-smnk: I always forget to revisit that...

lib/TPP/DecomposeLinalgOps.cpp
adam-smnk: The whole decomposition is reworked upstream `mlir/lib/Dialect/Linalg/Transforms/DecomposeLinalgOps.cpp` so, I tried to change as little as possible although I ended up modifying quite a few parts.
This is one of the original checks. I'm also not 100% sure about this comment but I just kept it as is.

chelini (approved):

adam-smnk:
No longer needed for tiling and packing, and we'll probably take a slightly different approach to handle converting multi op `linalg.generic`s to TPP. Closing for now.

PR #340, opened 2023-02-23
Introduce bufferization pass

chelini:


rengolin (approved):

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -91,18 +91,7 @@ struct DefaultTppPasses : public DefaultTppPassesBase<DefaultTppPasses> {
...
-    buffOpts.allowReturnAllocs = true;
-    buffOpts.bufferizeFunctionBoundaries = true;
-    buffOpts.functionBoundaryTypeConversion =
-        bufferization::LayoutMapOption::IdentityLayoutMap;
-    pm.addPass(bufferization::createOneShotBufferizePass(buffOpts));
-    pm.addPass(bufferization::createDropEquivalentBufferResultsPass());
-    pm.addNestedPass<func::FuncOp>(
-        bufferization::createFinalizingBufferizePass());
-    // Clean up after bufferization.
-    pm.addPass(bufferization::createBufferDeallocationPass());
```
adam-smnk, line 104: This should be added either to the bufferization pass or to the default pipeline after bufferization.

chelini: is the post-processing in Bufferize.cpp

adam-smnk: Probably the default pipeline. But tests seems happy so I can add it separately with some test case.

adam-smnk: Nevermind, I'm just blind. Another `createBufferDeallocationPass` is already present at the end of the default pipeline and there is a test case in `DefaultPipeline/mixed.mlir` that verifies buffer deallocation.
No need for any changes.

PR #342, opened 2023-02-25
Initial version of multi-head-attention layer from BERT

nhasabni:
Initial version of multi-head-attention layer from BERT.. @rengolin Please check if this matches your expectations. I am yet to add CHECKs for `tpp-opt`. I am hitting an error mentioned below.

```
$ nhasabni_tpp-sandbox/build/bin/tpp-opt test/Models/multi-head-attention.mlir -default-tpp-passes -expand-strided-metadata
test/Models/multi-head-attention.mlir:125:12: error: 'tpp.identity' op fails to verify broadcasting rules
    %102 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d3, d1, d2)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%98 : tensor<32x8x2x64xf32>) outs(%101 : tensor<32x2x64x8xf32>) {
           ^
test/Models/multi-head-attention.mlir:125:12: note: see current operation: "tpp.identity"(%62, %63) : (memref<8x64xf32, affine_map<(d0, d1)[s0] -> (d0 * 128 + d1 + s0)>>, memref<64x8xf32, affine_map<(d0, d1)[s0] -> (d0 * 8 + d1 + s0)>>) -> ()
```


test/Models/multi-head-attention.mlir
```diff
@@ -0,0 +1,261 @@
...
+    
+    %transformer_layer_0_self_attention_attention_output_bias = arith.constant dense<"0x04A1A03C2E49A03DFA9BB13D185120BEACB29EBDA5D207BEF8871ABE834D5DBC9A376A3C57370DBEE7EC14BCDE5B193E74D4083DDE9589BED54E703B9E0D87BD4523AF3D3552FFBD9F6E3DBEDA3C9CB927EFCD3C96BBA9BC641A523D760BE0BD57CC0E3EEADD9B3C4C893FBDE425AA3D52D533BDA69A99BDF0C6F4BDCB6906BEA278913D3FEFD9BD72F49ABD706A56BEC0A63DBDC8F3FD3D78842A3B15303FBE8F5F303EFD97313DF8DD09BE7E65703EDB0512BCA5FE343DE5F0DE3DDBE6073DE2D96CBD876CB3BCC615A33E9E00053D0E569F3C7549EE3D3C9EDC3C56EF0DBED56E8D3DE73EEB3DD083D83CFCABC83D684FF43DCB837ABD381A91BC0C2923BD5A7A18BE0C8A0C3E5FAF3B3D54D277BDFA79393BB2E2963CE56564BE57D2163E921CD13D752ED0BD9A8FBABC113F0FBED601EE3C29C4893D2B27583E069FA33D16558C3D1342D8BD5018C53DBA1E2CBD15515E3D0D7E11BEF763D4BD82ADDCBD5319573DAADEAABDB0E90D3D574F983D249D98BDD4ED673D7E9DAABDACC460BD1D3B88BCDBB7AB3D0C21D53DBADCED3C9BE6F9BD3B57E73D75E78BBC930A3D3DF64F6D3DADDF5FBDE11AC8BD191B223E36F68E3CEF0C15BD1B21C93DF3A44DBD20F8A53DE667343BAD7FA53CB946503D4AE137BE6EAD473C95B4CA3C08FB7A3CFFA7453EA46EF03D71DFC23C85B0713D2581BBBD80EA1C3D47E3903D049B0ABD"> : tensor<128xf32>
```
rengolin, line 55: For this example, we probably shouldn't use pre-trained weights yet (it's too large!), as we care more about ops than results here. We're building tensor initialization support that we can use with this model later to get actual results.

adam-smnk:
> I am hitting an error mentioned below.
> 
> ```
> $ nhasabni_tpp-sandbox/build/bin/tpp-opt test/Models/multi-head-attention.mlir -default-tpp-passes -expand-strided-metadata
> test/Models/multi-head-attention.mlir:125:12: error: 'tpp.identity' op fails to verify broadcasting rules
>     %102 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d0, d3, d1, d2)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%98 : tensor<32x8x2x64xf32>) outs(%101 : tensor<32x2x64x8xf32>) {
>            ^
> test/Models/multi-head-attention.mlir:125:12: note: see current operation: "tpp.identity"(%62, %63) : (memref<8x64xf32, affine_map<(d0, d1)[s0] -> (d0 * 128 + d1 + s0)>>, memref<64x8xf32, affine_map<(d0, d1)[s0] -> (d0 * 8 + d1 + s0)>>) -> ()
> ```

It looks like a problem with the `isTppIdentity` mapping logic. The above generic op represents a transpose which shouldn't be converted to an identity. We probably should add a check that validates operand dimensions before we map it to a tpp op.

test/Models/multi-head-attention.mlir
nhasabni: Sounds good. I can remove pre-trained weights. I believe tensor initialization won't initialize weight tensors, right? So I will need to initialize them in the model anyway (may be with `const dense<value>`)

rengolin: For now, `const dense<1.0>` splat should be fine. I'm working on a `tpp-run` option that will replace constant splat in the kernel with random values.

nhasabni:
@rengolin Please check. I've addressed your comment and have added all the CHECKs.

rengolin:
> It looks like a problem with the `isTppIdentity` mapping logic. The above generic op represents a transpose which shouldn't be converted to an identity. We probably should add a check that validates operand dimensions before we map it to a tpp op.

I'm getting the same error on my MLP generator. This is clearly a bug that needs to be fixed.

rengolin:
> As a side note, we do not seem to lower `linalg.batch_matmul` to TPP..

Hm, that's an obvious thing to do. I'll create a ticket for that, too, but that's for later.

We just need to fix the `tpp.identity` issue for the tests to go green so we can merge this PR.

test/Models/multi-head-attention.mlir
```diff
@@ -0,0 +1,375 @@
...
+    // CHECK: scf.parallel (%[[arg1:.*]], %[[arg2:.*]]) = (%[[c0]], %[[c0]]) to (%[[c32]], %[[c2]]) step (%[[c1]], %[[c1]]) {
+    // CHECK: %[[cast_51:.*]] = memref.cast %[[subview:.*]] : memref<8x8xf32, #[[map7]]> to memref<*xf32>
+    // CHECK-NEXT: %[[cast_52:.*]] = memref.cast %[[subview_49:.*]] : memref<8x8xf32, #[[map7]]> to memref<*xf32>
+    // CHECK-NEXT: %[[cast_53:.*]] = memref.cast %[[subview_50:.*]] : memref<8x8xf32, #[[map7]]> to memref<*xf32>
+    // CHECK-NEXT: func.call @xsmm_binary_invoke(%[[c1_i64]], %[[ret]], %[[cast_51]], %[[cast_52]], %[[cast_53]]) : (i64, i64, memref<*xf32>, memref<*xf32>, memref<*xf32>) -> ()
+    // CHECK-NEXT: scf.yield
+    // CHECK-NEXT: }
+    //
+
+    // Encoder 1 - not sure what this block is.. looks like some form of activation
```
rengolin, line 236: I think this is the feed-forward layer after the self-attention

rengolin (commented):
This is a long IR and hard to read. From the little I understand of self-attention, it doesn't quite match what I expected, but it could also be me reading it wrong.

My understanding is that the input (`KQV`) is split into the number of heads initially, then concurrent "heads", then concat into the feed-forward (With relu), to a linear, to a softmax. Here, IIUC, only `KQ` are split into multiple heads, while `V` isn't?

Is this taken from a "self-attention" or the "encoder-decoder-attention"?

rengolin (approved):

PR #343, opened 2023-02-25
Random initialization of tensors

### LINKED ISSUE #229: Random Initialization in Benchmarks/Tests

rengolin:
Wrote a simple class structure to initialize tensors in various ways, including the existing `dense<1.0>`, some odd patterns and two random splats.

The random init creates a `dense<0x...>` with random values with a library that can be used by both `tpp-run` and the C++ benchmarks.

Using the same seed, we can make sure we generate the same tensors on both sides.

Fixes #229

### DIFF
--- a/include/TPP/TensorInit.h
+++ b/include/TPP/TensorInit.h
@@ -1,0 +1,134 @@
+#ifndef TPP_RUN_TENSORINIT_H
+#define TPP_RUN_TENSORINIT_H
+
+//===- TensorInit.h - MLIR Tensor Initialization --------------------------===//
+//
+// Initializes tensors for kernel input/output handling with some reasonable
+// distribution to allow for layout testing (reorder, pad) without vanishing
+// or exploding values at the end of a large model (0.0 ~ 1.0).
+//
+//===----------------------------------------------------------------------===//
+
+#include "mlir/IR/BuiltinOps.h"
+
+#include <algorithm>
+#include <random>
+#include <vector>
+
+/// Base class
+/// Assumes float (32) as base type.
+/// TODO: Add a template parameter if/when we support double.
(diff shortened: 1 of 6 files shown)


rengolin:
@KavithaTipturMadhu I'm going to use this on my [mlp code](https://github.com/rengolin/tpp-mlir/tree/mlp/utils/mlp-gen)

KavithaTipturMadhu (approved):

PR #349, opened 2023-02-28
Use print() instead of dump()

### LINKED ISSUE #350: unresolved symbol for llvm::Module::dump() when llvm-project is built as Release
Some translation units (e.g., `tpp-run.cpp` or `MLIRBench.cpp`) rely on llvm::Module::dump() which seems to cause an unresolved symbol when `llvm-project` was built with `-DCMAKE_BUILD_TYPE=Release`.

Perhaps one would use `print(llvm::outs())` or protect this method with some preprocessor symbol.

rengolin:
LLVM Dump methods are not compiled on release builds, so we should really use `print(errs)` instead. Also changing the actual printing of IR to stdout with `print(outs)`.

Fixes #350

### DIFF
--- a/tpp-run/MLIRBench.cpp
+++ b/tpp-run/MLIRBench.cpp
@@ -370,7 +370,7 @@ LogicalResult MLIRBench::finalize(bool dumpMLIR) {
   }
 
   if (dumpMLIR)
-    module->dump();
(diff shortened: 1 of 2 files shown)


hfp (approved):
Looks good to me!

PR #355, opened 2023-03-02
Relax consumer selection in tile and fuse

chelini:
We used to require each consumer to have a single user, this is a bit too restrictive, and we may lose fusion opportunities (i.e., the last relu does not get fused as it has multiple users - return and print). So change the candidate walk by going bottom up, not requiring a single user. The bottom-up traversal allows us to insert candidates differently and drop the user check; we know if an op is already part of a fusion cluster if it is already visited.

### DIFF
--- a/test/Passes/tile-and-fuse.mlir
+++ b/test/Passes/tile-and-fuse.mlir
@@ -273,3 +273,424 @@ func.func @matmul_sequence_fusion(%arg0: tensor<32x32xf32>, %arg1: tensor<32x32x
 // CHECK-NEXT: }
 // CHECK: scf.yield %{{.+}} : tensor<32x32xf32>
 // CHECK-NEXT: }
+
+// -----
+
+#map = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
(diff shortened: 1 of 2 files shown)


test/Passes/tile-and-fuse.mlir
```diff
@@ -273,3 +273,424 @@ func.func @matmul_sequence_fusion(%arg0: tensor<32x32xf32>, %arg1: tensor<32x32x
...
+  // CHECK:       ^bb0(
+  // CHECK-NEXT:    arith.addf
+  // CHECK: linalg.generic
+  // CHECK-SAME:  indexing_maps = [#[[MAP3]], #[[MAP3]]]
+  // CHECK-SAME:  iterator_types = ["parallel", "parallel"]
+  // CHECK:       ^bb0(
+  // CHECK-NEXT:    arith.maxf
+  // CHECK: }
+  // CHECK: }
+  %c0 = arith.constant 0: index
```
KavithaTipturMadhu, line 689: This part can go, all it does is print 4x4 slice of the result. I added it to verify results manually in my benchmark.

KavithaTipturMadhu (approved):

test/Passes/tile-and-fuse.mlir
chelini: Actually, let's keep it. The last relu was not fusing because it has multiple users (the extract_slice and the return). So in part the this last step exercise what we want to test.

PR #356, opened 2023-03-02
Modularize default pipeline

adam-smnk:
Modularizes the default pipeline by splitting it into sub-passes. The TPP lowering now consists of multiple major stages.
Additionally, smaller helper passes are added.

All new passes are registered and available for automated testing as well as manual debugging and testing.

### DIFF
--- a/test/Models/resnet50-bottleneck-block.mlir
+++ b/test/Models/resnet50-bottleneck-block.mlir
@@ -1,9 +1,9 @@
 // RUN: tpp-opt %s -default-tpp-passes -expand-strided-metadata | \
 // RUN: FileCheck %s
 
-// R_UN: tpp-run %s -n 10 \
-// R_UN:         -print -e resnet50_bottleneck_block -entry-point-result=void | \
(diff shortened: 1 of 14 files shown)


test/Models/resnet50-bottleneck-block.mlir
```diff
@@ -4,6 +4,10 @@
 // RUN: tpp-run %s -n 10 \
 // RUN:         -print -e resnet50_bottleneck_block -entry-point-result=void | \
 // RUN: FileCheck %s -check-prefix=EXEC
+// XFAIL:*
```
rengolin, line 7: Can we just comment the execution run lines instead? So we still run the IR check and make sure it doesn't change?

Making a TODO linking to the issue is still a good idea!

rengolin (approved):
One small nit on the test and looks good to me! Thanks!

test/Models/resnet50-bottleneck-block.mlir
adam-smnk: Good point, we XFAIL is too broad.

PR #357, opened 2023-03-02
Fix mapping for `tpp.identity`

chelini:
The mapping logic for `tpp.identity` was too naive. So now we check the affine maps and require them to be projected permutation. Additionally, we ensure before mapping that the linalg op verifies "our" broadcasting rules and rejects candidate ops that do not satisfy these requirements. Current mapping is still not optimal, and more effort needs to be spent on mapping and broadcast support.

### DIFF
--- a/lib/TPP/Dialect/Tpp/TppUtils.cpp
+++ b/lib/TPP/Dialect/Tpp/TppUtils.cpp
@@ -132,6 +132,12 @@ bool hasCopySemantics(linalg::LinalgOp linalgOp) {
     return false;
   if ((linalgOp->getNumOperands() != 2) || (linalgOp.getNumDpsInputs() != 1))
     return false;
+  if (!allIndexingsAreProjectedPermutation(linalgOp))
+    return false;
+  AffineMap outputIndexingMap =
+      linalgOp.getMatchingIndexingMap(linalgOp.getDpsInitOperand(0));
+  if (!outputIndexingMap.isIdentity())
+    return false;
   return hasOnlyOp<linalg::YieldOp>(linalgOp->getRegion(0));
 }
 
@@ -254,7 +260,8 @@ bool hasMappingToTppConditions(linalg::GenericOp linalgOp) {
     return false;
   if (linalgOp.hasTensorSemantics())
     return false;
-  return hasStaticShape(linalgOp);
+  return hasStaticShape(linalgOp) &&
+         allIndexingsAreProjectedPermutation(linalgOp);
 }
 
 static bool hasOneUser(Value val) {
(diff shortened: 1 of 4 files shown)


lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -355,6 +362,39 @@ static bool hasMaxfZeroOp(linalg::LinalgOp linalgOp) {
   return false;
 }
 
+MatchBroadcastRuleResult verifyTppIdentityBroadcastingRules(Type inputType,
+                                                            Type outputType) {
+  // input scalar, just return.
+  if (!inputType.isa<ShapedType>())
```
rengolin, line 368: common up `inputType.cast<ShapedType>()`

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -363,8 +403,7 @@ bool isTppAdd(linalg::GenericOp linalgOp) {
     return false;
   if (!allOperandsHaveSameShapeAndStrides(linalgOp->getOperands().getTypes()))
     return false;
-  return allIndexingsAreProjectedPermutation(linalgOp) &&
-         hasOnlyOp<arith::AddFOp>(linalgOp.getRegion());
+  return hasOnlyOp<arith::AddFOp>(linalgOp.getRegion());
```
rengolin, line 406: Why this change?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -385,7 +423,12 @@ bool isTppIdentity(linalg::GenericOp linalgOp) {
     return false;
   if (!isUnaryOp(linalgOp))
     return false;
-  return hasCopySemantics(linalgOp);
+  if (!hasCopySemantics(linalgOp))
+    return false;
+  auto res = verifyTppIdentityBroadcastingRules(
+      linalgOp.getDpsInputOperand(0)->get().getType(),
```
rengolin, line 429: `linalgOp.getOperand(0).getType()`?

test/Conversion/LinalgToTpp/linalg-to-tpp.mlir
```diff
@@ -556,3 +556,171 @@ func.func @relu_mapping_strided(%i: memref<4x4xf32>, %o: memref<4x4xf32, strided
...
+
+// -----
+
+#map = affine_map<(d0, d1) -> ()>
+#map1 = affine_map<(d0, d1) -> (d0, d1)>
+
+// CHECK-LABEL: func.func @empty_map_identity
+// CHECK-SAME: %[[ARG0:.+]]: f32, %[[ARG1:.+]]: memref<8x32xf32>
+func.func @empty_map_identity(%arg0 : f32, %arg1: memref<8x32xf32>) {
+  // CHECK: ins(%[[ARG0]] : f32) out(%[[ARG1]] : memref<8x32xf32>)
```
rengolin, line 642: What about tpp.identity? No checks?

rengolin:
This PR needs to also fix the MHA code in #342

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: We use projectedPermutation for each tpp. So now we check directly in `hasMappingToTppConditions`

test/Conversion/LinalgToTpp/linalg-to-tpp.mlir
chelini: forgot while typing.

rengolin (approved):
Just tested on my MLP branch and it works (ie. doesn't match the wrong generic as a tpp.identity), so I'm happy.

The MHA also seems to work, but the checks are now different and the PR may need updating. @nhasabni make sure that's correct.

PR #359, opened 2023-03-03
Update LLVM

chelini:
Import some propagation fixes. Update `transform-convolutions.mlir` as bubble up in propagation is now more conservative. This choice has been made to avoid re-computation in larger models.


test/Dialect/Transform/transform-convolutions.mlir
```diff
@@ -249,8 +249,6 @@ func.func @walk(%arg0: tensor<1x1x64x64xf32>, %arg1: tensor<3x3x64x64xf32>, %arg
   // CHECK-NEXT:   scf.for {{.*}} = %[[C0]] to %[[C56]] step %[[C1]] iter_args
   // CHECK-NEXT:     scf.for {{.*}} = %[[C0]] to %[[C56]] step %[[C1]] iter_args
   // CHECK-NEXT:       scf.for {{.*}} = %[[C0]] to %[[C32]] step %[[C1]] iter_args
-  // CHECK: linalg.generic
```
rengolin, line 252: What is the end result of this change? Is this a regression in functionality (potential performance)? Do we need to open an issue to handle this later?

rengolin (approved):

test/Dialect/Transform/transform-convolutions.mlir
chelini: We are more conservative on the propagation as it was introducing too much recomputation. Specifically, when we bubble up the pack we require a single users for now. In this case the identity has two users: the unpack and the generic, thus stop to propagate. A solution would be to forward the input of the unpack to 'unpack' into the destination of the identity, but this is not implemented yet, and I still don't know what implications (i.e., bufferization) it will have.

PR #360, opened 2023-03-03
MLP MLIR Generator

### LINKED ISSUE #341: MLP generator
To compare results with libxsmm-dnn, we need to create a similar IR that would be converted to the same network in the DNN side. We need an MLP MLIR generator (`mlp-gen`) and a way to pass that MLIR through `tpp-opt` and/or `tpp-run`.

Here's an example of the command line they use:
```
./mlp_example ${ITER} ${MB} ${FUSE} ${PASS} 32 32 32 3584 1024 512 2048 512
```
Where:
* `${ITER}` means number of iterations to run (after warmup)
* `${MB}` is the MiniBatch size (ex. 256)
* `${FUSE}` is `0` (None), `1` (Bias), `2` (ReLU), `3` (Bias+ReLU)
* `${PASS}` is either `F` (Forward), `B` (Backward), `A` (Both)
* `32 32 32` is the tile sizes for the blocking (`N`, `C`, `K`). This needs to be passed to `tpp-opt`.
* `3584...512` is the size of the layers: input + hidden + output. This command has 5 (1 + 3 + 1).

More details on the command line parsing: https://github.com/libxsmm/libxsmm-dnn/blob/main/tests/mlp/mlp_example.c#L128

We need a script that will take that command line, call the right tools with the right options, save all intermediate states (original IR, optimized IR, input, output) and call both libxsmm and tpp-mlir sides.

@alheinecke

rengolin:
MLP MLIR Generator

Generator for multiple layers of MLP model, with bias add and ReLU,
selectable mini batch, layer sizes, etc. Supports FP32 and BF16, any
number of layers (>=2), etc.

Flags for emitting softmax, use bias as matmul accumulation, general
blocking and packing. To VNNI pack, pipe to `tpp-opt -pack-vnni`.

To execute, pipe to `tpp-run -e entry -entry-point-result=void -print`.

To replicate the libxsmm-dnn arguments, there's a `generate.sh` script
that support similar arguments. For example:
```
$ ./utils/mlp-gen/generate.sh 10 64 3 F 32 32 32 64 64 64
```
Generates a model with MB=64, tiling factors 32, layers=128 in FP32,
passes through `tpp-opt` and then `tpp-run`, saving the intermediate
MLIR states.

Note: This implementation isn't complete or robust by any means and
should not be relied upon for assessing our code generation quality at
the moment. But our goal should be to make this generator our go-to
solution for making sure our code generation quality is on par with
existing solutions (libxsmm/OneDNN).

Fixes #341


alheinecke:
I don't see that it can generate blocked & VNNI layouts. This is crucial.

rengolin:
> I don't see that it can generate blocked & VNNI layouts. This is crucial.

`tpp-opt` does not yet have `pack-matmul` in the pipeline. That's why I mentioned in the commit message to use `tpp-opt` in the middle. 

I can add that to the script if necessary, but `tpp-opt` isn't stable enough to receive generic IR and do the right thing every time, so I didn't do that by default. That's also why I save the intermediate results in MLIR files, so we can do more stuff with it.

rengolin:
Right, we want this tool to be not only a generator for the "perfect" IR, but also for the "naive" IR that the compiler needs to match. This will also help us test the default pipeline, making sure any IR we generate here can be optimally executed in hardware, or _at the very least_ not crash the compiler.

Right now, all we have is the "naive" IR. (your number 4 above)

The things we discussed offline were:
1. Generate packed versions (block / VNNI) directly, for input and weights, via a flag. (Assume VNNI is on by default for FP16) (your number 2/3 above)
2. Add a flag to not pack inputs at generation time, but add `pack` / `unpack` instructions for the input. (your number 1 above)
3. Don't emit `softmax` by default, use a flag to enable it.
4. Allow 2 layers only (no hidden layer, no bias add, no relu) for matmul performance.
5. Add flags for common optimizations, ex. move bias add to matmul init (replacing zeroes).

Right now, it's a bit of a pain to rebase because I have moved some stuff from `TppUtils` around and others from `MLIRBench` into a new file, so anything that touches those files (the former is particularly popular this week) makes my rebases complicated.

So we can merge this now and I create a new issue to track further changes, or I can keep this PR open until they're all done, with the cost of added rebase conflicts. Whatever's best for everyone.

rengolin:
I've done all but the unpacked input option, and I think it should be fine for now.

I've added a few tests on possible combinations, they all seem to work. The packing of tensors isn't using the same analysis as the pass, so it may not be right for all possibilities, but for what we have now, it works.

I think we should merge and continue working on it in-tree, because there are some changes that need others to look at and experiment to find out, and I think we'll be more productive doing that in-tree than just me, out-of-tree.

@alheinecke

rengolin:
@KavithaTipturMadhu Can you have a look?

rengolin:
Ok folks, this is a good seed and the rebases are getting unwindely. Can we merge this now and continue the work in tree?

adam-smnk:
Some issues I noticed:
- `mlp-gen` generates the same constants for any value of `seed`
- a model that crashes tpp-run:
`./bin/mlp-gen  --float-width=16 --seed=14 --mini-batch=64 --layers=64,64,64 --tiles=32,32,32`
Note that this works with `f32` or `bf16` with only 2 layers instead of 3.

rengolin:
> * `mlp-gen` generates the same constants for any value of `seed`

Yeah, I noticed it too. Looking into.

> * a model that crashes tpp-run:
>   `./bin/mlp-gen  --float-width=16 --seed=14 --mini-batch=64 --layers=64,64,64 --tiles=32,32,32`
>   Note that this works with `f32` or `bf16` with only 2 layers instead of 3.

What crashes? Generation? Execution?

Also, not all tile sizes work with all layer sizes. I'll check.

adam-smnk:
> What crashes? Generation? Execution?

Execution when the generated model is passed as input tpp-run. The same combination of layer and tile sizes works with f32. But if it makes no sense for bp16 then it's ok.

adam-smnk (approved):
Code overall looks fine to me. The current issues and limitation don't look like showstoppers to me and they can be addressed in the future in follow-up issues.
While I tested the generator a bit, I haven't used it extensively. Probably others should double check that the generated code is within expectations for MLP especially compared to models tested in oneDNN.

PR #363, opened 2023-03-06
Default pipeline - linalg to loops

adam-smnk:
Adds `linalg-to-loops` option to the default pipeline and `tpp-run` to skip linalg conversion to TPP and go directly to loops for testing and validation purposes.

### DIFF
--- a/test/Passes/DefaultPipeline/linalg-to-loops.mlir
+++ b/test/Passes/DefaultPipeline/linalg-to-loops.mlir
@@ -1,5 +1,49 @@
 // RUN: tpp-opt %s -default-tpp-passes="linalg-to-loops" -split-input-file | FileCheck %s
 
+// Direct linalg lowering to loops should leave all TPP operations untouched.
+// CHECK-NOT: func.func private @xsmm_
+// CHECK: func.func @tpp_ops(
+// CHECK-SAME:  %[[ARG0:[^ ]+]]: memref<3x3xf32>,
+// CHECK-SAME:  %[[ARG1:[^ ]+]]: memref<3x3xf32>,
+// CHECK-SAME:  %[[ARG2:[^ ]+]]: memref<1x1xf32>)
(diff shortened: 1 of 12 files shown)


chelini (approved):
`git-clang-format HEAD~1` allows to run clang format only on the lines you touched.

adam-smnk:
> `git-clang-format HEAD~1` allows to run clang format only on the lines you touched.

I format on save so, I just let it do whatever it wants 😜

adam-smnk:
@rengolin as discussed, linalg-to-loops now also skips all TPP mapping steps. I also moved `MLIRBench` options to a separate struct as they started to grow in number and I think it's a bit cleaner this way.

test/Passes/DefaultPipeline/linalg-to-loops.mlir
```diff
@@ -0,0 +1,237 @@
...
+
+// Direct linalg lowering to loops should leave all TPP operations untouched.
+// CHECK-NOT: func.func private @xsmm_
+// CHECK: func.func @tpp_ops(
+// CHECK-SAME:  %[[ARG0:[^ ]+]]: memref<3x3xf32>,
+// CHECK-SAME:  %[[ARG1:[^ ]+]]: memref<3x3xf32>,
+// CHECK-SAME:  %[[ARG2:[^ ]+]]: memref<1x1xf32>)
+func.func @tpp_ops(%arg0: memref<3x3xf32>, %arg1: memref<3x3xf32>, %arg2: memref<1x1xf32>) {
+  // CHECK: tpp.identity
+  // CHECK: tpp.add
```
rengolin, line 11: I wasn't expecting to see tpp ops at all here...

adam-smnk: It is a test case where we explicitly input TPP ops to check that they remain unchanged when `linalg-to-loops` is enabled.
Just to prevent a regression in case TPP mapping remains disabled but somehow TPP lowering slips into this mode.

rengolin: Makes sense!

rengolin (approved):

PR #370, opened 2023-03-06
Fix mapping to relu 

chelini:
Add an extra operand (OperandInfo) to `isTppRelu` that returns the input and output value. Drop `isUnaryOp` and `isBinaryOp` since this is not always the case - a linalg.generic representing a relu can also be ternary. The non-constant operand (the input of tpp relu) is obtained by looking at the linalg region and extracting the OpOperand tied to the block argument that represent a non constant value. We could follow a similar approach for other tpps, only looking a the scalar operation inside the region and get the input/output operands tied to the block arguments.


lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -343,22 +345,41 @@ static bool isUnaryOp(linalg::GenericOp linalgOp) {
...
-  }
+  Operation *regionOp = &genOp.getRegion().front().front();
+  auto maxfOp = cast<arith::MaxFOp>(regionOp);
+  Value maxfLhs = maxfOp.getLhs();
+  Value maxfRhs = maxfOp.getRhs();
+
+  if (isZeroTensor(maxfLhs) && isZeroTensor(maxfRhs))
+    return false;
+
+  if (isZeroTensor(maxfLhs)) {
```
rengolin, line 364: This could have been a lambda passed with alternating arguments

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -343,22 +345,41 @@ static bool isUnaryOp(linalg::GenericOp linalgOp) {
...
-      if (isZeroTensor(maxfOp.getLhs()) || isZeroTensor(maxfOp.getRhs()))
-        return true;
-    }
-  }
+  Operation *regionOp = &genOp.getRegion().front().front();
+  auto maxfOp = cast<arith::MaxFOp>(regionOp);
+  Value maxfLhs = maxfOp.getLhs();
+  Value maxfRhs = maxfOp.getRhs();
+
+  if (isZeroTensor(maxfLhs) && isZeroTensor(maxfRhs))
```
rengolin, line 361: This looks like a silly special case, and to be honest, `maxf(0.0, 0.0)` is still "technically" a ReLU.

In any case, lowering that to TPP would be faster than defaulting to loops, so I think it's best to just let this one slide into the cases below.

rengolin (commented):
Some inline comments, but otherwise looks good.

rengolin:
> If we believe this is the right direction. I can re-work the other tpp mapping to follow this approach.

I was thinking about that, too. Your fix looks good for the existing ops, but as we discussed last week, as the number of TPP ops increase, this pick & choose approach won't scale. 

What we really need is a more generic framework that can map to any element-wise, including relu, gelu, tanh, etc. But this can / should be done once we have a better design of the TPP dialect.

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: Perhaps mathematically it counts as a ReLU if you squint hard enough. But in reality it is just a zero fill of the output buffer so, I imagine a simple memset should be faster than TPP.

adam-smnk: *than `tpp.relu`. Although, okay - I admit I'm not 100% sure about this claim. And there is a TPP Zero operation which could be better.
Anyway, if we apply `tpp.relu` on a random uninitialized buffer, we might end up with a different result than zero initialization.

rengolin: > on a random uninitialized buffer, we might end up with a different result than zero initialization

"undefined behaviour is undefined" :smile: 

> And there is a TPP Zero operation which could be better.

Another reason why mapping one by one, running all comparators for all options won't scale.

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: +1 on unifying/rewriting these two if cases into one

adam-smnk (approved):
lgtm

PR #373, opened 2023-03-08
Multiple MLIR dump stages in tpp-run

### LINKED ISSUE #365: Split `tpp-run --dump-mlir` into early/late dumps
We should split the `tpp-run --dump-mlir` into early/late IR, so that we can see the optimized kernel IR and the generated main function, before lowering to LLVM.

The current one is the early one, while the late one is just before [the first LLVM dialect conversion](https://github.com/plaidml/tpp-mlir/blob/main/tpp-run/MLIRBench.cpp#L396).

@adam-smnk @KavithaTipturMadhu

rengolin:
Allow to dump MLIR from different stages of the compilation process while running it:
 * Early: After main code generation, before optimization
 *  Late: After optimization, before LLVM dialect conversion
 *  LLVM: After LLVM dialect conversion, before LLVM IR lowering

Dumping the LLVM IR is a separate flag and should remain so. For debug purposes we may want to dump both.

Fixes #365

### DIFF
--- a/test/Integration/tpp-run-splat-memref.mlir
+++ b/test/Integration/tpp-run-splat-memref.mlir
@@ -1,8 +1,8 @@
-// RUN: tpp-run %s -e entry -entry-point-result=void -print-mlir=early | \
+// RUN: tpp-run %s -e entry -entry-point-result=void -print-mlir=early 2>&1 | \
 // RUN: FileCheck %s --check-prefix=SPLAT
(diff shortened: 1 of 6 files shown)


rengolin:
@adam-smnk I shall change this PR once #356 is merged to use your stages.

adam-smnk (approved):
Nice solution +1

Issue #369, opened 2023-03-06
Move all passes to the default pipeline


rengolin:
We have a number of passes that are not connected to the default pipeline. While the default pipeline was created after some of those passes, it's about time we make sure all of our work can be used directly without magic flags. We also need to demonstrate all our passes interoperate with all our other passes in a non-destructive way.

The work here is to:
1. Go through every pass that is not in the default pipeline yet.
2. Find the best place for it to be in the default pipeline and move it there.
3. Discuss appropriate flags and options and how to pas this from `tpp-opt` and `tpp-run`.
4. Add/move execution test to the Integration directory, using `tpp-run` and compare outputs.

There may be a few passes that do not belong in the default pipeline. For example, the transform pass, or #367. For those cases, we need a clear and reasonable reason not to include and document it in the header file (so that we don't come back here again).

@chelini @adam-smnk @KavithaTipturMadhu

adam-smnk:
> > `element-wise-fusion` which is no longer needed by any other pass and can be left out of the default pipeline for now.
> 
> #400 shows that this might not be true on all cases.

Fair point. It can be still beneficial to fuse linalg ops we don't lower to TPP.
I'll have to see where it could be added in the pipeline. Maybe in the `post-processing` or a separate `post-opt` pass as this fusion doesn't really fit to any of the existing pipeline stages.

adam-smnk:
TODO: add `combine-tpp` to the default pipeline when the pass itself is fixed
The pass be added at the end of `TppConversionPass` after all other ops have already been lowered to TPP.

rengolin:
Is there anything left over from the previous passes? If not, I guess we can close this for now.

PR #380, opened 2023-03-09
Default pipeline - add packing

### LINKED ISSUE #361: Add `pack-matmul` and `pack-vnni` to the default pipeline
When creating the MLP generator (#360), I noticed the matmuls weren't being packed on `tpp-opt --default-tpp-passes`. It seems those passes aren't on our standard pass.

We need to add them to the default pipeline so that we can automatically optimise MLP without the need of running `tpp-opt` twice.

We may need to pass the tile sizes to the pass, so this could be an extra option on `tpp-opt` and `tpp-run`?

@adam-smnk

adam-smnk:
Adds various existing packing passes to the default pipeline.
These passes are run with either their own default blocking factors or values hardcoded for the default pipeline.
Additionally, other support passes are added to improve packing performance such as pack propagation or folding.

The default matmul packing is disabled in some benchmarks in which the pass highly reduces performance.

Fixes #361 
Work towards #369

### DIFF
--- a/test/Passes/DefaultPipeline/tpp-mapping.mlir
+++ b/test/Passes/DefaultPipeline/tpp-mapping.mlir
@@ -51,3 +51,39 @@ func.func @pack_vnni(%arg0: tensor<32x4x4xbf16>, %arg1: tensor<32x4x4xbf16>, %ar
 // CHECK:       linalg.transpose
 // CHECK:       tensor.insert_slice
 // CHECK: vnni.brgemm
+
+// -----
+
+func.func @pack_matmul(
+  %arg0: tensor<128x128xf32>, %arg1: tensor<128x128xf32>, %arg2: tensor<128x128xf32>)
+    -> tensor<128x128xf32> {
+  %0 = linalg.matmul  ins(%arg0, %arg1: tensor<128x128xf32>, tensor<128x128xf32>)
+                     outs(%arg2: tensor<128x128xf32>)
+    -> tensor<128x128xf32>
+  return %0 : tensor<128x128xf32>
+}
+
+// CHECK-LABEL: func.func @pack_matmul(
+// CHECK-NOT: linalg.matmul
+// Generalized pack of the first input
+// CHECK: scf.for
+// CHECK:   scf.for
+// CHECK:     tensor.extract_slice
+// CHECK:     linalg.transpose
+// CHECK:     tensor.insert_slice
+// Generalized pack of the second input
+// CHECK: scf.for
+// CHECK:   scf.for
+// CHECK:     tensor.extract_slice
+// CHECK:     linalg.transpose
+// CHECK:     tensor.insert_slice
+// Generalized pack of the output
+// CHECK: scf.for
+// CHECK:   scf.for
+// CHECK:     tensor.extract_slice
+// CHECK:     linalg.transpose
+// CHECK:     tensor.insert_slice
+// Packed matmul
+// CHECK: linalg.generic
+// CHECK:   arith.mulf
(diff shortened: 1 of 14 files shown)


rengolin (commented):
I'm surprised this didn't change any of the existing tests, especially in the `Models` directory. Perhaps we should convert all integration tests to the default pass sooner than later.

adam-smnk:
> I'm surprised this didn't change any of the existing tests, especially in the `Models` directory. Perhaps we should convert all integration tests to the default pass sooner than later.

Don't worry, some models do fail now. That is why I need to common up dispatch calls before I start updating like 100 of the existing dispatch checks 😜

test/Passes/DefaultPipeline/tpp-mapping.mlir
```diff
@@ -1,13 +1,89 @@
...
 }
 
 // CHECK-LABEL: func.func @conv_to_matmul(
 // CHECK-NOT: linalg.conv_2d_nhwc_hwcf
-// CHECK: linalg.matmul
+// CHECK: scf.for
+// CHECK:   scf.for
+// CHECK:     scf.for
+// CHECK:       scf.for
+// CHECK:         tensor.extract_slice
```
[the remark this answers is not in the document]
adam-smnk: Added

adam-smnk:
Default matmul packing currently completely decimates benchmark performance so, it'll be disabled there.
However, `pack-matmul` will remain in the default pipeline as long term we still need this pass. In some cases, packing is necessary to enable other transformations such as VNNI packing. The results of matmul packing will remain tested in various passes tests' IR checks.

test/Passes/DefaultPipeline/tpp-mapping.mlir
```diff
@@ -1,13 +1,90 @@
-// RUN: tpp-opt %s -tpp-mapping -split-input-file | FileCheck %s
+// RUN: tpp-opt %s -tpp-mapping -cleanup -split-input-file | FileCheck %s
```
rengolin, line 1: Is this cleanup just for the test or is it really necessary? If the latter, then it should go in the pipeline after `createGeneralizeTensorPackAndUnPackPass`.

rengolin (commented):
We may have to break the `TppMappingPass` later, but it's ok for now.

rengolin:
> Default matmul packing currently completely decimates benchmark performance so, it'll be disabled there. However, `pack-matmul` will remain in the default pipeline as long term we still need this pass. In some cases, packing is necessary to enable other transformations such as VNNI packing. The results of matmul packing will remain tested in various passes tests' IR checks.

It may sound like a silly test, but that's a good canary to show us the side-effects of pack/tile transforms.

test/Passes/DefaultPipeline/tpp-mapping.mlir
adam-smnk: It is already present in the default pipeline as we run cleanup after every major stage i.e., mapping, lowering etc.
It is mainly for the test as it produces a bit nicer IR. But you're right, it should be in the mapping pass to limit number of possible combinations.

adam-smnk:
Canary performance regression due to matmul packing is now captured in #410.
Cause of performance regression in the bf16 MLP models is unknown. However, the compilation pipeline cannot be currently optimized for them due to the runtime issue #409.

rengolin (approved):
There are some performance regressions but we need to merge this before the dungeon. I'll approve this so we can do that, but we need to make sure we come back from the crash in performance by spotting and fixing all of the issues in the default pipeline (not manual runs, special flags, etc).

PR #382, opened 2023-03-10
Default pipeline - common up ops after LICM

### LINKED ISSUE #377: Hoist and common up `xsmm_dispatch` calls
The TPP to XSMM lowering always creates a pair of `dispatch` / `invoke` for each TPP op. Some of the `dispatch`es do get hoisted out of the loops, but if we have multiple dispatches with the same arguments, we can common them all up and reuse the same SSA function pointer value.

This avoids a call + a few checks and just lowers the indirect call to the `invoke` with the new data pointers.

adam-smnk:
Adds extra cleanup stage directly after LICM and before lowering any ops to function calls to allow CSE to eliminate common operations now that they are hoisted out of loops e.g., eliminate repeated XSMM dispatch calls with the same arguments.

Resolves #377

### DIFF
--- a/test/Models/multi-head-attention.mlir
+++ b/test/Models/multi-head-attention.mlir
@@ -33,17 +33,6 @@
 !multi_head_attention_output_tensor_t  = tensor<32x8x128xf32> // batch_size, embedding_size, seq_length
 !tensor_print_t = tensor<1x8xf32>
 
-
-// CHECK: func.func private @xsmm_binary_invoke(i64, i64, memref<*xf32>, memref<*xf32>, memref<*xf32>)
-// CHECK-DAG: func.func private @xsmm_binary_dispatch(i64, i64, i64, i64, i64, i64, i64, i64) -> i64 
-// CHECK-DAG: func.func private @xsmm_unary_invoke(i64, i64, memref<*xf32>, memref<*xf32>)
-// CHECK-DAG: func.func private @xsmm_unary_dispatch(i64, i64, i64, i64, i64, i64, i64) -> i64
-// CHECK-DAG: func.func private @xsmm_matmul_invoke(i64, i64, memref<*xf32>, memref<*xf32>, memref<*xf32>)
-// CHECK-DAG: func.func private @xsmm_matmul_dispatch(i64, i1, i64, i64, i64, i64, i64, i64) -> i64
-//
-// CHECK-LABEL: @multi_head_attention(
-// CHECK-SAME: %[[arg:.*]]: memref<32x8x128xf32>, %[[arg1:.*]]: memref<1x8xf32>) {
-//
 func.func @multi_head_attention(
         %input : !multi_head_attention_input_tensor_t, %output : !tensor_print_t) -> !tensor_print_t {
     %cst = arith.constant 0xFF800000 : f32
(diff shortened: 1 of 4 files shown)


adam-smnk:
Models tests are reworked a bit to make it easier to "automatically" generate CHECK statements using grep and a bitof find/replace magic. Imo, Models tests are still not easily maintainable but I don't have any better ideas for now.
Well... initially I tried to simplify these tests a bit but then FileCheck seemed too limited to accommodate that as, for example, there's no support for `CHECK-DAG-COUNT-N`.

test/Models/multi-head-attention.mlir
```diff
@@ -92,59 +71,36 @@ func.func @multi_head_attention(
...
     %expanded_22 = tensor.expand_shape %83 [[0, 1], [2, 3]] : tensor<256x128xf32> into tensor<32x8x2x64xf32>
     %84 = tensor.empty() : tensor<32x8x2x64xf32> 
     %85 = linalg.generic {indexing_maps = [affine_map<(d0, d1, d2, d3) -> (d2, d3)>, affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%transformer_layer_0_self_attention_key_bias : tensor<2x64xf32>) outs(%84 : tensor<32x8x2x64xf32>) {
     ^bb0(%in: f32, %out: f32):
       linalg.yield %in : f32
     } -> tensor<32x8x2x64xf32>
     %86 = tensor.empty() : tensor<32x8x2x64xf32>
-    //
-    // CHECK: %[[DISPATCH2:.+]] = call @xsmm_unary_dispatch(%[[C1]], %[[C2]], %[[C64]], %[[C64]], %[[C64]], %[[C1]], %[[C0]]) : ({{.+}}) -> i64
-    // CHECK: func.call @xsmm_unary_invoke(%{{.+}}, %[[DISPATCH2]], %{{.+}}, %{{.+}})
```
adam-smnk, line 108: Oops, I removed binary/unary checks

rengolin:
> Models tests are reworked a bit to make it easier to "automatically" generate CHECK statements using grep and a bitof find/replace magic. Imo, Models tests are still not easily maintainable but I don't have any better ideas for now. Well... initially I tried to simplify these tests a bit but then FileCheck seemed too limited to accommodate that as, for example, there's no support for `CHECK-DAG-COUNT-N`.

I think these models don't need thorough checks at all. We can let thorough checks to pass/integration tests.

So, here, just checking for either xsmm call or linalg op should be enough to show us what gets lowered and what doesn't.

rengolin:
> Yeah, ideally we'd check it some general statistics like overall number of XSMM calls, some individual loops checks to validate correct blocking etc. But pure FileCheck isn't really suitable for this kind of validation.

I see, that's why you mentioned `DAG+COUNT`. Given that those are all linear, just checking for the ops at the right places (no args) should be enough and not change unless there's new patterns matching (ex. commoning up dispatches, but those are easy to change).

test/Models/multi-head-attention.mlir
adam-smnk: Readded

adam-smnk: But I should add them in the two other models too

adam-smnk: Updated

adam-smnk:
New tests cases were generated using the following .sh script (uploaded as .txt because of github rules):
[generate_model_checks.txt](https://github.com/plaidml/tpp-mlir/files/10942743/generate_model_checks.txt)

Adjust original grep as needed.

adam-smnk:
> I see, that's why you mentioned `DAG+COUNT`. Given that those are all linear, just checking for the ops at the right places (no args) should be enough and not change unless there's new patterns matching (ex. commoning up dispatches, but those are easy to change).

There's definitely some balance between going in too far in either direction (too general vs too explicit) which we still have to figure out.
I know I'll have to update these tests again after packing is added to the pipeline. It'll be good test to see how the checks break again and how many have to be updated. If it is still quite problematic, I'll try to simplify the checks a bit further e.g., remove all args from invokes.

rengolin:
> If it is still quite problematic, I'll try to simplify the checks a bit further e.g., remove all args from invokes.

My concerns with this strategy are:
* The update script isn't checked in
* It's hard to validate the correctness of the update script. Regex isn't the easiest thing to parse.
* It's hard to look at the FileCheck output when the test breaks to see if the error is meaningful or just formatting
* Most of the time it will be the latter, and we can waste a lot of time (during development and review) looking at it
* This could encourage developers to just update the checks without looking
* It also encourages reviewers to hope the author did the checking and not bother

LLVM has a similar set of scripts (`update_*_checks.py` [here](https://github.com/llvm/llvm-project/tree/main/llvm/utils)), and while it's quite useful there, it's also maintained in-tree and used by a lot of different groups (less bias effect).

My proposal here is to not bother with any of that at the `Models` level, just check for basic calls, and to bother with that at the `Dialect`, `Convert` and `Integration` level, writing actual checks like this one.

The reasoning behind it is that we'll have to worry about a lot less checks, but still _the same_ checks. Larger models may have _"more of the same"_ and not necessarily increase signal **quality** only **quantity**.

I think your script is still worth having in the repo, for the other fine-tuning tests. Perhaps for a different PR. Also, perhaps a single perl script would be better than a bash script with repeated perl invocations on the same inline file.

rengolin (approved):
Regardless of what we do with the output, this PR looks good to me. We can update the results later and restart this discussion when they break again.

adam-smnk:
> I think your script is still worth having in the repo, for the other fine-tuning tests. Perhaps for a different PR. Also, perhaps a single perl script would be better than a bash script with repeated perl invocations on the same inline file.

Personally, I'm also not the biggest fan of this approach for the reasons you listed above.
I attached the script to PR mostly for future reference (just in case) as one possible option.
If we decide to go in the direction of having test scripts, it might be better to use a separate script entirely for high-level validation (like number of present XSMM calls) or maybe even use some other testing strategy altogether for validating larger models.

PR #383, opened 2023-03-10
Tpp Add Broadcast operation, combine tpps pass

KavithaTipturMadhu:


chelini:
> If we start creating fused ops we'll end up in combinatorial explosion. If you cannot get away without a new op, I suggest you create some `tpp.broadcast` and apply that to the right operand.
> 
> On the fused brgemm, I think we can match three TPP ops to a single XSMM and remove them all. Even if we do a hard-coded match for now, and generalize it later.

Not sure about `tpp.broadcast` how this op looks like? Will it allocate a new memref?

PR #329, opened 2023-02-15
NFC: Extract slice before brgemm for vnni

KavithaTipturMadhu:
There was an additional extract slice being introduced due to the map dimension check for VNNI. This patch fixes the issue.

PR #386, opened 2023-03-13
Fix brgemm rewrite to VNNI

chelini:
We still need to clarify this but my understanding was that Brgemm VNNI was 3 - 4 - 2. Meaning the first operand was a slice of size 3, the second a slice of size 4 and so on. Unclear why the affine map checks was looking for a number of results equal to 5. This fixes MLP example.

### DIFF
--- a/lib/TPP/RewriteToBatchReduceGemm.cpp
+++ b/lib/TPP/RewriteToBatchReduceGemm.cpp
@@ -180,8 +180,8 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
   SmallVector<Value> slicedOperands;
   for (OpOperand *operand : linalgOp.getDpsInputOperands()) {
     FailureOr<Value> slicedOperand;
-    AffineMap map = linalgOp.getMatchingIndexingMap(operand);
-    if (isVNNILoop && map.getNumResults() == 5) {
+    // In VNNI layout the second operand has size '4'.
+    if (isVNNILoop && operand->getOperandNumber() == 1) {
       slicedOperand = linalgx::utils::getSliceOperand(
           builder, operand, linalgOp, localIvs, valuesToUse, 4);
     } else {
--- a/test/Passes/pass-map-to-brgemm.mlir
+++ b/test/Passes/pass-map-to-brgemm.mlir
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -rewrite-to-brgemm | FileCheck %s
+// RUN: tpp-opt %s -rewrite-to-brgemm -split-input-file | FileCheck %s
 
 #map0 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
 #map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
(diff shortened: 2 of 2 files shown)


lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -180,8 +180,8 @@ getSlicedOperands(OpBuilder &builder, Location loc, ValueRange localIvs,
   SmallVector<Value> slicedOperands;
   for (OpOperand *operand : linalgOp.getDpsInputOperands()) {
     FailureOr<Value> slicedOperand;
-    AffineMap map = linalgOp.getMatchingIndexingMap(operand);
-    if (isVNNILoop && map.getNumResults() == 5) {
+    // In VNNI layout the second operand has size '4'.
+    if (isVNNILoop && operand->getOperandNumber() == 1) {
```
rengolin, line 184: Getting the order of the operands is even flakier than the number of results... There must be a better way of doing this?

chelini: Why? Looking at the op documentation the second operand must be 4d.

rengolin: What matters isn't the documentation, but the match that leads here.

If I'm reading correctly, `isLinalgOpVNNI` can match both ranks 4 and 5, which seems to be the only check before calling this function.

In what cases it is 4 or 5 and can we have tests for both?

chelini: Kavitha, will check tomorrow. My understanding was only 4.

rengolin: https://github.com/plaidml/tpp-mlir/blob/main/lib/TPP/RewriteToBatchReduceGemm.cpp#L213

alheinecke: The whole looks very C&P to me, why do we have 2 routines which are close to being identical for mapping to BRGEMM and VNNI BRGEMM. 
The line 213 @rengolin  is referring to should be removed, this is what I recently discussed with @KavithaTipturMadhu . The check for the rank of the inner dimension of the second operand being equal to the VNNI blocking factor should go into "CheckVNNIStruture".

I also had a look at this: https://github.com/plaidml/tpp-mlir/blob/9311ad0c727c83134c0b894bc608d9d7b6f1ba45/lib/TPP/Dialect/Tpp/TppUtils.cpp#L92 I think we should not name the MatMulBody but rather FMA or Linear.

alheinecke (commented):
In general, I think this file has a lot of code duplication for just mapping 2 BRGEMM variants where one (VNNI) as an auxiliary reduction dimension.

alheinecke:
In summary, this is was structure/access patter need to check:

- A has 3 inner dims [r1][p1][r2]
- B has 4 inner dimes where r3 is exactly the VNNI blocking factor and r4=r2/r3 without remainder [r1][r4][p2][r3] 
- C has 2 inner dims [p1][p2]

(except for the r3 shebang this is all the same checks for regular/flat BRGEMM, so as mentioned above, I'm not happy with the amount of duplication we have right now)

alheinecke:
Going through all PRs why was #329 never merged.... I assume because it was the similar partial fix as here?

KavithaTipturMadhu:
> In summary, this is was structure/accesspattern routines need to check (assuming row-major):
> 
> * A has 3 inner dims [r1][p1][r2]
> * B has 4 inner dimes where r3 is exactly the VNNI blocking factor and r4=r2/r3 without remainder [r1][r4][p2][r3]
> * C has 2 inner dims [p1][p2]
> 
> (except for the r3 shebang this is all the same checks for regular/flat BRGEMM, so as mentioned above, I'm not happy with the amount of duplication we have right now)
> 
> So, IMHO, the entire isLinAlgOpVNNI is not needed (
> 
> https://github.com/plaidml/tpp-mlir/blob/9311ad0c727c83134c0b894bc608d9d7b6f1ba45/lib/TPP/RewriteToBatchReduceGemm.cpp#L207
> 
> )

@alheinecke I have gotten rid of isLinAlgOpVNNI method and cleaned up some of the related checks in PR https://github.com/plaidml/tpp-mlir/pull/387.

lib/TPP/RewriteToBatchReduceGemm.cpp
KavithaTipturMadhu: I have renamed hasMatmulBody to hasFMABody in PR https://github.com/plaidml/tpp-mlir/pull/387

KavithaTipturMadhu:
I'll go ahead and approve this change because this is the same as the PR that I raised a few weeks ago https://github.com/plaidml/tpp-mlir/pull/329 without a test case.

KavithaTipturMadhu (approved):

rengolin (changes requested):
This is a partial fix, the whole fix is in #387. Let's close this and merge the other.

rengolin:
Closing as #387 supersedes it.

PR #387, opened 2023-03-14
VNNI Brgemm detection changes

KavithaTipturMadhu:


lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -92,18 +92,13 @@ static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
...
       maps.push_back(map.getMinorSubMap(4));
     } else {
       maps.push_back(map.getMinorSubMap(3));
     }
   }
   for (OpOperand *operand : linalgOp.getDpsInitOperands()) {
     AffineMap map = linalgOp.getMatchingIndexingMap(operand);
-    if (map.getNumResults() < 2)
-      return failure();
     maps.push_back(map.getMinorSubMap(2));
```
chelini, line 102: `maps.push_back(linalgOp.getMatchingIndexingMap(operand).getMinorSubMap(2));`

I think we can drop the temporary variable `map`.

chelini (commented):
Would you mind adding a test case. Maybe the one in https://github.com/plaidml/tpp-mlir/pull/386.

lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -121,6 +116,11 @@ static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
   // Expected access patterns of BRGEMM.
   expectedMaps = infer(
       {{r1, p4, r3}, {r1, r3.floorDiv(*blockingFactor), p5, r2}, {p4, p5}});
+  auto secondOperand = linalgOp.getDpsInputOperands()[1];
+  assert(
```
chelini, line 120: I think this assert should not be here. You are checking the maps in line 124.

include/TPP/Dialect/Tpp/TppUtils.h
```diff
@@ -43,7 +43,7 @@ bool hasTppMark(linalg::LinalgOp linalgOp);
 bool isMarkedWithTpp(linalg::LinalgOp linalgOp, const std::string &target);
 
 // Returns true if the linalg operation has a Matmul region.
-bool hasMatmulBody(linalg::LinalgOp linalgOp);
+bool hasFMABody(linalg::LinalgOp linalgOp);
```
chelini, line 46: Personally I don't like this rename. It gives the wrong impression that we are matching for FMA ops, which is not the case.

lib/TPP/RewriteToBatchReduceGemm.cpp
KavithaTipturMadhu: This is to check that the observed blocking factor is exactly equal to the expected blocking factor.

KavithaTipturMadhu:
> Would you mind adding a test case. Maybe the one in #386.

Rebased on top of #386 patch

include/TPP/Dialect/Tpp/TppUtils.h
alheinecke: your not matching MatMul either... just looking for Mul/Add is not a Matmul.

chelini: `hasAddMulBody`?

rengolin: `hasMulAddBody`? To get the order right? :smile:

lib/TPP/RewriteToBatchReduceGemm.cpp
rengolin: Can you at least common up the shape, so that the assert isn't multi-line?

include/TPP/Dialect/Tpp/TppUtils.h
rengolin: Oh, and don't forget to update the comment, too.

test/Passes/pass-map-to-brgemm.mlir
```diff
@@ -33,3 +33,30 @@ func.func @blocked_matmul(%arg0: tensor<4x16x32x32xf32>, %arg1: tensor<8x16x32x3
     } -> tensor<4x8x32x32xf32>
   return %1 :  tensor<4x8x32x32xf32>
 }
+
```
rengolin, line 36: Don't forget to add the tests requested in https://github.com/plaidml/tpp-mlir/pull/329#issuecomment-1430812339

rengolin (approved):

PR #390, opened 2023-03-14
Provide defaults for tile and fuse pass

chelini:
* Fix `isTppMatmul` in case the operation is a linalg.matmul

* Rework `isBlockedMatmul`. The idea is to provide a better way to understand if a given linalg.generic is a blocked matmul. Do do that we check operands, affine maps and classify the loops. Each loops can belong to three categories only I, J and K. The I loops are present in C and A but not in B and are required to be parallel. The J loops are present in C and B but not in A and must be parallels. The K loops are present in A and B but not in C, and are reductions.

* Provide defaults for tile and fuse pass. linalg.matmul are blocked with tile sizes of 32 along i and j. Blocked convolutions are tile and fused along the three outermost loops to expose a BRGEMM. Similarly, blocked matmul are tiled along the two outermost parallel loops.

### DIFF
--- a/lib/TPP/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/TileConsumerAndFuseProducers.cpp
@@ -117,11 +117,7 @@ struct ConvertToForAll : public OpRewritePattern<scf::ForOp> {
 };
 
 static bool isMatmulLike(Operation *op) {
-  if (isa_and_nonnull<linalg::MatmulOp>(op))
-    return true;
-  if (linalg::GenericOp maybeMatmul = dyn_cast_or_null<linalg::GenericOp>(op))
-    return linalgx::utils::isBlockedMatmul(maybeMatmul);
-  return false;
+  return linalgx::utils::isBlockedMatmul(op) || tpp::utils::isTppMatmul(op);
 }
 
 static bool isConvolutionLike(Operation *op) {
@@ -533,7 +529,20 @@ fuseWithEltwise(RewriterBase &rewriter, TilingInterface consumer,
   return tileAndFuseResult;
 }
 
-static SmallVector<int64_t> getDefaultTileSizes() { return {32, 32}; }
+static SmallVector<int64_t> getDefaultTileSizes(linalg::LinalgOp linalgOp) {
+  // Blocked convolutions are tiled and fused along the three outermost parallel
+  // loops to expose a BRGEMM.
+  if (linalgx::utils::isBlockedConvolution(linalgOp))
+    return {1, 1, 1};
+  // Matmuls are tiled and fused along i and j with 32.
+  if (tpp::utils::isTppMatmul(linalgOp))
+    return {32, 32};
+  // Blocked matmuls are tiled and fused along the two outermost parallel loops
+  // to expose a BRGEMM.
+  if (linalgx::utils::isBlockedMatmul(linalgOp))
+    return {1, 1};
+  return {32, 32};
+}
 
 static Operation *
 getLastFusableConsumer(linalg::LinalgOp linalgOp,
@@ -603,11 +612,15 @@ struct TileConsumerAndFuseProducers
     std::reverse(linalgOperations.begin(), linalgOperations.end());
     llvm::SmallDenseSet<Operation *> visitedConsumers;
     llvm::SmallDenseSet<Operation *> fusionRoots;
+    DenseMap<Operation *, SmallVector<int64_t>> defaultTiles;
     for (linalg::LinalgOp linalgOp : linalgOperations) {
       Operation *rootOp = linalgOp.getOperation();
       if (this->startFromLastFusableConsumer)
         rootOp = getLastFusableConsumer(linalgOp, visitedConsumers);
       fusionRoots.insert(rootOp);
+      defaultTiles[rootOp] = (this->tileSizes.empty())
+                                 ? getDefaultTileSizes(linalgOp)
+                                 : llvm::to_vector(this->tileSizes);
     }
 
     LLVM_DEBUG(llvm::dbgs() << "#fusionRoots: " << fusionRoots.size() << "\n");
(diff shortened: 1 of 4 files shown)


chelini:
Let me know if you want to split in multiple commits.

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -533,7 +529,20 @@ fuseWithEltwise(RewriterBase &rewriter, TilingInterface consumer,
...
 
-static SmallVector<int64_t> getDefaultTileSizes() { return {32, 32}; }
+static SmallVector<int64_t> getDefaultTileSizes(linalg::LinalgOp linalgOp) {
+  // Blocked convolutions are tiled and fused along the three outermost parallel
+  // loops to expose a BRGEMM.
+  if (linalgx::utils::isBlockedConvolution(linalgOp))
+    return {1, 1, 1};
+  // Matmuls are tiled and fused along i and j with 32.
+  if (tpp::utils::isTppMatmul(linalgOp))
+    return {32, 32};
```
adam-smnk, line 539: `pack-matmul` takes 3 values for its block-factors

chelini: Yes, but for tile and fuse only two are necessary

adam-smnk: Oh yes, sorry. That's completely unrelated - all good.

lib/TPP/TransformUtils.cpp
```diff
@@ -208,13 +208,83 @@ bool isBlockedConvolution(Operation *op) {
...
+  for (auto map : indexingMaps)
+    for (auto expr : map.getResults())
+      if (!expr.isa<AffineDimExpr>())
+        return false;
+
+  auto iteratorTypes = linalgOp.getIteratorTypesArray();
+
+  // Make sure all loops are characterized as one of:
+  // - I loop: present in C and A but not in B. I must be parallel.
+  // - J loop: present in C and B but not in A. J must be paralell.
```
adam-smnk, line 245:
```suggestion
  // - J loop: present in C and B but not in A. J must be parallel.
```

lib/TPP/TransformUtils.cpp
```diff
@@ -208,13 +208,83 @@ bool isBlockedConvolution(Operation *op) {
...
+}
+
 bool isBlockedMatmul(Operation *op) {
-  if (!isa<linalg::LinalgOp>(op))
+  auto linalgOp = dyn_cast<linalg::LinalgOp>(op);
+  if (!linalgOp)
     return false;
-  linalg::LinalgOp linalgOp = cast<linalg::LinalgOp>(op);
-  return tpp::utils::hasMatmulBody(linalgOp);
+  if (linalgOp.getNumDpsInputs() != 2 || linalgOp.getNumDpsInits() != 1)
```
rengolin, line 224: We need to stop using magic constants, but this is for another PR.

lib/TPP/TransformUtils.cpp
```diff
@@ -208,13 +208,83 @@ bool isBlockedConvolution(Operation *op) {
...
+  // - I loop: present in C and A but not in B. I must be parallel.
+  // - J loop: present in C and B but not in A. J must be paralell.
+  // - K loop: present in A and B but not in C. K must be reduction.
+
+  llvm::SmallDenseSet<unsigned> cDims = getPreservedDims(indexingMaps.back());
+  llvm::SmallDenseSet<unsigned> bDims = getPreservedDims(indexingMaps[1]);
+  llvm::SmallDenseSet<unsigned> aDims = getPreservedDims(indexingMaps.front());
+  llvm::SmallDenseSet<unsigned> allLoopDims;
+  for (auto cExpr : indexingMaps.back().getResults()) {
+    unsigned cDim = cExpr.cast<AffineDimExpr>().getPosition();
```
rengolin, line 253: Too much repetition here, we need to common these up.

We'll soon start looking more and more into maps to match TPP ops, so we need a stable API to get things like number and order of parallel/reduction loops, correct usage of affine map indices, etc.

For now, just commoning these up in a separate helper should do the trick.

rengolin (commented):
This pass is in the default pipeline right? I imagine it didn't change any tests because the default behaviour is extended, not changed, right?

lib/TPP/TransformUtils.cpp
chelini: I want to make sure that I am dealing with a matul so two inputs 1 output.

lib/TPP/TransformUtils.cpp
chelini: where is the repetition here? The for loops (i.e., line 252-286)?

lib/TPP/TransformUtils.cpp
rengolin: I get what you want, but we keep spreading those checks against magic constants all over the place and that's not good practice. It leads to either misunderstanding what they're checking or copying and pasting elsewhere and failing to update all copies of the same check.

My point is that we need to common those up: Checks, magic constants, semantics of what we expect of operations, etc. As we ramp up the matching for more TPP ops, this will get worse very quickly.

lib/TPP/TransformUtils.cpp
rengolin: Yes. There are three identical and one very similar loops here. This looks like an easy refactory.

chelini: sorry I don't get it. Can you please clarify what you have in mind? Each if uses different sets and check for different conditions.

chelini:
> This pass is in the default pipeline right? I imagine it didn't change any tests because the default behaviour is extended, not changed, right?

yes, it is.

lib/TPP/TransformUtils.cpp
rengolin: The logic for I,J,K loops are the same, you just have to pass the containers by reference, the dim and iterator type by value, and return boolean. At least these could be a local lambda.

rengolin: Discussed offline, it's fine.

adam-smnk (approved):
Looks fine to me after the improvements.

chelini:
Can this be merged @adam-smnk and @rengolin ?

rengolin (approved):
Yes, sorry, was reviewing this and got distracted.

PR #393, opened 2023-03-14
Updating XsmmRunnerUtils w.r.t MatMul changes

nhasabni:
Following changes:

1. Removing _f32 suffix from iree_xsmm_matmul APIs as datatype is an argument to these APIs as well as to tpp.matmul and xsmm.matmul. Datatype can now be passed all the way from MLIR to TPP to XSMM, without the need of datatype specific APIs.

2. Correcting the number of elements and types of params in both dispatch and invoke APIs for MatMul.

*Tests*: These changes work with lowering linalg.matmul into XSMM MatMul via IREE and executing it to produce correct output.


tpp-rt/XsmmRunnerUtils.h
```diff
@@ -71,14 +71,14 @@ extern "C" MLIR_RUNNERUTILS_EXPORT void _mlir_ciface_xsmm_brgemm_invoke(
 extern "C" MLIR_RUNNERUTILS_EXPORT int
 iree_xsmm_brgemm_dispatch_f32(void *context, void *params, void *reserved);
 extern "C" MLIR_RUNNERUTILS_EXPORT int
-iree_xsmm_matmul_dispatch_f32(void *context, void *params, void *reserved);
+iree_xsmm_matmul_dispatch(void *context, void *params, void *reserved);
```
rengolin, line 74: I imagine this will need to be done for brgemm as well.

rengolin:
> 1. Removing _f32 suffix from iree_xsmm_matmul APIs as datatype is an argument to these APIs as well as to tpp.matmul and xsmm.matmul. Datatype can now be passed all the way from MLIR to TPP to XSMM, without the need of datatype specific APIs.
> 2. Correcting the number of elements and types of params in both dispatch and invoke APIs for MatMul.

That's great! I imagine the API has changed since October, so this is a good change. Can you also test BRGEMM and get those through the same API?

> _Tests_: These changes work with lowering linalg.matmul into XSMM MatMul via IREE and executing it to produce correct output.

Would it be possible to add those tests to our IREE fork? In case we want to merge this, they will definitely require some tests.

tpp-rt/XsmmRunnerUtils.h
nhasabni: Done.

rengolin:
@nhasabni FYI, it's OK to have merge branches on PR, just make sure you squash and merge. We try to not have merge branches on main if possible.

PR #394, opened 2023-03-15
Fix propagation

chelini:
We did not compute the starting point of the tiled loops correctly. Use a set to keep track of the indexes of the tiled loops. Later on, if we hit in the set, we know that the point loop is two hops away, as we have the nest in the canonical form ABab.


chelini:
@adam-smnk Cannot use BF16 as I don't have a print function.

Issue #398, opened 2023-03-15
`test/Integration/pack-folding{1}.mlir` problems


rengolin:
A few issues:
 * it does not test the default pipeline, and the pass wasn't added to it
 * it cannot use tpp-run because of memory leaks
 * it should not create temporary files as they leave traces which could crash the next run

Also, as a TODO, it should use a dense<1.0> constant and run the init pass in (#367) once it's done.

@chelini

PR #399, opened 2023-03-15
Clean up integration tests

### LINKED ISSUE #368: Run all Integration tests with `tpp-run` and the default pipeline
Integration tests are meant to demonstrate we can reach a valid execution from different input IR through the same pipeline (the one that users will use).

If we pass different optimizations (ex. through `tpp-opt` into `mlir-cup-runner`), we're not testing the pipeline that users will use. 

Even if at some point, those passes are the same as what we have in the default pipeline, as we improve the pipeline, those tests will become obsolete. They may break for unrelated changes, they may not break when the default pipeline would, but we avoided adding "another test" on the default pipeline because we "already had one".

So, all Integration tests should take various IR files, pass them through the default pipeline, execute them with `tpp-run` and check the results are valid, either with `FileCheck` or with the check dialect.

The tasks here are:
1. Go through all our existing tests and make sure they adhere to these rules.
2. Make sure every pass we have is represented in the integration tests.
3. For new passes / change in passes, we need to add (or improve) at least one integration test to show our work.

@adam-smnk @chelini @KavithaTipturMadhu

rengolin:
We don't need to compare custom passes with the default pipeline now that all our integration tests are working with `tpp-run`. This PR cleans up all the old passes and moves the remaining usages of `mlir-cpu-runner` (except one, see #398, which will be handled separately).

Fixes #368

### DIFF
--- a/test/Integration/conv-to-matmul.mlir
+++ b/test/Integration/conv-to-matmul.mlir
@@ -1,21 +1,6 @@
-// RUN: tpp-opt %s -bufferize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -finalize-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts | \
-// RUN: mlir-cpu-runner \
-// RUN:  -e entry -entry-point-result=void  \
-// RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext | \
-// RUN: FileCheck %s
-//
-
+// This should really be in the passes directory, not here
 // RUN: tpp-opt %s -rewrite-conv-to-matmul-or-brgemm | FileCheck %s -check-prefix=IR 
-//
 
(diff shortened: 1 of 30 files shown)


test/Integration/conv-to-matmul.mlir
```diff
@@ -1,21 +1,6 @@
-// RUN: tpp-opt %s -bufferize -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -finalize-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts | \
-// RUN: mlir-cpu-runner \
-// RUN:  -e entry -entry-point-result=void  \
-// RUN: -shared-libs=%llvmlibdir/libmlir_c_runner_utils%shlibext | \
-// RUN: FileCheck %s
-//
-
+// This should really be in the passes directory, not here
```
adam-smnk, line 1: I think in general it's fine to have some small IR sanity checks in these tests.
Whether these already present are strictly necessary is another question.
Any more insights from your side?

adam-smnk (approved):
+1 - as discussed offline, integration tests have been stable for quite some time now using both tpp-run and mlir-cpu-running so, it is fine to switch fully to using the default pipeline only.

test/Integration/conv-to-matmul.mlir
rengolin: Nope, exactly what you said. Just a comment for the future.

PR #400, opened 2023-03-16
Adding BF16 MLP benchmarks

rengolin:
Adding some temporary, manually packed, BF16/VNNI benchmarks so we can start tracking.

Because of bugs in MLIR bufferization, I had to disable LSAN on those benchmakrs so they stop complaining about the temporary buffers leaked. Also had to XFAIL those tests, as I don't know if it's possible to add the environment to RUN lines (LLVM doens't seem to have this).

Of course this is bad, but we know what it is and how to fix.


rengolin:
image

test/Benchmarks/mlp-bf16-10layers-3584.mlir
```diff
@@ -0,0 +1,213 @@
+// RUN: tpp-opt %s -element-wise-fusion -tile-consumer-and-fuse-producers="tile-sizes=1,1 max-depth=2" |
```
adam-smnk, line 1: `element-wise-fusion` shouldn't be needed anymore with the updated tile-and-fuse pass.

rengolin:
@adam-smnk after I took out the `element-wise-fusion`, I got a small regression in perf:
image

rengolin:
Yup, went back up when I reverted:
image

benchmarks/benchmarks.json
```diff
@@ -13,19 +13,51 @@
...
+  }},
+  {
+  "mlp-vnni-3x1024": {
+    "mlir": {
+      "type": "MLIR",
+      "benchmark": "mlp-vnni-3layers-1536.mlir",
+      "flags": ["--disable-lsan"]
+    }
+  }},
+  {
```
adam-smnk, line 60: Small suggestion: mlp tests could be captured together by a high-level `"mlp": {` same as it is done for `matmul`. Imo, it makes the console output a bit nicer with less white-spaces.

rengolin:
@adam-smnk should be better now.

The plots are a bit broken but it just needs @hfp to update the name matches.

adam-smnk (approved):
I can't really comment on the mlp test models but looks good in terms of infrastructure.

adam-smnk:
Once this PR is merged, I will rebase and adjust packing PR and we should be good to go forward.

PR #401, opened 2023-03-16
Benchmarks - pass arg flags to tpp-run

adam-smnk:
Allows to pass optional tpp-run flags through benchmarks controller. This functionally lets benchmarks use high level tpp-run flags which influence IR generated by the runner.


adam-smnk:
@rengolin this change will allow us to fine tune the default pipeline in benchmarks. For example, to disable matmul packing for now.

adam-smnk:
> 

Actually, that was my only intention. I don't think FIleCheckParser can pick up `run-args` right now.

PR #402, opened 2023-03-16
Add debug options for bufferization

chelini:
This is intended to be used manually (i.e.,
-bufferize="test-analysis-only")


adam-smnk (approved):
Seems fine, it is only for debugging so we don't need any tests.

PR #405, opened 2023-03-17
Revert "Revert "Provide defaults for tile and fuse pass (#390)""

chelini:
Re-apply commit 2db7058b0c6b8a9e218b240ec99f2b8a2a42b94d.

`isBlockedMatmul` assumed all the maps to be permutation maps, but surprise, this is not the case for VNNI packing, where you have a floor-div expression with constant on the B operand. Specially treat the B operand by providing a walker that ensures each other dimension is an AffineDimExpr, and there is one floor-div dimension with constant. This dimension must be a reduction. The rest of the logic remains the same; each loop must be classified in the I, J, or K bucket. I are parallels, J is parallels, and K is reductions.


lib/TPP/TransformUtils.cpp
```diff
@@ -208,13 +209,159 @@ bool isBlockedConvolution(Operation *op) {
...
+  if (llvm::any_of(indexingMaps[1].getResults(),
+                   [&operandBWalker](AffineExpr expr) {
+                     return failed(operandBWalker.visit(expr));
+                   })) {
+    return false;
+  }
+
+  // Make sure all loops are characterized as one of:
+  // - I loop: present in C and A but not in B. I must be parallel.
+  // - J loop: present in C and B but not in A. J must be paralell.
```
KavithaTipturMadhu, line 311: Nit: parallel

KavithaTipturMadhu (approved):

Issue #384, opened 2023-03-13
Fusion and bufferization create extra allocs/copies.


chelini:
Read and writing tensor outside an scf.forall introduce extra copies and allocation. A possible solution is to use iter_args. Note this changes was addressed in https://github.com/llvm/llvm-project/commit/2ed7c3fd841db1ae2a1ae2c3df865b04a890bb0d but reverted in https://reviews.llvm.org/D134411. Why has been reverted is TDB.

``` 
// Simple MLP: I want to tile and fuse the two outermost parallel loops
// the tile factor I want to use is 1, 1.

#map = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d2, d4, d6)>
#map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d1, d2, d6 floordiv 2, d5, d3)>
#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d1, d4, d5)>
#map3 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
#map4 = affine_map<(d0, d1, d2, d3) -> (d1, d3)>
module {
  func.func @mlp(%arg0: tensor<8x48x32x32xbf16>, %arg1: tensor<48x48x16x32x2xbf16>, %arg2: tensor<1536xbf16>, %arg3: tensor<8x48x32x32xbf16>, %arg4: tensor<48x48x16x32x2xbf16>, %arg5: tensor<1536xbf16>, %arg6: tensor<8x48x32x32xbf16>, %arg7: tensor<48x48x16x32x2xbf16>, %arg8: tensor<1536xbf16>, %arg9: tensor<8x48x32x32xbf16>) -> tensor<8x48x32x32xbf16> {
    %cst = arith.constant 0.000000e+00 : bf16
    %0 = linalg.generic {indexing_maps = [#map, #map1, #map2], iterator_types = ["parallel", "parallel", "reduction", "reduction", "parallel", "parallel", "reduction"]} ins(%arg0, %arg1 : tensor<8x48x32x32xbf16>, tensor<48x48x16x32x2xbf16>) outs(%arg3 : tensor<8x48x32x32xbf16>) {
    ^bb0(%in: bf16, %in_0: bf16, %out: bf16):
      %3 = arith.mulf %in, %in_0 : bf16
      %4 = arith.addf %out, %3 : bf16
      linalg.yield %4 : bf16
[...]

Issue #407, opened 2023-03-17
`allOperandsHaveSameShapeAndStrides` too restrictive.


chelini:
```
scf.forall (%arg4, %arg5) = (0, 0) to (128, 512) step (32, 32) {
...
linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} ins(%alloc : memref<32x32xf32>) outs(%subview_2 : memref<32x32xf32, strided<[512, 1], offset: ?>>) {
      ^bb0(%in: f32, %out: f32):
        %0 = arith.maxf %in, %cst : f32
        linalg.yield %0 : f32
      }
```
This should just convert to a tpp.relu. I think the only requirement we need to satisfy is to have stride 1 in the fast-varying dimension.

PR #408, opened 2023-03-17
Enable better bufferization for tile and fuse

chelini:
Replace iterators of the outermost loop with region arguments of the innermost one. The changes avoid later `bufferization` passes to insert allocation within the body of the innermost loop.

Fix: #384

### DIFF
--- a/lib/TPP/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/TileConsumerAndFuseProducers.cpp
@@ -683,18 +683,18 @@ struct TileConsumerAndFuseProducers
 
     auto &ctx = getContext();
     {
-      // Patterns for scf.for
+      // Patterns for scf.for.
       RewritePatternSet patterns(&ctx);
       patterns.add<ReplaceIterArgs>(&ctx);
       (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
     }
 
     {
-      // Patterns for scf.forall
+      // Patterns for scf.forall.
       RewritePatternSet patterns(&ctx);
       if (this->useForAll)
         patterns.add<ConvertToForAll>(&ctx);
-      // fold unit-extent dims for linalg on tensors.
+      // Fold unit-extent dims for linalg on tensors.
       linalg::populateFoldUnitExtentDimsViaSlicesPatterns(patterns);
       tensor::populateMergeConsecutiveInsertExtractSlicePatterns(patterns);
       (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
--- a/lib/TPP/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/TileConsumerAndFuseProducers.cpp
@@ -32,6 +32,10 @@ using namespace mlir;
 
 namespace {
 
+// Replace the iter operand of the outermost loop with the region iter argument
+// of the innermost loop in the region of the innermost loop. This fix-up
+// destination passing style in tile-consumer-and-fuse-producers:
+// https://github.com/llvm/llvm-project/issues/61386
 struct ReplaceIterArgs : public OpRewritePattern<scf::ForOp> {
   using OpRewritePattern<scf::ForOp>::OpRewritePattern;
 
@@ -40,8 +44,8 @@ struct ReplaceIterArgs : public OpRewritePattern<scf::ForOp> {
     assert(outerFor.getNumIterOperands() == innerFor.getNumIterOperands() &&
            "expect same number of iter args");
     Block *block = &(*innerFor.getRegion().begin());
-    for (auto it :
-         llvm::zip(outerFor.getIterOperands(), innerFor.getRegionIterArgs())) {
+    for (auto it : llvm::zip_equal(outerFor.getIterOperands(),
+                                   innerFor.getRegionIterArgs())) {
       Value source = std::get<0>(it);
       Value target = std::get<1>(it);
       rewriter.replaceUsesWithIf(source, target, [&](OpOperand &use) {
--- a/lib/TPP/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/TileConsumerAndFuseProducers.cpp
@@ -32,6 +32,38 @@ using namespace mlir;
 
 namespace {
 
+struct ReplaceIterArgs : public OpRewritePattern<scf::ForOp> {
+  using OpRewritePattern<scf::ForOp>::OpRewritePattern;
+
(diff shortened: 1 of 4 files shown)


test/Passes/pass-mlp-fusion.mlir
```diff
@@ -8,20 +8,25 @@
...
+  // CHECK-SAME: out(%{{.*}} : memref<32x32xf32>)
   %2 = linalg.generic {indexing_maps = [#map2, #map3, #map4], iterator_types = ["parallel", "parallel", "reduction"]} ins(%arg0, %arg1 : tensor<128x256xf32>, tensor<256x512xf32>) outs(%1 : tensor<128x512xf32>) attrs =  {iterator_ranges = [128, 512, 256]} {
     ^bb0(%arg9: f32, %arg10: f32, %arg11: f32):
       %16 = arith.mulf %arg9, %arg10 : f32
       %17 = arith.addf %arg11, %16 : f32
       linalg.yield %17 : f32
   } -> tensor<128x512xf32>
-  // CHECK: tpp.relu ins(%{{.*}} : memref<32x32xf32, strided<[512, 1], offset: ?>>) out(%{{.*}} : memref<32x32xf32, strided<[512, 1], offset: ?>>)
   %c0 = arith.constant 0.0 : f32
+  // We don't map the relu anymore - as allOperandsHaveSameShapeAndStrides fails here.
```
rengolin, line 26: That should be fixed before merging this.

test/Passes/dps-in-tile-and-fuse.mlir
```diff
@@ -0,0 +1,80 @@
+// RUN: tpp-opt %s -tile-consumer-and-fuse-producers -cse | FileCheck %s
+
+#map = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d2, d4, d6)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d1, d2, d6 floordiv 2, d5, d3)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5, d6) -> (d0, d1, d4, d5)>
+#map3 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>
+#map4 = affine_map<(d0, d1, d2, d3) -> (d1, d3)>
+
+// This test check that destination-passing-style is respected by fusion.
+// %arg3 should not be used in the scf.forall region.
```
rengolin, line 10: You should CHECK-NOT for allocs and copies etc which should not be there anymore.

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -650,6 +682,12 @@ struct TileConsumerAndFuseProducers
       }
     }
 
+    // Pattern to replace iter args of the outer most loop with region args of
+    // the inner most one.
+    RewritePatternSet patterns1(&getContext());
```
rengolin, line 687: no silly variable names please.

why can't this be added in the pattern below?

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -33,6 +33,38 @@ using namespace mlir;
 
 namespace {
 
+struct ReplaceIterArgs : public OpRewritePattern<scf::ForOp> {
```
rengolin, line 36: please add a comment explaining what this pass is/does.

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -33,6 +33,38 @@ using namespace mlir;
...
+  void replaceIterArgs(PatternRewriter &rewriter, scf::ForOp outerFor,
+                       scf::ForOp innerFor) const {
+    assert(outerFor.getNumIterOperands() == innerFor.getNumIterOperands() &&
+           "expect same number of iter args");
+    Block *block = &(*innerFor.getRegion().begin());
+    for (auto it :
+         llvm::zip(outerFor.getIterOperands(), innerFor.getRegionIterArgs())) {
+      Value source = std::get<0>(it);
+      Value target = std::get<1>(it);
+      rewriter.replaceUsesWithIf(source, target, [&](OpOperand &use) {
```
rengolin, line 48: this looks awfully generic... without negative tests you can't be sure this won't hit patterns that it shouldn't

chelini: Why is generic here?

test/Passes/pass-mlp-fusion.mlir
rengolin: No, we cannot keep adding regressions to add features. We need to make forward progress. There are a number of issues that were created that way months ago and haven't been addressed yet.

lib/TPP/TileConsumerAndFuseProducers.cpp
rengolin: It seems this will replace the args of all loops in the contained function, even if they're not related to the pass you just ran, even if you didn't tile.

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -33,6 +33,38 @@ using namespace mlir;
 
 namespace {
 
+struct ReplaceIterArgs : public OpRewritePattern<scf::ForOp> {
+  using OpRewritePattern<scf::ForOp>::OpRewritePattern;
+
+  void replaceIterArgs(PatternRewriter &rewriter, scf::ForOp outerFor,
+                       scf::ForOp innerFor) const {
+    assert(outerFor.getNumIterOperands() == innerFor.getNumIterOperands() &&
```
rengolin, line 41: you could just use `zip_equal`

lib/TPP/TileConsumerAndFuseProducers.cpp
chelini: No the pattern works only on tiled loop. See attributes checks above.  If other loops have the same issue we should apply this pattern anyway.

rengolin (dismissed):
Please fix #407 before merging this.

test/Passes/pass-mlp-fusion.mlir
chelini: Will still fails. Once I rebase to the new flow, it should work.

adam-smnk (approved):
Current state looks good to me

chelini:
> Please fix #407 before merging this.

#407 has been fixed, and the PR is approved. I will land this and if you have more concerns I will fix in tree.

PR #419, opened 2023-03-23
Initial TPP dialect document

rengolin:
As discussed in the meetings, this is a first approach. Not complete, not perfect, but has enough of the ideas to serve as a good start. We should update this document when we agree to specific syntax/semantic changes, so that it serves as our official "spec".


docs/TPPDialect.md
```diff
@@ -0,0 +1,355 @@
...
+  %M = tpp.matmul ins(%A, %B, %C) : (tensor<>, tensor<>, tensor<>) -> tensor<>
+```
+
+All `memref` operations follow the `linalg` syntax of `ins` and `outs`.
+The `outs` values are an indication of where the output is written to.
+These are generated by the bufferization process, from the `tensor` level operations.
+```mlir
+  // Out-of-place ops output on new allocations
+  %buf = memref.alloc : memref<>
+  tpp.add ins(%arg0, %arg1) outs(%buf) : (memref<>, memref<>) -> memref<>
```
chelini, line 41: here `-> memref<>` no needed.

docs/TPPDialect.md
```diff
@@ -0,0 +1,355 @@
...
+GEMM operations must have compatible types (MxNxK, VNNI, etc.).
+
+### Ternary
+
+On `tensors`, ternary ops have three arguments and one return value.
+
+On `memrefs`, ternary operations have both `ins` and `outs`, with always three `ins` and one `outs` arguments.
+The `outs` can alias with any type-compatible `ins` argument (accumulation matrix).
+
+A ternary `matmul` is equivalent to `C = A x B + C`, with `C` as an input and a potential output.
```
adam-smnk, line 78: This can be split into two separate operations `matmul` and `add` so, no need for ternary variant.

docs/TPPDialect.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Binary
+
+On `tensors`, binary ops have two arguments and one return value.
+
+On `memrefs`, binary operations have both `ins` and `outs`, with always two `ins` and one `outs` arguments.
+On most binary ops, the `outs` can alias `ins` (ex. `a=a+b`), and `ins` can alias themselves (ex. `a=a+a`).
+
+The exception is `matmul`, where the output cannot alias the inputs.
+A binary `matmul` is equivalent to `C = A x B`, with `C` as a new `memref` allocation.
```
adam-smnk, line 65: We probably need to also add zero initialization to new buffer allocation or reuse.

docs/TPPDialect.md
```diff
@@ -0,0 +1,355 @@
...
+
+TPP uses `tensor` and `memref` as its default data type.
+The supported element types are `fp32` and `bf16`.
+
+For advanced operations (ex. higher-order), we have the need to return special sparse tensors.
+We're still investigating what representation to use or if the existing sparse `tensor` attributes are enough.
+
+We may create additional types, as needed.
+For example:
+ * `vnni18`, `vnni8` for VNNI shapes of `bf16` and `bf8`. Those are _packed_ floats, ex. `{ bf16, bf16 }`.
```
adam-smnk, line 102:
```suggestion
 * `vnni16`, `vnni8` for VNNI shapes of `bf16` and `bf8`. Those are _packed_ floats, ex. `{ bf16, bf16 }`.
```

adam-smnk: It also could be `vnni2` and `vnni4` based on the factors but I think it's better to keep 16 and 8 sizes equivalent to `bf` sizes.

docs/TPPDialect.md
rengolin: Which transforms this into a ternary op, reinforcing my point below. :)

docs/TPPDialect.md
rengolin: Hm, you're right. libxsmm uses `VNNI2` and `VNNI4` internally, so we should do the same.

docs/TPPDialect.md
adam-smnk: Agreed ^

docs/TPPDialect.md
adam-smnk: Good point. We can make these semantics more explicit with ternary.
The third operand could be either directly an output `C` matrix or a scalar value `β`. But as we probably aren't interested in supporting non-zero or non-one `β`. Passing output storage `C` makes more sense.

> If anything, we can probably drop the binary version and force matmul to have a zero initializer as the third element.

I think that's a good idea. We should support only single op format whenever possible.

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+
+The TPP dialect supports `tensor` types, converted directly from `linalg`, `tensor` and `tcp` operations, and have their own tiling and bufferization rules at `tensor` level.
+
+After bufferization, the dialect also supports `memref` types, where further grouping, transforms and annotation happen, in preparaiton for optimal `xsmm` lowering.
+
+## Optimizations
+
+Due to cache locality, instruction set availability and number of registers, there are various costs and benefits of picking particular combinations, which aren't immediately clear to the programmer.
+
+Our compiler needs to select the most appropriate combination that can be lowered to existing and optiomal `libxsmm` calls with the right blockng and tiling factors.
```
egeor, line 17: optiomal -> optional

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+
+TPP operations can carry attributes, their operands can be chained and grouped, forming trees and graphs, which can be mathed in different ways.
```
egeor, line 19: mathed -> matched

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+  tpp.add ins(%arg0, %arg1) outs(%buf) : (memref<>, memref<>)
+  // In-of-place ops output on input operands
+  tpp.matmul ins(%A, %B, %C) outs(%C) : (memref<>, memref<>, memref<>)
+```
+
+### Unary
+
+On `tensors`, unary ops have one argument and one return value.
+
+On `memrefs`, unary operations have only `outs` for non-transforms (neg, inv, exp, etc.).
```
egeor, line 50: Does it imply in-place semantics for these on memrefs?

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+When lowering a broadcast (ex. a `linalg.generic` with lower dimension argument and apporpriate affine map), we map to `tpp.broadcast` + the operation.
+
+### Binary
+
+On `tensors`, binary ops have two arguments and one return value.
+
+On `memrefs`, binary operations have both `ins` and `outs`, with always two `ins` and one `outs` arguments.
+On most binary ops, the `outs` can alias `ins` (ex. `a=a+b`), and `ins` can alias themselves (ex. `a=a+a`).
+
+Element wise operations must have the same types, with broadcast done as a separate op, per operand.
```
egeor, line 64: same types in this context means both tensor/memref types? Or meant to say same input sizes? I believe both of these should be the case

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+## Transpose
+Transpose a shape into new memory.
+Type must have the same rank, 2, dims flipped.
+```mlir
+  tpp.transpose ins(%0) outs(%1) : memref<MxNxTy> -> memref<NxMxTy> // TRANSFORM_NORM_TO_NORMT
+```
+Shiould be fused with the user(s).
+GEMM ops have transposed versions, we should use this op to annotate operands.
+
+## BF16 Pack / Unpack
```
egeor, line 175: Shall be renamed to Zip/Unzip per new API (currently a PR in libxsmm)

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+Split fp32 into bf16 halfs.
```
egeor, line 176: Unzip an fp32 tensor to two tensors with 16-bit entries in an elementwise fashion / Zip two tensors with 16-bit entries into an fp32 tensor in an elementwise fashion

docs/TPPDialect.md
rengolin: It does not. We discussed this a bit more and we'll common up the syntax with `ins` always present.

On `tensors`, there is no in-place. On `memrefs`, an op is in-place if the `outs` alias with at least one of the `ins`.

I'm updating this part to make that clear.

docs/TPPDialect.md
rengolin: Correct, everything needs to be identical: shapes, element sizes and tensor/memref. I'm updating this part to make that clear.

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+If operand's producer is `tpp.broadcast`, add appropriate `BCAST_*` flag to op.
+
+Generic out-of-place syntax:
+```mlir
+  %2 = memref.alloc : memref<NxMxTy>
+  tpp.add ins(%0, %1) outs(%2) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy> // ADD
+  tpp.sub ins(%0, %1) outs(%2) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy> // SUB
+  tpp.mul ins(%0, %1) outs(%2) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy> // MUL
+  tpp.div ins(%0, %1) outs(%2) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy> // DIV
+  tpp.muladd ins(%0, %1) outs(%2) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy> // MULADD
```
egeor, line 249: In libxsmm TPP we support muladd as binary because the output *must* be aliasing with one of the inputs (the additive one). In principle the muladd should be ternary in the generic fashion: d = a * b + c (whereas we really implement only c = a * b +c in libxsmm TPP)

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+  tpp.add ins(%0, %1) outs(%1) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy>
+  // In-place on the only argument, %2 alias %0 -> `a = a + a`, or `a += a`
+  tpp.add ins(%0, %0) outs(%0) : (memref<NxMxTy>, memref<NxMxTy>, memref<NxMxTy>) -> memref<NxMxTy>
+```
+
+## Matmul
+Matrix-multiply between two memrefs.
+Always out-of-place.
+Operands of the compatible type (M, N, K), broadcast by operator
+If operand's producer is `tpp.broadcast`, add appropriate `BCAST_*` flag to op.
```
egeor, line 257: The matmul op does not support brodcast input semantics in libxsmm TPP so eventually that should be lowered to an explicit bcast TPP on the corresponding input.

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+  tpp.zero(%2) : memref<MxNxTy>
+  tpp.matmul ins(%0, %1) outs(%2) : (memref<NxKxTy>, memref<KxMxTy>) -> memref<NxMxTy>
+```
+
+If the shape of the second argument has rank+1 and the inner dim is 2 or 4, this is a VNNI matmul:
+```mlir
+  tpp.matmul ins(%0, %1) outs(%2) : (memref<64x16xTy>, memref<8x32x2xTy>) -> memref<64x32xTy>
+```
+The output of a VNNI-packed matmul is **not** VNNI packed and needs no _unpacking_.
+
```
egeor, line 270: For future reference, we plan to support in GEMMs at the libxsmm TPP level with C being in either vnni-packed or flat format

docs/TPPDialect.md
```diff
@@ -0,0 +1,351 @@
...
+  tpp.zero(%2) : memref<MxNxTy>
+  tpp.matmul ins(%0, %1, %2) outs(%2) : (memref<NxKxTy>, memref<KxMxTy>) -> memref<NxMxTy>
+```
+
+# Quaternary Ops
+
+## Fused BRGEMM
+Batched reduce matrix-multiply between two memrefs.
+Operands of the compatible type (Batch, M, N, K), broadcast by operator
+If operand's producer is `tpp.broadcast`, add appropriate `BCAST_*` flag to op.
```
egeor, line 291: same comment wrt BCAST flag as above in matmul

egeor (approved):

docs/TPPDialect.md
rengolin: Done as an NFC commit after merge.

PR #428, opened 2023-03-27
MLP bf16 3 layer unpacked benchmark

adam-smnk:
List of changes:
- adds a new MLP benchmark model: bf16 3 layer 1024 unpacked
- adjusts existing MLP bf16 3 layer 1024 packed model to match operations of the unpacked version
- moves MLP vnni benchmark layer outputs to passed arguments to avoid explicit linalg.fill for better code generation

Low performance of the new unpacked MLP test is due to issues with pack propagation. When other PRs address that we should see improvements.

### DIFF
--- a/benchmarks/benchmarks.json
+++ b/benchmarks/benchmarks.json
@@ -32,7 +32,12 @@
     "bf16_3x1024_mlir": {
       "type": "MLIR",
       "benchmark": "mlp-bf16-3layers-1024.mlir",
-      "flags": [ "-run-args='-def-pack=0'" ]
+      "flags": []
+    },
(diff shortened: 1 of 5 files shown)


rengolin:
`bf16_3x1024_mlir` has dropped 50% performance

adam-smnk:
> `bf16_3x1024_mlir` has dropped 50% performance

It increased from ~17 to ~32 gflops. `bf16_3x1024_unpacked_mlir` has poor performance due to `pack`s still present after the pipeline. Everything else is unaffected.

rengolin (approved):

PR #433, opened 2023-03-28
benchmark/harness: fix error logic

hfp:


hfp:
The file `controller.py` is the broken piece which is [fixed](https://github.com/plaidml/tpp-mlir/pull/433/commits/5bdcb5042aa5f5891048ba977911cde70f645835) here.

hfp:
Now, `LIBXSMM_VERBOSE` is possible (writing to `stderr`). Without this PR, any non-empty `stderr` is assumed to be an error.

benchmarks/driver.py
```diff
@@ -81,8 +81,10 @@ def __init__(self, args, loglevel):
             for v in range(args.verbose - args.quiet):
                 self.extra_args.append("-v")
         # Set environment variables for dynamic loading (Linux and Mac)
-        os.putenv("LD_LIBRARY_PATH", f"{self.lib_dir}:{os.getenv('LD_LIBRARY_PATH')}")
-        os.putenv("DYLD_LIBRARY_PATH", f"{self.lib_dir}:{os.getenv('DYLD_LIBRARY_PATH')}")
+        for path in ["LD_LIBRARY_PATH", "DYLD_LIBRARY_PATH"]:
+            environ = [os.getenv(path)] if os.getenv(path) else []
+            environ.insert(0, self.lib_dir)  # prepend
+            os.environ[path] = ":".join(environ)
```
adam-smnk, line 87: Why is this better?
Just a general question as I'm not that familiar with python's environ

adam-smnk: Ahh ok it's just generalized the prepending, I got a bit confused by `path` thinking it's a file path.

adam-smnk (approved):
Looks good to me

rengolin (approved):
Awesome! I had this in my back burner but never had a strong reason to do it. Thanks!

benchmarks/driver.py
hfp: Indeed, a few lines more code but it avoids producing something `/path/to/xyz:` (an empty component after `:`).

PR #434, opened 2023-03-28
Check arch flags and skip bench

rengolin:
This makes running on laptops not crash with invalid TPP lowering. Using regular expression to allow for multiple patterns, for example `avx512.*`.


hfp (approved):
Yes, this is much needed - Thank you! Looks good to me. Of course, other platforms can be supported later as needed/requested (macOS, FreeBSD, etc).

PR #435, opened 2023-03-28
Pass build directory to harness

rengolin:
Before, the driver would find the programs but not the harness, when running benchmarks from an install / build dir in a separate place.

Current builds passing because we copied things by hand, but to disable that, we need to pass the build directory to the harness too.


hfp (approved):
Looks good to me... Thank you!

PR #437, opened 2023-03-29
Move leftover test/Benchmarks test to Integration

adam-smnk:
Real benchmark files now reside in `benchmarks/mlir` directory. The old leftovers in `test/Benchmarks` act as simple integration tests, thus, they are moved to central `Integration` directory.


rengolin:
We should also remove `-n 10` and the FLOPS comments to keep it simple.

rengolin (approved):

PR #443, opened 2023-03-30
Use matchers in linalg to tpp conversion

chelini:
We still provided isTpp* methods, but the implementation now uses the newly introduced structural matchers; this allows us to remove clutter and have a nice syntax to define what we want to match. Matchers are composed together (see `isTppOp` and `isTppBinary/Unary`).

We also update the operation traits for unary and binary.  All the unary and binary tpp ops implement `BroadcastableShape` and `UnitStrideInnerLoop`.


include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -35,6 +35,7 @@ def TppVNNIOperand : AnyTypeOf<[TppVNNIMemrefInput, AnyFloat]>;
 
 // Tpp operation trait - make sure the operand are broadcastable
 // following Numpy-style broadcast.
+//  https://numpy.org/doc/stable/user/basics.broadcasting.html#general-broadcasting-rules
```
rengolin, line 38: Stop adding these URLs, we do not follow them completely, and we'll remove all of those rules soon.

chelini: I was just moving it. But I can delete it.

include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -46,7 +47,8 @@ def UnitStrideInnerLoop : NativeOpTrait<"tpp::UnitStrideInnerLoop">;
 //===----------------------------------------------------------------------===//
 
 class Tpp_UnaryOp<string mnemonic, list<Trait> traits = []> :
-  Tpp_Op<mnemonic, !listconcat(traits, [])> {
+  Tpp_Op<mnemonic, !listconcat(traits, [BroadcastableShape,
```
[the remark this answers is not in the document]
chelini: Bringing tpp.broadcast will take time, and this can unlock some progress (i.e., Kavitha PR).

include/TPP/Dialect/Tpp/TppOps.td
rengolin: Better to remove any comment to the numpy broadcast, or it will pass the idea that we support it fully.

include/TPP/Dialect/Tpp/TppOps.td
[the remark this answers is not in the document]
rengolin: Kavitha's PR should be merged sooner than all this, we shouldn't keep that waiting for this discussion to unfold.

It's less painful to split the work with just what she needs now and then work on the rest than implement a whole set of broadcast support to ops that will be done later in the completely opposite direction.

Forward progress that needs backwards progress isn't really progress.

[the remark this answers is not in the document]
chelini: I don't think `BroadcastableShape` is a backward progress especially if the tpp.broadcast will implement the broadcast interface. We will need to drop the interface from the op, no big deal.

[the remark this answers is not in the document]
chelini: We can drop from the ops, but we loose verification pieces, if we are happy with that, I am fine. Matchers can still use the `check` methods. Adding the trait is just an additional verifier checks.

lib/TPP/Dialect/Tpp/TppOps.cpp
```diff
@@ -21,29 +21,7 @@ using namespace mlir::tpp;
...
-// TODO: use verifyCompatibleOperandBroadcast
-LogicalResult IdentityOp::verify() {
-  utils::MatchBroadcastRuleResult res =
-      utils::verifyTppIdentityBroadcastingRules(getInput().getType(),
-                                                getOutput().getType());
-  if (res != utils::MatchBroadcastRuleResult::Success)
-    return emitOpError(getMatchBroadcastRuleMessage(res));
-  return success();
-}
+LogicalResult IdentityOp::verify() { return success(); }
```
rengolin, line 24: you could just remove the verify command from the TD file.

test/Conversion/LinalgToTpp/linalg-to-tpp.mlir
```diff
@@ -83,9 +83,11 @@ func.func @add_mapping_brcst(%arg0: memref<3x3xf32>, %arg1: memref<1x3xf32>) {
 
 // -----
 
+// The output is not an identity we should not map this.
 #map = affine_map<(d0, d1) -> (d0, d1)>
 #map1 = affine_map<(d0, d1) -> (d1, d0)>
 func.func @add_mapping(%arg0: memref<4x4xf32>, %arg1: memref<4x4xf32>) {
+  // CHECK-NOT: tpp.add
```
rengolin, line 90: identity or add?

chelini: `identity` is overloaded. I mean `identity` map.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -223,225 +221,140 @@ static bool isZeroOp(Operation *defOp) {
...
-  auto firstOperandType = types[0].cast<MemRefType>();
-
-  // Step1. Validate rank.
-  int64_t rankOperand = firstOperandType.getRank();
-  for (Type currentType : types) {
-    if (!currentType.isa<MemRefType>())
+// Return true if the linalg.generic can be mapped to a tpp.relu.
+bool isTppRelu(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
+  // Callback to check relu-body.
+  auto hasReluBody = [=](Operation *op,
```
adam-smnk, line 303: I'd move it a separate function like `hasMulAddBody`

adam-smnk (approved):

PR #447, opened 2023-03-30
Do not allow scalar output in tpps

chelini:
Tpp are memory-to-memory operations and we should not allow a scalar as output memref. With this changes the tpp.relu and tpp.add verifier can be removed as we check operand shape directly in TD. Update tests and avoid checking operand prefix, this is automatically generate and there is not need for us to check.


adam-smnk (approved):
Good change. Later we can constrain it even further when we move to explicit broadcasting.

rengolin (approved):
We probably will have to change this on reduction, but it's safer this way for now.

PR #448, opened 2023-03-30
Improve loop conversion

chelini:
Common-up loop builder for `tpp.add` and `tpp.relu`. `convertBinaryTppToLoops` can be used for any operation that lowers to:
```
load A
load B
S = do an operation on A and B
store S in C
```
`tpp.identity` is refreshed a bit but cannot use `convertTppToLoops` as it is just a store operation.

### DIFF
--- a/lib/TPP/ConvertTppToLoops.cpp
+++ b/lib/TPP/ConvertTppToLoops.cpp
@@ -57,18 +57,15 @@ static SmallVector<Value> getInputOperandIvs(Value in, Value out,
   auto shapedOut = out.getType().dyn_cast<ShapedType>();
   assert(shapedOut.getRank() >= shapedIn.getRank());
 
-  // Handle rank 1 input.
   auto shapeIn = shapedIn.getShape();
-  if (shapedIn.getRank() == 1) {
-    if (shapeIn[0] == 1)
-      return {zero};
-    return {ivs[shapedOut.getRank() - 1]};
-  }
-
-  // Handle rank 2 input.
-  assert(shapedIn.getRank() == shapedOut.getRank());
   SmallVector<Value> newIvs;
   for (auto [idx, shapeOnDim] : llvm::enumerate(shapeIn)) {
+    // Handle rank 1 input.
+    if (shapedIn.getRank() == 1) {
+      if (shapeIn[0] == 1)
+        return {zero};
+      return {ivs[shapedOut.getRank() - 1]};
+    }
     if (shapeOnDim == 1)
       newIvs.push_back(zero);
     else
@@ -98,7 +95,7 @@ static inline Value getOne(RewriterBase &rewriter, Location loc) {
 }
 
 template <typename OpTy>
-static void convertBinaryTppToLoops(RewriterBase &rewriter, Location loc,
+static void convertBinaryToLoops(RewriterBase &rewriter, Location loc,
                                     Value lhs, Value rhs, Value out) {
   assert(out.getType().isa<ShapedType>());
   auto shape = out.getType().cast<ShapedType>().getShape();
--- a/lib/TPP/ConvertTppToLoops.cpp
+++ b/lib/TPP/ConvertTppToLoops.cpp
@@ -24,8 +24,33 @@ namespace {
 
 static bool isScalarVal(Value val) { return !val.getType().isa<ShapedType>(); }
 
-static SmallVector<Value> getLocalIvs(Value in, Value out, ValueRange ivs,
-                                      Value zero) {
+static inline Value getZero(RewriterBase &rewriter, Location loc) {
+  return rewriter.create<arith::ConstantIndexOp>(loc, 0);
+}
+
+static inline Value getOne(RewriterBase &rewriter, Location loc) {
+  return rewriter.create<arith::ConstantIndexOp>(loc, 1);
+}
+
+static inline SmallVector<Value> getShapeAsValues(RewriterBase &rewriter,
+                                                  Location loc,
+                                                  ArrayRef<int64_t> shape) {
+  SmallVector<Value> shapeAsValues;
+  for (int64_t shapeDim : shape) {
+    assert(shapeDim != ShapedType::kDynamic);
+    shapeAsValues.push_back(
+        rewriter.create<arith::ConstantIndexOp>(loc, shapeDim));
+  }
+  return shapeAsValues;
+}
+
+// Return the touched ivs by `in` based on `out` and `ivs`. The iteration
+// domain in tpp is defined by the output tensor. Assuming static shape we
+// can get the induction varibales used by `in` looking at the shape of `in`
+// and `out`. Exmaple:
+// out = [4, 3] in = [3] iv =  [i, j]. Will return j.
+static SmallVector<Value> getInputOperandIvs(Value in, Value out,
+                                             ValueRange ivs, Value zero) {
   assert(in.getType().isa<ShapedType>());
   assert(out.getType().isa<ShapedType>());
   auto shapedIn = in.getType().dyn_cast<ShapedType>();
@@ -73,18 +98,15 @@ static inline Value getOne(RewriterBase &rewriter, Location loc) {
 }
 
 template <typename OpTy>
-static void convertTppToLoops(RewriterBase &rewriter, Location loc, Value lhs,
-                              Value rhs, Value out) {
+static void convertBinaryTppToLoops(RewriterBase &rewriter, Location loc,
+                                    Value lhs, Value rhs, Value out) {
   assert(out.getType().isa<ShapedType>());
   auto shape = out.getType().cast<ShapedType>().getShape();
   int64_t rank = shape.size();
-  SmallVector<Value> ubs;
-  for (int64_t shapeDim : shape)
-    ubs.push_back(rewriter.create<arith::ConstantIndexOp>(loc, shapeDim));
-  Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+  Value zero = getZero(rewriter, loc);
+  SmallVector<Value> ubs = getShapeAsValues(rewriter, loc, shape);
   SmallVector<Value> lbs(rank, zero);
-  Value one = rewriter.create<arith::ConstantIndexOp>(loc, 1);
-  SmallVector<Value> steps(rank, one);
+  SmallVector<Value> steps(rank, getOne(rewriter, loc));
   (void)scf::buildLoopNest(
       rewriter, loc, lbs, ubs, steps,
       [&](OpBuilder &b, Location loc, ValueRange localIvs) {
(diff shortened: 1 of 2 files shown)


lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,75 @@ using namespace mlir::tpp;
 
 namespace {
 
+static bool isScalarVal(Value val) { return !val.getType().isa<ShapedType>(); }
```
rengolin, line 25: is there a similar one elsewhere? candidate for utils?

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,75 @@ using namespace mlir::tpp;
...
+
+static SmallVector<Value> getLocalIvs(Value in, Value out, ValueRange ivs,
+                                      Value zero) {
+  assert(in.getType().isa<ShapedType>());
+  assert(out.getType().isa<ShapedType>());
+  auto shapedIn = in.getType().dyn_cast<ShapedType>();
+  auto shapedOut = out.getType().dyn_cast<ShapedType>();
+  assert(shapedOut.getRank() >= shapedIn.getRank());
```
rengolin, line 33: that's not a reasonable requirement for a function with the generic name of `getLocalIvs`. There is logic here that isn't explicit in the function name / comments.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,75 @@ using namespace mlir::tpp;
...
+
+  // Handle rank 1 input.
+  auto shapeIn = shapedIn.getShape();
```
rengolin, line 36: This looks like it could be covered on the loop below.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,75 @@ using namespace mlir::tpp;
...
+template <typename OpTy>
+static void convertTppToLoops(RewriterBase &rewriter, Location loc, Value lhs,
+                              Value rhs, Value out) {
+  assert(out.getType().isa<ShapedType>());
+  auto shape = out.getType().cast<ShapedType>().getShape();
+  int64_t rank = shape.size();
+  SmallVector<Value> ubs;
+  for (int64_t shapeDim : shape)
+    ubs.push_back(rewriter.create<arith::ConstantIndexOp>(loc, shapeDim));
+  Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
```
rengolin, line 64: This pattern of zeroes and ones is *very* common, candidate to common up.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,75 @@ using namespace mlir::tpp;
...
+    if (shapeOnDim == 1)
+      newIvs.push_back(zero);
+    else
+      newIvs.push_back(ivs[idx]);
+  }
+  return newIvs;
+}
+
+template <typename OpTy>
+static void convertTppToLoops(RewriterBase &rewriter, Location loc, Value lhs,
```
rengolin, line 56: This is specific to binary ops, but the name doesn't reflect that.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -74,70 +100,32 @@ struct ConvertTppAddOp : public OpRewritePattern<AddOp> {
...
     SmallVector<Value> ubs;
-    size_t rank = identityOp.getOutput().getType().cast<MemRefType>().getRank();
-    ArrayRef<int64_t> shapeOutput =
-        identityOp.getOutput().getType().cast<MemRefType>().getShape();
-    for (size_t idx = 0; idx < rank; idx++) {
-      Value dim =
-          rewriter.create<arith::ConstantIndexOp>(loc, shapeOutput[idx]);
-      ubs.push_back(dim);
-    }
+    for (int64_t shapeDim : shape)
```
rengolin, line 113: This is also another candidate for common up

lib/TPP/ConvertTppToLoops.cpp
chelini: Looks like we use this only when converting from tpp to loops.

lib/TPP/ConvertTppToLoops.cpp
chelini: Will change name and add more comment.

lib/TPP/ConvertTppToLoops.cpp
chelini: I see this as an early exit. If we are dealing with rank 1 we have only two choices: zero or the right-most ivs.

lib/TPP/ConvertTppToLoops.cpp
chelini: done.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,98 @@ using namespace mlir::tpp;
 
 namespace {
 
+static bool isScalarVal(Value val) { return !val.getType().isa<ShapedType>(); }
```
adam-smnk, line 25: Shouldn't this be `return isa<ScalarType>()`?

lib/TPP/ConvertTppToLoops.cpp
rengolin: I get it, but it's a convoluted early exit. It made me think the cases were different and I had to go through to make sure they're not. This loop will run at most twice, an early exit here doesn't help much.

lib/TPP/ConvertTppToLoops.cpp
rengolin: I'm pondering on the utility of this function... given this is a really cheap and short check... It makes me thing there's more to it, then I look at the function, and there isn't...

adam-smnk: Oh it seems to be spir-v specific. Nevermind.

chelini: It is used in multiple places. I think this is why is an helper function.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,95 @@ using namespace mlir::tpp;
...
+  return shapeAsValues;
+}
+
+// Return the touched ivs by `in` based on `out` and `ivs`. The iteration
+// domain in tpp is defined by the output tensor. Assuming static shape we
+// can get the induction varibales used by `in` looking at the shape of `in`
+// and `out`. Exmaple:
+// out = [4, 3] in = [3] iv =  [i, j]. Will return j.
+static SmallVector<Value> getInputOperandIvs(Value in, Value out,
+                                             ValueRange ivs, Value zero) {
```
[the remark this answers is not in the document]
adam-smnk: Alright, I'm fine with either approach 👍

lib/TPP/ConvertTppToLoops.cpp
rengolin: You don't need to. If `shapeIn` has rank 1, the loop will only run once and `newIvs` will only have one element, `ivs[0]`. Returning `newIvs` should do the trick regardless of the rank.

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -22,49 +22,95 @@ using namespace mlir::tpp;
 
 namespace {
 
+static bool isScalarVal(Value val) { return !val.getType().isa<ShapedType>(); }
+
+static inline Value getZero(RewriterBase &rewriter, Location loc) {
+  return rewriter.create<arith::ConstantIndexOp>(loc, 0);
+}
+
+static inline Value getOne(RewriterBase &rewriter, Location loc) {
```
rengolin, line 31: Can you use `BuilderUtils.h::getConstIndex()` instead?

lib/TPP/ConvertTppToLoops.cpp
rengolin: What I meant is: I don't think the code is large enough to be worth being replaced by a function call.
```
if (!val.getType().isa<ShapedType>())
// vs
if (isScalarVal(val))
```

This is only used in the same pattern:
```
Value scalarLhs =
            isScalarVal(lhs)
                ? lhs
                : b.create<memref::LoadOp>(
                      loc, lhs, getInputOperandIvs(lhs, out, localIvs, zero));
```
and could very easily have been:
```
Value scalarLhs =
            val.getType().isa<ShapedType>()
                ? b.create<memref::LoadOp>(
                      loc, lhs, getInputOperandIvs(lhs, out, localIvs, zero))
                : lhs;
```

rengolin (commented):
There are still some unresolved comments and the PR is gathering conflicts. Is this active work or has it been superseded by other PRs?

lib/TPP/ConvertTppToLoops.cpp
chelini: Looks like build a constant with unknown location.

rengolin: It does, we don't need a location for compiler constants.

chelini: No, it doesn't: https://github.com/plaidml/tpp-mlir/blob/1b7913f209b2154e876eef2f8c0855fb31cea6bf/lib/TPP/BuilderUtils.cpp#L23 We build constant op with unknown loc.

chelini:
> There are still some unresolved comments and the PR is gathering conflicts. Is this active work or has it been superseded by other PRs?

The goos olut

> There are still some unresolved comments and the PR is gathering conflicts. Is this active work or has it been superseded by other PRs?

The right solution, not that we have the tpp interface is to have a single pattern and separate the conversion to loops based on binary/ternary and unary., but will require a bit of effort.

lib/TPP/ConvertTppToLoops.cpp
rengolin: I agreed with you: "It does (build a constant with unknown location)". :smile: 

We don't care about the location of compiler-generated constants.

chelini: I am not completely sure why we don't care. We build a constant operation with no location. I did not see this pattern in MLIR. We can modify the method to accept a location too.

rengolin: > I am not completely sure why we don't care.

The main purpose of file locations is stepping through the code while debugging. 

When stepping through a line in the model (say, Python or C++), you can have multiple HW operations per IR instruction and multiple IR instructions per line of code. All operations need to be tagged with the same LOC as the program line, so that the debugger knows when to stop (breakpoint, single-step, etc).

Constants, however, are generally not operations. They're stored in the binary (some data section, constant pools), or converted to immediate values on instructions, or just optimised away. There's no way you can "set a breakpoint" or "step-through" a constant declaration, unless this becomes an actual operation (say, offload a CPU constant into GPU memory), which only really happens to tensors.

These particular constants are indices, integers and floats that will become part of instructions. A loop induction variable range check (usually immediate or zero check), a float zero-fill, which will be lowered as `XOR` or `MOV reg, zero`, etc.

None of them are relevant to the debugging experience, and therefore, need no location.

More generally, for actual ops, by blinding attaching "some" location to random constants and ops, we may be changing the semantics of what the debugger is expected to do.

A one-to-many rewrite can easily just use the original location to all the generated ops. A many-to-many rewrite needs to make sure what location is going where, as they may be in different (non-sequential) code lines. 

Ultimately, an op without location won't break the debugger more than a wrong location op. But none of that applies to compiler-generated constants.

> I did not see this pattern in MLIR.

Generic MLIR passes cannot assume where their constants came from. We can.

> We can modify the method to accept a location too.

If you do this, then make sure to default to `UnknownLoc`, so that it doesn't change the existing uses.

chelini:
Not more relevant for now.

PR #451, opened 2023-03-30
Fixing the assert check for iree_input_tensor

nhasabni:


runtime/XsmmRunnerUtils.cpp
```diff
@@ -531,13 +531,14 @@ typedef struct {
 
 // Based on the code
 // https://mlir.llvm.org/doxygen/TypeConverter_8cpp_source.html#l00283
-// allocatedPtr and alignedPtr should be same. Also, the offset needs to be 0.
-// We add these assumptions as asserts.
+// the offset needs to be 0. We add this assumption as assert.
+// We will remove it if we find a violation of this assumption.
+//
+// We do not enforce the constraint that allocatedPtr and alignedPtr are same.
```
chelini, line 537: interesting is this what is done in IREE?

chelini:
thanks!

runtime/XsmmRunnerUtils.cpp
nhasabni: Thanks. I think it comes from MLIR, not from IREE.

PR #453, opened 2023-03-31
Refactor structured matchers API

adam-smnk:
Adds a more generic selection tag that can be used to specify which operands, dimensions etc. are targeted by given predicate.
Also, updates region predicate API to be consistent with the rest.


rengolin (commented):
Nice!

chelini (approved):
Nice

PR #460, opened 2023-04-04
Update README.md

alheinecke:
fixed LLVM config issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ export PATH=$CUSTOM_LLVM_ROOT/bin:$PATH
 
 # Build
 cmake -G Ninja ../llvm \
-   -DLLVM_ENABLE_PROJECTS=mlir;openmp \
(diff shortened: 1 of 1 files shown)


chelini (approved):

alheinecke:
@chelini : is there way to only build openmp and not clang? It complained when for me when just adding the quotes that a clang as dependency is missing... however now the files to compile got from 3900 to 6400 :-(

chelini:
> @chelini : is there way to only build openmp and not clang? It complained when for me when just adding the quotes that a clang as dependency is missing... however now the files to compile got from 3900 to 6400 :-(

I am not aware of a way to build OpenMP standalone without Clang. We can use ccache to avoid re-building all the files.

alheinecke:
OK, it's fine if there is now way to skip... I'm just trying to new install in my 8 core VM and an installation from scratch of TPP-MLIR is now 30+ min :-)

PR #462, opened 2023-04-10
Finish the FP32 compute type fix

rengolin:
This propagates the FP32 compute type to all operations, and takes into account that simple copies don't need to up/down cast.


rengolin:
Do we need the same exceptions for other transforms, like `NORM_TO_NORMT`?

rengolin:
> for now, yes, all TRANSFORM TPP need compute = input = output ... however, I think we should set it to LIBXSMM_DATATYPE_IMPLICIT. LIBXMM itself needs to checked :-(

Right, for now I'm just adding a helper function for the types. Later we can do the implicit typing.

alheinecke (approved):

PR #465, opened 2023-04-11
Adding XSMM-DNN Benchmarks

rengolin:
Adding a libxsmm-dnn build to CMake (like libxsmm), and exposing the `xsmm_dnn_mlp` program to the benchmarks.

This allows us to run four xsmm-dnn benchmarks to compare with our MLIR performance: `{ bf16, fp32 } x { matmul-64, mlp-layer }`.

LSAN detects leaks on xsmm-dnn, so I'm forcing leak-san disabled for all benchmarks. This is also needed for some MLIR runs, so it should be fine here.

Finally, added `taskset` to all runs to make them more stable.


rengolin:
> > LSAN detects leaks on xsmm-dnn, so I'm forcing leak-san disabled for all benchmarks.
> 
> I will take a look and fix LIBXSMM-DNN.

If the fix isn't trivial, we can start here without LSAN and then re-enable later when you fix it.

hfp (approved):
Looks good to me. We eventually want to revisit pinning benchmarks providing more options or user-control. Right now, we assume the driver is sequential and with no particular multicore runtime. For example, pinning to the 4th core/thread potentially masks available threads for instance in OpenMP to one thread only (which should be fine for now).

rengolin:
> Looks good to me. We eventually want to revisit pinning benchmarks providing more options or user-control. Right now, we assume the driver is sequential and with no particular multicore runtime. For example, pinning to the 4th core/thread potentially masks available threads for instance in OpenMP to one thread only (which should be fine for now).

Exactly! We'll need to add support for multi-core once we have a benchmark that makes use of OpenMP, but I think they still need to run sequentially. I don't want to `taskset` multiple benchmarks on the same machine, even with NUMA and high core count, as interactions with the kernel (ex. `malloc` and `free`) can still interfere.

PR #467, opened 2023-04-12
Random initialization of f64 tensors

adam-smnk:
Adds support for random initialization of 'f64' data type splats.

Also, adds extra utility functions, slightly cleans up current initialization logic, and adds f64 test cases.


lib/TPP/TensorInit.cpp
```diff
@@ -70,29 +70,35 @@ DenseElementsAttr TensorInit::get(ShapedType shape) {
...
   buffer.push_back(llvm::APFloat(value));
-  if (type == DataType::BF16)
-    toBF16(buffer.back());
+  convertType(buffer.back());
 }
 
-float TensorInit::at(size_t index) {
-  return buffer[index].convertToFloat();
+void TensorInit::convertType(llvm::APFloat &value) {
+  switch (type) {
```
rengolin, line 82: Perhaps a check if the type is already correct? I'm guessing the cast is more expensive than a check?

rengolin (commented):
We're not planning to support `f64` as a data type for input/weight tensors any time soon, what's the rationale behind this?

adam-smnk:
> We're not planning to support `f64` as a data type for input/weight tensors any time soon, what's the rationale behind this?

I need this for integration test of `perf.mean` and new `perf.median` ops which operate on `f64` time deltas.

rengolin:
> I need this for integration test of `perf.mean` and new `perf.median` ops which operate on `f64` time deltas.

Thought so. Sounds good.

rengolin (approved):

PR #468, opened 2023-04-13
Perf dialect - `perf.median` operation

adam-smnk:
Adds a new Perf dialect op `perf.median`, which allows user to compute median value of the measured time deltas.
The new op provides an alternative to the existing `perf.mean`. Due to more complex implementation logic, the median computation relies on runtime support.

Additionally, `perf.bench` optional arguments `args` are renamed  to `iter_args` to be consistent with `scf.for` naming as the two args are semantically identical.

### DIFF
--- a/include/TPP/Dialect/Perf/PerfOps.td
+++ b/include/TPP/Dialect/Perf/PerfOps.td
@@ -273,6 +273,44 @@ def Perf_MeanOp : Perf_Op<"mean", []> {
 }
 
 //===----------------------------------------------------------------------===//
+// MedianOp
+//===----------------------------------------------------------------------===//
+
+def Perf_MedianOp : Perf_Op<"median", []> {
+  let summary = "Compute median value.";
+  let description = [{
+    The `perf.median` operation computes median value of the provided
+    time deltas.
+
+    Example:
+
+    ```mlir
+
+    %deltas = perf.bench (%n) {
+      ... // benchmarked code
+    } -> memref<?xf64>
+
+    %mean = perf.median(%deltas : memref<?xf64>) : f64
+
+    ```
+  }];
+
+  let arguments = (ins RankedOrUnrankedMemRefOf<[F64]>:$input);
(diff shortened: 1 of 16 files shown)


adam-smnk:
`tpp-run` can optionally output median instead of mean of the measured time deltas using `-timer-median`. However, after testing this option for a bit, it is not always correct or beneficial approach as it mostly depends on the distribution and variance of the measured deltas. Thus, by default we still stick to using mean and it is just an optional utility.

rengolin:
> `tpp-run` can optionally output median instead of mean of the measured time deltas using `-timer-median`. However, after testing this option for a bit, it is not always correct or beneficial approach as it mostly depends on the distribution and variance of the measured deltas. Thus, by default we still stick to using mean and it is just an optional utility.

I'm wondering about the value of this, tbh.

It doesn't mean the most likely nor the expected value on long runs (for greatly skewed distributions), so won't reliably get rid of outliers. I'm also not sure the problem is in the mean itself and not in the resolution of the clock, as we're getting deviations as large as the mean on smaller benchmarks.

Not to mention the addition of runtime components, etc, which increase maintenance costs... :/

I think a better option is to just have a `perf.bench` (additional) mode (via flag?) where the timer is put around the loop and dividing the number by N, but that means moving all allocations outside of the loop and reusing it inside of the loop, which may not be trivial.

adam-smnk:
@rengolin I agree, it ended up not being really helpful. I'm fine with ultimately dropping this if you also can't see any good use cases.
I'm a bit skeptical about the clock resolution being the problem but I noticed that measurements are also affected by allocation and copies introduced by bufferization. This can happen when the original kernel is at the tensor level then wrapped with `perf.bench` and only bufferized later which can create new ops within the `perf.bench` region. I'm preparing another tweak that tries to address this issue.

adam-smnk:
Let's give up on the `perf.median` for now. I'll push the `args` refactor separately.

PR #469, opened 2023-04-13
Remove parallel loop fusion

KavithaTipturMadhu:
Parallel fusion used to fuse all the scf parallels into one, however we want barrier syncs between brgemms for functional correctness.


adam-smnk (approved):
This upstream pass is not crucial for performance or enabling other passes.
No issue removing it.

PR #470, opened 2023-04-13
Adding CI scripts to be used in CI

rengolin:
These scripts reproduce the README recipe as well as the CI recipes and should be used in build YAML files instead of direct scripting on the UI.

There are three separate scripts:
 * deps.sh: Check basic dependencies (lit, ninja, cmake)
 * cmake.sh: runs the CMake command with all the options
 * build.sh: builds and optionally checks, installs, benchmarks

These scripts should be called from a build-manager YAML file (Buildkite, ADO, Jenkins) with the appropriate parameters. Multiple calls to the same script should be harmless and preferred, when separating into different steps (build, check, install, ...).

With these, we can already replace our main builds (`tpp-mlir` and `tpp-benchmarks`), with the YAML contents from our experimental `tpp-scripts` build.

Later on, we'll move the YAML files inside the `tpp-mlir` repository and the build infrastructure will have no more hard-coded logic, meaning we can change it on different branches, but also use different build infrastructure (ex. Buildkite and Github Actions and ADO) at the same time.

### DIFF
--- a/scripts/ci/build.sh
+++ b/scripts/ci/build.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+#
+# Builds the project (tests, installs, post-install)
+
+SCRIPT_DIR=$(realpath $(dirname $0))
+
+die_syntax() {
+  echo "Syntax: $0 -b BLD_DIR [-c] [-i] [-p] [-B]"
+  echo ""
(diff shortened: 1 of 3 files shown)


hfp (approved):
Please merge!

PR #472, opened 2023-04-13
Make `tpp.matmul` and `tpp.brgemm` ternary

chelini:
Make `tpp.matmul` and `tpp.brgemm` ternary at the memref level, meaning three input and one output operand. Update XSMM lowering for `tpp.matmul` and `tpp.brgemm`, providing custom dispatch and invoking calls as LIBXSMM does. A follow up PR will clean-up `xsmm.ternary` and provide more verification for xsmm dialect, that at the moment has zero verification. Similarly, aliasing check for `tpp.matmul` and `tpp.brgemm` for the right-most operand and the output will be added separately.


include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -9,8 +9,8 @@
 #ifndef TPP_XSMM_OPS
 #define TPP_XSMM_OPS
 
-include "XsmmDialect.td"
-include "XsmmAttr.td"
+include "TPP/Dialect/Xsmm/XsmmDialect.td"
+include "TPP/Dialect/Xsmm/XsmmAttr.td"
```
rengolin, line 13: Why is this necessary?

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -140,6 +141,34 @@ def Xsmm_UnaryOp : Xsmm_Op<"unary", []> {
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// MatmulOp
+//===----------------------------------------------------------------------===//
+
+def Xsmm_MatmulOp : Xsmm_Op<"matmul"> {
+  let summary = "matmul call operation.";
+  let arguments = (ins Xsmm_DataType:$dataType, Variadic<XsmmMemRef>:$inputs);
```
rengolin, line 150: I don't think we should be using `Variadic` here, since the number of arguments doesn't change.

This may have less human-written boilerplate, but it passes the wrong idea to the compiler and to whoemever reads this later. It wouldn't be a lot more code to have a clear number of arguments for each op type.

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -226,4 +256,34 @@ def Xsmm_UnaryDispatchOp : Xsmm_Op<"unary.dispatch",[Pure]> {
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// MatmulDispatchOp
+//===----------------------------------------------------------------------===//
+
+// TODO: all the `flags` should be a list.
+class Xsmm_MatmulAndBrgemmDispatchOp<string mnemonic, list<Trait> traits = []> :
```
rengolin, line 264: I'd call it `GEMMLike`?

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -226,4 +256,34 @@ def Xsmm_UnaryDispatchOp : Xsmm_Op<"unary.dispatch",[Pure]> {
...
+  Xsmm_Op<mnemonic, !listconcat(traits, [Pure])> {
+    let arguments = (ins DenseI64ArrayAttr:$inputs,
```
rengolin, line 266: Would be good to have an explanation of what the attributes refer to. Also for the other ops.

runtime/XsmmRunnerUtils.cpp
```diff
@@ -119,8 +119,8 @@ extern "C" void _mlir_ciface_xsmm_matmul_invoke(const libxsmm_datatype dtype,
 }
 
 extern "C" int64_t _mlir_ciface_xsmm_matmul_dispatch(
-    const libxsmm_datatype dtype, const bool isVNNI, int64_t m, int64_t n,
-    int64_t k, int64_t lda, int64_t ldb, int64_t ldc) {
+    const libxsmm_datatype dtype, int64_t m, int64_t n, int64_t k, int64_t lda,
+    int64_t ldb, int64_t ldc, const libxsmm_gemm_flags flag) {
```
rengolin, line 123:
```suggestion
    int64_t ldb, int64_t ldc, const libxsmm_gemm_flags flags) {
```

runtime/XsmmRunnerUtils.cpp
```diff
@@ -134,13 +134,11 @@ extern "C" int64_t _mlir_ciface_xsmm_matmul_dispatch(
   libxsmm_blasint k_int = k;
 
   libxsmm_gemm_shape l_shape;
-  libxsmm_bitfield l_flags;
-  if (isVNNI) {
+  libxsmm_bitfield l_flags = LIBXSMM_GEMM_FLAG_NONE;
+  if (flag == LIBXSMM_GEMM_FLAG_VNNI_B) {
```
rengolin, line 138:
```suggestion
  if (flag && LIBXSMM_GEMM_FLAG_VNNI_B) {
```

runtime/XsmmRunnerUtils.cpp
```diff
@@ -393,12 +390,9 @@ _mlir_ciface_xsmm_brgemm_dispatch(const libxsmm_datatype dtype, bool isVNNI,
   libxsmm_blasint stride_b = ldb * k * typeSize;
 
   libxsmm_gemm_shape l_shape;
-  libxsmm_bitfield l_flags;
-  if (isVNNI) {
+  libxsmm_bitfield l_flags = LIBXSMM_GEMM_FLAG_NONE;
+  if (flag == LIBXSMM_GEMM_FLAG_VNNI_B) {
```
rengolin, line 394:
```suggestion
  if (flag && LIBXSMM_GEMM_FLAG_VNNI_B) {
```

rengolin (commented):
I think your title to the PR is backwards... You're moving `matmul` and `brgemm` away from `ternary` no?

runtime/XsmmRunnerUtils.cpp
```diff
@@ -593,18 +587,18 @@ extern "C" int iree_xsmm_brgemm_dispatch(void *context, void *params,
...
     int64_t ldc;
+    const libxsmm_gemm_flags flag;
   } xsmm_brgemm_dispatch_t;
   xsmm_brgemm_dispatch_t *p = (xsmm_brgemm_dispatch_t *) params;
-  p->address = _mlir_ciface_xsmm_brgemm_dispatch((libxsmm_datatype) p->dtype,
-                                                 p->vnni, p->m, p->n, p->k,
-                                                 p->lda, p->ldb, p->ldc);
+  p->address =
+      _mlir_ciface_xsmm_brgemm_dispatch((libxsmm_datatype)p->dtype, p->m, p->n,
+                                        p->k, p->lda, p->ldb, p->ldc, p->flag);
```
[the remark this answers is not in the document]
chelini: Currently is a single 'flag'. Do you still prefer 'flags'?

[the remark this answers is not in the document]
rengolin: Yes, because this will eventually be for all flags.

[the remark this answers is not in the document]
chelini: ok

include/TPP/Dialect/Xsmm/XsmmOps.td
chelini: Keep things consistent with MLIR and how we do in other files.

include/TPP/Dialect/Xsmm/XsmmOps.td
chelini: Went for MatmulLike as the op is called Matmul

include/TPP/Dialect/Xsmm/XsmmOps.td
chelini: added.

alheinecke:
why doing it piece-meal? Let's add percision for at least in, out, compute. Transpose, etc.

runtime/XsmmRunnerUtils.cpp
rengolin: We still need to check against the individual flag (as there may be more), so `==` won't do.

chelini: Currently, this is fine as this is actually a single flag. We need to extend `Xsmm_GemmFlags` to accept at list of enums instead of a single 1. Unclear how to do in .td

runtime/XsmmRunnerUtils.cpp
adam-smnk: You probably want bitwise AND here:
` if (flag & LIBXSMM_GEMM_FLAG_VNNI_B) {`
If it still complains about implicit conversion, you might need to throw in `== 1` to check if it is set.

chelini: I will try to OR prior to conversion to func. Will take a bit of time.

chelini: yes, agree. There is not OR atm.

rengolin: > You probably want bitwise AND here

D'oh! That's what I meant. Sorry. Updated the comment.

include/TPP/Dialect/Xsmm/XsmmOps.td
chelini: Yes, but this would require custom builders as the assembly format dictates the builder signature. I can do it but this will PR will likely grow even more.

runtime/XsmmRunnerUtils.cpp
chelini: currently is a single flag

rengolin: I don;'t think the AND and the OR need to be done at the same time. You can easily check for the AND here and do the OR to set the flags later.

KavithaTipturMadhu (approved):

Issue #474, opened 2023-04-13
Convert `linagl.generic { mul + add }` to `linalg.matmul` before blocking & tiling


rengolin:
Our block and tile pass only looks at named ops, and upstream models emit generics, so we need to convert them to named ops before we try to block and fuse, otherwise we won't get decent performance from most upstream code.

This also helps our unpacked benchmarks, which is one of our main priorities right now.

@chelini

rengolin:
Also, this needs to be in the default pipeline to be closed as done.

PR #476, opened 2023-04-14
Convert linalg.generic to linalg.matmul before blocking

chelini:
Enable conversion from linalg.generic to linalg.matmul before blocking. MLP test in linalg-tensor.mlir changed. Specifically, 1) we use BRGEMM instead of GEMM, 2) the identity cannot be pulled into the fused loops as its output is packed to go as input in the matmul.


adam-smnk (approved):
LGTM - it just re-enables existing logic

adam-smnk:
> 2) the identity cannot be pulled into the fused loops as its output is packed to go as input in the matmul

I guess this means that long term we should change the packing strategy to sth more flexible?

chelini:
> Does this really convert generic into matmul? Or just detect generics as matmul?

It converts.

chelini:
> Can you add a `gemm-fp32-3layers-1024-unpacked.mlir` to benchmarks to make sure it has the same performance as the packed version?
> 
> Just copy and paste changing the shapes to unpacked should do.

Added. Each input is packed (7 inputs) and then there is a final unpack. We cannot do better than this here.

benchmarks/benchmarks.json
```diff
@@ -22,6 +22,12 @@
       "flags": [],
       "extensions": [ "avx2" ]
     },
+    "fp32_3x1024_mlir": {
```
rengolin, line 25: you have to change the name

rengolin:
> Added. Each input is packed (7 inputs) and then there is a final unpack. We cannot do better than this here.

Ok, so this will be horrible because of the multiple packs. We need to add a benchmark that has only one packing for the input, with the rest being constants inside the kernel, so that they're packed at compile time. This can be left for another PR.

benchmarks/benchmarks.json
chelini: done.

benchmarks/benchmarks.json
```diff
@@ -22,6 +22,12 @@
       "flags": [],
       "extensions": [ "avx2" ]
     },
+    "unpacked_fp32_3x512.mlir": {
```
rengolin, line 25: the other "unpacked" benchmark name is after the type, perhaps keep the same style would help.

rengolin:
Does this fixes #474?

PR #477, opened 2023-04-17
Custom parser and printer for `MatmulLike` dispatch op

chelini:
Implement a custom parser and printer for `xsmm.matmul.dispatch` and `xsmm.brgemm.dispatch` ops. Currently only for matmul and brgemm but all the dispatch ops will follow. The custom parser and printer is necessary to preserver the pretty printing of the enum, which otherwise would get printed as integer. I did not find a way to get the same result using a custom assembly form.

### DIFF
--- a/lib/TPP/Dialect/Xsmm/XsmmOps.cpp
+++ b/lib/TPP/Dialect/Xsmm/XsmmOps.cpp
@@ -25,7 +25,7 @@ static ParseResult parseEnum(EnumClass &value, OpAsmParser &parser) {
     return failure();
   auto flagAttr = symbolizeEnum<EnumClass>(flag);
   if (!flagAttr)
-    return parser.emitError(loc, "invalid enum") << flag;
+    return parser.emitError(loc, "invalid enum ") << flag;
   value = *flagAttr;
   return success();
 }
@@ -33,7 +33,11 @@ static ParseResult parseEnum(EnumClass &value, OpAsmParser &parser) {
 static ParseResult parserImpl(OpAsmParser &parser, OperationState &result) {
   auto &builder = parser.getBuilder();
   // Parse the input
-  result.addAttribute("inputs", DenseI64ArrayAttr::parse(parser, Type{}));
+  DenseI64ArrayAttr kindAttr;
+  if (parser.parseCustomAttributeWithFallback(kindAttr, Type{}, "inputs",
+                                              result.attributes)) {
+    return failure();
+  }
 
   if (parser.parseKeyword("flags") || parser.parseEqual() ||
       parser.parseLParen())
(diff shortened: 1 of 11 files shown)


lib/TPP/Dialect/Xsmm/XsmmOps.cpp
```diff
@@ -16,3 +16,82 @@
...
+
+template <typename EnumClass>
+static ParseResult parseEnum(EnumClass &value, OpAsmParser &parser) {
+  StringRef flag;
+  auto loc = parser.getCurrentLocation();
+  if (parser.parseKeyword(&flag))
+    return failure();
+  auto flagAttr = symbolizeEnum<EnumClass>(flag);
+  if (!flagAttr)
+    return parser.emitError(loc, "invalid enum") << flag;
```
adam-smnk, line 28:
```suggestion
    return parser.emitError(loc, "invalid enum ") << flag;
```

adam-smnk: Fixes error print formatting

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
```diff
@@ -16,3 +16,82 @@
...
+  value = *flagAttr;
+  return success();
+}
+
+static ParseResult parserImpl(OpAsmParser &parser, OperationState &result) {
```
adam-smnk, line 33: Missing `parseOptionalAttrDict` for `attr-dict`

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
```diff
@@ -16,3 +16,82 @@
...
+  auto &builder = parser.getBuilder();
+  // Parse the input
+  result.addAttribute("inputs", DenseI64ArrayAttr::parse(parser, Type{}));
```
adam-smnk, line 36: nit: this hits assert on any unexpected token - `parseCustomAttributeWithFallback` (or sth else) would be more graceful

adam-smnk (approved):
Some minor comments, otherwise, looks good

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
chelini: Thanks, do we need it?

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
chelini: You mean `DenseI64ArrayAttr::parse`?

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
adam-smnk: Some generic passes could attach attributes. I think it's standard for all ops to support attr-dict.

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
adam-smnk: Yes

PR #479, opened 2023-04-17
Naive tpp bufferization

chelini:
- Unary op: Attempt to bufferize in place if the input type matches with the output type, otheriwse bufferize out-of-place.

- Binary op: Attempt to bufferize in place if the output type matchers either the lhs input or the rhs input, otherwise bufferize out-of-place.

- Ternary op: bufferize on the rhs input operand.

### DIFF
--- a/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
+++ b/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
@@ -1,0 +1,366 @@
+//===- BufferizableOpInterfaceImpl.cpp - Impl. of BufferizableOpInterface -===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Dialect/Tpp/TppDialect.h"
+#include "TPP/Dialect/Tpp/TppOps.h"
+#include "mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h"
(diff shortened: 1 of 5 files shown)


chelini:
@KavithaTipturMadhu let me know if this is blocking you in any way. We can merge, or I can keep this in house for a bit.

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,372 @@
...
+    FailureOr<Value> allocBuffer = getBuffer(rewriter, *alloc, options);
+    if (failed(allocBuffer))
+      return failure();
+    rewriter.create<OpTy>(loc, *buffer, *allocBuffer);
+    replaceOpWithBufferizedValues(rewriter, op, *allocBuffer);
+    return success();
+  }
+  // In-place bufferization.
+  rewriter.create<OpTy>(loc, *buffer, *buffer);
+  replaceOpWithBufferizedValues(rewriter, op, *buffer);
```
adam-smnk, line 52: In case the input tensor has users after the currently bufferized op, are there any safeguards to prevent in-place bufferization?
For example:
```
%0 = unary %A
%1 = matmul %0, %val
%2 = matmul  %A, %val  // unary needs out-of-place bufferization due to data dependency
```

chelini: Bufferization will insert a copy

chelini: I think I have an example somewhere in the test, will look for it

adam-smnk: You're right, I see some cases  with `add` where copy is inserted 👍

KavithaTipturMadhu (approved):

PR #481, opened 2023-04-17
Bump llvm

chelini:
- Some improvement in propagation

- Matcher dialect

Fix tppCombine pass. We cannot get the result of an op with no result. This was returning a garbage pointer before this:

https://github.com/llvm/llvm-project/commit/13b21971fbf5cd47dc3acb0c593aae8ae980c8d8


adam-smnk (approved):

chelini:
Thanks @adam-smnk will keep here 1 days more to give people time to build LLVM. Will land tomorrow lunch time.

rengolin:
> Let's wait until we're sure LLVM isn't at fault in #484, or we might hide the problem again and it could be harder to fix later.

Let's merge some other PRs and if they're all green, we can merge this one today.

rengolin (approved):
Should be fine by now. Thanks for waiting.

PR #482, opened 2023-04-17
Validate unary/binary and ternary dispatch

chelini:
Add verification for `input` argument. `input` are the leading dimensions and sizes passed to the library expect them to be positive and bounded to some value based on the op type.


rengolin:
Please do not merge PRs without approval, nor with broken tests. We allow bypassing the rules for special reasons, this wasn't one of them.

chelini:
> 

Ops, was in the NFC one and I did not see it.

PR #483, opened 2023-04-17
Use custom printer and parser for unary/binary and ternary

chelini:
- Use custom printer and parser to align with matmul and brgemm op

- Simplify dispatch creation in conversion from Xsmm to Func. All the dispatch but Quaternary can share same conversion functions

Quaternary is the only one that is not updated. Will follow up if we need it.

### DIFF
--- a/lib/TPP/ConvertXsmmToFunc.cpp
+++ b/lib/TPP/ConvertXsmmToFunc.cpp
@@ -339,12 +339,36 @@ static func::CallOp buildDispatchCall(Location loc,
   return call;
 }
 
-// TODO: We can derive all the dispatch from the same base class and use
-// this method. Left for future work as require changes on the td.
+template <typename OpTy,
+          typename = std::enable_if_t<
+              std::is_same<OpTy, xsmm::UnaryDispatchOp>::value ||
+              std::is_same<OpTy, xsmm::BinaryDispatchOp>::value ||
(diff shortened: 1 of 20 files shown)


lib/TPP/ConvertXsmmToFunc.cpp
```diff
@@ -339,12 +339,36 @@ static func::CallOp buildDispatchCall(Location loc,
...
+                    SmallVectorImpl<Value> &dispatchOperands,
+                    SmallVectorImpl<Type> &dispatchOperandTypes) {
+  /* do nothing */
+}
+
+void addKindOperand(RewriterBase &rewriter, BrgemmDispatchOp dispatchOp,
+                    SmallVectorImpl<Value> &dispatchOperands,
+                    SmallVectorImpl<Type> &dispatchOperandTypes) {
+  /* do nothing */
+}
```
adam-smnk, line 367: Are these needed? There's an if type check before calling this function anyway

adam-smnk (approved):

lib/TPP/ConvertXsmmToFunc.cpp
chelini: Yes, needed. Otherwise the compiler complains. Need to see specialization for GEMM and BRGEMM

PR #485, opened 2023-04-18
Improve bufferization for binary and ternary ops

chelini:
- Binary: Avoid bufferize on constant-like operation. If an operand is detected to be constant we cannot write to it.

- Ternary: If C is zero filled we can add only a 'write' effect this allows to reuse C without introducing further copies.

### DIFF
--- a/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
+++ b/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
@@ -9,6 +9,8 @@
 #include "TPP/Dialect/Tpp/TppDialect.h"
 #include "TPP/Dialect/Tpp/TppOps.h"
 #include "mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/Operation.h"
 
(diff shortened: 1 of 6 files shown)


test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -341,3 +341,54 @@ func.func @test_mlp_bf16_3_layer_1024(%arg0: tensor<256x1024xbf16>,
...
+  %4 = tpp.add (%3: tensor<256x1024xbf16>, %cst1: tensor<256x1024xbf16>) -> tensor<256x1024xbf16>
+  %5 = tpp.relu (%4: tensor<256x1024xbf16>) -> tensor<256x1024xbf16>
+  %6 = tpp.matmul (%5: tensor<256x1024xbf16>, %cst: tensor<1024x1024xbf16>,
+                   %zero: tensor<256x1024xbf16>) -> tensor<256x1024xbf16>
+  %7 = tpp.add (%6: tensor<256x1024xbf16>, %cst1: tensor<256x1024xbf16>) -> tensor<256x1024xbf16>
+  %8 = tpp.relu (%7: tensor<256x1024xbf16>) -> tensor<256x1024xbf16>
+  return %8 : tensor<256x1024xbf16>
+}
+
+// CHECK-NOT: memref.alloc
```
adam-smnk, line 377: This check should be lower after label checks

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -191,13 +235,17 @@ static bool bufferizesToMemoryWriteBinaryImpl(Operation *op,
   // If the rhs can bufferize in place with the result return true.
   if (opOperand.getOperandNumber() == 1 &&
       opOperand.get().getType() == op->getResult(0).getType() &&
-      !isConstantVal(opOperand.get()))
+      !isConstantVal(opOperand.get())) {
     return true;
+  }
+
   // If the lhs can bufferize in place with the result return true. Note that
   // if both can bufferize with the result we select the rhs first.
```
adam-smnk, line 243: Does it make any difference whether we choose lhs or rhs?

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -126,11 +128,53 @@ struct IdentityBufferizationInterface
...
+      auto globalOp =
+          dyn_cast_or_null<memref::GlobalOp>(SymbolTable::lookupSymbolIn(
+              symbolTableOp, getGlobalOp.getNameAttr()));
+      if (!globalOp)
+        return false;
+      if (globalOp.getConstantInitValue())
+        return true;
+    }
+  }
+  return matchPattern(val, m_Constant());
```
rengolin, line 176: Nit: The shortest check should go first, so I'd change this to:
```
  auto toTensorOp = val.getDefiningOp<bufferization::ToTensorOp>();
  if (!tensorOp)
    return matchPattern(val, m_Constant());
  ...
```
https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code
"Aim to reduce indentation where possible when it doesn’t make it more difficult to understand the code."

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -254,6 +302,12 @@ struct AddBufferizationInterface
 // Ternary
 //===----------------------------------------------------------------------===//
 
+static bool isZeroFilled(Value val) {
```
rengolin, line 305: We have a `isZero` somewhere, this should go there.

Avoid creating little static helpers where there's a more generic function available.

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -278,10 +332,14 @@ static LogicalResult bufferizeTernaryOp(Operation *op, RewriterBase &rewriter,
 }
 
 // Helper function to bufferize ternay operations.
-// All operands bufferize to memory reads.
 static bool bufferizesToMemoryReadTernaryImpl(Operation *op,
                                               OpOperand &opOperand,
                                               const AnalysisState &state) {
+  // If the rhs input operand is zeroFilled, the access is not read/write
```
rengolin, line 338: This isn't `rhs` but the third (accumulation) operand. 'lhs` and `'rhs` only make sense for binary ops.

test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -341,3 +341,54 @@ func.func @test_mlp_bf16_3_layer_1024(%arg0: tensor<256x1024xbf16>,
   %2 = tpp.relu (%1: tensor<256x1024xbf16>) -> tensor<256x1024xbf16>
   return %2 : tensor<256x1024xbf16>
 }
+
+// -----
+
+// CHECK: test_zero
+// CHECK-SAME: %[[ARG0:.+]]: memref<5x5xf32>
+func.func @test_zero(%arg0: tensor<5x5xf32>) -> tensor<5x5xf32> {
```
rengolin, line 349: Can you add either `CHECK-NOT: alloc` or `CHECK: alloc` to make sure we keep the behaviour over future changes?

chelini: in `test_zero`? If we emit an alloc the CHECK will still fail.

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
chelini: No, but we need to pick one. If both are fine we always pick the rhs.

adam-smnk: Ah okay I see the need for it now 👍

chelini:
Any further comment here?

test/Dialect/Tpp/tpp-bufferization.mlir
chelini: done in: https://github.com/plaidml/tpp-mlir/pull/485/commits/1531fd621045320d5d625a5095fff26836441d9a

Issue #484, opened 2023-04-17
Release check-all failing on `Passes/pass-mlp-packed.mlir` - probably bufferization related


rengolin:
Intermittent bug in tpp-benchmark (Release build + check-all). @chelini @adam-smnk @KavithaTipturMadhu 

First seen here:
https://buildkite.com/intel/tpp-benchmark/builds/187#01877eef-b13d-4bb2-95a9-39a859ec8838

Then:
https://buildkite.com/intel/tpp-benchmark/builds/213#01878f46-f50b-4b9c-836f-0f68fb72fbac
https://buildkite.com/intel/tpp-benchmark/builds/215#01878f5c-60f7-4d67-8e98-e082d42b98cf
https://buildkite.com/intel/tpp-benchmark/builds/217#01878f63-6dfc-4ff9-83ac-13f64cfad887
https://buildkite.com/intel/tpp-benchmark/builds/218#01878f67-e42d-4b91-8be2-2ea4602ed9f6

Check-all fails with what looks like a SIGSEGV:
```
  | FAIL: TPP_OPT :: Passes/pass-mlp-packed.mlir (128 of 149)
  | ******************** TEST 'TPP_OPT :: Passes/pass-mlp-packed.mlir' FAILED ********************
  | Script:
  | --
  | : 'RUN: at line 1';   /tmp/tpp-mlir-clang/bin/tpp-opt /nfs_home/buildkite-slurm/builds/pcl-tiergarten-13/intel/tpp-benchmark/test/Passes/pass-mlp-packed.mlir -pack-matmul="block-factors=32,32,32" -propagate-pack-and-unpack -canonicalize -tile-consumer-and-fuse-producers="tile-sizes=1,1" -canonicalize -generalize-tensor-pack-unpack -bufferize -canonicalize -convert-linalg-to-tpp -rewrite-to-brgemm -convert-linalg-to-tpp \| /data/nfs_home/buildkite-slurm/builds/tpp/llvm/9cffb2f74114a870024ca6c9e675bdf0e2a6d0fb/bin/FileCheck /nfs_home/buildkite-slurm/builds/pcl-tiergarten-13/intel/tpp-benchmark/test/Passes/pass-mlp-packed.mlir
  | --
  | Exit Code: 2
  |  
  | Command Output (stderr):
  | --
  | PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
  | Stack dump:
[...]

Issue #454, opened 2023-03-31
Crash in `pass-bias-matmul-relu.mlir` test


adam-smnk:
A random crash occurred in the `test/Passes/pass-bias-matmul-relu.mlir` test while running a routine `tpp-benchmark` CI job (no SANs, release build) on the `main` branch.
The crash is not reliably reproducible, however, we should still investigate the trace and try to figure out what caused it.
From a quick first look, it looks that the crash occurred at the bufferization stage.

See full dump below or in the CI job:
https://buildkite.com/intel/tpp-benchmark/builds/104#01873793-d2ef-4387-8b0f-9f01e36725e5

adam-smnk:
```
[202/203] Running the regression tests
  | -- Testing: 148 tests, 112 workers --
  | Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.
  | FAIL: TPP_OPT :: Passes/pass-bias-matmul-relu.mlir (144 of 148)
  | ******************** TEST 'TPP_OPT :: Passes/pass-bias-matmul-relu.mlir' FAILED ********************
  | Script:
  | --
  | : 'RUN: at line 1';   tpp-opt test/Passes/pass-bias-matmul-relu.mlir -pack-matmul="block-factors=32,32,32" -propagate-pack-and-unpack -canonicalize -tile-consumer-and-fuse-producers -generalize-tensor-pack-unpack -empty-tensor-to-alloc-tensor -one-shot-bufferize="bufferize-function-boundaries allow-return-allocs" -canonicalize  -drop-equivalent-buffer-results -finalizing-bufferize -canonicalize -convert-linalg-to-tpp -rewrite-to-brgemm -convert-linalg-to-tpp \| FileCheck test/Passes/pass-bias-matmul-relu.mlir
  | --
  | Exit Code: 2
  |  
  | Command Output (stderr):
  | --
  | PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
  | Stack dump:
[...]

PR #487, opened 2023-04-18
Check result of getPerfectlyNestedLoops

rengolin:
This may help #484, which was first reported in #454 and just after #408 was merged, which makes use of this function and replacing iterArgs with [size-1] array without making sure size > 0.

Either way, it's a safe code to add anyway.


chelini (approved):
Thanks.

PR #489, opened 2023-04-19
Split `pass-bias-matmul-relu.mlir` and `pass-mlp-packed.mlir`

chelini:
Make sure that the tests checks only single passes. Remove these tests after splitting. See issue https://github.com/plaidml/tpp-mlir/issues/484


rengolin (approved):
Thanks!

PR #490, opened 2023-04-19
Enable linalg to tpp conversion at tensor level

chelini:
Currently, only for Brgemm and Gemm (named). Also, changed behaviour for `TppLoweringPass`. It should only deal with linalg to tpp conversion and not pull in passes like `createConvertForAllToParallelOpPass` or `createRewriteToBatchReduceGemmPass`. These passes need to run after bufferization to preserve parallelism, while we may want to run `TppLoweringPass` at the tensor level to check the conversion.

### DIFF
--- a/test/Passes/DefaultPipeline/tpp-conversion.mlir
+++ b/test/Passes/DefaultPipeline/tpp-conversion.mlir
@@ -46,27 +46,3 @@ func.func @vnni_dialect(%arg0: memref<4x256x512xbf16>,
 // CHECK: tpp.vnni_brgemm
 // CHECK-NOT: vnni.matmul
 // CHECK: tpp.vnni_matmul
-
-// -----
-
-func.func @matmul_lowering(%arg0: tensor<8x9xf32>,
-                           %arg1: tensor<9x8xf32>, %arg2: tensor<8x8xf32>) -> tensor<8x8xf32> {
-  %0 = linalg.matmul ins(%arg0, %arg1: tensor<8x9xf32>, tensor<9x8xf32>)
-                     outs(%arg2: tensor<8x8xf32>) -> tensor<8x8xf32>
-  return %0 : tensor<8x8xf32>
-}
-
-// CHECK-LABEL: matmul_lowering
-// CHECK: tpp.matmul
-
-// -----
-
-func.func @brgemm_lowering(%arg0: tensor<3x5x4xf32>, %arg1: tensor<3x4x5xf32>,
-                          %arg2: tensor<5x5xf32>) -> tensor<5x5xf32> {
-  %0 = linalg.batch_reduce_matmul ins(%arg0, %arg1: tensor<3x5x4xf32>, tensor<3x4x5xf32>)
-                                  outs(%arg2: tensor<5x5xf32>) -> tensor<5x5xf32>
-  return %0 : tensor<5x5xf32>
-}
-
-// CHECK-LABEL: brgemm_lowering
-// CHECK: tpp.brgemm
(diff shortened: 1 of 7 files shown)


test/Passes/DefaultPipeline/tpp-conversion.mlir
```diff
@@ -1,27 +1,5 @@
 // RUN: tpp-opt %s -tpp-conversion -split-input-file | FileCheck %s
 
-#map0 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
-#map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
-#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
-
-// Rewriting to brgemm is done after bufferization which is why it is not part of 'tpp-mapping'.
-// See the default pipeline in 'DefaultTppPasses' for more details.
-func.func @generic_to_brgemm(%arg0: tensor<4x16x32x32xf32>, %arg1: tensor<8x16x32x32xf32>, %arg2: tensor<4x8x32x32xf32>) -> tensor<4x8x32x32xf32> {
```
rengolin, line 9: What's wrong with this test? It should still work, no?

adam-smnk:
> Also, changed behaviour for `TppLoweringPass`
*for `TppConversionPass`

Without `createRewriteToBatchReduceGemmPass` we have nothing to generate `tpp.brgemm`. The lowering from `linalg.generic` should be extended with `isTppBrgemm`.

test/Passes/DefaultPipeline/tpp-conversion.mlir
chelini: I removed from `tpp-conversion` two passes: `createConvertForAllToParallelOpPass` and `createRewriteToBatchReduceGemmPass`. Both of these require to run bufferization first. `createConvertForAllToParallelOpPass` will not work as scf.parallel works only at memref while `createRewriteToBatchReduceGemmPass` will emit `scf.for` and not `scf.parallel` at the tensor level. So this test does not work anymore as we don't rewrite to brgemm at tensor level (also the test comment what misleading - we are at tensor level and we map to brgemm).

chelini:
> > Also, changed behaviour for `TppLoweringPass`
> 
> *for `TppConversionPass`
> 
> Without `createRewriteToBatchReduceGemmPass` we have nothing to generate `tpp.brgemm`. The lowering from `linalg.generic` should be extended with `isTppBrgemm`.

We used to run `createRewriteToBatchReduceGemmPass` twice. One before bufferization (this one) and anther one after bufferization https://github.com/plaidml/tpp-mlir/blob/5532e6bdfa995e46a8112ffc2ebf77a5585163e7/lib/TPP/DefaultTppPasses.cpp#L438)

chelini:
> > Also, changed behaviour for `TppLoweringPass`
> 
> *for `TppConversionPass`
> 
> Without `createRewriteToBatchReduceGemmPass` we have nothing to generate `tpp.brgemm`. The lowering from `linalg.generic` should be extended with `isTppBrgemm`.

If we want to run `createRewriteToBatchReduceGemmPass` at tensor level would be nice to extend the pass to emit scf.forAll like we did for fusion, then we can preserve parallel semantics. If we use linalg utils this is not possible because linalg emit only scf.for and scf.paralle.

adam-smnk:
> We used to run `createRewriteToBatchReduceGemmPass` twice. One before bufferization (this one) and anther one after bufferization

It being present twice was a mistake which I'd always forget to remove but it was pretty harmless. But both instances were after bufferization. Only `TppMappingPass` operates on tensors.

chelini:
> 

Yes, sorry but we run `tpp-mapping` on tensor stuff too (see removed test).

test/Passes/DefaultPipeline/tpp-conversion.mlir
adam-smnk: It's fine to remove it. Conversion of linalg.generic to tpp.brgemm is still covered by the next test case.

chelini: done.

adam-smnk (approved):

PR #491, opened 2023-04-19
Enable linalg to tpp conversion at tensor level - part 2

chelini:
Enable all the rest. Relax constraint on the matchers and allow tensor semantics. Replicate some memref test in `linalg-to-tpp-tensor.mlir`.

### DIFF
--- a/lib/TPP/ConvertLinalgToTpp.cpp
+++ b/lib/TPP/ConvertLinalgToTpp.cpp
@@ -66,8 +66,6 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
 
   LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
                                 PatternRewriter &rewriter) const override {
-    if (!linalgOp.hasBufferSemantics())
-      return rewriter.notifyMatchFailure(linalgOp, "Expect buffer semantics");
     if (!tpp::utils::hasStaticShape(linalgOp))
       return rewriter.notifyMatchFailure(
           linalgOp, "Expect static shape when mapping to tpp");
--- a/lib/TPP/ConvertLinalgToTpp.cpp
+++ b/lib/TPP/ConvertLinalgToTpp.cpp
@@ -9,6 +9,7 @@
 #include "TPP/Dialect/Tpp/TppOps.h"
 #include "TPP/Dialect/Tpp/TppUtils.h"
 #include "TPP/Passes.h"
+#include "TPP/TransformUtils.h"
 #include "TPP/Transforms.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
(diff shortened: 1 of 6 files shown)


lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -66,8 +66,6 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
 
   LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
                                 PatternRewriter &rewriter) const override {
-    if (!linalgOp.hasBufferSemantics())
```
rengolin, line 69: Do we want to restrict this rewrite (later) to only operate at tensor level?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -175,9 +174,8 @@ static bool isTppUnaryOp(linalg::GenericOp linalgOp) {
 }
 
 // Returns true if the linalg.generic maps to a tpp.gemm.
+// TODO: we may want to remove now.
```
rengolin, line 177: Yes, this looks like it could just call the `linalgx` function directly.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: Imo, lowering is simple and generic enough that we can support both tensors and memrefs.
Probably many transformation and optimizations will be restricted to tensor but that should be fine.

rengolin: I'm concerned that this will allow us to "choose" at which level we want to lower, and then end up with two semantics in one implementation. In the past we ended up with code that has `hasBufferSemantics` checks everywhere, and then it gets hard to know what works with what (`N:M` problem).

We only want to lower once. Tensor, or Memref. So, if it's already lowered at Tensor, we should have no more linalg left overs at Memref level. If we do, it was introduced by some other pass and perhaps we shouldn't play with it?

Regardless, in compilers, it's always safer to block unintentional usage than allow "because it might be useful later". Rule of thumb is that it rarely is and when it is, we can't use the same code.

adam-smnk: >  In the past we ended up with code that has hasBufferSemantics checks everywhere

Tpp ops exist in both tensor and memref worlds, we have to perform this validation anyway if a pass expects to work only on tensors or memrefs.

Restricting lowering only to tensors seems a bit artificial. But if supporting lkowering both for tensors and memref adds significant maintenance overhead then it would be reasonable to constrain it.

> We only want to lower once.

I agree, hopefully we won't need double lowering in the default pipeline, and I guess most of our future passes will require tpp on tensors.

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: I know that in the past we didn't run `isTppOp` for matmul (don't remember why anymore) but shouldn't this be included here?

adam-smnk (approved):

lib/TPP/Dialect/Tpp/TppUtils.cpp
chelini: Here the history: 
In the past we used not to check memref abstraction when mapping to TPP. Then we fix with: https://github.com/plaidml/tpp-mlir/commit/09815cc477df600caf70b331cfdc3cf9ce773fc9 and today since we want to handle tensor too we "revert" back. Does this answer your question? I assume at some point `isTPPMatmul` will re-surface back if we want to handle beta = 0. Now tpp.matmul = linalg.matmul but this will change soon.

Issue #414, opened 2023-03-21
Implement unary transform TPP operations


rengolin:
We currently only have one transform, `tpp.identity` which is a mixed bag. We decided to go with more explicit operations.

All TPP transforms are out-of-place operations, so they need the `ins` and `outs` to not be the same memref.

Ops to create:
```mlir
  tpp.zero ins(%0) outs(%0) : memref<MxNxTy> // XOR op

  tpp.copy ins(%0) outs(%1) : (memref<MxNxTy>, memref<MxNxTy>) // COPY

  tpp.broadcast ins(%0) outs(%1) : (memref<2x1xf32>, memref<2x4xf32>) // COPY + BCAST_COL
  tpp.broadcast ins(%0) outs(%1) : (memref<2xf32>, memref<4x2xf32>) // COPY + BCAST_ROW
  tpp.broadcast ins(%0) outs(%1) : (memref<1xf32>, memref<4x2xf32>) // COPY + BCAST_SCALAR
  tpp.broadcast ins(%0) outs(%1) : (memref<f32>, memref<4x2xf32>) // COPY + BCAST_SCALAR
  tpp.broadcast ins(%0) outs(%1) : (f32, memref<4x2xf32>) // COPY + BCAST_SCALAR

  tpp.reduce_add ins(%0) outs(%1) : (memref<MxNxTy>, memref<MxTy>) // REDUCE_X_OP_ADD + REDUCE_ROWS
  tpp.reduce_max ins(%0) outs(%1) : (memref<MxNxTy>, memref<NxTy>) // REDUCE_X_OP_MAX + REDUCE_COLS
  ...

  tpp.transpose ins(%0) outs(%1)  : memref<MxNxTy> -> memref<NxMxTy> // TRANSFORM_NORM_TO_NORMT

  tpp.vnni_pack ins(%0) outs(%1)  : memref<MxNxTy> -> memref<NxMxTy> // TRANSFORM_NORM_TO_VNNI2
```

These are required to implement `tensor.pack` and `tensor.unpack` in TPP (#290).

@chelini @adam-smnk

rengolin:
Table-Gen boilerplate here https://github.com/rengolin/tpp-mlir/tree/unary

rengolin:
With the upstream movement of https://reviews.llvm.org/D153422 and https://reviews.llvm.org/D153421, we may want to add all of those ops to `linalg` instead.

rengolin:
[...]

PR #492, opened 2023-04-20
`tpp.zero` lowering

adam-smnk:
Adds lowering infrastructure to and from `tpp.zero` op.

List of changes:
- XSMM unary kind 'zero'
- Lowering of linalg.generic -> tpp.zero
- Lowering of linalg.fill -> tpp.zero
- Lowering of tpp.zero -> xsmm.unary
- Lowering of tpp.zero to loops
- WithSingleOp predicate capture rules relaxed to allow linalg.yield with a value from a region outside of linalg op region
- TppOp utility functions is*nary

Work towards #414


include/TPP/Dialect/Tpp/TppInterface.td
```diff
@@ -82,6 +82,45 @@ def TppOpInterface : OpInterface<"TppOp"> {
...
+      /*args=*/(ins),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        auto opInputs = ValueRange{$_op.getInputs()};
+        return opInputs.size() == 2;
+      }]
+    >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Return true if the op is ternary.
```
rengolin, line 114: We're dropping `ternary` from the lingo. It's now going to be `unary`, `binary`, `gemm`, `brgemm` (and maybe `brgemm_ext`).

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -35,31 +35,55 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
   LogicalResult rewriteToTppOp(linalg::GenericOp linalgOp,
                                PatternRewriter &rewriter) const {
     SmallVector<Value> operands;
+    if (tpp::utils::isTppZero(linalgOp, &operands)) {
+      assert((operands.size() != 0 && operands.size() <= 2) &&
```
rengolin, line 39: Nit: `size > 0 && size <= 2` looks easier to understand.

I had to pause and think if `size` was signed or unsigned to see if there could be a problem.

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -35,31 +35,55 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
...
+      // Only take the output, the other operand might be a constant.
+      rewriter.replaceOpWithNewOp<tpp::ZeroOp>(linalgOp,
+                                               operands[operands.size() - 1],
+                                               operands[operands.size() - 1]);
+      return success();
+    }
+
+    // Clear operands as the previous checks might have captured some operations
+    // that are no longer relevant for the next check.
+    operands.clear();
```
chelini, line 51: Do we have false positive? We should capture right before returning `true`. Can you share a test cases for this?

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -35,31 +35,55 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
...
                                PatternRewriter &rewriter) const {
     SmallVector<Value> operands;
+    if (tpp::utils::isTppZero(linalgOp, &operands)) {
+      assert((operands.size() != 0 && operands.size() <= 2) &&
+             "zero expects one or two operands");
+
+      // Only take the output, the other operand might be a constant.
+      rewriter.replaceOpWithNewOp<tpp::ZeroOp>(linalgOp,
+                                               operands[operands.size() - 1],
+                                               operands[operands.size() - 1]);
```
rengolin, line 45: I'm assuming one op is at tensor level and two at memref level, but both have to be the same.

Here, we ignore the first, without checking it. What if they're different? If they verify forbids, I still think we should add an `assert` and help developers see that it's an error.

Also, if they're identical, just taking the `[0]` element would do, without the need to have `size - 1`.

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -110,6 +134,30 @@ struct ConvertMatmulToTpp : public OpRewritePattern<linalg::MatmulOp> {
   }
 };
 
+// Convert a linalg.matmul to a tpp.matmul.
```
rengolin, line 137: copy pasta comment

test/Conversion/TppToLoops/tpp-to-loops.mlir
```diff
@@ -53,6 +53,22 @@ func.func @relu_to_loops(%arg0: memref<3x3xf32>, %arg1: memref<3x3xf32>) {
 
 // -----
 
+// CHECK: func.func @zero_to_loops(
+// CHECK-SAME:  %[[ARG0:.+]]: memref<3x3xf32>) {
+func.func @zero_to_loops(%arg0: memref<3x3xf32>) {
```
chelini, line 58: Can we please add a test case for the scalar case? Do we also support broadcasting? The PR i have open should handle simple broadcasting till we have a proper tpp op.

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -613,8 +611,36 @@ struct ConvertTppReluOp : public OpRewritePattern<tpp::ReluOp> {
...
+  }
+};
+
+struct ConvertTppZeroOp : public OpRewritePattern<tpp::ZeroOp> {
+  using OpRewritePattern<tpp::ZeroOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(tpp::ZeroOp zeroOp,
+                                PatternRewriter &rewriter) const override {
+    Type outputType = zeroOp.getInputs()[0].getType();
+    assert(outputType.isa<MemRefType>() && "expect a memref type");
```
chelini, line 625: I think we may want to have this kind of thing in the interface. So that we can assert:
` assert(zeroOp.hasBufferSemantics())`.

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -110,6 +134,30 @@ struct ConvertMatmulToTpp : public OpRewritePattern<linalg::MatmulOp> {
...
+struct ConvertFillToTpp : public OpRewritePattern<linalg::FillOp> {
```
chelini, line 138: Do we test this pattern?

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -35,31 +35,55 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
   LogicalResult rewriteToTppOp(linalg::GenericOp linalgOp,
                                PatternRewriter &rewriter) const {
     SmallVector<Value> operands;
+    if (tpp::utils::isTppZero(linalgOp, &operands)) {
+      assert((operands.size() != 0 && operands.size() <= 2) &&
+             "zero expects one or two operands");
```
chelini, line 40: this comment trips my mind a bit: "zero expects one or two" -> "tpp.zero op expects.."

include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -164,8 +164,8 @@ def Tpp_ZeroOp : Tpp_UnaryOp<"zero"> {
     
     ```mlir
     
-    // out-of-place - memref abstraction.
-    tpp.zero ins(%0: memref<2x2xf32>) outs(%1: memref<2x2xf32>)
+    // in-place - memref abstraction.
+    tpp.zero ins(%0: memref<2x2xf32>) outs(%0: memref<2x2xf32>)
```
chelini, line 168: I assume the in-place should be the only valid combo?

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -71,6 +71,30 @@ struct ConvertTppAddOp : public OpRewritePattern<AddOp> {
...
+  assert(tppOp.isUnary() && "Expect tpp unary op");
+
+  Location loc = tppOp.getLoc();
+
+  SmallVector<Value> ubs;
+  size_t rank = tppOp.getOutput().getType().cast<MemRefType>().getRank();
+  ArrayRef<int64_t> shapeOutput =
+      tppOp.getOutput().getType().cast<MemRefType>().getShape();
+  for (size_t idx = 0; idx < rank; idx++) {
+    Value dim = rewriter.create<arith::ConstantIndexOp>(loc, shapeOutput[idx]);
```
chelini, line 86: I would avoid creating the temp

lib/TPP/ConvertTppToLoops.cpp
```diff
@@ -209,7 +240,8 @@ struct ConvertTppMatmulOp : public OpRewritePattern<MatmulOp> {
     ArrayRef<int64_t> shapeB = matmulOp.getMemRefInputType(1).getShape();
     ArrayRef<int64_t> shapeA = matmulOp.getMemRefInputType(0).getShape();
     if (shapeB.size() == 3)
```
chelini, line 242: multiple lines we should use { }

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -613,8 +611,36 @@ struct ConvertTppReluOp : public OpRewritePattern<tpp::ReluOp> {
...
+
+    MemRefType outputMemRef = outputType.cast<MemRefType>();
```
chelini, line 627: can we use the interface method here?

test/Conversion/TppToLoops/tpp-to-loops.mlir
adam-smnk: Actually to support scalar we'd need to also allow scalars in the output as only in-place computation is allowed for tpp.zero.
Supporting that is questionable in the first place so actually there should be no scalar case. I'll remove it.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: At this stage we're looking at `linalg.generic` operands not `tpp.zero` operands.
This takes into account the `@zero_fill_arg` test case (`linalg-to-tpp-memref.mlir`) where the two `linalg.generic` operands are: lhs a zero constant `arith.constant 0.0` and rhs a `memref`.
In case like `@zero_fill_const` only one `linalg.generic` operand is returned that is a `memref` from its `outs`. So we need to only take the output which will always be the last item in `operands`.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: There was but you're right that we shouldn't pollute `operands` on false match. Adapted the code to handle that at isTpp* level.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: Yes, I added `@linalg_fill_*` test cases.

lib/TPP/ConvertTppToLoops.cpp
adam-smnk: I'm not sure what would be the alternative?

test/Conversion/TppToLoops/tpp-to-loops.mlir
chelini: ok perfect, not scalar on the output pls.

lib/TPP/ConvertLinalgToTpp.cpp
rengolin: I get it but we still need to make sure they're the same, as the semantics isn't the same if not.

I also think the assert above should really not be there. 

It should just not be a match.

lib/TPP/ConvertTppToLoops.cpp
rengolin: IIUC, this isn't a temp, it's just indices to the loop.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: But the operands of `linalg.generic` don't have to be the same for to represent zero fill. As long as we fill a 1D or a 2D buffer with zero values that is equivalent to a `tpp.zero`.

lib/TPP/ConvertTppToXsmm.cpp
adam-smnk: Added where possible.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: We allow similar case with relu it's just this operand matching complexity is hidden somewhere else.
I'll just shift it to TppUtils as it is for relu to make this conversion logic cleaner.

rengolin: In the examples you have, the output memref is always the last one. What if it isn't? What if there are two inputs? Or two outputs? You still have two operands and you'll always take the last. The code you wrote matches your tests, yes, but it can also match other cases that are not `tpp.zero`.

At the very least, it passes the information that it's "the last operand" that is relevant, when actually, it's "the destination memref". If we make that clear, it's not only harder to match the wrong thing here, but it also helps developers to understand what the constraints are.

My suggestion is to:
1. move all checks to match a `tpp.zero` inside `isTppZero`, including all asserts and op checks.
2. create a `tpp.zero` with the operand that is **guaranteed** to be the output memref.

lib/TPP/ConvertLinalgToTpp.cpp
```diff
@@ -35,33 +35,51 @@ struct ConvertGenericOpToTpp : public OpRewritePattern<linalg::GenericOp> {
...
+      // Only take the output, the other operand might be a constant.
+      rewriter.replaceOpWithNewOp<tpp::ZeroOp>(linalgOp,
+                                               operands[operands.size() - 1],
+                                               operands[operands.size() - 1]);
+      return success();
+    }
+
     if (tpp::utils::isTppIdentity(linalgOp, &operands)) {
-      assert(operands.size() == 2 && "Expect two operands");
+      if (operands.size() != 2)
```
rengolin, line 50: Same comment here, the `operands.size()` should be inside the match above and just "not match" and continue.

lib/TPP/ConvertLinalgToTpp.cpp
adam-smnk: > What if there are two inputs? Or two outputs?

Then it fails the definition of a tpp unary op. Or the definition of an tpp op all together if it is really weird generic.
But I agree here that it is unclear at the conversion level. These assumptions are tied to `isTppZero`, I'll shift all the validation logic there.

lib/TPP/ConvertLinalgToTpp.cpp
rengolin: Similar to all other existing ones... We shouldn't fail the compilation process if we match only half-way through, we should just not match and continue.

adam-smnk: I agree the internal if check shouldn't be there. Its place is in the util matcher. I'd still keep the asserts just as a double check. If the assert ever triggers, it means the matcher is too generic.

rengolin: We can do this at another commit, but we *have* to move to that style or we'll keep generating broken code.

rengolin (approved):

PR #497, opened 2023-04-21
Skip default builder for tpp

chelini:
We cannot call default builder as we need to set operand segement sizes in different way if we are in tensor or memref land.

### DIFF
--- a/include/TPP/Dialect/Tpp/TppOps.td
+++ b/include/TPP/Dialect/Tpp/TppOps.td
@@ -85,7 +85,8 @@ class Tpp_UnaryOp<string mnemonic, list<Trait> traits = []> :
                        Variadic<TppOutputOperand>:$outputs);
   let results = (outs Variadic<TppTensor>:$results);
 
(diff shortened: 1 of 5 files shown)


include/TPP/Dialect/Tpp/TppOps.td
```diff
@@ -85,7 +85,8 @@ class Tpp_UnaryOp<string mnemonic, list<Trait> traits = []> :
                        Variadic<TppOutputOperand>:$outputs);
   let results = (outs Variadic<TppTensor>:$results);
 
-  let hasCustomAssemblyFormat = 1; 
+  let hasCustomAssemblyFormat = 1;
+  let skipDefaultBuilders = 1;
```
adam-smnk, line 89: We don't need that for other ops?

chelini: forgot.

adam-smnk (approved):

PR #499, opened 2023-04-21
Rename tpp and xsmm matmul ops

chelini:
Rename `tpp.matmul` to `tpp.gemm` and `xsmm.matmul` to `xsmm.gemm`. This better reflects the naming used
in LIBXSMM. `tpp.gemm` assumes beta = 1.


chelini:
depends on #498

test/Dialect/Tpp/tpp-invalid.mlir
```diff
@@ -310,14 +309,6 @@ func.func @tpp_relu_invalid_number_of_operands(%arg0: tensor<2x2xf32>) -> tensor
 
 // -----
 
-func.func @tpp_matmul_invalid_number_of_operands(%arg0: tensor<2x2xf32>) -> tensor<2x2xf32> {
```
chelini, line 313: this test was duplicated.

adam-smnk: Isn't the other one on memref?

chelini: Good catch, sorry. Added back

rengolin:
> For optimization purposes we will need a new `tpp.matmul` where the beta equals 0, thus rename the current `tpp.matmul` to `tpp.gemm`.

We don't need `tpp.matmul`, we'll have `tpp.group { tpp.zero + tpp.gemm }` which is the same thing without the need of a new op.

> `tpp.gemm` now assumes beta = 1.

It shouldn't.

rengolin:
> > `tpp.gemm` now assumes beta = 1.
> 
> It shouldn't.

Correction: it should. This is the safest option, and would generate less efficient code if it actually executed `tpp.zero` first.

Update to beta = 0 if there's a `tpp.zero` before can be done as an optimization.

Temporarily we could have a flag (which would be a replacement from `tpp.zero+tpp.gemm -> tpp.gemm { beta = 0 }` anyway. 

But this should *NOT* be the long term solution, as we'll have `tpp.group` for the very reason to not need to create too many ops and attributes.

PR #500, opened 2023-04-21
Improve bufferization for tpp dialect

chelini:
- We cannot return no-alias if we decide to bufferize on the lhs operand
since the rhs is constant. `alloc_must_escape` tests this behavior.
Other tests are added to make sure we can handle tiling optimizations.
- Drop ill-formed bufferization test, we should not pass buffer DPS-style 
for now.
- Add support for in-place bufferization for binary ops.
- Fix crash when the operand is not a tensor type.

### DIFF
--- a/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
+++ b/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
@@ -22,8 +22,8 @@ namespace mlir {
 namespace tpp {
 namespace {
 
-FailureOr<Value> getMemrefBuffer(RewriterBase &rewriter, Value toBufferize,
-                                 BufferizationOptions const &options) {
+FailureOr<Value> getBufferOrScalar(RewriterBase &rewriter, Value toBufferize,
+                                   BufferizationOptions const &options) {
   // nothing to bufferize.
   if (!isa<ShapedType>(toBufferize.getType()))
     return toBufferize;
@@ -41,7 +41,7 @@ static LogicalResult bufferizeUnaryOp(Operation *op, RewriterBase &rewriter,
   auto unaryOp = cast<OpTy>(op);
   auto loc = unaryOp.getLoc();
   FailureOr<Value> buffer =
-      getMemrefBuffer(rewriter, unaryOp.getInputs()[0], options);
+      getBufferOrScalar(rewriter, unaryOp.getInputs()[0], options);
   if (failed(buffer))
     return failure();
   // Out-of-place bufferization.
@@ -51,7 +51,7 @@ static LogicalResult bufferizeUnaryOp(Operation *op, RewriterBase &rewriter,
                                      /*escape=*/true, options, /*copy=*/false);
     if (failed(alloc))
       return failure();
-    FailureOr<Value> allocBuffer = getMemrefBuffer(rewriter, *alloc, options);
+    FailureOr<Value> allocBuffer = getBufferOrScalar(rewriter, *alloc, options);
     if (failed(allocBuffer))
       return failure();
     rewriter.create<OpTy>(loc, *buffer, *allocBuffer);
@@ -194,11 +194,11 @@ static LogicalResult bufferizeBinaryOp(Operation *op, RewriterBase &rewriter,
   auto binaryOp = cast<OpTy>(op);
   auto loc = binaryOp.getLoc();
   FailureOr<Value> lhsBuffer =
-      getMemrefBuffer(rewriter, binaryOp.getInputs()[0], options);
+      getBufferOrScalar(rewriter, binaryOp.getInputs()[0], options);
   if (failed(lhsBuffer))
     return failure();
   FailureOr<Value> rhsBuffer =
-      getMemrefBuffer(rewriter, binaryOp.getInputs()[1], options);
+      getBufferOrScalar(rewriter, binaryOp.getInputs()[1], options);
   if (failed(rhsBuffer))
     return failure();
   // Out-of-place bufferization.
@@ -212,7 +212,7 @@ static LogicalResult bufferizeBinaryOp(Operation *op, RewriterBase &rewriter,
                                      /*escape=*/true, options, /*copy=*/false);
     if (failed(alloc))
       return failure();
-    FailureOr<Value> allocBuffer = getMemrefBuffer(rewriter, *alloc, options);
+    FailureOr<Value> allocBuffer = getBufferOrScalar(rewriter, *alloc, options);
     if (failed(allocBuffer))
       return failure();
     rewriter.create<OpTy>(loc, ValueRange{*lhsBuffer, *rhsBuffer},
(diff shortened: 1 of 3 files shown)


chelini:
depends on #498.

test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -422,3 +381,127 @@ func.func @brgemm_in_loops(%arg0: tensor<4x16x32x32xf32>, %arg1: tensor<8x16x32x
...
+      tensor.parallel_insert_slice %11 into %arg3[%arg1, %arg2] [32, 32] [1, 1] : tensor<32x32xf32> into tensor<256x1024xf32>
+    }
+  }
+  return %8 : tensor<256x1024xf32>
+}
+
+// -----
+
+// CHECK-LABEL: alloc_must_escape
+func.func @alloc_must_escape(%arg0: tensor<256x1024xf32>) -> tensor<256x1024xf32> {
```
adam-smnk, line 496: Could you add a small comment what it tests? I'll forget tomorrow and will be confused when it breaks 😅

adam-smnk: Maybe a word why it would inject dealloc in the first place.

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -134,6 +134,8 @@ struct ZeroBufferizationInterface
                                                     tpp::ZeroOp> {
   bool bufferizesToMemoryRead(Operation *op, OpOperand &opOperand,
                               const AnalysisState &state) const {
+    // TODO: here we may want to return false as tpp.zero
+    // as only "write" effects.
```
rengolin, line 138:
```suggestion
    // has only "write" effects.
```

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -260,17 +262,23 @@ getAliasingOpResultsBinaryImpl(Operation *op, OpOperand &opOperand,
...
+      !isConstantVal(opOperand.get())) {
     return {{op->getOpResult(0), BufferRelation::Equivalent,
              /*isDefinite=*/true}};
+  }
   // If the lhs can bufferize in place with the result return the lhs. Note
   // that if both can bufferize with the result we select the rhs first.
   if (opOperand.getOperandNumber() == 0) {
-    if (op->getOpOperand(1).get().getType() == op->getResult(0).getType())
+    // The lhs does not alias with op result if we can bufferize on the rhs.
+    if (op->getOpOperand(1).get().getType() == op->getResult(0).getType() &&
```
rengolin, line 273:
```suggestion
    if (op->getOpOperand(1)->getType() == op->getResult(0).getType() &&
```

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -260,17 +262,23 @@ getAliasingOpResultsBinaryImpl(Operation *op, OpOperand &opOperand,
...
+        !isConstantVal(op->getOpOperand(1).get())) {
       return {};
-    if (opOperand.get().getType() == op->getResult(0).getType())
+    }
+    // We cannot bufferize on rhs, lhs alias opResult.
+    if (opOperand.get().getType() == op->getResult(0).getType()) {
```
rengolin, line 278:
```suggestion
    if (opOperand->getType() == op->getResult(0).getType()) {
```

test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -422,3 +381,130 @@ func.func @brgemm_in_loops(%arg0: tensor<4x16x32x32xf32>, %arg1: tensor<8x16x32x
 }
 
 // CHECK: tpp.brgemm ins(%{{.+}} : memref<16x32x32xf32, strided<[1024, 32, 1], offset: ?>>, %{{.+}} : memref<16x32x32xf32, strided<[1024, 32, 1], offset: ?>>, %{{.+}} : memref<32x32xf32, strided<[32, 1], offset: ?>>) outs(%{{.+}} : memref<32x32xf32, strided<[32, 1], offset: ?>>)
+
+// -----
+
+func.func @mlp_single_layer(%arg0: tensor<256x1024xf32>) -> tensor<256x1024xf32> {
```
rengolin, line 387: No bf16 tests? Removed above but not added elsewhere?

test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -422,3 +381,130 @@ func.func @brgemm_in_loops(%arg0: tensor<4x16x32x32xf32>, %arg1: tensor<8x16x32x
...
+    }
+  }
+  return %8 : tensor<256x1024xf32>
+}
+
+// -----
+
+// CHECK-LABEL: alloc_must_escape
+// The kernel allocates memory via tensor.empty. The allocation need
+// to survive, meaning that bufferization should not introduce a
```
rengolin, line 497:
```suggestion
// to survive because it's a return value, meaning that bufferization should not introduce a 
```

test/Dialect/Tpp/tpp-bufferization.mlir
chelini: This was due to a bug in `getAliasingOpResultsBinaryImpl` which was not setting proper alias information, thus bufferization was inserting the dealloc. I made explicit in the IR what we expect from bufferization.

test/Dialect/Tpp/tpp-bufferization.mlir
chelini: We don't need to test multiple data type. The test I removed was wrong as it was in DPS style.

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -134,6 +134,8 @@ struct ZeroBufferizationInterface
                                                     tpp::ZeroOp> {
   bool bufferizesToMemoryRead(Operation *op, OpOperand &opOperand,
                               const AnalysisState &state) const {
+    // TODO: here we may want to return false as tpp.zero
+    // has only "write" effects.
```
adam-smnk, line 138: Sounds correct to me, can it just return false now or does it break sth?

test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -422,3 +381,134 @@ func.func @brgemm_in_loops(%arg0: tensor<4x16x32x32xf32>, %arg1: tensor<8x16x32x
...
+
+// -----
+
+// CHECK-LABEL: alloc_must_escape
+// The kernel allocates memory via tensor.empty. The allocation need to survive
+// because it's a return value, meaning that bufferization should not introduce
+// a dealloc.
+func.func @alloc_must_escape(%arg0: tensor<256x1024xf32>) -> tensor<256x1024xf32> {
+  // CHECK-NOT: memref.dealloc
+  // CHECK: memref.alloc
```
adam-smnk, line 501:
```suggestion
  // CHECK-NOT: memref.dealloc
  // CHECK: memref.alloc
  // CHECK-NOT: memref.dealloc
```

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
chelini: not a ptr type, we cannot use `->`.

rengolin: Sorry, mistook `shared_ptr` semantics. Ignore me.

lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -22,6 +22,14 @@ namespace mlir {
 namespace tpp {
 namespace {
 
+FailureOr<Value> getMemrefBuffer(RewriterBase &rewriter, Value toBufferize,
```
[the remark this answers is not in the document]
chelini: yes, it can fails. It is a tiny wrapper around the bufferization `getBuffer`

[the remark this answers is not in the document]
rengolin: Sorry, my comment was about the function name `getMemrefBuffer`. It doesn't always.

Perhaps something like `getBufferOrScalar` would be better.

adam-smnk (approved):

PR #505, opened 2023-04-24
Allow random behaviour without user seed

rengolin:
Generates a new seed (with localtime) instead. This allows for easily running random benchmarks without having to worry about propagating the seed through the multiple harnesses.

Once the harness can handle seeds (and use the same for all different tools), it can just pass `-seed` and it will behave accordingly.

### DIFF
--- a/tools/tpp-run/tpp-run.cpp
+++ b/tools/tpp-run/tpp-run.cpp
@@ -138,6 +138,11 @@ MLIRBench::PrintStage parsePrintStage(StringRef stage) {
 // so we can modify the IR with the needed wrappers
 static LogicalResult prepareMLIRKernel(Operation *op,
                                        JitRunnerOptions &options) {
+  // Randon options need seed
+  if (!seed && (splatRandom || initType == "random" || initType == "normal")) {
+    seed = std::time(0);
+  }
+
   // Benchmark object
   MLIRBenchConfig config(seed, tppToLoops, linalgToLoops,
                          parseTensorInitType(initType));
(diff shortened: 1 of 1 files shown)


tools/tpp-run/tpp-run.cpp
```diff
@@ -138,6 +138,11 @@ MLIRBench::PrintStage parsePrintStage(StringRef stage) {
 // so we can modify the IR with the needed wrappers
 static LogicalResult prepareMLIRKernel(Operation *op,
                                        JitRunnerOptions &options) {
+  // Randon options need seed
+  if (!seed && (splatRandom || initType == "random" || initType == "normal")) {
+    seed = std::time(0);
```
adam-smnk, line 143: I like the idea of input randomization by default especially as it prevent these fake optimizations where we pass single dense tensor for all arguments.
But wouldn't it be more predictable to have fixed default seed? I feel like this type of randomness on top of random initialization might be asking for trouble. On the other hand, it technically increases testing coverage but it might be tricky to debug.

adam-smnk: Do we log the seed that is used?

rengolin: > Do we log the seed that is used?

No. I wanted to do this but it messes up the output, which we rely on for the benchmarks. If we want to control the seed we need to do that at the harness/driver level, in which case it's trivial to log (and we should do every time).

> But wouldn't it be more predictable to have fixed default seed?

Here, I'm not looking for predictable, as you said yourself...

> it technically increases testing coverage

which is what I'm aiming at at this stage. However...

> it might be tricky to debug

Bear in mind that this will only be enable in the case a user asks for `-splat-to-random` and *not* `-seed N`, which gives user intention pretty clearly that they don't care.

If we use this in our benchmark CI, it could be a problem. But right now, we're not having that problem (need to debug issues related to random initialization) because we're not randomly initializing anything in the benchmark. 

So the next logical step is to make that a reality in the driver/harness. But right now, it's better to have it random when you ask for it, rather than force seed propagation.

adam-smnk: Good points, the strategy is sound and the turning knobs are still exposed for easy control of testing behavior.
Let's go with this approach.

adam-smnk (approved):
Looks good. Thanks for the extra insight, the testing strategy seems solid.

PR #508, opened 2023-04-24
Tpp-run support for AArch64 systems

### LINKED ISSUE #501: clean aarch64 support for tpp-mlir
I have updated (after successfully testing) the README.md with instructions on how to build tpp-mlir on aarch64 machines. However, some test fail, when using tpp-run, simply as tpp-run forces right now x86 settings (x86 triple and specific x86 machines). 

I'm not sure if there is a clean way to only force skylake-avx2 as default, when the target is actually x86-64. A simple fix is to guard all the x86-64 specific stuff in tpp-run.cpp by `#if defined(__x86_64__)` .

rengolin:
Fixes #501

Adding changes that made it pass, `check-all` and `benchmarks` targets on my AArch64 system.


tools/tpp-run/tpp-run.cpp
```diff
@@ -91,19 +91,37 @@ llvm::cl::opt<unsigned>
...
+#endif
 
 // Target CPU name
 // Default skylake is old enough to be relevant for most cases
 llvm::cl::opt<std::string>
     cpuName("cpu", llvm::cl::desc("CPU name (sapphirerapids, alderlake, etc)"),
+#if defined(__x86_64__)
             llvm::cl::init("skylake"));
+#elif defined(__aarch64__)
+            llvm::cl::init("cortex-a53"));
```
alheinecke, line 109: Is this is pretty old aarch64 CPU, such that everything works on Graviton, M1/M2 etc?

alheinecke (approved):
Minor questions on the cortex A53: we don't really care about which ARM chip we use for loop optimization. It should be just a reasonable aarch64 vanilla 8.1 / 8.2 without any fancy extensions such that we are sure it runs everywhere out of the box, e.g. M1/M2 or Graviton 2/3. 

In general: why not using host for the CPU JIT and only adjust the FPU?

tools/tpp-run/tpp-run.cpp
rengolin: It's the baseline. This is just the default in case we don't pass anything. Code generated for A53 will work in pretty much on every v8/v9 core.

For GVT we may want to pass the argument to define the right core, to make use of SVE when available. But we don't want to rely on non-TPP code-gen for these benchmarks anyway, so it should be fine anyway.

rengolin:
> Minor questions on the cortex A53: we don't really care about which ARM chip we use for loop optimization. It should be just a reasonable aarch64 vanilla 8.1 / 8.2 without any fancy extensions such that we are sure it runs everywhere out of the box, e.g. M1/M2 or Graviton 2/3.

Correct.

> In general: why not using host for the CPU JIT and only adjust the FPU?

This code is using the `llvm::Target`, it doesn't recognise `host` as a valid CPU. IIRC, Clang is the one replacing "host" with the appropriate name. If we pass the CPU empty (""), then it selects the minimum common denominator and it won't accept the common FPUs (because Pentium didn't have anything past SSE).

Getting the CPU name isn't trivial at this level. Intel flags are per family, not model name. On Arm, a lot of cores don't even have names in `cpuinfo`, or even `cpuinfo` at all.

Using "skylake" for x86 is a hack, to make it easier to set an FPU while not setting the CPU. In theory, if you pass both CPU/FPU, then it should do the right thing every time.

alheinecke (approved):

PR #509, opened 2023-04-24
Add shaped GEMM benchmarks, 256x1025x1024

rengolin:
Also adds a new Buildkite job "TPP-performance" which is scheduled every week for now due to the length of these benchmarks. We can also trigger them manualy on PRs using `pull/#/head` as branch name.
    
Also adds the OMP tests with higher iterations to get more confidence than the PR/CI builds.
    
Problems:
     * N is still 1000 for now, we'll increase as we get more comfortable
     * ~~MLIR BF16 matmul is completely broken, needs fixing before merge~~
     * Reports are not showing on each tab (@hfp)


alheinecke:
Please also add GEMM with BiasAdd and ReLU

rengolin:
> Please also add GEMM with BiasAdd and ReLU

This is just the infrastructure part. The rest of the benchmarks will come later. For now I just want to make sure we enable the performance job on `main` and have something running there. 

Once this is in, it will be really easy to add lots of other builds soon after by more than just me.

Issue #488, opened 2023-04-18
[RFC] Benchmark kernels structure


adam-smnk:
We need to align in terms of how we write our test kernels, especially input IR that serves as our performance benchmarks (`benchmarks/mlir/*.mlir`) and larger models we use to manually evaluate quality of our passes (`test/Models`).
Currently, these MLIR IR files are a collection of random tests written at different point of our tensor compiler development. As our assumptions and constantly evolve, the test kernel also have to be updated and adapted to present times.

Here, I want to define and show some examples of how these test IR kernel should be structured such that:
1. We do not keep chasing unrealistic edge cases and complex IR structures that are unlikely to be even generated by automated front-ends.
2. Performance results represent realistic numbers i.e., if some memory allocations for intermediate results are unavoidable, then we have to pay the price and it should reflected in the measurements.
3. The test kernels are well defined which ensures semantic correctness (they represent exactly and only what we want to measure) and the IR is less susceptible to upstream changes (e.g., an explicit `alloc` will always* remain in its place - *subject to some optimizations that specifically target `alloc`s).

### Kernel inputs and outputs
All arguments of a kernel function (inputs) should be treated as constant/read-only values. No operation should write to any of the function arguments and the function arguments should not be used for `outs` of any up. This also ensures that bufferization does not use kernel arguments as output buffers at memref level.
The benchmarking framework (part of `tpp-run`) will still automatically allocate the input buffers based on the kernel arguments and random initialize them when requested.

All the significant results of a kernel function (outputs) e.g., values to be printed for `FileCheck` or just the final computed value, should have their buffers allocated within the kernel and be returned at the end of the kernel via `func.return`/`return` op. The destination passing should not be used for the kernel outputs.
This applies to both tensors and memrefs. If there are multiple values of interest, they all should be returned from a kernel which will result in multi-result function call `%out:N = func.call`.
The benchmarking framework (part of `tpp-run`) will automatically handles lifetime of the returned buffers. Note that currently there is no support for multi-return kernels but it can be added when needed.

#### An example of single-return kernel
```
func.func @entry(%A: tensor<48x96xf32>, %B: tensor<96x64xf32>) -> tensor<48x64xf32> {
  %0 = tensor.empty() : tensor<48x64xf32>
  %C = linalg.matmul ins(%A, %B: tensor<48x96xf32>, tensor<96x64xf32>) outs(%0: tensor<48x64xf32>) -> tensor<48x64xf32>
[...]

Issue #409, opened 2023-03-17
`tpp-opt` | `tpp-run` infinite execution


adam-smnk:
While optimizing benchmark pipelines as a part of #380, I ran across an issue with executing generate IR.
The same problem has been reproduced on the main branch (c886b3b).

The following pipeline using in `mlp-bf16-10layers-3584.mlir` benchmarks works correctly:
`./bin/tpp-opt -element-wise-fusion -tile-consumer-and-fuse-producers="tile-sizes=1,1 max-depth=2" ../test/Benchmarks/mlp-bf16-10layers-3584.mlir | ./bin/tpp-run -e entry -entry-point-result=void`
However, when the default pipeline is run by `tpp-opt` before `tpp-run`:
`./bin/tpp-opt -element-wise-fusion -tile-consumer-and-fuse-producers="tile-sizes=1,1 max-depth=2" -default-tpp-passes ../test/Benchmarks/mlp-bf16-10layers-3584.mlir | ./bin/tpp-run -e entry -entry-point-result=void`
the generated IR when executed does not seem terminate at all and continuously allocates memory - I had to kill the execution once it allocated over 60GB of RAM. For comparison, valid test run consumes ~9GB.

The same issue occurs when the default pipeline is disabled in `tpp-run` to avoid double compilation.
The problem currently seems to arise when compiling current `mlp-bf16-*` benchmarks.

adam-smnk:
The issue is caused by creating `memref.global`s for all the 30 test kernel input arguments.
The exact root cause is a bit unclear at the moment but once the input IR with many large globals is passed to jitter, their initialization consumes large amounts of memory. And these memory appears to be persistent for the lifetime of the runner.

The difference in the behavior between `tpp-run` and `tpp-opt | tpp-run` was due to the fact that by running `tpp-run` alone CSE would substitute most of the kernel arguments leaving only 3 globals. By running `tpp-opt` first with bufferization causes `tpp-run` to allocate 30 separate global buffers that cannot be folded away. This leads to high RAM usage.

The execution itself is not infinite, however, large RAM usage causes significant slow downs or crashes. Due to this issue, `memref.global`s are not suitable for passing large input or output data buffers. A long term solution will be provided by #271.

Closing this issue.

PR #510, opened 2023-04-24
Writable benchmark tensor arguments

adam-smnk:
Generated tensors used for benchmark kernel function arguments are now writable.
This enables use of destination passing for benchmark kernels and the benchmarking framework behavior consistent for both memrefs and tensors. Making tensor globals writable avoids creation of allocs and copies for arguments which are written into by a kernel function.

Additionally, 10 layer MLP benchmark is refactored to minimize number of passed kernel arguments due to high runtime memory consumption (see issue #409).

Work towards #488

### DIFF
--- a/lib/TPP/BuilderUtils.cpp
+++ b/lib/TPP/BuilderUtils.cpp
@@ -1,4 +1,5 @@
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
@@ -6,6 +7,8 @@
 #include "mlir/IR/BuiltinOps.h"
 
 #include "TPP/BuilderUtils.h"
+
+#include <optional>
 
 namespace mlir {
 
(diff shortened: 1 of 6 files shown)


adam-smnk:
Note that the alloc+copy operations inserted by bufferization into the timing loop are actually correct.
The main issue is how our kernels are structured. This PR primarily make benchmarks behavior consistent between memrefs and tensors, and lowers overhead of destination passing in kernels if it is absolutely necessary. Although, we should avoid it in general.

I will follow up with another PR with further kernel refactoring in style of MHA and the 10 layer MLP refactor from this PR.

adam-smnk:
> how can we avoid alloc and single thread copy in the timing loop? Is this what you mean with the follow-up PR?

This PR eliminates these allocs and copies.
It's just that currently our benchmarks are poorly written and not representative. We shouldn't be passing 10+ arguments and temporary buffers should be allocated (and zero initialized when needed) within the kernels.

alheinecke:
got it

alheinecke:
seems to work for GEMM, but not MLP:
Benchmark: openmp
gemm_fp32_3x1024_omp_4_mlir :   325.748 +-     7.381 gflops
mlp_fp32_3x1024_omp_4_mlir  :   213.624 +-     3.068 gflops
gemm_fp32_3x1024_omp_8_mlir :   648.787 +-    13.544 gflops
mlp_fp32_3x1024_omp_8_mlir  :   423.876 +-     8.763 gflops
gemm_fp32_3x1024_omp_16_mlir:  1229.691 +-     9.473 gflops
mlp_fp32_3x1024_omp_16_mlir :   810.904 +-    15.483 gflops

lib/TPP/BuilderUtils.cpp
```diff
@@ -67,8 +69,17 @@ Value getConstIndex(OpBuilder &builder, int value) {
 }
 
 Value createDenseTensor(OpBuilder &builder, TensorInitType initType,
-                        TensorType type, int seed) {
+                        TensorType type, int seed,
+                        std::optional<ModuleOp> module) {
   auto unkLoc = builder.getUnknownLoc();
+
+  if (module) {
```
rengolin, line 76: I'm not sure I understand what the module means here... Is it just a differentiation between tensor and memref on the TypeSwitch?

adam-smnk: Fair point, it was poor abstraction. Cleaned it up now.

rengolin (approved):

PR #511, opened 2023-04-24
Increase benchmark iterations for stability

rengolin:
Should have done this a long time ago. Isn't meant to fix anything, but we need to start running these benchmarks more seriously.

This will increase the benchmark job's time, but it's needed. We can clean up some benchmarks if that becomes a problem.


rengolin:
This may be too slow. I was discussing with @hfp earlier about having two benchmark jobs: a quick one for PRs and a long one, once a day, at night.

We can have the short (with `-n 1000`) plus all shapes we want for the larger benchmarking effort.

The quick one can spot obvious mistakes, while the long one can help us keep track longer term.

I'll create a new job and move things around.

rengolin:
Dropping in favour of a separate benchmark job

PR #512, opened 2023-04-25
Split benchmarks

rengolin:
This PR splits the benchmarks into multiple configs and allow the driver to load up all of them for the CI job, with a shorter iteration flag to track basic runs on stable machines (CLX non-turbo).

Main changes:
* Moves configurations into a `config` directory, with `ci` as the base run and `omp` as the scalability checks.
* Adds more OMP runs, including a single-threaded, so we can compare on different machines.
* Allows to override the iterations at the driver level, so we can run PR CI at lower iter and nightly CI at longer ones.
* Can read multiple config files, allowing bundling of existing benchmarks into one CI job.
* Removes the `CPPRun` that was useless. `XSMM-DNN` is the new default now and if we need anything else, we should extend that one.

Future changes:
* Add a new job, for nightly/weekly runs, with `-n 1000` or more on varied shapes, as separate jobs (so they run in parallel)


rengolin:
A few problems spotted:
```
ERROR Benchmark gemm_bf16_dp2_mlir, run gemm_bf16_dp2_3x1024_single_mlir produced no output, can't verify results
ERROR Error executing tpp-run: OMP: Warning #234: OMP_NUM_THREADS: Invalid symbols found. Check the value "".
OMP: Error #101: Out of heap memory.
Command: /tmp/tpp-mlir-clang/bin/tpp-run -n 10 -e entry --entry-point-result=void --print=0 -def-parallel
```

Also, on my machine, the XSMM-DNN don't scale at all, even with `OMP_NUM_THREADS`, I'm probably building it wrong.

rengolin:
Good results: CI time doesn't explode.

rengolin:
> Also, on my machine, the XSMM-DNN don't scale at all, even with `OMP_NUM_THREADS`, I'm probably building it wrong.

Fixed.

rengolin:
> ERROR Error executing tpp-run: OMP: Warning #234: OMP_NUM_THREADS: Invalid symbols found. Check the value "".

Fixed.

Should be good to go now. @adam-smnk

benchmarks/config/omp/mlir-bf16.json
```diff
@@ -5,29 +5,29 @@
       "type": "MLIR",
       "benchmark": "gemm-bf16_dp2-3layers-1024.mlir",
       "environment": {},
-      "flags": [ "-run-args='-def-parallel'" ],
```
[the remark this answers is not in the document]
rengolin: Not only crashes but makes OpenMP run out of stack... :astonished: :exploding_head:

CMakeLists.txt
```diff
@@ -77,11 +78,14 @@ add_subdirectory(runtime)
...
   add_subdirectory(test)
-  add_subdirectory(benchmarks)
 
   set(BENCHMARK_DIR "${PROJECT_SOURCE_DIR}/benchmarks")
-  add_custom_target(benchmarks ${BENCHMARK_DIR}/driver.py -v --build ${PROJECT_BINARY_DIR}
-                    DEPENDS bench_mlp bench_matmul tpp-opt tpp-run xsmm_dnn_mlp
+  set(CONFIG_DIR "${BENCHMARK_DIR}/config")
+  # Run a small set of benchmarks with small iterations on a stable machine (non-turbo) to track simple performance
+  # A separate CI job should track longer running and larger benchmarks elsewhere
+  add_custom_target(benchmarks ${BENCHMARK_DIR}/driver.py -v --build ${PROJECT_BINARY_DIR} -n 100
```
adam-smnk, line 86: I know it's for testing now but we want to keep it `-n 100` should probably go to config JSON files.

adam-smnk (approved):

cmake/modules/openmp.cmake
```diff
@@ -0,0 +1,10 @@
+option(USE_OpenMP "Use OpenMP" ON)
```
chelini, line 1: Why enable OpenMP as default? I would do the opposite (noticed as I don't have openMP).

rengolin: Because CPU performance without parallelisation makes no sense. :)

You should install OpenMP.

chelini: I agree, but people can just want to clone the repo and look at the compiler, I don't think they should install OpenMP for this.

chelini: I see this in reverse: People that work on compiler can set the openMP option, while people that are curios usually don't, and loose interest soon. In any case, let's at least add this requirement in the readme.

chelini: Still I would mention the requirement in the readme.

rengolin: Done!

chelini: Thank you.

PR #513, opened 2023-04-25
Move `tensor.pack` simplification in its own pass (NFC)

chelini:
Avoid entanglement. Also not all the patterns were running simplification.
Make simplification and canonicalization of tensor.pack a standalone pass.

### DIFF
--- a/lib/TPP/DefaultTppPasses.cpp
+++ b/lib/TPP/DefaultTppPasses.cpp
@@ -273,9 +273,8 @@ private:
     // mess up tensor producer-consumer chains used for analysis in the
     // following passes.
     pm.addPass(createPropagatePackUnPackPass());
-    pm.addPass(createCanonicalizerPass());
     pm.addPass(createConstantFoldPackPass());
-    pm.addPass(createCanonicalizerPass());
+    pm.addPass(createSimplifyAndCanonicalizePackPass());
 
     pm.addPass(createTileConsumerAndFuseProducersPass());
     pm.addPass(createCleanupPass());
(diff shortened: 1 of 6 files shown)


adam-smnk:
I'm probably missing sth. Why is this needed or beneficial?

chelini:
> I'm probably missing sth. Why is this needed or beneficial?

The pass or the split?

adam-smnk:
> The pass or the split?

The pass.

chelini:
> > The pass or the split?
> 
> The pass.

It runs `tensor::populateSimplifyTensorPack` upstream simplification and canonicalization.

adam-smnk:
Fair enough, I'd just call it `-simplify-pack`. I don't think canonicalization has to be explicitly mentioned.
Otherwise, seems fine to me.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -273,9 +273,8 @@ struct TppMappingPass : public TppMappingBase<TppMappingPass>,
     // mess up tensor producer-consumer chains used for analysis in the
     // following passes.
     pm.addPass(createPropagatePackUnPackPass());
-    pm.addPass(createCanonicalizerPass());
     pm.addPass(createConstantFoldPackPass());
-    pm.addPass(createCanonicalizerPass());
+    pm.addPass(createSimplifyAndCanonicalizePackPass());
```
rengolin, line 277: So, IIUC, this canonicalizes the packing and propagation of all operations (matmul, conv, etc) done so far?

chelini: Yes, it runs canonicalization for `tensor.pack` and `tensor.unpack` plus some simplification. Added more doc on the td file.

rengolin: Thanks! LGTM.

PR #518, opened 2023-04-26
Add more GEMM benchmarks

adam-smnk:
Adds `quick-perf` build target to quickly run performance benchmarks on local machine.
Adds more GEMM benchmarks of MxNxK shapes:
  * 256x1024x4096
  * 256x4096x1024
  * 128x1024x4096
  * 128x4096x1024
  * 128x1024x1024


rengolin (approved):
LGTM, Thanks!

Can you add the string join for the existing benchmark target as well, before merge?

PR #519, opened 2023-04-26
attempt to fix unary compute type

alheinecke:


rengolin (approved):
I can't test this yet, which is probably why I got it wrong in the first place. 😅

Feel free to merge when you have reached a solution that you can test.

PR #521, opened 2023-04-26
Restrict tensor pack folding to few cases

chelini:
Previously, the following IR:

```
%cst = arith.constant 0 : i64
%empty = tensor.empty() : tensor<3xi64>
%fill = linalg.fill ins(%cts : i64) outs(%empty : tensor<3xi64>) -> tensor<3xi64>
pack(%fill)

```
was folded into:

```
%cts = arith.constant dense<0> : i64
```

This is too aggressive because it may lead to a situation where the writable output of a given operation is replaced with a read-only constant. Similarly,

```
func.func @main(%arg0: tensor<3x3xf32>) -> tensor<1x1x3x3xf32>) {
   %cst = arith.constant 0 : i64
   %pack (%cst) into %arg0
   return %pack
}
```
is not allowed to fold as the user intention is unclear. Is `%arg0` the final destination?

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -271,6 +271,11 @@ def ConstantFoldPack : Pass<"constant-fold-pack", "ModuleOp"> {
     Reduce pack overhead by folding tensor.pack into constant tensors.
   }];
   let constructor = "mlir::tpp::createConstantFoldPackPass()";
+  let options = [
+    Option<"foldConstant", "fold-constant", "bool",
+           /*default=*/"false",
+           "Fold pack(cst) into a packed cst">
+  ];
 }
 
 def ElementWiseFusion : Pass<"element-wise-fusion", "func::FuncOp"> {
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -271,11 +271,6 @@ def ConstantFoldPack : Pass<"constant-fold-pack", "ModuleOp"> {
     Reduce pack overhead by folding tensor.pack into constant tensors.
   }];
   let constructor = "mlir::tpp::createConstantFoldPackPass()";
-  let options = [
-    Option<"foldConstant", "fold-constant", "bool",
-           /*default=*/"false",
-           "Fold pack(cst) into a packed cst">
-  ];
 }
 
 def ElementWiseFusion : Pass<"element-wise-fusion", "func::FuncOp"> {
--- a/lib/TPP/ConstantFoldPack.cpp
+++ b/lib/TPP/ConstantFoldPack.cpp
@@ -163,19 +163,15 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
     rewriter.replaceOpWithNewOp<arith::ConstantOp>(packOp, newDense);
   }
 
-  void foldPackIntoCstWithFill(RewriterBase &rewriter, tensor::PackOp packOp) {
+  void foldPackIntoFill(RewriterBase &rewriter, tensor::PackOp packOp) {
     OpBuilder::InsertionGuard guard(rewriter);
     Value sourcePack = packOp.getSource();
     auto fillOp = sourcePack.getDefiningOp<linalg::FillOp>();
     if (!fillOp)
       return;
-    auto cstOp = fillOp.getInputs()[0].getDefiningOp<arith::ConstantOp>();
-    if (!cstOp)
-      return;
-
-    rewriter.setInsertionPoint(fillOp);
-    rewriter.replaceOpWithNewOp<arith::ConstantOp>(
-        packOp, DenseElementsAttr::get(packOp.getDestType(), cstOp.getValue()));
+    rewriter.setInsertionPoint(packOp);
+    rewriter.replaceOpWithNewOp<linalg::FillOp>(packOp, fillOp.getInputs()[0],
+                                                packOp.getDest());
   }
 
   void runOnOperation() override {
@@ -183,9 +179,8 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
     IRRewriter rewriter(&getContext());
     module->walk(
         [&](tensor::PackOp packOp) { foldPackIntoSplatCst(rewriter, packOp); });
-    module->walk([&](tensor::PackOp packOp) {
-      foldPackIntoCstWithFill(rewriter, packOp);
-    });
+    module->walk(
+        [&](tensor::PackOp packOp) { foldPackIntoFill(rewriter, packOp); });
     module->walk(
         [&](tensor::PackOp packOp) { foldPackIntoCst(rewriter, packOp); });
   }
--- a/lib/TPP/ConstantFoldPack.cpp
+++ b/lib/TPP/ConstantFoldPack.cpp
@@ -177,12 +177,19 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
   void runOnOperation() override {
     auto module = getOperation();
     IRRewriter rewriter(&getContext());
-    module->walk(
-        [&](tensor::PackOp packOp) { foldPackIntoSplatCst(rewriter, packOp); });
+    // Enable folding constant and tensor.pack if the user requests it, as it
+    // changes the semantics of the program. `tensor.pack` is a writable
+    // operation while constant is not; thus, we don't know the user intent.
+    // Does the user want a writable tensor or a constant one?
+    if (foldConstant) {
+      module->walk([&](tensor::PackOp packOp) {
+        foldPackIntoSplatCst(rewriter, packOp);
+      });
+      module->walk(
+          [&](tensor::PackOp packOp) { foldPackIntoCst(rewriter, packOp); });
+    }
     module->walk(
         [&](tensor::PackOp packOp) { foldPackIntoFill(rewriter, packOp); });
-    module->walk(
-        [&](tensor::PackOp packOp) { foldPackIntoCst(rewriter, packOp); });
   }
 };
 
--- a/lib/TPP/ConstantFoldPack.cpp
+++ b/lib/TPP/ConstantFoldPack.cpp
@@ -48,11 +48,22 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
   }
 
   void foldPackIntoCst(RewriterBase &rewriter, tensor::PackOp packOp) {
+    // Bail out if the user uses pack as a writable operation
+    // (i.e., the destination is not a tensor.empty).
+    if (!packOp.getDest().getDefiningOp<tensor::EmptyOp>())
+      return;
     OpBuilder::InsertionGuard guard(rewriter);
     auto cstAndAttribute = getDenseAttributeAndConstant(packOp);
     if (failed(cstAndAttribute))
       return;
     auto [cstOp, oldDense] = *(cstAndAttribute);
+    // Happy path, splat constant.
+    if (oldDense.isSplat()) {
+      auto newDense = oldDense.reshape(packOp.getDestType());
+      rewriter.setInsertionPoint(cstOp);
+      rewriter.replaceOpWithNewOp<arith::ConstantOp>(packOp, newDense);
+      return;
+    }
     LLVM_DEBUG(llvm::dbgs()
                << "NUM ELEMENT: " << oldDense.getNumElements() << "\n");
     const int64_t bytes =
@@ -149,20 +160,6 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
     rewriter.replaceOpWithNewOp<arith::ConstantOp>(packOp, newDense);
   }
 
-  void foldPackIntoSplatCst(RewriterBase &rewriter, tensor::PackOp packOp) {
-    OpBuilder::InsertionGuard guard(rewriter);
-    auto cstAndAttribute = getDenseAttributeAndConstant(packOp);
-    if (failed(cstAndAttribute))
-      return;
-    auto [cstOp, oldDense] = *(cstAndAttribute);
-    if (!oldDense.isSplat())
-      return;
-    auto newDense = oldDense.reshape(packOp.getDestType());
-
-    rewriter.setInsertionPoint(cstOp);
-    rewriter.replaceOpWithNewOp<arith::ConstantOp>(packOp, newDense);
-  }
-
   void foldPackIntoFill(RewriterBase &rewriter, tensor::PackOp packOp) {
     OpBuilder::InsertionGuard guard(rewriter);
     Value sourcePack = packOp.getSource();
@@ -177,8 +174,6 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
   void runOnOperation() override {
     auto module = getOperation();
     IRRewriter rewriter(&getContext());
-    module->walk(
-        [&](tensor::PackOp packOp) { foldPackIntoSplatCst(rewriter, packOp); });
     module->walk(
         [&](tensor::PackOp packOp) { foldPackIntoFill(rewriter, packOp); });
     module->walk(
--- a/lib/TPP/ConstantFoldPack.cpp
+++ b/lib/TPP/ConstantFoldPack.cpp
@@ -177,19 +177,12 @@ struct ConstantFoldPack : public ConstantFoldPackBase<ConstantFoldPack> {
   void runOnOperation() override {
     auto module = getOperation();
     IRRewriter rewriter(&getContext());
-    // Enable folding constant and tensor.pack if the user requests it, as it
-    // changes the semantics of the program. `tensor.pack` is a writable
-    // operation while constant is not; thus, we don't know the user intent.
-    // Does the user want a writable tensor or a constant one?
-    if (foldConstant) {
-      module->walk([&](tensor::PackOp packOp) {
-        foldPackIntoSplatCst(rewriter, packOp);
-      });
-      module->walk(
-          [&](tensor::PackOp packOp) { foldPackIntoCst(rewriter, packOp); });
-    }
+    module->walk(
+        [&](tensor::PackOp packOp) { foldPackIntoSplatCst(rewriter, packOp); });
     module->walk(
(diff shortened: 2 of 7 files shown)


rengolin:
> This is too aggressive because it may lead to a situation where the writable output of a given operation is replaced with a read-only constant.

Ouch! We just fixed the benchmarks by doing the opposite transform. :)

rengolin (commented):
Technically, `pack(constant)->empty` isn't the same as a packed constant either. The original program might have intended (as we do in the benchmarks) to have a constant fill a writeable tensor so we can use it on outputs later.

The pattern that fits this pass would be `empty+fill+empty+pack` or `pack(fill(empty)) -> empty` to become just `empty+fill_packed`, because it remains writable, as expected.

Where were these examples from?

chelini:
> 

I see this pattern a bit everywhere also in MHA: https://github.com/plaidml/tpp-mlir/blob/85aceb35c5306a73960810951b9636eff74aa5c9/test/Models/multi-head-attention.mlir#L88

rengolin:
> I see this pattern a bit everywhere also in MHA:
> 
> https://github.com/plaidml/tpp-mlir/blob/85aceb35c5306a73960810951b9636eff74aa5c9/test/Models/multi-head-attention.mlir#L88

Right, this is clearly wrong and the PR here fixes it, that's good.

Where do you see the other pattern in the same test?

chelini:
> > I see this pattern a bit everywhere also in MHA:
> > https://github.com/plaidml/tpp-mlir/blob/85aceb35c5306a73960810951b9636eff74aa5c9/test/Models/multi-head-attention.mlir#L88
> 
> Right, this is clearly wrong and the PR here fixes it, that's good.
> 
> Where do you see the other pattern in the same test?

the pattern `tensor.pack(cst)` is for the input operands of your GEMM. For example, if you have a splat tensor you don't want to pack.

chelini:
> > the pattern `tensor.pack(cst)` is for the input operands of your GEMM. For example, if you have a splat tensor you don't want to pack.
> 
> If this is from hand crafted code, then I suggest we review the code and drop this transform, because it's equally dangerous.
> 
> We cannot change the constness of values with transforms.

Here, we are packing a constant, not a fill. This pattern works also on non-splatted constants, and proved beneficial when I was looking at Resnet. Why you believe this pattern changes the constness of a value?

rengolin:
> Here, we are packing a constant, not a fill. This pattern works also on non-splatted constants, and proved beneficial when I was looking at Resnet. Why you believe this pattern changes the constness of a value?

Technically, we're not packing the constant, we're packing the `empty`, which might not be treated as constant by the users, no matter what the original reason for it to be put there.

Since there's no information of the original intention in the current state of the IR, we can't assume.

### Use Case

In the benchmarks we fixed a bufferization problem by introducing a copy of a constant into an empty so that the tensor passed to the kernel is writable, and the compiler doesn't need to generate allocations+copy+dealloc inside the main loop.

Pseudo code of the original:
```mlir
%0 = arith.constant
loop {
  // writing to %0 inside the kernel will generate allocs before
  %temp = alloc
  %copy = copy ins(%0) outs(%temp)
  kernel(%temp) 
  dealloc(%temp)
}
```

So we do:
```mlir
%0 = arith.constant
%1 = tensor.empty
%2 = copy inst(%0) outs(%1)
loop {
  kernel(%2) // <- %2 is writable, no allocs necessary
}
```

Now say we pack the values:
```mlir
%0 = arith.constant
%1 = tensor.empty
%2 = copy inst(%0) outs(%1)
%3 = tensor.empty
%4 = tensor.pack ins(%2) outs(%3) // writable
loop {
  kernel(%4) // <- %4 is writable, no allocs necessary}
```

And we realise that a `pack` is essentially a `copy` with some reshape, so we elide the copy:
```mlir
%0 = arith.constant
%1 = tensor.empty
%2 = tensor.pack inst(%0) outs(%1)
loop {
  kernel(%2) // <- %2 is writable, no allocs necessary
}
```

And now you simplify the `pack`:
```mlir
%0 = arith.constant <packed_shape>
loop {
  // writing to %0 inside the kernel will generate allocs before
  %temp = alloc
  %copy = copy ins(%0) outs(%temp)
  kernel(%temp) 
  dealloc(%temp)
}
```
and you're back with the allocs again, because you're trying to write to a constant value.

chelini:
> > Here, we are packing a constant, not a fill. This pattern works also on non-splatted constants, and proved beneficial when I was looking at Resnet. Why you believe this pattern changes the constness of a value?
> 
> Technically, we're not packing the constant, we're packing the `empty`, which might not be treated as constant by the users, no matter what the original reason for it to be put there.

Did not get this, why are we packing the empty? What i have in mind is something like:

```
%cst = arith.constant
%empty = tensor.empty
%pack = tensor.pack %cst into %empty
```
tensor.empty for me here just carries the shape, %pack will return the packed constant.

> 
> Since there's no information of the original intention in the current state of the IR, we can't assume.
> 
> ### Use Case
> In the benchmarks we fixed a bufferization problem by introducing a copy of a constant into an empty so that the tensor passed to the kernel is writable, and the compiler doesn't need to generate allocations+copy+dealloc inside the main loop.
> 
> Pseudo code of the original:
> 
> ```mlir
> %0 = arith.constant
> loop {
>   // writing to %0 inside the kernel will generate allocs before
>   %temp = alloc
>   %copy = copy ins(%0) outs(%temp)
>   kernel(%temp) 
>   dealloc(%temp)
> }
> ```
> 
> So we do:
> 
> ```mlir
> %0 = arith.constant
> %1 = tensor.empty
> %2 = copy inst(%0) outs(%1)
> loop {
>   kernel(%2) // <- %2 is writable, no allocs necessary
> }
> ```
> 
> Now say we pack the values:
> 
> ```mlir
> %0 = arith.constant
> %1 = tensor.empty
> %2 = copy inst(%0) outs(%1)
> %3 = tensor.empty
> %4 = tensor.pack ins(%2) outs(%3) // writable
> loop {
>   kernel(%4) // <- %4 is writable, no allocs necessary}
> ```
> 
> And we realise that a `pack` is essentially a `copy` with some reshape, so we elide the copy:
> 
> ```mlir
> %0 = arith.constant
> %1 = tensor.empty
> %2 = tensor.pack inst(%0) outs(%1)
> loop {
>   kernel(%2) // <- %2 is writable, no allocs necessary
> }
> ```
> 
> And now you simplify the `pack`:
> 
> ```mlir
> %0 = arith.constant <packed_shape>
> loop {
>   // writing to %0 inside the kernel will generate allocs before
>   %temp = alloc
>   %copy = copy ins(%0) outs(%temp)
>   kernel(%temp) 
>   dealloc(%temp)
> }
> ```
> 
> and you're back with the allocs again, because you're trying to write to a constant value.

Understand the concern. Some thinking:
1. In the use case you have propagation will bubble-up the pack before the copy. We then pack the constant, and copy the packed constant into the copy 'empty' buffer, so we preserve the "write" semantics.
2. Indeed we need to pay attention of not folding pack and copy. Perhaps we can add an attribute to `tensor.pack` (i.e., no fold) to control if we want to fold it or not.

rengolin:
> Did not get this, why are we packing the empty? What i have in mind is something like:
> 
> ```
> %cst = arith.constant
> %empty = tensor.empty
> %pack = tensor.pack %cst into %empty
> ```
> 
> tensor.empty for me here just carries the shape, %pack will return the packed constant.

It does, but it's also a writable tensor, so ops that use the result are technically free to write to it.

The intended semantics is "here's a shape where you'll splat that constant, so it should be a constant splat".  But the IR operation doesn't carry the constant semantics, and it's now technically writable.

Bufferization will look at that and assume it's writable, so won't create temporaries if a later op writes to it. If after that you convert back to a constant, now the IR writes to a constant tensor, and that's undefined behaviour at best. LLVM will have a lot of fun with it, in the worst possible sense.

This is the problem or creating a dialect that doesn't have a strict semantics and doesn't strictly follow the SSA design principles. You end up in a situation where you can't prove safety. If a pass _assumes_ safety and does something technically illegal, it might get away with it on the "cases we have tests for". 

But as we integrate with other passes, we might get their illegal assumptions, and they might get ours, and that leads to code generation bugs that are hard to debug.


> Understand the concern. Some thinking:
> 
> 1. In the use case you have propagation will bubble-up the pack before the copy. We then pack the constant, and copy the packed constant into the copy 'empty' buffer, so we preserve the "write" semantics.
> 2. Indeed we need to pay attention of not folding pack and copy. Perhaps we can add an attribute to `tensor.pack` (i.e., no fold) to control if we want to fold it or not.

It's generally not safe to fix a code generation bug/UB with an optimization strategy. We may try as much as we want, but corner cases always go through and we go back debugging code-gen bugs for days.

I think the safer bet is to forbid anything that can technically be unsafe and when we see a problem generated by this strategy, we try to fix the root cause of it, not the symptom.

chelini:
> > Did not get this, why are we packing the empty? What i have in mind is something like:
> > ```
> > %cst = arith.constant
> > %empty = tensor.empty
> > %pack = tensor.pack %cst into %empty
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > tensor.empty for me here just carries the shape, %pack will return the packed constant.
> 
> It does, but it's also a writable tensor, so ops that use the result are technically free to write to it.
> 
> The intended semantics is "here's a shape where you'll splat that constant, so it should be a constant splat". But the IR operation doesn't carry the constant semantics, and it's now technically writable.
> 
> Bufferization will look at that and assume it's writable, so won't create temporaries if a later op writes to it. If after that you convert back to a constant, now the IR writes to a constant tensor, and that's undefined behaviour at best. LLVM will have a lot of fun with it, in the worst possible sense.
> 
> This is the problem or creating a dialect that doesn't have a strict semantics and doesn't strictly follow the SSA design principles. You end up in a situation where you can't prove safety. If a pass _assumes_ safety and does something technically illegal, it might get away with it on the "cases we have tests for".
> 
> But as we integrate with other passes, we might get their illegal assumptions, and they might get ours, and that leads to code generation bugs that are hard to debug.
> 
> > Understand the concern. Some thinking:
> > 
> > 1. In the use case you have propagation will bubble-up the pack before the copy. We then pack the constant, and copy the packed constant into the copy 'empty' buffer, so we preserve the "write" semantics.
> > 2. Indeed we need to pay attention of not folding pack and copy. Perhaps we can add an attribute to `tensor.pack` (i.e., no fold) to control if we want to fold it or not.
> 
> It's generally not safe to fix a code generation bug/UB with an optimization strategy. We may try as much as we want, but corner cases always go through and we go back debugging code-gen bugs for days.
> 
> I think the safer bet is to forbid anything that can technically be unsafe and when we see a problem generated by this strategy, we try to fix the root cause of it, not the symptom.

Ok, I need to think about this `pack(cst)`. I don't have a good solution; perhaps adding an attribute (or property) as "non-writable"? We want for sure packing constants. If we are too worried about this `pack(cst)`, we can disable it for now.

rengolin:
> Ok, I need to think about this `pack(cst)`. I don't have a good solution; perhaps adding an attribute (or property) as "non-writable"? We want for sure packing constants. If we are too worried about this `pack(cst)`, we can disable it for now. 

Let's be conservative. Disable anything we can't prove correct. 

With regards to constant attributes, this needs to be an upstream discussion. But I don't think we have enough data to back up any such discussion, so let's just wait until we have a real problem.

chelini:
> > Ok, I need to think about this `pack(cst)`. I don't have a good solution; perhaps adding an attribute (or property) as "non-writable"? We want for sure packing constants. If we are too worried about this `pack(cst)`, we can disable it for now.
> 
> Let's be conservative. Disable anything we can't prove correct.
> 
> With regards to constant attributes, this needs to be an upstream discussion. But I don't think we have enough data to back up any such discussion, so let's just wait until we have a real problem.

Disabled `pack(cst)`. It is still available via a user-specified option.

rengolin:
> Disabled `pack(cst)`. It is still available via a user-specified option.

No, this is an incorrect transform, we don't want it even with a flag. Please disable it completely and remove from the code.

Worst still having a flag, Some tests will have it (when you add) and others won't (default) and behaviour will be different, sometimes broken sometimes not.

chelini:
> > Disabled `pack(cst)`. It is still available via a user-specified option.
> 
> No, this is an incorrect transform, we don't want it even with a flag. Please disable it completely and remove from the code.
> 
> Worst still having a flag, Some tests will have it (when you add) and others won't (default) and behaviour will be different, sometimes broken sometimes not.

I don't agree. Remove this pass make

> 

I can't entirely agree. I may want as a user to constant fold `pack(cst)` because I know what I am doing. Disabling this pass would block progress in propagation in my opinion, how can we experiment with packing if we cannot constant fold weights and bias?. However, I agree that should not be able by default as the user may want a writable tensor out of the packing, it is also true the opposite, I may want to constant fold the packing.

rengolin:
> I can't entirely agree. I may want as a user to constant fold `pack(cst)` because I know what I am doing. 

Compiler passes run on any code not just the one you have in tests. If you want to run an illegal pass because "you know what you're doing" in this particular example, you'll open up to do the same on things "you don't know what you're doing", and that leads to crashes or worse, bad code generation.

> Disabling this pass would block progress in propagation in my opinion, how can we experiment with packing if we cannot constant fold weights and bias?.

As I said, we need to fix the problem at the source, not the symptom. Where does this code come from? Can we canonicalize it earlier? Can we lower it to something else?

> However, I agree that should not be able by default as the user may want a writable tensor out of the packing (it is still questionable to me this statement), it is also true the opposite, I may want to constant fold the packing.

Using flags to enable illegal transformations is worse. You're adding a requirement to your local experimentation that cannot exist in the main repo and may lead you to the wrong overall solution. If you can't use a particular transform long term, you have to find another way.

In this PR, you've added it to some tests and not others. You'll run with it locally and not add this to the default pipeline. Other developers won't use the flag and won't see the same transforms you do. This strategy does not work alone, and is counter productive in a group.

For other non-illegal transformations, you can experiment locally, but once you push to the main repository it has to work for everyone else. It's either in the default pass pipeline or it doesn't exist. But illegal transforms should never exist in the main repository, especially not under flags or special cases.

chelini:
> > I can't entirely agree. I may want as a user to constant fold `pack(cst)` because I know what I am doing.
> 
> Compiler passes run on any code not just the one you have in tests. If you want to run an illegal pass because "you know what you're doing" in this particular example, you'll open up to do the same on things "you don't know what you're doing", and that leads to crashes or worse, bad code generation.
> 
> > Disabling this pass would block progress in propagation in my opinion, how can we experiment with packing if we cannot constant fold weights and bias?.
> 
> As I said, we need to fix the problem at the source, not the symptom. Where does this code come from? Can we canonicalize it earlier? Can we lower it to something else?

Propagation exposes possible folding for constants using packing. So we could run right after it, in an atomic way. Would this help? 

> 
> > However, I agree that should not be able by default as the user may want a writable tensor out of the packing (it is still questionable to me this statement), it is also true the opposite, I may want to constant fold the packing.
> 
> Using flags to enable illegal transformations is worse. You're adding a requirement to your local experimentation that cannot exist in the main repo and may lead you to the wrong overall solution. If you can't use a particular transform long term, you have to find another way.
> 
> In this PR, you've added it to some tests and not others. You'll run with it locally and not add this to the default pipeline. Other developers won't use the flag and won't see the same transforms you do. This strategy does not work alone, and is counter productive in a group.
> 
> For other non-illegal transformations, you can experiment locally, but once you push to the main repository it has to work for everyone else. It's either in the default pass pipeline or it doesn't exist. But illegal transforms should never exist in the main repository, especially not under flags or special cases.

chelini:
> > > I can't entirely agree. I may want as a user to constant fold `pack(cst)` because I know what I am doing.
> > 
> > 
> > Compiler passes run on any code not just the one you have in tests. If you want to run an illegal pass because "you know what you're doing" in this particular example, you'll open up to do the same on things "you don't know what you're doing", and that leads to crashes or worse, bad code generation.
> > > Disabling this pass would block progress in propagation in my opinion, how can we experiment with packing if we cannot constant fold weights and bias?.
> > 
> > 
> > As I said, we need to fix the problem at the source, not the symptom. Where does this code come from? Can we canonicalize it earlier? Can we lower it to something else?
> 
> Propagation exposes possible folding for constants using packing. So we could run right after it, in an atomic way. Would this help?
> 
> > > However, I agree that should not be able by default as the user may want a writable tensor out of the packing (it is still questionable to me this statement), it is also true the opposite, I may want to constant fold the packing.
> > 
> > 
> > Using flags to enable illegal transformations is worse. You're adding a requirement to your local experimentation that cannot exist in the main repo and may lead you to the wrong overall solution. If you can't use a particular transform long term, you have to find another way.

I still struggle to understand why is illegal: you start from a constant and you end up with a packed constant. 

> > In this PR, you've added it to some tests and not others. You'll run with it locally and not add this to the default pipeline. Other developers won't use the flag and won't see the same transforms you do. This strategy does not work alone, and is counter productive in a group.
> > For other non-illegal transformations, you can experiment locally, but once you push to the main repository it has to work for everyone else. It's either in the default pass pipeline or it doesn't exist. But illegal transforms should never exist in the main repository, especially not under flags or special cases.

rengolin:
> I still struggle to understand why is illegal: you start from a constant and you end up with a packed constant.

As I explained in my examples, the constant gets stored in a non-constant `empty` tensor, which can be used for writing.

Your argument is that the empty tensor is only used for shape, that the value is still constant, but the operation itself does not carry that semantics.

If this was SSA syntax, it would be easier:
```mlir
  // %pck is definitely a constant
  %cst = arith.constant dense<1.0> : <128x128xf32>
  %pck = tensor.pack [...] %cst : <4x32x8x16xf32>

  // %0 is not a constant, %pck cannot be proven constant
  %cst = arith.constant dense<1.0> : <128x128xf32>
  %0 = tensor empty : <4x32x8x16xf32>
  %pck = tensor.pack [...] %cst into %0 : <128x128xf32> -> <4x32x8x16xf32>
```

rengolin:
> Propagation exposes possible folding for constants using packing. So we could run right after it, in an atomic way. Would this help?

I'm not sure. Let's discuss this offline.

chelini:
> > Propagation exposes possible folding for constants using packing. So we could run right after it, in an atomic way. Would this help?
> 
> I'm not sure. Let's discuss this offline.

Sounds good. I pinged you on chat.

rengolin (approved):

PR #522, opened 2023-04-26
Update GEMM benchmarks blocking factors

adam-smnk:
Blocks existing GEMM performance benchmarks with factors from PARLOOPER.
Also, fixes DNN sizes for GEMM 128x1024x4096.


rengolin:
Did the performance improve significantly?

PR #523, opened 2023-04-26
[XSMM] Align priting for invoke and dispatch (NFC)

chelini:
MLIR uses `snake_case` for IR. We followed this convention when re-working dispatch 
operations.


chelini:
> There are still some left-over `dataType`s in there, why change these and not those?

The rest comes from the quaternary op. I am not updating it as I have some follow-up that remove it.

rengolin (approved):

PR #524, opened 2023-04-26
Updated LIBXSMM/DNN and revised TPP-benchmark plot

hfp:


scripts/buildkite/tpp-benchmark.yml
```diff
@@ -19,6 +19,8 @@ steps:
               $${BUILDKITE_BUILD_CHECKOUT_PATH}/scripts/buildkite/build_tpp.sh'"
     env:
       LOGRPTSUM: "mlir"
-      #LOGRPTFMT: "svg"
+      LOGRPTFMT: "svg"
+      LOGRPTFMT: "svg"
```
rengolin, line 23: Twice?

hfp: Ohh, I did not save in the editor... let me update.

rengolin (approved):
Should be fine to merge whenever you fix the report. Thanks!

hfp:
Buildkite applies CSS to an img component of the log with maximum height hard-coded to 500px.

hfp:
There is no way to show results/report with more than 500px in height (without feature request to Buildkite). I will add multi-format support for the reporting tool like PNG/SVG and PDF, upload PDF with higher height, and only show PNG/SVG limited to 500px in height.

hfp:
Updated PR will have URL embedded into log navigating to PDF-report with best details/readability.

hfp:
> Updated PR will have URL embedded into log navigating to PDF-report with best details/readability.

This is reflected by the updated PR. However, we will also need an updated recipe to upload artifacts. Uploading artifacts cannot be done as part of a running job because our Slurm jobs run on compute nodes without agent.

PR #526, opened 2023-04-27
Fully connected benchmarks

adam-smnk:
Adds fully connected (FC) benchmarks - same shapes and tile sizes as for GEMM kernels.


adam-smnk:
Let's see what numbers we'll get, kernels are written such that they bufferize in-place i.e., no extra allocations, everything goes into output argument.

chelini:
> Let's see what numbers we'll get, kernels are written such that they bufferize in-place i.e., no extra allocations, everything goes into output argument.

I assume we do tile + fuse, right?

adam-smnk:
> I assume we do tile + fuse, right?

Yes, we end up with a single parallel loop with all invokes inside.

rengolin:
Some odd results for `fc_1024x352x512_fp32_dnn`, as odd as for `matmul_1024x352x512_fp32_dnn`, so probably some copied bug?

rengolin (approved):

benchmarks/config/fc/1024x352x512.json
```diff
@@ -0,0 +1,87 @@
...
+      "benchmark": "xsmm_dnn_mlp",
+      "environment": { "OMP_NUM_THREADS": "1" },
+      "flags": [ "100", "1024", "3", "F", "64", "64", "32", "0", "512", "352" ],
+      "extensions": [ "(avx2|asimd)" ]
+    },
+    "fc_fp32_omp_16_dnn": {
+      "type": "XSMM-DNN",
+      "benchmark": "xsmm_dnn_mlp",
+      "environment": { "OMP_NUM_THREADS": "16", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
+      "flags": [ "100", "1024", "3", "F", "64", "64", "32", "0", "512", "352" ],
```
adam-smnk, line 15: TODO: fix packing for this dim

PR #528, opened 2023-04-28
Default pipeline - remove BufferDeallocationPass

adam-smnk:
Removes 'BufferDeallocationPass' from the default pipeline. Currently, the pass creates deallocation in invalid places as it does not process correctly memory casts as aliases, which results in invalid buffer lifetime analysis.

Also, lack of automatic memory deallocation promotes safer memory management where users have to explicitly deallocate their buffers.

### DIFF
--- a/test/Passes/DefaultPipeline/mixed.mlir
+++ b/test/Passes/DefaultPipeline/mixed.mlir
@@ -46,31 +46,6 @@ module @predict_function  {
 
 // -----
 
-// CHECK: func.func @buffer_dealloc(
-// CHECK-SAME:  %[[ARG0:.+]]: memref<512x128xf32>,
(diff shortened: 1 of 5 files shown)


chelini (approved):

test/Passes/DefaultPipeline/mixed.mlir
```diff
@@ -116,6 +91,8 @@ func.func @heap_to_stack(%A: memref<4x8xf32>,
   memref.copy %0, %C : memref<4x4xf32> to memref<4x4xf32>
 
   // CHECK-NOT: memref.dealloc
```
chelini, line 93: I think we should remove this check?

adam-smnk: This one tests heap to stack conversion so it replaces alloc+dealloc with alloca.

PR #529, opened 2023-04-28
Bump LLVM

chelini:
Why: Incorporated some fix on the propagation pass:

Changes:
- pre-load dialect is retired. Our passes need now to load the dialect they use.

- affine has its own namespace

- drop useless return

- Attribute -> TypedAttr implicit conversion does not happen anymore

- Use callback in the transform dialect instead of free function, looks cleaner.

- There is something off with transform.FuseOp. It crashed with a SEGV if we tile and fuse all the relus together.  Need more investigation.

- Switch TPP and XSMM to use properties.


rengolin:
Rebase on the current `main` to get the GCC tests and make sure we're still passing. Otherwise, LGTM, thanks!

rengolin (approved):

PR #540, opened 2023-05-11
Move benchmarks to `mlp-gen`

### LINKED ISSUE #533: Use `mlp-gen` for benchmarks
We're benchmarking a lot of matmul/fully-connected kernels, but we generated the MLIR files with some scripts / by hand, when we already have an MLP generator.

We should replace all (MLP-related) MLIR files in `benchmarks/mlir` with calls to `mlp-gen` with the appropriate flags, piping the result to `tpp-run`.

@adam-smnk

adam-smnk:
Replaces matmul and fc generated benchmark files with runtime IR generator using `mlp-gen`.

* Adds a new benchmark type 'IR-GEN' to allow running MLIR IR created by any external generator.
* Extends benchmarks harness to accept STDIN input.
* Extends `mlp-gen` with matmul and fully connected kernel generators.
* Aligns `mlp-gen` packing with `pack-matmul`.
* Improves  `mlp-gen` bias and relu IR generation to avoid temporary buffers.
* Adds option to  `mlp-gen`  to apply VNNI packing at the time of IR generation.
* Adds metadata generation to `mlp-gen` to attach to kernels required 'RUN:' strings and total flops count.

Fixes #533


rengolin:
We're not running the long benchmark jobs on PR from forks to avoid overload, but we do build on branches in the main repo, so when you change the performance benchmarks you should use a direct branch, not a fork branch, to make sure they run every update.

adam-smnk:
> We're not running the long benchmark jobs on PR from forks to avoid overload, but we do build on branches in the main repo, so when you change the performance benchmarks you should use a direct branch, not a fork branch, to make sure they run every update.

Ah right, good point. I'll move it and open new PR.

Issue #389, opened 2023-03-14
Remove VNNI dialect


rengolin:
This was meant as a stop gap, not a full blown dialect, and we need to keep the conversion between linalg and TPP simple, and this distinction is getting out of hand (#385).

This is not trivial work, that's why we created the dialect in the first place, but it's about time we replace it with some proper linalg infrastructure.

@KavithaTipturMadhu @chelini @alheinecke

rengolin:
As seen in #45, the VNNI dialect uses variadic results and assembly format, which is what crashed GCC.

Once we kill VNNI, `perf.bench` will be the only other op we use that, which will be fixed in #406.

Issue #392, opened 2023-03-14
Deduplicate `lib/TPP/RewriteToBatchReduceGemm.cpp`


rengolin:
Currently, the batch-reduce GEMM rewrite has an almost complete duplicate of code for standard and VNNI packing, which isn't right. This was created in a way to expose the pattern for an upstream proposal that didn't go through, and now we need to cleanup.

The previous code has a bad pattern of having `isVNNI` in every call, and branching it for every check (num of inputs, outputs, iterators, dimensions, gemm loops, etc). So, going back to what we had before isn't good either.

For now, we only have 2 packing types, but once we support 8-bit data types, there will be a third. If new extensions come along that require their own packing, we'll need a simple way to replicate it.

My proposal is to create a matcher class, that commons up all properties for each packing type. The base class would have private fields that record the expected shapes (number of operands, dims, rank, etc) and methods for checking ops against each of those properties. Derived classes just need a constructor that initialises the fields with the right attributes.

Each existing check on those properties, not only in the BRGEMM rewrite but everywhere, become simply a matter of passing the op to the checker. Example:

```c++
// Unpacked ops
class PackingProperties {
  size_t numLoops;
  ...
public:
  bool checkNumLoops(LinalgOp &op) {
    return (op.numLoops() == numLoops;
  }
  ...
  bool verify(LinalgOp &op) {
    return checkNumLoops(op) && ...;
  }
};
// Basic packed ops
...
// VNNI packed ops
class VNNIProperties : public PackingProperties {
  VNNIProperties() : numLoops(4), ... {}
};
...
FailureOr someRewrite(LinalgOp op, PackingProperties config) {
  if (!config.verify(op))
    return failure();
  ...
}
...
// Standard rewrites pass standard configs
PackingRewrite::runOnOperation(Operation *op) {
  someRewrite(op, StandardPacking());
}
...
// VNNI rewrites pass VNNI configs
VNNIPackingRewrite::runOnOperation(Operation *op) {
  someRewrite(op, VNNIPacking());
}
...
// We can even make it parametric
GenericRewrite::runOnOperation(Operation *op) {
  auto type = op->getReturnType().getElementType();
  someRewrite(op, type.isBF16() ? VNNIPacking() : StandardPacking());
}
```

This can be used for other decisions, not just packing, but anything we want to match. We could have one for packing, one for TPP semantics (that checks affine maps, iterators and translates to attributes and flags), etc.

@chelini

rengolin:
[...]

PR #546, opened 2023-05-16
Retire VNNI dialect

chelini:
Retire VNNI dialect and VNNI TPP operations.

History: We introduced the VNNI dialect before TPP on tensor. Tensor semantics was needed as packing happens at this abstraction. Now that we have TPP on tensor, we no longer need the VNNI dialect.

Annoyance: We use an explicit VNNI layout (i.e., the layout is reflected on the tensor/memef shape); this complicates the verification for the ops and the lowering to XSMM and Loops. Pack VNNI will be sinked to memref in later commits.

Fixes: #389


lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -270,195 +228,14 @@ struct ConvertTppBrgemmOp : public OpRewritePattern<tpp::BrgemmOp> {
...
-    invokeOperands.push_back(dispatched);
-    invokeOperands.append(brgemmOp->getOperands().begin(),
-                          brgemmOp->getOperands().end());
-    invokeOperands.push_back(batchDim);
-    rewriter.replaceOpWithNewOp<xsmm::QuarternaryOp>(brgemmOp, dtype, attr,
-                                                     invokeOperands);
-    return success();
+    // TODO: Handle the conversion. Was not tested and questionable.
+    // Will follow-up on this.
+    return failure();
```
rengolin, line 238: It'd be good to understand that the point was here before dropping it. Being untested, it may just mean we forgot to add tests. @KavithaTipturMadhu

lib/TPP/Dialect/Tpp/TppOps.cpp
```diff
@@ -268,34 +270,68 @@ void AddOp::getEffects(
...
+    // VNNI layout: [K/VNNI][J][VNNI]
+    if (shape[0] * shape[shape.size() - 1] != dimI || shape[1] != dimJ)
+      return operation->emitOpError("operand 1 fails to verify expected shape");
+    return success();
+  }
+  if (shape.size() != 2) {
+    return operation->emitOpError()
+           << "operand 1 expects rank 2, but got: " << shape.size() << "\n";
+  }
+  if (shape[0] != dimI || shape[1] != dimJ)
```
rengolin, line 298: This check could be moved up to avoid the extra indentation on the VNNI case.

lib/TPP/Dialect/Tpp/TppOps.cpp
```diff
@@ -268,34 +270,68 @@ void AddOp::getEffects(
...
+static LogicalResult
+validateVnniGemmOperand(OpTy operation, ArrayRef<int64_t> shape,
+                        Type elementType, int dimI, int dimJ) {
+  if (shape.size() == 3) {
+    if (!elementType.isBF16()) {
+      return operation->emitOpError() << "operand 1 invalid element type for "
+                                         "VNNI layout expect bf16, but got: "
+                                      << elementType << "\n";
+    }
+    if (shape[shape.size() - 1] !=
```
rengolin, line 283: No point in using `size - 1` given that the `shape.size()` is guaranteed to be 3.

lib/TPP/Dialect/Tpp/TppOps.cpp
```diff
@@ -321,30 +357,54 @@ void GemmOp::getEffects(
 // BrgemmOp
 //===----------------------------------------------------------------------===//
 
-static bool verifyBRGemmShape(ShapedType shapedA, ShapedType shapedB,
-                              ShapedType shapedC) {
-  return !(shapedB.getRank() != 3 || shapedC.getRank() != 2 ||
-           shapedA.getRank() != 3);
-}
+template <typename OpTy>
+static LogicalResult verifyBgemmOperands(OpTy operation) {
```
rengolin, line 361: Can you common `gemm` and `brgemm` verification, at least for a subset of the checks? They look very similar.

lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -28,7 +28,7 @@ using namespace mlir;
 
 #define DEBUG_TYPE "mlir-rewrite-to-brgemm"
 
-// Look for [p ... p] brgemm [r p p r r]
+// Look for [p ... p] brgemm [r r p p r]
```
rengolin, line 31: both `checkVNNIStructure` and `checkStructure` are very similar and could be de-duplicated.

lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -108,7 +108,7 @@ static LogicalResult checkVNNIAccessPatterns(linalg::LinalgOp linalgOp) {
   SmallVector<AffineMap> expectedMaps;
   bindDims(linalgOp.getContext(), r1, r2, p4, p5, r3);
   auto blockingFactor =
-      vnni::utils::getVNNIBlockingFactor(linalgOp->getOperands()[0].getType());
+      vnni::utils::getVnniBlockingFactor(linalgOp->getOperands()[0].getType());
```
rengolin, line 111: Both `checkVNNIAccessPatterns` and `checkAccessPatterns` are very similar and could be de-duplicated.

lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -288,8 +288,11 @@ mlir::linalgx::rewriteToBRGemmOp(RewriterBase &rewriter,
             : builder.create<linalg::BatchReduceMatmulOp>(
                   loc, ValueRange{slicedOperands[0], slicedOperands[1]},
                   slicedOperands[2]);
-    tensorResults = insertSlicesBack(builder, loc, linalgOp, slicedOperands,
-                                     brgemm->getResults());
+    tensorResults =
```
rengolin, line 291: both `rewriteToBRGemmOp` and `rewriteToBrGemmVnniOp` need to be de-duplicated.

lib/TPP/RewriteToBatchReduceGemm.cpp
```diff
@@ -434,7 +438,6 @@ struct RewriteToBatchReduceGemm
     patterns
         .add<RewriteToBatchReduceGemmImpl, RewriteToBatchReduceGemmVnniImpl>(
             patterns.getContext());
```
rengolin, line 440: Please de-duplicate `RewriteToBatchReduceGemmVnniImpl` too.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -444,7 +444,7 @@ mlir::linalgx::packVNNIMatmulOp(RewriterBase &rewriter,
 
   Location loc = matmulOp.getLoc();
   auto blockingFactor =
-      vnni::utils::getVNNIBlockingFactor(matmulOp.getInputs()[1].getType());
+      vnni::utils::getVnniBlockingFactor(matmulOp.getInputs()[1].getType());
```
rengolin, line 447: We also need to de-duplicate `packMatmulOpImpl` and `packVNNIMatmulOp`

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -515,7 +515,7 @@ mlir::linalgx::packVNNIMatmulOp(RewriterBase &rewriter,
   return replacementOp;
 }
 
-FailureOr<vnni::BRGemmOp>
+FailureOr<tpp::BrgemmOp>
 mlir::linalgx::packVNNIBRGemmOp(RewriterBase &rewriter,
```
rengolin, line 519: We don't seem to have a `packBRGemmOp`, is that expected?

lib/TPP/VNNIUtils.cpp
```diff
@@ -28,10 +31,17 @@ Optional<int64_t> getVNNIBlockingFactor(Type type) {
 }
 
 bool isBF16Type(Type type) {
-  if (!isa<ShapedType>(type))
+  auto elementType = getElementTypeOrSelf(type);
+  return elementType.isBF16();
+}
+
+bool isInVnniLayout(MemRefType memref) {
```
rengolin, line 38: Have you checked that any VNNI-related static local function to other files you create/use could be moved here?

test/BF16/Integration/mlp-all.mlir
```diff
@@ -1,10 +1,10 @@
-// RUN: tpp-opt %s -pack-matmul="block-factors=32,32,32" -pack-vnni  -generalize-tensor-pack-unpack -bufferize -rewrite-to-brgemm -convert-vnni-to-tpp -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-check-to-loops -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -finalize-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts |\
+// RUN: tpp-opt %s -pack-matmul="block-factors=32,32,32" -pack-vnni  -generalize-tensor-pack-unpack -bufferize -rewrite-to-brgemm -canonicalize -convert-tpp-to-xsmm -convert-xsmm-to-func -canonicalize -convert-check-to-loops -convert-linalg-to-loops -convert-vector-to-scf -convert-scf-to-cf -expand-strided-metadata -lower-affine -convert-arith-to-llvm -convert-vector-to-llvm -finalize-memref-to-llvm -arith-expand -convert-math-to-llvm -convert-func-to-llvm -reconcile-unrealized-casts |\
```
rengolin, line 1: This needs to change to `tpp-opt --default-tpp-passes` instead and run with `tpp-run`.

test/Conversion/TppToXsmm/tpp-to-xsmm-quarternary.mlir
```diff
@@ -1,4 +1,5 @@
 //RUN: tpp-opt %s --combine-tpp --convert-tpp-to-xsmm | FileCheck %s
+// XFAIL:*
```
rengolin, line 2: Nope, we can't XFAIL tests without a proper review. @KavithaTipturMadhu @adam-smnk

rengolin (commented):
Long needed, thanks! I have a few comments, including the de-duplication of a lot of VNNI specific code that was copy & pasted over the last few months and needs to go, too (see #392).

test/Conversion/TppToXsmm/tpp-to-xsmm-quarternary.mlir
chelini: Here to be honest I don't understand the test and now combine-tpp runs at tensor, so now sure what to do with this..

adam-smnk: The test might need to be rewritten but we still need to check that now `tpp.brgemm` with VNNI input gets lowered to `xsmm.quarternary`.

rengolin:
> memref<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1,VNNI)>

This looks easy at first, but can get complicated if composing multiple packs on each other, ie. the complexity to check for shapes will move to the complexity to check for affine maps at the last step (xsmm conversion).

The solution we discussed (and agreed) on the meeting in March was to use TPP operations to encode semantics:
```mlir
  %A = ... : tensor<128x256xbf16>
  %Ap = tensor.pack %A ... : tensor<4x8x32x32xbf16>
  %B = ... : tensor<256x512xbf16>
  %Bp = tensor.pack %B ... : tensor<8x16x32x32xbf16>
  %Bv = tpp.vnni_pack %Bp : tensor<8x16x16x32x2xbf16>
  %Cp = ... : tensor<4x16x32x32xbf16>
  %3 = tpp.gemm %Ap, %Bv -> %Cp : tensor<4x16x32x32xbf16>
```

Since `tpp.gemm`'s second operand's producer is `tpp.vnni_pack` (and the shape is appropriate), this is VNNI packed.

OTOH, here:
```mlir
  ...
  %3 = tpp.gemm %Ap, %Bp -> %Cp : tensor<4x16x32x32xbf16>
```

The operand's producer is `tensor.pack` (and the shape is appropriate), so this is NOT VNNI packed.

An `extract_memref` of a `cast` of a `tpp.vnni_pack` is _still_ VNNI packed, so checks on producers need to follow through.

test/Conversion/TppToXsmm/tpp-to-xsmm-quarternary.mlir
chelini: Ok, I actually reenabled it few commits later. If you don't mind I will do this after we fix also the quaternary stuff, as I am actually changing it.

adam-smnk:
@rengolin One potential problem I see with introducing yet another custom pack operation like `tpp.vnni_pack` is that we might need to duplicate or recreate the propagation logic of the upstream `tensor.pack` which should be able to also handle VNNI packing in one go. 
Alternatively, a sth like `tpp.cast to vnni` from already prepacked tensor might be simpler or maybe other metadata-based approach. We should start by looking at actual IR examples before we pick the final solution.

test/Conversion/TppToXsmm/tpp-to-xsmm-quarternary.mlir
chelini: It is unclear what this tests what to do. If it checks only the lowering from tpp to xsmm for the quaternary, which is already covered or the combined pass. The conversion should be tested in isolation and it is exercised in `tpp-to-xsmm-quarternary.mlir`. To me we can delete it.

lib/TPP/Dialect/Tpp/TppOps.cpp
chelini: Was to avoid magic numbers.

lib/TPP/Dialect/Tpp/TppOps.cpp
chelini: Sorry, did not get. Where you want to move it? line 289?

rengolin:
> One potential problem I see with introducing yet another custom pack operation like `tpp.vnni_pack` is that we might need to duplicate or recreate the propagation logic of the upstream `tensor.pack` which should be able to also handle VNNI packing in one go. Alternatively, a sth like `tpp.cast to vnni` from already prepacked tensor might be simpler or maybe other metadata-based approach. We should start by looking at actual IR examples before we pick the final solution.

As discussed in march, the `tpp.vnni_pack` operation is both "operation" and "annotation":
```mlir
  // Runtime packing from inputs
  %B = ... : tensor<256x512xbf16>
  // This is an operation
  %Bp = tensor.pack %B ... : tensor<8x16x32x32xbf16>
  // This is an operation (in-shape != out-shape, out-shape is VNNI)
  %Bv = tpp.vnni_pack %Bp : tensor<8x16x16x32x2xbf16>

  // Constant packed at compile time
  %B = ... : tensor<8x16x16x32x2xbf16>
  // This is an annotation (in-shape == out-shape, both VNNI)
  %Bv = tpp.vnni_pack %B : tensor<8x16x16x32x2xbf16>
```

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: The problem here is that `packVNNIMatmulOp` assume already a packed matmul. Currently VNNI requires first a "traditional" pack.

chelini:
> > One potential problem I see with introducing yet another custom pack operation like `tpp.vnni_pack` is that we might need to duplicate or recreate the propagation logic of the upstream `tensor.pack` which should be able to also handle VNNI packing in one go. Alternatively, a sth like `tpp.cast to vnni` from already prepacked tensor might be simpler or maybe other metadata-based approach. We should start by looking at actual IR examples before we pick the final solution.
> 
> As discussed in march, the `tpp.vnni_pack` operation is both "operation" and "annotation":
> 
> ```mlir
>   // Runtime packing from inputs
>   %B = ... : tensor<256x512xbf16>
>   // This is an operation
>   %Bp = tensor.pack %B ... : tensor<8x16x32x32xbf16>
>   // This is an operation (in-shape != out-shape, out-shape is VNNI)
>   %Bv = tpp.vnni_pack %Bp : tensor<8x16x16x32x2xbf16>
> 
>   // Constant packed at compile time
>   %B = ... : tensor<8x16x16x32x2xbf16>
>   // This is an annotation (in-shape == out-shape, both VNNI)
>   %Bv = tpp.vnni_pack %B : tensor<8x16x16x32x2xbf16>
> ```

Right but if it is an operation you need to plug it into the upstream propagation.

chelini:
> > > One potential problem I see with introducing yet another custom pack operation like `tpp.vnni_pack` is that we might need to duplicate or recreate the propagation logic of the upstream `tensor.pack` which should be able to also handle VNNI packing in one go. Alternatively, a sth like `tpp.cast to vnni` from already prepacked tensor might be simpler or maybe other metadata-based approach. We should start by looking at actual IR examples before we pick the final solution.
> > 
> > 
> > As discussed in march, the `tpp.vnni_pack` operation is both "operation" and "annotation":
> > ```mlir
> >   // Runtime packing from inputs
> >   %B = ... : tensor<256x512xbf16>
> >   // This is an operation
> >   %Bp = tensor.pack %B ... : tensor<8x16x32x32xbf16>
> >   // This is an operation (in-shape != out-shape, out-shape is VNNI)
> >   %Bv = tpp.vnni_pack %Bp : tensor<8x16x16x32x2xbf16>
> > 
> >   // Constant packed at compile time
> >   %B = ... : tensor<8x16x16x32x2xbf16>
> >   // This is an annotation (in-shape == out-shape, both VNNI)
> >   %Bv = tpp.vnni_pack %B : tensor<8x16x16x32x2xbf16>
> > ```
> 
> Right but if it is an operation you need to plug it into the upstream propagation. I also would prefer a cast here, just to keep the 2d shape and avoid complexity in the verifier/lowering.

lib/TPP/Dialect/Tpp/TppOps.cpp
rengolin: I thought so, but `-1` is still magic number. :)

lib/TPP/Dialect/Tpp/TppOps.cpp
rengolin: Yes, make the `shape == 2` the first case, because it's one line check, and the `shape == 3` the remainder, because there are many checks.

rengolin:
> Right but if it is an operation you need to plug it into the upstream propagation. I also would prefer a cast here, just to keep the 2d shape and avoid complexity in the verifier/lowering.

We can easily have two ops here: `tpp.pack_vnni` (the operation) and `tpp.vnni_operand` (the annotation). 

Technically, the annotation isn't a cast, just an annotation, so I wouldn't call it `cast` because the in/out types are identical.

test/Conversion/TppToXsmm/tpp-to-xsmm-quarternary.mlir
rengolin: Asserting that the test is covered elsewhere and removing the test is also a valid outcome. So is splitting the test into smaller steps (like we did previously) while making sure we don't lose test coverage.

adam-smnk:
> > Right but if it is an operation you need to plug it into the upstream propagation. I also would prefer a cast here, just to keep the 2d shape and avoid complexity in the verifier/lowering.
> 
> We can easily have two ops here: `tpp.pack_vnni` (the operation) and `tpp.vnni_operand` (the annotation).
> 
> Technically, the annotation isn't a cast, just an annotation, so I wouldn't call it `cast` because the in/out types are identical.

Yup, I had the annotation style in mind. It should be sufficient for us to start with this option and we'll see if we need the operation variant at all.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: I was looking at OneDNN (sc_data_format.hpp) and basically with @adam-smnk we were discussing something along the line:
```
func.func @gemm(%arg0: tensor<>, %arg1: tensor<>, %arg2: tensor<>) -> tensor<> {
    %i = arith.constant 32 : index
    %k = arith.constant 32 : index
    // We cannot use only attributes as the blocking factors can be values.
    %0 = layout.reorder %arg0 [KCSck2c, [%c, %k]] : tensor<256x512> -> tensor<256x512, #KCSck2c> // 2 packs
    %1 = layout.reorder %arg1 [KCSck2c, [32, 32]] : tensor<256x512> -> tensor<256x512, #KCSck2c>
    %2 = layout.reorder %arg2 [KCSck2c, [32, 32]] : tensor<256x512> -> tensor<256x512, #KCSck2c>
    /*** some ops ***/
    %4 = layout.reorder %3 [KCSck2c, [32, 32]] : tensor<256x512, #KCSck2c> -> tensor<256x512>
    return %4: tensor<256x512>
}
```
This would allow the compiler to have an analysis pass to decide the optional layout for each operation and simplify the packing for VNNI, and company.

chelini:
> 

We would like to have the in and out be different such that we can fold the VNNI dimension:
`%0 = vnni.cast %packed_tensor : %tensor<16x32x2xbf16> ->  %vnni_tensor<32x32xbf16>` this would simplify our life in multiple places: 
1. RewriteToBrgemm, currently, is hard to deduplicate as VNNI assumes specific affine maps and loop nest, and we emit different ops (linalg.brgemm, vnni.brgemm). Maybe @KavithaTipturMadhu has some suggestion on how to improve here.
2. tpp.gemm and tpp.brgemm verification, we need to handle 3d tensor in the verifier. Today, it is "fine" as we VNNI pack only one operand, but if we VNNI pack A and C, it would be a bit messy.
3. detection of blocked matmul before tile and fuse; we again need to look at the specific access map generated from VNNI packing. Basically having a custom handling for B.
4. XSMM lowering, here again, we need to consider a possible VNNI layout for B. 

If we move forward with the annotation/cast, that seems the most promising thing to do, I have some questions:
1. Where the annotation/cast should live, VNNI dialect? I don't see this being specific to TPP.
2. If we tile and fuse a linalg operation with a vnni_tensor<32x32xbf16>, the iteration domain (assuming identity maps) will still be a 32x32 and not a 16x32x2. I saw that when we pack matmul using the non-vnni packing, you want to change the iteration domain so that you can tile by one the IB/JB and map to BRGEMM, but maybe this is not the case for VNNI as we fold it as a GEMM flag into the library. Do you have any thoughts here?
3. How a `%vnni_tensor<32x32xbf16>` should lower to?

rengolin:
> We would like to have the in and out be different such that we can fold the VNNI dimension: `%0 = vnni.cast %packed_tensor : %tensor<16x32x2xbf16> -> %vnni_tensor<32x32xbf16>` this would simplify our life in multiple places:

This creates multiple problems:
1. The result type lies about the shape to MLIR. It's not even the right shape for VNNI (should be `<16x32xVbf16>`)
2. We have to introduce new types, handle new types and control conversions between those types and regular ones.
3. This only "works" with the consumer IF the consumer has a VNNI flag that accepts that input, not all ops do.
4. VNNI is one of many packing types that exist today, and more will come in the future.
5. This either creates a generic cast (where your confusing code will end up at) or lots of casts (where your confusing code will end up at).

Basically you're moving matching shapes complexity to matching custom types, from handling odd shaped tensors to handling all possible conversions between tensor types, from understanding odd affine maps to understanding wrong affine maps.

This is not a good direction.

The proposed way to deal with this last year was to move all VNNI matching logic to a set of helper functions which scale linearly with the number of extensions and can easily be composed with more generic matchers via some enum.

The proposed way to use annotation is just for the sake of the `tpp` dialect, with the shapes clear to the rest of MLIR and if invalid IR is generated, the process stop and does not generate broken code. This is the only safe thing to do.

> If we move forward with the annotation/cast, that seems the most promising thing to do, I have some questions:

From now on I'm talking about operation/annotation, not cast.

> 1. Where the annotation/cast should live, VNNI dialect? I don't see this being specific to TPP.

This is definitely a TPP thing. There is no VNNI dialect anymore and we don't want to create one dialect for every type packing extension.

> 2. If we tile and fuse a linalg operation with a vnni_tensor<32x32xbf16>, the iteration domain (assuming identity maps) will still be a 32x32 and not a 16x32x2. I saw that when we pack matmul using the non-vnni packing, you want to change the iteration domain so that you can tile by one the IB/JB and map to BRGEMM, but maybe this is not the case for VNNI as we fold it as a GEMM flag into the library. Do you have any thoughts here?

This is another problem creating by lying about the shape. Let's just not do that.

> 3. How a `%vnni_tensor<32x32xbf16>` should lower to?

Nothing, it should not exist.

chelini:
> 1. The result type lies about the shape to MLIR. It's not even the right shape for VNNI (should be `<16x32xVbf16>`)

Well this is depend what you have in mind when talking about layout. It can be a logical thing like a map, not necessarily a tensor type.

> 2\. We have to introduce new types, handle new types and control conversions between those types and regular ones.

True.

> 3\. This only "works" with the consumer IF the consumer has a VNNI flag that accepts that input, not all ops do.

The cast would be a no-op, it work also for consumer that do now support VNNI, as we still pack.

> 4\. VNNI is one of many packing types that exist today, and more will come in the future.

Indeed. I think current infrastructure is brittle. Handling different shapes add complexity everywhere from verifier to lowering, this is why we have different flows in rewrite to brgemm, bringing them together is hard (i.e., different affine maps matching and loop structure).

> 5\. This either creates a generic cast (where your confusing code will end up at) or lots of casts (where your confusing code will end up at).

No sure what this means. I create a cast from a VNNI operand.

chelini:
To be clear I am not supporting the cast for me something like: tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI>) would be fine as well as long as we keep the 2d nature of the tensor.

rengolin:
> > 1. The result type lies about the shape to MLIR. It's not even the right shape for VNNI (should be `<16x32xVbf16>`)
> 
> Well this is depend what you have in mind when talking about layout. It can be a logical thing like a map, not necessarily a tensor type.

A logical thing (if you mean just `tensor<32x32xbf16>`) would be worse, because now that's a know type to MLIR with a different semantics. WE know the difference, the rest of MLIR doesn't. This creates an impedance mismatch between assumptions and exposes risks of bad code generation.


> > 3. This only "works" with the consumer IF the consumer has a VNNI flag that accepts that input, not all ops do.
> 
> The cast would be a no-op, it work also for consumer that do now support VNNI, as we still pack.

Yes, but it creates a tensor with the original shape, and non-TPP ops do not look at the producer for packing info. If tensor manipulation ops (extract/insert) are created around for some other pass, these will be the wrong shapes. 

We cannot lie to the compiler.

> 
> > 4. VNNI is one of many packing types that exist today, and more will come in the future.
> 
> Indeed. I think current infrastructure is brittle. Handling different shapes add complexity everywhere from verifier to lowering, this is why we have different flows in rewrite to brgemm, bringing them together is hard (i.e., different affine maps matching and loop structure).

No, we have different flows because we didn't want to pay the cost of introducing that (required) complexity early on. 

We need to handle all the shapes we support, and that includes multiple levels of packing. If we cannot afford not to cope with those shapes, as this is the whole point of having an abstraction.

We need to find a way to cleanly handle different shapes and maps. The proposals above may "look" clean from the IR point of view, but they are not from the implementation and rest of the compiler point of view. You keep assuming a clean IR is a clean implementation and that is far from the truth, often the opposite is true.

If we pay the cost now for VNNI, we'll see the patterns in code, and when we need to add others, we'll be in a better place to judge. Right now we do not have a clear vision because we do not have a clear (even if messy) implementation and we do not have a second packing type.

Making such a high-level decision now is unwise. Let's introduce boilerplate in a secluded place, treat the common code as if the types and maps are "compatible" and not have shape/map specific decisions there.

> > 5. This either creates a generic cast (where your confusing code will end up at) or lots of casts (where your confusing code will end up at).
> 
> No sure what this means. I create a cast from a VNNI operand.

For more pacing types, you'll need to create more casts. Either one cast for all packing (which will have the shape problem we have now) or one cast for each packing type (where the code will end up split in multiple files).

> To be clear I am not supporting the cast for me something like: tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI>) would be fine as well as long as we keep the 2d nature of the tensor.

We have discussed this above, this does not simplify the picture in the same way.

chelini:
> > > 1. The result type lies about the shape to MLIR. It's not even the right shape for VNNI (should be `<16x32xVbf16>`)
> > 
> > 
> > Well this is depend what you have in mind when talking about layout. It can be a logical thing like a map, not necessarily a tensor type.
> 
> A logical thing (if you mean just `tensor<32x32xbf16>`) would be worse, because now that's a know type to MLIR with a different semantics. WE know the difference, the rest of MLIR doesn't. This creates an impedance mismatch between assumptions and exposes risks of bad code generation.
> 
> > > 3. This only "works" with the consumer IF the consumer has a VNNI flag that accepts that input, not all ops do.
> > 
> > 
> > The cast would be a no-op, it work also for consumer that do now support VNNI, as we still pack.
> 
> Yes, but it creates a tensor with the original shape, and non-TPP ops do not look at the producer for packing info. If tensor manipulation ops (extract/insert) are created around for some other pass, these will be the wrong shapes.
> 
> We cannot lie to the compiler.
> 
> > > 4. VNNI is one of many packing types that exist today, and more will come in the future.
> > 
> > 
> > Indeed. I think current infrastructure is brittle. Handling different shapes add complexity everywhere from verifier to lowering, this is why we have different flows in rewrite to brgemm, bringing them together is hard (i.e., different affine maps matching and loop structure).
> 
> No, we have different flows because we didn't want to pay the cost of introducing that (required) complexity early on.
> 
> We need to handle all the shapes we support, and that includes multiple levels of packing. If we cannot afford not to cope with those shapes, as this is the whole point of having an abstraction.
> 
> We need to find a way to cleanly handle different shapes and maps. The proposals above may "look" clean from the IR point of view, but they are not from the implementation and rest of the compiler point of view. You keep assuming a clean IR is a clean implementation and that is far from the truth, often the opposite is true.
> 
> If we pay the cost now for VNNI, we'll see the patterns in code, and when we need to add others, we'll be in a better place to judge. Right now we do not have a clear vision because we do not have a clear (even if messy) implementation and we do not have a second packing type.
> 
> Making such a high-level decision now is unwise. Let's introduce boilerplate in a secluded place, treat the common code as if the types and maps are "compatible" and not have shape/map specific decisions there.
> 
> > > 5. This either creates a generic cast (where your confusing code will end up at) or lots of casts (where your confusing code will end up at).
> > 
> > 
> > No sure what this means. I create a cast from a VNNI operand.
> 
> For more packing types, you'll need to create more casts. Either one cast for all packing (which will have the shape problem we have now) or one cast for each packing type (where the code will end up split in multiple files).
> 
> > To be clear I am not supporting the cast for me something like: tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI>) would be fine as well as long as we keep the 2d nature of the tensor.
> 
> We have discussed this above, this does not simplify the picture in the same way.

I did not explain myself clearly enough, let me try again. I am proposing to have the VNNI layout as a logical layout in the tensor, something like: `tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI)>` or `vnni2<32x32xbf16>`. I think this will help us in different places in the codebase from verifier to rewrite-to-brgemm. Keeping the VNNI explicitly in the tensor shape complicated the code a bit and prevent further simplification in my opinion (i.e., unclear to me how to simplify rewrite-to-brgemm specifically the affine maps matching). I propose to connect the logical layout with a cast operation, for example:
```
%0 = tensor.pack [2] [0] : tensor<32x32xbf16> -> tensor<16x32x2xbf16> // do the actual packing
%1 = vnni.cast tensor<16x32x2xbf16> to tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI)> // cast to keep the shape 2d
%2 = use(%1)
```
Every consumer of `%1` will use the type: `tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI)> ` and this is fine also for other dialects like linalg, memref normalization will expand the type to 3d further down the lowering, while for us it will fold as a library flag.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: We never packed BRGEMM

rengolin:
> ```
> %0 = tensor.pack [2] [0] : tensor<32x32xbf16> -> tensor<16x32x2xbf16> // do the actual packing
> %1 = vnni.cast tensor<16x32x2xbf16> to tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI)> // cast to keep the shape 2d
> %2 = use(%1)
> ```
> 
> Every consumer of `%1` will use the type: `tensor<32x32xbf16, affine_map<(d0, d1) -> (d0/VNNI, d1, VNNI)> ` and this is fine also for other dialects like linalg, memref normalization will expand the type to 3d further down the lowering, while for us it will fold as a library flag.

This is exactly what I understood.

To reiterate, the problems are:
1. You need a `vnni` dialect. Then you'll need a dialect for each new type packing. Though this could be a generic `tensor` cast, but you need to encode multiple semantics and upstream, which is moving the complexity elsewhere.
2. You need to compose affine maps on packf-of-packs, which is not simpler to understand that their final shapes.
3. You can't guarantee you'll have a perfect chain of packs (there may be other ops in the middle) so you can't use producer/consumer to determine the steps taken to the final affine map (as you can't for shapes either).
4. The shape analysis code you "save" will be "spent" in affine map analysis, neither have guaranteed provenance.
5. It does not differentiate between extensions that have the same final packing shape/map, but may require different function flags.

A simpler solution is to just have an annotation to the operand:
```mlir
  // From:
  %0 = tensor.pack [2] [0] ... { vnni } : tensor<16x32x2xbf16>
  // Note the `vnni` attribute. Without it, we don't know that this is a VNNI pack.
  // This is equivalent of having a tpp.vnni_pack that does the same thing.
  // We cannot assume VNNI without it.
  %2 = use(%1)

  // To:
  %0 = ... : tensor<16x32x2xbf16>
  %1 = tpp.vni_operand %0 : tensor<16x32x2xbf16>
  %2 = use(%1)
```

The verification of the 3D shape in `use` is the same for all operations, so only needs to be done once. All TPP tile ops are 2D or 3D depending on whether it's VNNI or not. New extensions will have different shapes, or different operand annotations even with the same shape.

lib/TPP/VNNIUtils.cpp
chelini: grepping a bit seems all clean.

test/Conversion/TppToXsmm/tpp-to-xsmm-quarternary.mlir
chelini: Test is already covered. See new exmaples for the fusion pass and we have the conversion in`tpp-to-xsmm-quaternary.mlir`.

lib/TPP/ToBlockLayoutAndBack.cpp
rengolin: Why do we pack VNNI brgemm and not standard brgemm?

We don't want to have VNNI specific functions at this level, because the end result is that they're very similar but different. We need a way that doesn't make them different, relying on composition (template, pimpl, etc).

lib/TPP/Dialect/Tpp/TppOps.cpp
```diff
@@ -268,36 +270,95 @@ void AddOp::getEffects(
...
-    return emitOpError("fails to verify operands dimensions mismatch");
+template <typename OpTy>
+static LogicalResult
+validateVnniGemmOperand(OpTy operation, ArrayRef<int64_t> shape,
+                        Type elementType, int dimI, int dimJ) {
+
+  if (shape.size() == 2 && (shape[0] != dimI || shape[1] != dimJ))
+    return operation->emitOpError("operand 1 fails to verify expected shape");
+
+  if (shape.size() == 3) {
```
rengolin, line 281: You just need to bail here if `size != 3`.

lib/TPP/Dialect/Tpp/TppOps.cpp
```diff
@@ -268,36 +270,95 @@ void AddOp::getEffects(
...
+      return operation->emitOpError() << "operand 1 invalid VNNI layout expect "
+                                         "inner dims to be 2 or 4, but got: "
+                                      << shape[shape.size() - 1] << "\n";
+    }
+    // VNNI layout: [K/VNNI][J][VNNI]
+    if (shape[0] * shape[2] != dimI || shape[1] != dimJ)
+      return operation->emitOpError("operand 1 fails to verify expected shape");
+    return success();
+  }
+  if (shape.size() != 2) {
```
rengolin, line 297: Early checks should go to the beginning of the function.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: No idea, why it is like this.

chelini:
> 

Ack.

lib/TPP/ToBlockLayoutAndBack.cpp
rengolin: @KavithaTipturMadhu can you shed light on this?

chelini:
> I have resolved the further simplification requests due to our recent discussion. We'll need to re-write the packing to avoid VNNI layout until the last moment, so there won't be the need to any of those VNNI checks this high anyway.

Perfect. What do you think is still missing in this PR?

rengolin:
> Perfect. What do you think is still missing in this PR? Happy to iterate more if needed.

The one question about `ConvertTppFusedBrgemmOp` to @KavithaTipturMadhu. I want to make sure how we proceed. That's removing both VNNI and non-VNNI variants, so I'm not sure it's not cutting too deep.

Perhaps we just need a test to check it, or that won't be necessary after we move the VNNI pass to the end, so we can remove.

chelini:
> > Perfect. What do you think is still missing in this PR? Happy to iterate more if needed.
> 
> The one question about `ConvertTppFusedBrgemmOp` to @KavithaTipturMadhu. I want to make sure how we proceed. That's removing both VNNI and non-VNNI variants, so I'm not sure it's not cutting too deep.
> 
> Perhaps we just need a test to check it, or that won't be necessary after we move the VNNI pass to the end, so we can remove.

We could rework it bottom-up, here my suggestion: https://github.com/plaidml/tpp-mlir/issues/552

chelini:
> 

@KavithaTipturMadhu any idea here? Would like to move this forward.

Issue #430, opened 2023-03-28
GPU runner


rengolin:
We want to try our passes on Intel GPUs, but we need some special sauce to glue our passes with GPU fiddling to get it offloaded onto devices.

We can start with simple recipes and then move to a unified `tpp-gpu-run` just like our `tpp-run`.

@hfp @adam-smnk

adam-smnk:
CUDA support was added to `tpp-run` in #547.
Vulkan support #573 has been parked for now due to lacking server infrastructure.

Closing this issue as we now have at least one viable path to execute IR on GPU.
Further work will be done in separate issues.

PR #547, opened 2023-05-17
Initial GPU support

adam-smnk:
Adds initial GPU pipeline infrastructure together with support for basic GPU lowering flow targeting CUDA.

The new features are disabled by default. When enabled, the new GPU compilation flow is integrated into `tpp-run` together with necessary extra MLIR runtime libraries.
Currently, the GPU pipeline is separate from the default pipeline. The two pipelines are not designed to produce the same results even when the input IR has no GPU operations.

Work towards #430


adam-smnk:
The GPU support is disabled by default as it requires matching LLVM build with GPU features enabled.

The pipeline has been tested locally both manually and using the added tests.
I am also able to observe some small loads showing on my Nvidia GPU when running these test kernels.

One major obstacle present is sharing memory between host and the device. The current upstream assumptions on the memory model fails to work for tested GPUs: laptop RTX A3000 and desktop RTX 3080.
This will have to be addressed separately through a new `gpux` dialect op like `gpu.alloc` - TBD

adam-smnk:
While there is upstream `gpu.alloc` op, it had no lowering infrastructure the last time I tested it (or maybe I was doing it wrong).
We might need an op that supports pointer translation like `cuMemHostGetDevicePointer` as `gpu.host_register` does not work for me.

lib/TPP/GPU/GpuPipeline.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+private:
+  void constructPipeline() override {
+    pm.clear();
+
+    GpuType gpuType = parseGpuOption(this->gpuBackend);
+
+    // Add no passes when no GPU is selected.
+    if (gpuType == GpuType::None)
+      return;
```
rengolin, line 91: What is the actual case when someone selects `USE_GPU=1` and `gpu_type=none`?

Do we even need that option?

lib/TPP/GPU/GpuPipeline.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    // Map and lower ops to GPU-compatible format.
+    pm.addNestedPass<func::FuncOp>(createConvertLinalgToParallelLoopsPass());
+    pm.addNestedPass<func::FuncOp>(createGpuMapParallelLoopsPass());
+    pm.addNestedPass<func::FuncOp>(createParallelLoopToGpuPass());
```
rengolin, line 96: As expected, there's no TPP here.

Once we do have it, do we just combine the passes (using the TPP pass builder), making this a GPU-but-not-TPP specific code (therefore `lib/GPU` makes sense), or do we add them directly (therefore `lib/TPP/GPU` makes sense).

lib/TPP/GPU/GpuToCuda.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    if (failed(runPipeline(pm, module)))
+      return signalPassFailure();
+  }
+
+private:
+  void constructPipeline() override {
+    pm.clear();
+
+#ifdef TPP_GPU_ENABLE
```
rengolin, line 71: I don't like this, but I understand why you did it.

I think a cleaner solution is to not even try to build the library if the GPU support is disabled and then you don't have to `ifdef` anything inside the code.

IIUC, this will clash with `tpp-run` when trying to conditionally add the GPU option, so not a perfect solution, still need figuring that out.

lib/TPP/GPU/README.md
```diff
@@ -0,0 +1,75 @@
...
+-DMLIR_ENABLE_CUDA_RUNNER=ON \
+-DMLIR_ENABLE_CUDA_CONVERSIONS=ON \
+-DMLIR_ENABLE_SPIRV_CPU_RUNNER=ON \
+-DMLIR_ENABLE_VULKAN_RUNNER=ON
+```
+Full build command:
+```sh
+cd llvm-project/build
+
+cmake -G Ninja ../llvm \
```
rengolin, line 21: I would not have this twice. Just the `cmake` command line is enough.

tools/tpp-run/MLIRBench.cpp
```diff
@@ -375,8 +381,13 @@ LogicalResult MLIRBench::finalize(PrintStage print) {
   if (print == PrintStage::Early)
     passManager.addPass(createPrintIRPass());
 
-  // Apply the default preprocessing pass
-  passManager.addPass(tpp::createDefaultTppPass(tppToLoops, linalgToLoops));
+  if (defGpuBackend != "none") {
```
rengolin, line 384: I'd initialize this as `""` and not allow `none`.

lib/TPP/GPU/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+add_mlir_library(TPPGPU
```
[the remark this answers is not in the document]
adam-smnk: I would assume that everything is TPP-centric. The current GPU infrastructure is pretty generic but it will evolve to be more tailor-made for our specific needs.
It is just a subdirectory to keep every GPU related in one place for convenience.

lib/TPP/GPU/GpuPipeline.cpp
adam-smnk: You can enable GPU support but choose to still target only CPU.
The explicit `none` is a bit silly so it is removed now.

lib/TPP/GPU/GpuPipeline.cpp
```diff
@@ -0,0 +1,127 @@
...
+    pm.addNestedPass<func::FuncOp>(createCleanupPass());
+
+    // Create GPU kernels.
+    pm.addPass(createGpuKernelOutliningPass());
+
+    // Lower GPU ops to the chosen GPU backend.
+    if (gpuType == GpuType::Cuda)
+      pm.addNestedPass<gpu::GPUModuleOp>(createGpuToCudaPass());
+    else
+      assert(false && "Unsupported GPU backend");
```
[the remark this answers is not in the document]
adam-smnk: Reorganized as suggested

lib/TPP/GPU/GpuPipeline.cpp
adam-smnk: My intention is to gather GPU related logic in this directory instead of top level `TPP` simply for convenience. No other special meaning.
In the future, I'd add `ConvertTppToGpu` etc. to `TPP/GPU`.

lib/TPP/GPU/GpuToCuda.cpp
adam-smnk: In general, most passes are always availables and to support tablegen generated code we need to provide minimum boilerplate. If we don't build anything, we just shift this problem to callers to figure out (more `ifdef`s) when `createGpuToCudaPass` is available.

Macros wouldn't be necessary if all passes followed that logic but `createGpuSerializeToCubinPass` is not available without `MLIR_GPU_TO_CUBIN_PASS_ENABLE` so we need to somehow handle that.
I tried a few alternative approaches but IMO this solution i.e., minimal pass skeleton that does nothing when GPU is disable, is the simplest.

tools/tpp-run/tpp-run.cpp
```diff
@@ -311,6 +311,14 @@ int main(int argc, char **argv) {
   llvm::InitializeNativeTargetAsmPrinter();
   llvm::InitializeNativeTargetAsmParser();
 
+  // Initialize LLVM NVPTX backend.
+#ifdef TPP_GPU_ENABLE
```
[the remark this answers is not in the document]
adam-smnk: Moved it to a separate utility to hide the macros in one place.

lib/TPP/GPU/GpuToCuda.cpp
rengolin: Let's go with this for now and decide how to improve later.

rengolin:
I've added a new CI loop callet `TPP-GPU` which sources the `scripts/buildkite/tpp-gpu.yml` file. Please replicate the same structure for the other builds, but with GPU support. 

This probably means:
* Copy `tpp-mlir.yml` to `tpp-gpu.yml` adding some `GPU=1` selecting NVidia node (only clang build is relevant)
* Add support for GPU variable in `scripts/buildkite/build_tpp.sh`
* Propagate that to `scripts/ci/cmake.sh` and `scripts/ci/build.sh` including finding the correct `nvcc`

All jobs must continue to pass (without GPU flag), and the new job should trigger inside an NVidia node.

For now, the `TPP-GPU` loop only triggers on branch push for branches named `gpu-*`, so once the PR is green otherwise, just push is to `upstream/gpu-something` and it will trigger the GPU build. 

Once that's green, we can merge and enable the GPU build for PRs as well as any branch push.

scripts/buildkite/build_tpp.sh
```diff
@@ -62,10 +68,24 @@ then
...
+CUDA_DRIVER=/usr/lib64/libcuda.so.1
+if [ ! -f "${CUDA_DRIVER}" ]; then
+  if [ "${GPU}" ]; then
+    echo "GPU support requires full CUDA driver to be present"
+    exit 1
+  else
+    # When GPU is not used, create link to CUDA stubs to satify dynamic linker.
+    echo "Creating links to CUDA stubs"
+    ln -s ${CUDATOOLKIT_HOME}/lib64/stubs/libcuda.so ${BLD_DIR}/lib/libcuda.so.1
+    ln -s ${BLD_DIR}/lib/libcuda.so.1 ${BLD_DIR}/lib/libcuda.so
```
adam-smnk, line 81: TODO: see if tpp-run can be successfully linked directly with CUDA lib to avoid these runtime hacks

scripts/buildkite/tpp-gpu.yml
```diff
@@ -0,0 +1,24 @@
...
+  #  - SRUN
+  LLVMROOT: "${TPPROOT}/llvm"
+  INSTALL_PREFIX: "${TPPROOT}/compiler"
+  NPROCS_LIMIT_LINK: "8"
+  #LIBXSMM_VERBOSE: 4
+  #LIBXSMMFETCH: 1
+  TPP_CLANGVER: "16.0.4"
+  TPP_GCCVER: "9.5.0" # Use an older version to match CUDA requirements
+  TPP_CUDAVER: "11.5"
+  ASAN_OPTIONS: "protect_shadow_gap=0:replace_intrin=0:detect_leaks=0:${ASAN_OPTIONS}"
```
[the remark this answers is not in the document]
rengolin: why not just add the comment directly to the yml?

rengolin (approved):

test/GPU/gpu-pipeline.mlir
```diff
@@ -0,0 +1,40 @@
+// RUN: tpp-opt %s -gpu-pipeline=gpu=cuda | FileCheck %s --check-prefix=CUDA
```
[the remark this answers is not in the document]
adam-smnk: There will be more gpu targets coming so it's a setup for future

scripts/buildkite/build_tpp.sh
adam-smnk: What I had in mind didn't work anyway. We'll have to stick with this approach for now.

scripts/buildkite/tpp-gpu.yml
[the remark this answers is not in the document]
adam-smnk: added now

adam-smnk:
Waiting for one more test pass on TPP-GPU and then pr should be good to go.

PR #548, opened 2023-05-17
Mobilebert stripped

chelini:
Taken from IREE


rengolin (commented):
This is really nice, but it's too big and has no testing in it.

I think it's time we create a separate repo with just MLIR models dumped from places and have a harness that builds `tpp-mlir` and runs it through all models. (like https://github.com/plaidml/mlir-generator)

This way we can easily just add more models without having to plug them to our testing infrastructure, or having to strip-clean them or anything.

For IR checks, the tests need to be a lot smaller (like the ones @nhasabni has created), so that we can check code generation on a more manageable way.

PR #556, opened 2023-05-19
Simplify constant pack folding test to compilation only

### LINKED ISSUE #398: `test/Integration/pack-folding{1}.mlir` problems

rengolin:
Relies on the compiler simplifying the packed constant and the expected packed comparison into a single constant and figuring out that the subtraction must be zero. It will only be zero if the compiler got the constant packing correct.

Fixes #398


test/Passes/pack-folding.mlir
```diff
@@ -0,0 +1,33 @@
+// RUN: tpp-opt %s -constant-fold-pack -canonicalize | FileCheck %s
```
chelini, line 1: would `-cse` be enough?

rengolin: Instead of canonicalize? Doesn't matter much here, to be honest...

Issue #561, opened 2023-05-22
Create a "target description" class for target-specific decisions


rengolin:
Today, we make wrong packing decisions based on types (ex. bf16 always means vnni) instead of target support.

We also make a [compile-time decision](https://github.com/plaidml/tpp-mlir/blob/main/lib/TPP/VNNIUtils.cpp#L24) about the packing shapes, which is also wrong in general.

Since we're lowering into library calls, we need a link into target libraries to tell us what they support, not just what the targets actually support.

A set of classes that return supported packing types, shapes, extensions would do the trick. Once lowered to our `tpp` dialect, the lowering directly matches the supported architecture.

rengolin:
https://github.com/libxsmm/libxsmm/blob/main_stable/src/libxsmm_cpuid_x86.c#L521

rengolin:
Discussing this with @chelini and @hfp, the issue here is to separate "libxsmm" from "x86" and "arm", for the compiler decisions (tiling, loop order, passes) and to pass the info down to the targets below. We do not, however, want to re-implement what LLVM already gives us.

LLVM has target descriptors for all supported CPUs and we already create them in `tpp-run` via the `--cpu` flag (like clang). But LLVM itself does not know how to detect the host's arch. That logic is performed by the Clang driver and we really do *not* want to replicate that.

LIBXSMM has target detection (via cpuid, etc) and uses it for its run time JIT compilation, but the API is limited and may not be present in a compiler that does not use libxsmm (but lowers directly). It also may not be 1:1 with what MLIR/LLVM represents, even for CPUs, so it wouldn't be a generic approach that could one day be upstreamed.

@chelini proposed we use https://github.com/pytorch/cpuinfo, which is the same as [IREE](https://github.com/openxla/iree/tree/main/third_party) uses, and is a generic enough library that could work. However, it would have the same problem as libxsmm regarding upstreaming. Since we already depend on libxsmm, we should try to use it until we can't.

But this is just the host detection part. We also need a way to convey target information (via `-cpu` flags) and pass that down to the MLIR compiler (for tiling) and the LLVM compiler (for vectorization) and the library implementation (for example MKL) if we intermix compiler generated code with ninja written library kernels.

alheinecke:
I strong agree with:
> However, it would have the same problem as libxsmm regarding upstreaming. Since we already depend on libxsmm, we should try to use it until we can't.
[...]

PR #565, opened 2023-05-24
Fix TODO to use libxsmm for VNNI check (NFC)

rengolin:
Does not change the semantics, just makes sure to use a libxsmm call instead of macros.

Working towards #561


hfp (approved):
Go ahead and merge/rebase. I would not bet the API used is rock-solid/final but good enough. More important, there is a starting point now to use LIBXSMM there.

PR #566, opened 2023-05-25
Include extra libxsmm headers

adam-smnk:
Adds extra libxsmm include directory to address missing libxsmm_version.h header.

### DIFF
--- a/lib/TPP/CMakeLists.txt
+++ b/lib/TPP/CMakeLists.txt
@@ -59,6 +59,7 @@ add_mlir_library(MLIRTPP
 target_include_directories(MLIRTPP
   PUBLIC
     ${XSMM_INCLUDE_DIRS}
+    ${XSMM_INCLUDE_DIRS}/../src/template
     $<BUILD_INTERFACE:${TPP_GEN_INCLUDE_DIR}>
     $<BUILD_INTERFACE:${TPP_MAIN_INCLUDE_DIR}>
 )


rengolin (approved):

hfp:
This is an unintentional fix and it is related to building LIBXSMM using an external build system. I will look at this case a bit deeper on LIBXSMM's side. The unintentional/wrong part is about including from `src/template`.

rengolin:
I actually get the same error on my machine...

nhasabni:
I am getting a configure error while configuring IREE .. 

```
CMake Error in third_party/tpp-mlir/lib/TPP/CMakeLists.txt:
  Target "MLIRTPP" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/nfs_home/nhasabni/other/tensor_compiler/iree-clang-build/_deps/xsmm-src/include"

  which is prefixed in the build directory.
```

nhasabni:
This change has fixed the problem for me - 

```
diff --git a/lib/TPP/CMakeLists.txt b/lib/TPP/CMakeLists.txt
index a1b13da..0e84108 100644
--- a/lib/TPP/CMakeLists.txt
+++ b/lib/TPP/CMakeLists.txt
@@ -58,8 +58,8 @@ add_mlir_library(MLIRTPP
 
 target_include_directories(MLIRTPP
   PUBLIC
-    ${XSMM_INCLUDE_DIRS}
-    ${XSMM_INCLUDE_DIRS}/../src/template
+    $<BUILD_INTERFACE:${XSMM_INCLUDE_DIRS}>
+    $<BUILD_INTERFACE:${XSMM_INCLUDE_DIRS}/../src/template>
     $<BUILD_INTERFACE:${TPP_GEN_INCLUDE_DIR}>
     $<BUILD_INTERFACE:${TPP_MAIN_INCLUDE_DIR}>
 )
```

rengolin:
That looks like the right fix. Forgive my ignorance, this is what I should have done from the beginning... :/

Send a PR so we can merge.

nhasabni:
No problem.

hfp:
This is definitely the wrong solution since LIBXSMM's `src/template/version.h` is no valid C or C++ code.

I will find a fix for LIBXSMM to better enable this use-case. Though, LIBXSMM supports (1) to be built with its own build system, (2) to be built with a 3rd part build system such as CMake and omitting configure/static instantiation of code, (3) to be build as header-only with C++ compiler, and (4) to be built as header-only using a C compiler.

PR #568, opened 2023-05-30
Fix IREE-side XsmmRunnerUtils

nhasabni:
These changes with the latest changes in IREE work for all the models including BERT.

CMake change is needed as we now enable XSMM symbol resolver (i.e., TPP import) via IREE's Executable Plugin mechanism. With this change, the symbol resolver can be built separately from IREE.

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -11,7 +11,7 @@ if (NOT TPP_INSIDE_IREE)
   target_compile_definitions(tpp_c_runner_utils PRIVATE mlir_c_runner_utils_EXPORTS)
 else()
   add_library(tpp_c_runner_utils
-    STATIC
+    SHARED
     XsmmRunnerUtils.cpp
     PerfRunnerUtils.cpp
   )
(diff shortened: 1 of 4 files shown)


runtime/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ if (NOT TPP_INSIDE_IREE)
   target_compile_definitions(tpp_c_runner_utils PRIVATE mlir_c_runner_utils_EXPORTS)
 else()
   add_library(tpp_c_runner_utils
-    STATIC
+    SHARED
```
chelini, line 14: Why shared here?

chelini (commented):
Thanks, overall good to me. Just one comment.

runtime/CMakeLists.txt
rengolin: Because for IREE plugin it needs to be a shared object that gets loaded at runtime.

We need to change the CMake files too (on a separate PR), to remove references to IREE directly and just separate between STATIC (for tpp-mlir stack) and SHARED (for plugins to frameworks).

rengolin (approved):
Looks good, thanks! Let's get this merged so we can complete the IREE merge without patches on our repo there.

nhasabni:
Thanks all!

PR #570, opened 2023-06-01
Fix for TPP configure error in IREE

nhasabni:
This is for the change in PR 566.


rengolin (approved):

nhasabni:
Thanks @rengolin for quick approval!

PR #572, opened 2023-06-07
Tensor init integers

adam-smnk:
Adds support for initialization of integer tensors.

The original tensor init logic is refactored and split into separate concrete implementations for floats and integers.
Currently, only signless integers (i8, i32 etc.) are supported as they are the most commonly used types when working with integers.


rengolin (approved):
This will be necessary for the MHA generator and benchmarks, thanks!

PR #573, opened 2023-06-07
GPU runner vulkan support

adam-smnk:
Extends GPU backend with Vulkan support which can then target many different devices: CPU, Nvidia GPU, Intel GPU etc.

Work towards #430


adam-smnk:
CI infrastructure currently lacks working Vulkan runtime setup. Furthermore, upstream SPIR-V support is lacking and further external dependencies would be needed to have truly viable pipeline.
Closing this PR for now. It will be revisited when server backend is ready.

PR #576, opened 2023-06-08
Naive TPP to GPU lowering

adam-smnk:
Adds a simple TPP to GPU lowering through parallel loops and standard MLIR infrastructure.

The `tpp-to-loops` pass is extended with optional use of parallel loops. The parallelization is disable by default as primary use of this pass is debugging when producing the simplest possible IR is desired.
The new lowering path is integrated into the GPU pipeline to allow moving TPP ops into GPU devices.

### DIFF
--- a/include/TPP/Passes.h
+++ b/include/TPP/Passes.h
@@ -83,6 +83,8 @@ std::unique_ptr<OperationPass<func::FuncOp>> createConvertLinalgToTppPass();
 std::unique_ptr<OperationPass<func::FuncOp>>
 createConvertLinalgToTppPass(bool, bool, ArrayRef<int64_t> tiles = {});
 std::unique_ptr<OperationPass<func::FuncOp>> createConvertTppToLoopsPass();
+std::unique_ptr<OperationPass<func::FuncOp>>
+createConvertTppToLoopsPass(bool parallel);
 std::unique_ptr<OperationPass<ModuleOp>> createConvertXsmmToFuncPass();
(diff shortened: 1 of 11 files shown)


include/TPP/Passes.h
```diff
@@ -83,6 +83,8 @@ std::unique_ptr<OperationPass<func::FuncOp>> createConvertLinalgToTppPass();
 std::unique_ptr<OperationPass<func::FuncOp>>
 createConvertLinalgToTppPass(bool, bool, ArrayRef<int64_t> tiles = {});
 std::unique_ptr<OperationPass<func::FuncOp>> createConvertTppToLoopsPass();
+std::unique_ptr<OperationPass<func::FuncOp>>
```
rengolin, line 86: Why a new one? Perhaps just adding the `parallel` flag (default false) would do?

adam-smnk: The defaults should have only one origin and that is the tablegen. Adding default argument values at the header only adds maintenance burden as now you have to sync both .td and .h.
(guilty of that, I'll have to clean it up one day 😅 )

rengolin: Then why not add the option/default to the tablegen?

adam-smnk: It is there but I still need a constructor that allows me to pass a custom value.

adam-smnk: Discussed offline, let's keep it consistent with our other passes.
TODO: cleanup double API, add a comment to the header

rengolin (approved):
Thanks!

PR #577, opened 2023-06-08
Adds fused linalg addf + maxf detection

### LINKED ISSUE #571: Recognize special addf+maxf fused linalg generic
In the past we approached the fused generic by trying to decompose it (#359, #339), but due to the complexities of identifying such cases and the fact that we're undoing fusion only to redo later on, we decided to go with a `tpp.group` operation that collects all scalar operations inside as a chain of operations on tiles.

However, `tpp.group` isn't going to be implemented soon and we now have a case of a matmul followed by bias add and relu (through IREE) that fuses the two element-wise ones too early, so when we lower to XSMM, we don't recognize their patterns.

Instead of reviving the decompose pass or rushing the implementation of `tpp.group` I propose special treatment for `addf+maxf` lowering to `tpp.add+tpp.relu`. Once `tpp.group` implementation can replace that lowering, we remove it from the code.

The rationale behind adding this _hack_ is that this is a key property of the main models we're working on and not recognizing them means half of our work goes down the drain, let alone performance crashing.

It should be trivial and isolated to have a post-TPP pass to recognize stray matches, which are all clear candidates for future cleaning.

rengolin:
This adds a generic detection for multiple chained ops in a linalg body to allow for fused element-wise addf + maxf that are generated by IREE.

It introduces a generic mechanism that duplicated a bit of the one-op matcher but that will eventually replace no-op and one-op matcher with a generic multi-op (zero or more) detector.

This can later be used for matmul (mulf + addf), groups (bcast + gemm + bias + relu) and equations.

Fixes #571

### DIFF
--- a/lib/TPP/IR/StructuredOpMatcher.cpp
+++ b/lib/TPP/IR/StructuredOpMatcher.cpp
@@ -239,16 +239,17 @@ bool tpp::structured_match::withOpChainImpl(
     // At least one operand must come from args or a previous op
     bool consumesValueFromChain = false;
     for (auto operand : innerOp.getOperands()) {
-      if (chainedValues.contains(operand) && capturedOperands) {
+      if (chainedValues.contains(operand)) {
         // First add to the captured
         auto ba = dyn_cast<BlockArgument>(operand);
-        if (ba && ba.getParentBlock() == linalgOp.getBlock()) {
+        if (capturedOperands && ba &&
+            ba.getParentBlock() == linalgOp.getBlock()) {
           capturedOperands->push_back(linalgOp.getMatchingOpOperand(ba)->get());
         }
+        // Then erase from the set
+        chainedValues.remove(operand);
+        consumesValueFromChain = true;
       }
-      // Then erase from the set
-      chainedValues.remove(operand);
-      consumesValueFromChain = true;
     }
 
     // Operation isn't in the chain
--- a/lib/TPP/IR/StructuredOpMatcher.cpp
+++ b/lib/TPP/IR/StructuredOpMatcher.cpp
@@ -201,6 +201,80 @@ bool tpp::structured_match::WithSingleOpImpl::withSingleOpImpl(
   return true;
 }
 
+// FIXME: This is a generalization of the method above and will eventually replace
+// the matcher for both no-op (yield) and one op (add, max).
+bool tpp::structured_match::withOpChainImpl(
+    Region *region, Operation *op, SmallVectorImpl<Value> *capturedOperands,
+    SmallVectorImpl<TypeCheckFunc>& typeChecks) {
+
+  // Numer of ops includes yield
+  ptrdiff_t numOps = typeChecks.size();
+
+  // Basic checks
+  if (!isa<linalg::GenericOp>(op))
+    return false;
+  auto linalgOp = cast<linalg::GenericOp>(op);
+  if (!region->hasOneBlock())
+    return false;
+  auto& block = region->front();
+  if (std::distance(block.begin(), block.end()) != (numOps + 1))
+    return false;
+  if (linalgOp.getNumDpsInits() != 1)
+    return false;
+
+  // Add generic arguments to the list of chained values
+  llvm::SmallSetVector<Value, 4> chainedValues;
+  for (auto arg: block.getArguments()) {
+    chainedValues.insert(arg);
+  }
+
+  // Check on the inner chain of operations in the right order.
+  // Make sure all operands are used and chained
+  for (auto [check, innerOp]: llvm::zip_first(typeChecks, block.getOperations())) {
+    // Must be right op in right order
+    if (!check(&innerOp))
+      return false;
+
+    // At least one operand must come from args or a previous op
+    bool consumesValueFromChain = false;
+    for (auto operand: innerOp.getOperands()) {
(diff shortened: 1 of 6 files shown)


adam-smnk:
The fused linalg detection crashes on memref.

rengolin:
> The fused linalg detection crashes on memref.

Hm, I didn't test on memref... I thought we weren't doing it anymore...

Do we even want to care about memrefs from now on?

adam-smnk:
> Do we even want to care about memrefs from now on?

The default pipeline still operates on memref tpps. It couldn't be moved to tensor due to some VNNI ops, I think it should be possible right now but I haven't revisited it yet.

rengolin:
> The default pipeline still operates on memref tpps. It couldn't be moved to tensor due to some VNNI ops, I think it should be possible right now but I haven't revisited it yet.

VNNI has been removed, we should move the default pipeline to tensors ASAP.

test/Conversion/LinalgToTpp/linalg-to-tpp-tensor.mlir
```diff
@@ -324,3 +324,136 @@ func.func @scalar_input(%arg0: tensor<f32>, %arg1: tensor<4x4xf32>) -> tensor<4x
...
+  // CHECK: tpp.zero
+  // CHECK: tpp.gemm
+  %empty = tensor.empty() : tensor<128x512xf32>
+  %c0 = arith.constant 0.0 : f32
+  %fill = linalg.fill ins(%c0 : f32) outs(%empty : tensor<128x512xf32>) -> tensor<128x512xf32>
+  %0 = linalg.generic
+            {indexing_maps = [#map2, #map3, #map4], iterator_types = ["parallel", "parallel", "reduction"]}
+            ins(%arg0, %arg1 : tensor<128x256xf32>, tensor<256x512xf32>)
+            outs(%fill : tensor<128x512xf32>)
+            attrs =  {iterator_ranges = [128, 512, 256]} {
```
adam-smnk, line 385: Are the `attrs` relevant in any way?

lib/TPP/IR/StructuredOpMatcher.cpp
```diff
@@ -201,6 +201,80 @@ bool tpp::structured_match::WithSingleOpImpl::withSingleOpImpl(
   return true;
 }
 
+// FIXME: This is a generalization of the method above and will eventually replace
+// the matcher for both no-op (yield) and one op (add, max).
+bool tpp::structured_match::withOpChainImpl(
+    Region *region, Operation *op, SmallVectorImpl<Value> *capturedOperands,
+    SmallVectorImpl<TypeCheckFunc>& typeChecks) {
+
+  // Numer of ops includes yield
```
adam-smnk, line 210:
```suggestion
  // Number of ops includes yield
```

adam-smnk: Does it really include yield if it requires `numOps + 1` later?

lib/TPP/IR/StructuredOpMatcher.cpp
```diff
@@ -201,6 +201,80 @@ bool tpp::structured_match::WithSingleOpImpl::withSingleOpImpl(
...
+      return false;
+
+    // At least one operand must come from args or a previous op
+    bool consumesValueFromChain = false;
+    for (auto operand: innerOp.getOperands()) {
+      if (chainedValues.contains(operand)) {
+        // First add to the captured
+        if (capturedOperands) {
+          if (auto ba = dyn_cast<BlockArgument>(operand)) {
+            if (ba.getParentBlock() == linalgOp.getBlock()) {
```
adam-smnk, line 245: nit: it'd be easier to read with reduced nesting depth

lib/TPP/IR/StructuredOpMatcher.cpp
rengolin: No it did not. I think it did in a previous iteration and I didn't update the comment. Since it's only used once, I changed it to include again.

lib/TPP/IR/StructuredOpMatcher.cpp
rengolin: Yeah, this was a result of iteration development, simplifying it.

test/Conversion/LinalgToTpp/linalg-to-tpp-tensor.mlir
rengolin: Probably not, this is just copy & paste. Will remove.

lib/TPP/IR/StructuredOpMatcher.cpp
```diff
@@ -201,6 +201,77 @@ bool tpp::structured_match::WithSingleOpImpl::withSingleOpImpl(
...
+      if (chainedValues.contains(operand) && capturedOperands) {
+        // First add to the captured
+        auto ba = dyn_cast<BlockArgument>(operand);
+        if (ba && ba.getParentBlock() == linalgOp.getBlock()) {
+          capturedOperands->push_back(linalgOp.getMatchingOpOperand(ba)->get());
+        }
+      }
+      // Then erase from the set
+      chainedValues.remove(operand);
+      consumesValueFromChain = true;
```
adam-smnk, line 251: I think this is missing some check. `consumesValueFromChain` is set to true on every iteration.

PR #578, opened 2023-06-09
Remove heap-to-stack pass

adam-smnk:
Removes 'heap-to-stack' pass.

The pass was originally introduced as a hack to address performance issues in some of the benchmarks. However, it is a hacky approach and generally leads to huge stack usage. Thus, its benefits are limited.
With the improved code generation and better benchmark kernels, this pass is no longer needed.


rengolin (approved):
Looks fine, this was not a very successful experiment. :)

adam-smnk:
It looks that the performance deviation is minimal. The results of the repeated run of TPP-performance pipeline also seem fine.

Issue #375, opened 2023-03-08
Cannot lower 1d tpp.identity to XSMM


chelini:
To reproduce:
```
func.func @main(%arg0: memref<8xf32, strided<[1]>>, %arg1: memref<8xf32, strided<[1]>>) {
  tpp.identity ins(%arg0 : memref<8xf32, strided<[1]>>) out(%arg1 : memref<8xf32, strided<[1]>>)
  return
}
```
The tpp.identity does not lower to xsmm.

rengolin:
Is this due to some missing BROADCAST flag in the runtime?

chelini:
> 

No, we bail out if the output is not 2d: https://github.com/plaidml/tpp-mlir/blob/4f9666a2598310abb4c21dc61db97f150715cb5a/lib/TPP/ConvertTppToXsmm.cpp#L372

rengolin:
This issue should be unnecessary, now that we'll retire `tpp.identity` and will have specific `zero`, `copy` and `broadcast` unary ops.

chelini:
This comes from old decisions to lower 1d and 2d tile operations. Should we just lower 2d tile operations?

chelini:
We can close, but we would need to rework tpp mapping to avoid 1d ops. We can simplify loop lowering and xsmm if we allow 2d tiles only. Do we all agree on the direction?

chelini:
See: https://github.com/plaidml/tpp-mlir/pull/553/commits/0b97a55b3cb108bde13de1dcf0a12b68359d392c

PR #579, opened 2023-06-09
Add predicate to match operand rank

chelini:
Commit 0ff739123309ac8e1ab68f6dbb3dadce9f1ee27f restricted tpp operation to 2d output, but the matchers to convert from linalg to tpp were not constrained, leading to verification errors. Introduce a matching predicate that allows constraining the rank of the operand, specifically the tpp matcher will filter only inputs that have rank 0 (scalar), 1 or 2. In contrast, for output, only rank 2 is allowed.

### DIFF
--- a/include/TPP/IR/StructuredOpMatcher.h
+++ b/include/TPP/IR/StructuredOpMatcher.h
@@ -58,7 +58,7 @@ struct MatchRange : public MatchSelector {
 // Callable object to check if the number of loops in `op` satisfies `fun`.
 struct NumOfLoops {
   NumOfLoops() = delete;
-  NumOfLoops(std::function<bool(size_t)> fun) : fun(fun){};
+  explicit NumOfLoops(std::function<bool(size_t)> fun) : fun(fun){};
 
   bool operator()(Operation *op) const {
     if (auto linalgOp = dyn_cast_or_null<linalg::LinalgOp>(op)) {
@@ -74,7 +74,7 @@ struct NumOfLoops {
 // `fun`.
 struct HasMap {
   HasMap() = delete;
-  HasMap(std::function<bool(AffineMap)> fun) : fun(fun){};
+  explicit HasMap(std::function<bool(AffineMap)> fun) : fun(fun){};
 
   bool operator()(OpOperand *operand, Operation *op) const {
     if (auto linalgOp = dyn_cast_or_null<linalg::LinalgOp>(op)) {
@@ -113,6 +113,26 @@ struct HasStaticShape {
         return false;
     return true;
   }
+};
+
+// Callable object to verify `operand` to have a rank in `ranks`.
+struct HasRank {
+  HasRank() = delete;
+  explicit HasRank(std::initializer_list<int64_t> ranks) : ranks(ranks){};
+
+  bool operator()(OpOperand *operand, Operation *op) const {
+    auto operandType = operand->get().getType();
+    if (!isa<ShapedType>(operandType))
+      return llvm::is_contained(ranks, HasRank::SCALAR);
+    int64_t rank = operandType.cast<ShapedType>().getRank();
+    return llvm::any_of(
+        ranks, [=](int64_t expectedRank) { return expectedRank == rank; });
+  }
+
+  // There are multiple way to represent a scalar: f32, tensor<f32>.
+  // SCALAR means f32.
+  static constexpr int64_t SCALAR = -1;
+  std::vector<int64_t> ranks;
 };
 
 // Callable object to check if the input is equal to specified `value`.
--- a/lib/TPP/TestMatchers.cpp
+++ b/lib/TPP/TestMatchers.cpp
@@ -159,6 +159,28 @@ void testTppIdentity(FunctionOpInterface funcOp) {
   });
 }
 
+void testTppRank(FunctionOpInterface funcOp) {
+  // clang-format off
+  auto matcherRank1 =
+    StructuredOpMatcher::make<linalg::GenericOp>()
+      .input(MatchAll(), HasRank({1}));
+  auto matcherRank2 =
+    StructuredOpMatcher::make<linalg::GenericOp>()
+      .input(MatchAll(), HasRank({2}));
+  auto matcherScalar = 
+    StructuredOpMatcher::make<linalg::GenericOp>()
+      .input(MatchAll(), HasRank({HasRank::SCALAR}));
+  // clang-format on
+  funcOp->walk([&](linalg::LinalgOp linalgOp) {
+    if (matcherRank1.match(linalgOp))
+      llvm::outs() << "match rank 1\n";
+    if (matcherRank2.match(linalgOp))
+      llvm::outs() << "match rank 2\n";
+    if (matcherScalar.match(linalgOp))
+      llvm::outs() << "match scalar\n";
+  });
+}
+
 void TestStructuralMatchers::runOnOperation() {
   auto f = getOperation();
   llvm::outs() << f.getName() << "\n";
@@ -176,6 +198,8 @@ void TestStructuralMatchers::runOnOperation() {
     testInterfaces(f);
   if (f.getName() == "test_tpp_identity")
     testTppIdentity(f);
+  if (f.getName() == "test_rank")
+    testTppRank(f);
 }
 
 namespace mlir {
--- a/test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
+++ b/test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
@@ -679,3 +679,23 @@ func.func @linalg_fill_3d(%arg0: memref<2x8x32xf32>) -> memref<2x8x32xf32> {
   linalg.fill ins(%cst : f32) outs(%arg0 : memref<2x8x32xf32>)
   return %arg0 : memref<2x8x32xf32>
 }
+
+// -----
+
+#map = affine_map<(d0, d1) -> ()>
+#map1 = affine_map<(d0, d1) -> (d0, d1)>
+
+// CHECK-LABEL: scalar_input
+func.func @scalar_input(%arg0: memref<f32>, %arg1: memref<4x4xf32>) {
+  // CHECK-NOT: tpp.add
+  linalg.generic {
+    indexing_maps = [#map, #map1],
+    iterator_types = ["parallel", "parallel"]} 
+    ins(%arg0: memref<f32>)
+    outs(%arg1: memref<4x4xf32>) {
+      ^bb0(%in: f32, %out: f32):
+        %0 = arith.addf %in, %out : f32
+        linalg.yield %0 : f32
+  }
+  return
+}
--- a/test/Passes/test-matchers.mlir
+++ b/test/Passes/test-matchers.mlir
@@ -250,3 +250,63 @@ func.func @test_tpp_identity(%arg0: memref<3xf32>, %arg1: memref<5x3xf32>) {
   } 
   return
 }
+
+#map13 = affine_map<(d0, d1) -> (0)>
+#map14 = affine_map<(d0, d1) -> (d0, d1)>
+#map15 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+#map16 = affine_map<(d0, d1) -> ()>
+
+// CHECK-LABEL: test_rank
+func.func @test_rank(%arg0: memref<1xf32>, 
+                     %arg1: memref<4x4xf32>, %arg2: memref<4x4x4xf32>, 
+                     %arg3: f32) {
+  // CHECK: match rank 1
+  // CHECK-NOT: match rank 2
+  // CHECK-NOT: match scalar
+  linalg.generic {
+    indexing_maps = [#map13, #map14],
+    iterator_types = ["parallel", "parallel"]} 
+    ins(%arg0: memref<1xf32>)
+    outs(%arg1: memref<4x4xf32>) {
+      ^bb0(%in: f32, %out: f32):
+        %0 = arith.addf %in, %out : f32
+        linalg.yield %0 : f32
+  }
+  // CHECK: match rank 2
+  // CHECK-NOT: match rank 1
+  // CHECK-NOT: match scalar
+  linalg.generic {
+    indexing_maps = [#map14, #map14],
+    iterator_types = ["parallel", "parallel"]} 
+    ins(%arg1: memref<4x4xf32>)
+    outs(%arg1: memref<4x4xf32>) {
+      ^bb0(%in: f32, %out: f32):
+        %0 = arith.addf %in, %out : f32
+        linalg.yield %0 : f32
+  }
+  // CHECK: match scalar
+  // CHECK-NOT: match rank 2
+  // CHECK-NOT: match rank 1
+  linalg.generic {
+    indexing_maps = [#map16, #map14],
+    iterator_types = ["parallel", "parallel"]} 
+    ins(%arg3: f32)
+    outs(%arg1: memref<4x4xf32>) {
+      ^bb0(%in: f32, %out: f32):
+        %0 = arith.addf %in, %out : f32
+        linalg.yield %0 : f32
+  }
+  // CHECK-NOT: match scalar
+  // CHECK-NOT: match rank 1
+  // CHECK-NOT: match rank 2
+  linalg.generic {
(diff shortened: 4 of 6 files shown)


include/TPP/IR/StructuredOpMatcher.h
```diff
@@ -115,6 +115,26 @@ struct HasStaticShape {
...
+    auto operandType = operand->get().getType();
+    if (!isa<ShapedType>(operandType))
+      return llvm::is_contained(ranks, HasRank::SCALAR);
+    int64_t rank = operandType.cast<ShapedType>().getRank();
+    return llvm::any_of(
+        ranks, [=](int64_t expectedRank) { return expectedRank == rank; });
+  }
+
+  // There are multiple way to represent a scalar: f32, tensor<f32>.
+  // SCALAR means f32.
```
rengolin, line 133: This is a weird requirement... 

"scalars" can be:
* `f32`, `bf16`, `i64` etc.
* `memref<f32>`, `memref<bf16>`, `memref<i64>` etc.
* `memref<1xf32>`, `memref<1xbf16>`, `memref<1xi64>` etc.
* `memref<1x1xf32>`, `memref<1x1xbf16>`, `memref<1x1xi64>` etc.
* etc

include/TPP/IR/StructuredOpMatcher.h
```diff
@@ -115,6 +115,26 @@ struct HasStaticShape {
...
+  static constexpr int64_t SCALAR = -1;
```
rengolin, line 134: What does `-1` means here?

lib/TPP/TestMatchers.cpp
```diff
@@ -159,6 +159,20 @@ void testTppIdentity(FunctionOpInterface funcOp) {
   });
 }
 
+void testTppRank(FunctionOpInterface funcOp) {
+  // clang-format off
+  auto matcher =
+    StructuredOpMatcher::make<linalg::GenericOp>()
+      .input(MatchAll(), HasRank({HasRank::SCALAR, 1}));
```
rengolin, line 166: This is very little testing...

adam-smnk:
Why should we constrain elementwise ops to 2D outputs?

test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
```diff
@@ -679,3 +679,23 @@ func.func @linalg_fill_3d(%arg0: memref<2x8x32xf32>) -> memref<2x8x32xf32> {
...
+#map = affine_map<(d0, d1) -> ()>
+#map1 = affine_map<(d0, d1) -> (d0, d1)>
+
+// CHECK-LABEL: scalar_input
+func.func @scalar_input(%arg0: memref<f32>, %arg1: memref<4x4xf32>) {
+  // CHECK-NOT: tpp.add
+  linalg.generic {
+    indexing_maps = [#map, #map1],
+    iterator_types = ["parallel", "parallel"]} 
+    ins(%arg0: memref<f32>)
```
rengolin, line 694: Here, your input is `memref<f32>` but in the comment you say `SCALAR means f32`.

You don't have enough tests to make sure all shapes are understood.

include/TPP/IR/StructuredOpMatcher.h
adam-smnk: I think if the input has form of `memref<1x1xf32>` etc. then it just falls under standard rank requirements i.e., 1D or 2D.
But it would be nice to support `memref<f32>`.

chelini:
> Why should we constrain elementwise ops to 2D outputs?

this what we discussed last time. It does not make sense to lower vector operations to tpp. We only lower 2d tile.

test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
chelini: SCALAR means f32, not `memref<f32>`. The latter is not handled.

include/TPP/IR/StructuredOpMatcher.h
chelini: represents a scalar (i.e. f32 or non shaped type).

include/TPP/IR/StructuredOpMatcher.h
chelini: indeed, memref<1x1xf32> is a 2d tensor. It does not make sense to handle this in the matchers.

rengolin: In other places of the code (broadcasting) we treat `1x1xf32` like a scalar.

We need to ask the question of what it the intention of this matcher: to match against the expectation or the literal. If it is the literal, we probably need another matcher to find the effective dimension (intention).

So, why is this matcher being added, and what are the surrounding matchers that will be changed/added/removed because of this?

include/TPP/IR/StructuredOpMatcher.h
rengolin: Well, that's the obvious answer. My point is, why `-1`? Why not `0`?

Why have the special constant when the others are just integers with the number of dims?

Why not just construct this with `{ 0, 1, 2 }` (which is perfectly obvious to me) instead of `{HasRank::SCALAR, 1, 2}`?

lib/TPP/TestMatchers.cpp
rengolin: Why does `matcher0` have both `SCALAR` and `1`?

lib/TPP/TestMatchers.cpp
```diff
@@ -159,6 +159,28 @@ void testTppIdentity(FunctionOpInterface funcOp) {
...
 }
 
+void testTppRank(FunctionOpInterface funcOp) {
+  // clang-format off
+  auto matcher0 =
+    StructuredOpMatcher::make<linalg::GenericOp>()
+      .input(MatchAll(), HasRank({HasRank::SCALAR, 1}));
+  auto matcher1 =
+    StructuredOpMatcher::make<linalg::GenericOp>()
+      .input(MatchAll(), HasRank({2}));
```
rengolin, line 169: Why is `matcher1` actually matches `2` when it reports "match 1"?

lib/TPP/TestMatchers.cpp
```diff
@@ -159,6 +159,28 @@ void testTppIdentity(FunctionOpInterface funcOp) {
...
+  auto matcher2 = 
+    StructuredOpMatcher::make<linalg::GenericOp>()
```
rengolin, line 171: Why is `matcher2` actually matches `SCALAR` when it reports "match 2"?

lib/TPP/TestMatchers.cpp
```diff
@@ -159,6 +159,28 @@ void testTppIdentity(FunctionOpInterface funcOp) {
...
+      .input(MatchAll(), HasRank({HasRank::SCALAR}));
+  // clang-format on
+  funcOp->walk([&](linalg::LinalgOp linalgOp) {
+    if (matcher0.match(linalgOp))
+      llvm::outs() << "match 0\n";
+    if (matcher1.match(linalgOp))
+      llvm::outs() << "match 1\n";
+    if (matcher2.match(linalgOp))
+      llvm::outs() << "match 2\n";
```
rengolin, line 180: What if it doesn't match anything, how can you grep the output?

test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
rengolin: In your implementation, yes. My question is why are we doing this?

Why is it not handled?

test/Passes/test-matchers.mlir
```diff
@@ -250,3 +250,63 @@ func.func @test_tpp_identity(%arg0: memref<3xf32>, %arg1: memref<5x3xf32>) {
...
+#map13 = affine_map<(d0, d1) -> (0)>
+#map14 = affine_map<(d0, d1) -> (d0, d1)>
+#map15 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+#map16 = affine_map<(d0, d1) -> ()>
+
+// CHECK-LABEL: test_rank
+func.func @test_rank(%arg0: memref<1xf32>, 
+                     %arg1: memref<4x4xf32>, %arg2: memref<4x4x4xf32>, 
+                     %arg3: f32) {
+  // CHECK: match 0
```
rengolin, line 263: These match output makes no sense. Please always abide to the "principle of least surprise".

rengolin (commented):
I think the matcher is needed and we just need to get the details right.

A wider question is: where is this IR coming from? You're making decisions of what is important and what's not from an IR that is here hand-written.

It would be good to understand what's the origin and its representation in the wild.

alheinecke:
I haven't looked at the code, and just saw some comments: TPP output can be a vector or a scalar, when use reduce TPP. I saw the comment that TPP output is 2D, hence I'm bringing this up.

test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
chelini: the support of memref<f32> never got implemented. Probably "we" was me at the beginning.

lib/TPP/TestMatchers.cpp
chelini: done.

include/TPP/IR/StructuredOpMatcher.h
chelini: Thinking a bit more here. It is better to keep them separate until we support memref<f32> or tensor<f32> otherwise it is unclear what { 0 } means. Should we match f32 or memref<f32>/tensor<f32>?

include/TPP/IR/StructuredOpMatcher.h
rengolin: Exactly and we don't want to have to do that because bufferization may get it wrong. IF whatever comes from input as `memref<f32>` comes as an `f32`, then it should be fine to accept it as `0D`. 

This is why I'm asking: where is this coming from? What are you seeing that isn't shared in this PR?

test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
rengolin: This is not clear to me what you mean and what support we're talking about. It will depend on what input IR we have, so the question is: where is this coming from and how can we support the types we need without adding more ops to the IR.

include/TPP/IR/StructuredOpMatcher.h
chelini: I am just trying to be pragmatic here. `tensor<1x1xf32>` has rank 2 even if it is a scalar. If we want to allow these kind of scalar we need an additional pass before to collapse. We should technically be able to handle both `f32` and `memref<f32`, the latter is not implemented but we can add support.

include/TPP/IR/StructuredOpMatcher.h
chelini: It is not all types, just shapedType vs scalarType. -1 represents a scalar type. Both input are representative: `f32` and `memref<f32>`.

lib/TPP/TestMatchers.cpp
chelini: this is fixed.

include/TPP/IR/StructuredOpMatcher.h
rengolin: What is the input you're working on?

test/Conversion/LinalgToTpp/linalg-to-tpp-memref.mlir
rengolin: What the spec allows or not depends on whether they appear in the wild or not. If they do, it would be simpler to support them than to force canonicalization.

But without knowing where this problem is coming from, it's hard to tell if the spec needs changing or not.

rengolin:
> > Why should we constrain elementwise ops to 2D outputs?
> 
> this what we discussed last time. It does not make sense to lower vector operations to tpp. We only lower 2d tile.

I'm not sure what you mean by this. 

Any TPP operation that has a reduction dimension will have a rank lower than 2D, so we can't just block all TPP operations to have 2D outputs.

Once we have a more fine-grained approach (more TPP ops, broadcast/reduce as special operand operations), we can restrict "computing" operations on 2D input/output, but some other operations will still need to support multiple ranks.

For now, with the broadcast as a property of each TPP operation, we need inputs < 2D. When we add reduction (soon), we'll need the same thing.

Hopefully, we'll be able to add broadcast/reduction operations before that, so we go back restricting "compute" operations on 2D tiles, but that definitely is not a property of _all_ TPP operations.

chelini:
> > > Why should we constrain elementwise ops to 2D outputs?
> > 
> > 
> > this what we discussed last time. It does not make sense to lower vector operations to tpp. We only lower 2d tile.
> 
> I'm not sure what you mean by this.
> 
> Any TPP operation that has a reduction dimension will have a rank lower than 2D, so we can't just block all TPP operations to have 2D outputs.
> 
> Once we have a more fine-grained approach (more TPP ops, broadcast/reduce as special operand operations), we can restrict "computing" operations on 2D input/output, but some other operations will still need to support multiple ranks.
> 
> For now, with the broadcast as a property of each TPP operation, we need inputs < 2D. When we add reduction (soon), we'll need the same thing.
> 
> Hopefully, we'll be able to add broadcast/reduction operations before that, so we go back restricting "compute" operations on 2D tiles, but that definitely is not a property of _all_ TPP operations.

@rengolin we discussed during our last 1 to 1 and we agreed to restricted the matching here the discussion (https://github.com/plaidml/tpp-mlir/issues/375).

rengolin:
> > > this what we discussed last time. It does not make sense to lower vector operations to tpp. We only lower 2d tile.
> 
> @rengolin we discussed during our last 1 to 1 and we agreed to restricted the matching here the discussion (#375).

Not quite the same statements here, I think we have miscommunicated.

We want to lower compute tile operations to 2D tiles because this is optimal code gen on modern vector architectures. But we still must support non-2D input for broadcast, reduction, tiling, etc.

Our agreement in March was that we want to restrict ops like `add`, `relu` to 2D tiles and have `broadcast` and `reduce` operations to/from 0D/1D.

This means we can't for **all** TPP operations to output 2D tiles, as this will break reduction operations.

The main thing I want to know it: where is this IR coming from? What is the requirement here?

include/TPP/IR/StructuredOpMatcher.h
chelini: The input (test case in `linalg-to-tpp-memref/tensor.mlir` with `memref<f32>` was provided by Adam when moving the default pipeline to tensor. The fact that we need to handle `f32` is because we do have tests that require to handle scalars (https://github.com/plaidml/tpp-mlir/blob/5c19269f0fcdaddb009a9337de258fa5457bbb8d/test/Integration/copy.mlir#L60).

rengolin: That's an artificial IR from a test, I'm not sure what's the relevance of it. We need to understand what is the meaning of `memref<f32>`, ie. what to front-ends mean when they lower to this shape and now `f32`.

chelini: Deciding to support `f32` and/or `memref<f32>` can be addresses in another PR. This PR intends not to allow `memref<f32>` as tpp verifier requires `f32` to specify a scalar and not `memref<f32>`. The change is needed when moving the default pipeline to tpp on tensor.

rengolin: The solution may be to add a check for the particular case of `memref<f32>` instead of forcing all TPP ops to output 2D shapes or to create an artificial rule about all of their inputs.

It'd be much simpler to just bail on `memref<f32>` and leave a comment like `TODO: check if this is a common pattern and what to do about it`. If this isn't coming from any ingress IR, then it's not high priority to fix this now.

rengolin (commented):
Speaking about this offline we've reached an agreement: The checks are good to have but the current state of the dialect makes them not very helpful. Once we have more ops and they move into a more stable semantics (ex. broadcast/reduce/tile split), we can add stronger checks for each type.

For now, we just need to make sure we don't accept 0D tensor/memref (`<f32>`) neither on inputs nor outputs. Scalars, 1D and 2D tensors/memrefs are fine.

In summary: Change the `output` match to be the same as `input`.

rengolin (approved):

Issue #580, opened 2023-06-09
Move default pipeline to TPP on tensors


adam-smnk:
To complete the work of TPP on tensors, the conversion from linalg to TPP has to be moved before bufferization in the default pipeline.

The [current WIP attempt](https://github.com/adam-smnk/tpp-mlir/tree/def-pipe-tpp-tensors) exposed a few issues that have to be solved before the pipeline can start operating on TPP on tensors:
* ~outdated heap-to-stack tests - the pass should be removed~ ; done in #578 
* unsupported `xsmm.fused_brgemm.dispatch` without broadcast (`binary_flags = (none)`) generated by `mlir-gen.mlir` test - the issue due to libxsmm codegen ; #582 
* missing `tpp.fused_brgemm` lowering to loops ; #591 
* new failures due to sanitizers - memory leaks present in tests
  * buffers returned from functions called by test kernel lead to leaks ; #599 
  * TPP bufferization leads to memory leaks - #586
  * incorrect bufferization of chained ternary ops ; #592 
* invalid linalg to TPP mapping on tensors with `tensor<f32>`  type e.g., `copy.mlir` or `tpp-run.mlir` test - the linalg is not matched on memrefs but matched on tensors and created TPP op validation fails ; #579 
* `rewrite-to-brgemm` and `rewriteConvToMatmul` is no longer parallelized - `scf.parallel` is generated for memrefs but only `scf.for` for tensors - this impacts parallelization as certain kernels rely on this logic to create `scf.parallel` loops which get later picked up by OMP rewriting ; #590

PR #582, opened 2023-06-09
Handle 'tpp.fused_brgemm' without broadcasting

adam-smnk:
Adds a workaround to enable working lowering of `tpp.fused_brgemm` without XSMM `bcast_col_in0` binary flag.
Additionally, quarternary ops testing is improved.

Due to current limitation of LIBXSMM https://github.com/libxsmm/libxsmm/issues/766, code generation fails when the specific binary flag is not present.
After TPP on tensors support was introduced, `tpp.fused_brgemm` ops which violate this limitation started showing up. To address the issue, offending fused ops are split into separate XSMM operations which do not have this limitations.

Work towards #580

### DIFF
--- a/include/TPP/Dialect/Tpp/TppUtils.h
+++ b/include/TPP/Dialect/Tpp/TppUtils.h
@@ -16,6 +16,7 @@
 namespace mlir {
 class TypeRange;
 class Value;
+class PatternRewriter;
 
 namespace linalg {
 class LinalgOp;
@@ -24,6 +25,8 @@ class YieldOp;
 } // end namespace linalg
 
 namespace tpp {
+class FusedBrgemmOp;
+
 namespace utils {
 
 // Returns true if the linalg operation is marked with 'target'.
@@ -74,6 +77,9 @@ bool isValConstZero(Value val);
 
 // Returns true if the op defining `val` represents a zero filled tensor.
 bool isZeroTensor(Value val);
+
+// Splits fused op into its individual components.
+void splitFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp, PatternRewriter &rewriter);
 
 } // namespace utils
 } // namespace tpp
--- a/include/TPP/Dialect/Tpp/TppUtils.h
+++ b/include/TPP/Dialect/Tpp/TppUtils.h
@@ -78,8 +78,11 @@ bool isValConstZero(Value val);
 // Returns true if the op defining `val` represents a zero filled tensor.
 bool isZeroTensor(Value val);
 
-// Splits fused op into its individual components.
-void splitFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp, PatternRewriter &rewriter);
+// Splits fused op into its individual components and returns created
+// operations.
+// The passed fused op remains untouched.
+SmallVector<Operation *> splitFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp,
+                                      PatternRewriter &rewriter);
 
 } // namespace utils
 } // namespace tpp
--- a/include/TPP/Dialect/Tpp/TppUtils.h
+++ b/include/TPP/Dialect/Tpp/TppUtils.h
@@ -82,8 +82,8 @@ bool isZeroTensor(Value val);
 // Temporary workaround for:
 // https://github.com/libxsmm/libxsmm/issues/766
 // TODO: Move into tpp-to-loops as a private helper.
-void splitAndReplaceFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp,
-                            PatternRewriter &rewriter);
+LogicalResult splitAndReplaceFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp,
+                                     PatternRewriter &rewriter);
 
 } // namespace utils
 } // namespace tpp
--- a/include/TPP/Dialect/Tpp/TppUtils.h
+++ b/include/TPP/Dialect/Tpp/TppUtils.h
@@ -78,11 +78,12 @@ bool isValConstZero(Value val);
 // Returns true if the op defining `val` represents a zero filled tensor.
 bool isZeroTensor(Value val);
 
-// Splits fused op into its individual components and returns created
-// operations.
-// The passed fused op remains untouched.
-SmallVector<Operation *> splitFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp,
-                                      PatternRewriter &rewriter);
+// Splits and replaces fused op with its individual components.
+// Temporary workaround for:
+// https://github.com/libxsmm/libxsmm/issues/766
+// TODO: Move into tpp-to-loops as a private helper.
+void splitAndReplaceFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp,
+                            PatternRewriter &rewriter);
 
 } // namespace utils
 } // namespace tpp
--- a/lib/TPP/ConvertTppToXsmm.cpp
+++ b/lib/TPP/ConvertTppToXsmm.cpp
@@ -291,10 +291,13 @@ struct ConvertTppFusedBrgemmOp : public OpRewritePattern<tpp::FusedBrgemmOp> {
     // binary is supported. Split into separate operations if other flags
     // are present.
     // TODO: remove the split once LIBXSMM is fixed.
+    auto isBiasAdd = brgemmOp.getBinaryKind() == tpp::FusedBinaryOpKind::ADD;
     auto binaryFlag = getBinaryFlags(rewriter, brgemmOp)[0]
                           .cast<xsmm::BinaryFlagsAttr>()
                           .getValue();
-    if (binaryFlag != xsmm::BinaryFlags::BCAST_COL_IN_0) {
+    auto isBitSet = static_cast<uint64_t>(binaryFlag) &
+                    static_cast<uint64_t>(xsmm::BinaryFlags::BCAST_COL_IN_0);
+    if (isBiasAdd && !isBitSet) {
       tpp::utils::splitFusedOp(brgemmOp, rewriter);
       rewriter.eraseOp(brgemmOp);
       return success();
--- a/lib/TPP/ConvertTppToXsmm.cpp
+++ b/lib/TPP/ConvertTppToXsmm.cpp
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "TPP/Dialect/Tpp/TppOps.h"
+#include "TPP/Dialect/Tpp/TppUtils.h"
 #include "TPP/Dialect/Xsmm/XsmmEnum.h"
 #include "TPP/Dialect/Xsmm/XsmmOps.h"
 #include "TPP/Passes.h"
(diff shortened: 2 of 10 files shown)


lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -157,9 +157,8 @@ static bool isTppUnaryOp(linalg::GenericOp linalgOp) {
 // Return true if the linalg.generic can be mapped to a tpp.add.
 bool isTppAdd(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
   using namespace tpp::structured_match;
-  auto addMatcher =
-      StructuredOpMatcher::make<linalg::GenericOp>()
-          .region(MatchOne(0), WithSingleOp<arith::AddFOp>(operands));
+  auto addMatcher = StructuredOpMatcher::make<linalg::GenericOp>().region(
```
rengolin, line 160: avoid formatting only changes as this conflicts with other in-flight PR.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -272,6 +270,34 @@ bool isTppZero(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
   return true;
 }
 
+void splitFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp, PatternRewriter &rewriter) {
```
rengolin, line 273: Can we just not fuse them (for now) in the first place?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -272,6 +270,34 @@ bool isTppZero(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
...
+  assert(fusedBrgemmOp.hasBufferSemantics() &&
+         "tpp.fused_brgemm expects a memref type");
+
+  Location loc = fusedBrgemmOp.getLoc();
+
+  // Split the fused op into individual operations.
+  auto ins = fusedBrgemmOp.getInputs();
+  auto out = fusedBrgemmOp.getOutput();
+  rewriter.create<tpp::BrgemmOp>(loc, ValueRange{ins[0], ins[1], ins[2]}, out);
```
rengolin, line 282: You need to replace or you'll end up with an extra op?

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: You can still have a stand-alone op, addressing the issue at the time of lowering ensures that `tpp.fused_brgemm` always works.
Anyway, I will need the split anyway to implement lowering to loops.

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: I let the caller handle the cleanup.

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: We should really add some hooks to enforce consistent formatting.

adam-smnk: #583

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: Hmm I think I'm fine with either approach. Do you think erasing the original op within `splitFusedOp` is more intuitive?

TODO: the function should return a SmallVector with created ops

adam-smnk: TODO done

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: Fixed formatting

include/TPP/Dialect/Tpp/TppUtils.h
```diff
@@ -75,6 +78,12 @@ bool isValConstZero(Value val);
 // Returns true if the op defining `val` represents a zero filled tensor.
 bool isZeroTensor(Value val);
 
+// Splits fused op into its individual components and returns created
+// operations.
+// The passed fused op remains untouched.
```
rengolin, line 83: Add a FIXME/TODO here to explain why this shouldn't be permanent not a pattern to repeat.

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -285,6 +286,24 @@ struct ConvertTppFusedBrgemmOp : public OpRewritePattern<tpp::FusedBrgemmOp> {
...
+    // is fused.
+    // TODO: remove the split once LIBXSMM is fixed.
+    auto isBiasAdd = brgemmOp.getBinaryKind() == tpp::FusedBinaryOpKind::ADD;
+    auto binaryFlag = getBinaryFlags(rewriter, brgemmOp)[0]
+                          .cast<xsmm::BinaryFlagsAttr>()
+                          .getValue();
+    auto isBitSet = static_cast<uint64_t>(binaryFlag) &
+                    static_cast<uint64_t>(xsmm::BinaryFlags::BCAST_COL_IN_0);
+    if (isBiasAdd && !isBitSet) {
+      (void)tpp::utils::splitFusedOp(brgemmOp, rewriter);
```
rengolin, line 302: There is a single use which doesn't use the return values, why return them?

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -272,6 +272,47 @@ bool isTppZero(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
...
+  assert(fusedBrgemmOp.hasBufferSemantics() &&
+         "tpp.fused_brgemm expects a memref type");
+
+  SmallVector<Operation *> splitOps;
+  Location loc = fusedBrgemmOp.getLoc();
+
+  // Split the fused op into individual operations.
+  auto ins = fusedBrgemmOp.getInputs();
+  auto out = fusedBrgemmOp.getOutput();
+  Operation *brgemm = rewriter.create<tpp::BrgemmOp>(
```
rengolin, line 286: This function creates new operations but requires the caller to remove the old one, I don't see a case where this would be beneficial over erasing it here, and it's not needed in the one case where you use it.

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -272,6 +272,47 @@ bool isTppZero(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
...
+    Operation *relu = rewriter.create<tpp::ReluOp>(loc, out, out);
+    splitOps.push_back(relu);
+    break;
+  }
+  case tpp::FusedUnaryOpKind::NONE: {
+    break;
+  }
+  }
+
+  return splitOps;
```
rengolin, line 313: There are no early exits, no branches, nothing that would require careful control flow analysis to only remove the old op in special cases. You can just insert the removal here.

lib/TPP/ConvertTppToXsmm.cpp
rengolin, line 302: You're adding new ops with the rewriter without setting the insertion point. Is that guaranteed to be at the right place?

test/BF16/Integration/xsmm-quarternary-bf16.mlir
```diff
@@ -1,16 +1,17 @@
...
     flags = (vnni_b) binary_flags = (bcast_col_in0) unary_flags = (none) data_type = bf16
-  xsmm.fused_brgemm(data_type = bf16, %func, %arg0, %arg1, %arg2, %arg3, %c16_i64) : (i64, memref<64x4x4xbf16>, memref<64x2x4x2xbf16>, memref<4xbf16>, memref<4x4xbf16>, i64) -> ()
+  xsmm.fused_brgemm(data_type = bf16, %func, %arg0, %arg1, %arg2, %arg3, %c16_i64) : (i64, memref<64x4x4xbf16>, memref<64x2x4x2xbf16>, memref<4x4xbf16>, memref<4xbf16>, i64) -> ()
+
+  %threshold = arith.constant 0.0 : bf16
+  %outVal = arith.constant 66.0 : bf16
+  %trueOut = memref.alloc(): memref<4x4xbf16>
+  linalg.fill ins(%outVal : bf16) outs(%trueOut : memref<4x4xbf16>)
+  check.expect_almost_eq(%trueOut, %arg2, %threshold): memref<4x4xbf16>, memref<4x4xbf16>, bf16
+
```
rengolin, line 15: Why are you dropping the output check?

lib/TPP/Dialect/Tpp/TppUtils.cpp
adam-smnk: Reworked the function

test/BF16/Integration/xsmm-quarternary-bf16.mlir
adam-smnk: It was a useless check. By default we print the last function argument `%arg3` which in this case is used as constant bias values. The computation results are stored in `%arg2`. The check op covers the output validation.

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -285,6 +286,23 @@ struct ConvertTppFusedBrgemmOp : public OpRewritePattern<tpp::FusedBrgemmOp> {
...
+    // is fused.
+    // TODO: remove the split once LIBXSMM is fixed.
+    auto isBiasAdd = brgemmOp.getBinaryKind() == tpp::FusedBinaryOpKind::ADD;
+    auto binaryFlag = getBinaryFlags(rewriter, brgemmOp)[0]
+                          .cast<xsmm::BinaryFlagsAttr>()
+                          .getValue();
+    auto isBitSet = static_cast<uint64_t>(binaryFlag) &
+                    static_cast<uint64_t>(xsmm::BinaryFlags::BCAST_COL_IN_0);
+    if (isBiasAdd && !isBitSet) {
+      tpp::utils::splitAndReplaceFusedOp(brgemmOp, rewriter);
```
rengolin, line 302: you can make `splitAndReplaceFusedOp` return `LogicalResult` and just:
```
return tpp::utils::splitAndReplaceFusedOp(brgemmOp, rewriter);
```

lib/TPP/Dialect/Tpp/TppUtils.cpp
```diff
@@ -272,6 +272,40 @@ bool isTppZero(linalg::GenericOp linalgOp, SmallVectorImpl<Value> *operands) {
   return true;
 }
 
+void splitAndReplaceFusedOp(tpp::FusedBrgemmOp fusedBrgemmOp,
+                            PatternRewriter &rewriter) {
+  assert(fusedBrgemmOp.hasBufferSemantics() &&
```
rengolin, line 277: This should be a `return failure();` as we may have this pattern as we move to tensors.

rengolin (approved):

Issue #586, opened 2023-06-12
TPP bufferization does not deallocate


adam-smnk:
Discovered while going through memory leaks mentioned by sanitizers in #580.
The example snippet is extracted from model generated with:
`mlir-gen --kernel=mlp --seed=123 --mini-batch=10 --layers=10,10,10 --bias-acc --softmax`

It appears that TPP bufferization does not handle correctly and/or prevents automatic deallocation of allocated memrefs during bufferization.
A minimal reproduction snippet:
```
#map3 = affine_map<(d0, d1) -> (d0, d1)>
#map4 = affine_map<(d0, d1) -> (d0, 0)>

func.func @entry(%arg0: tensor<10x10xf32>, %arg1: tensor<10x1xf32>, %arg2: tensor<10x10xf32>) -> tensor<10x10xf32> {
    %10 = tensor.empty() : tensor<10x10xf32>
    %11 = linalg.generic {indexing_maps = [#map4, #map3], iterator_types = ["parallel", "parallel"]} ins(%arg1 : tensor<10x1xf32>) outs(%10 : tensor<10x10xf32>) {
    ^bb0(%in: f32, %out: f32):
      linalg.yield %in : f32
    } -> tensor<10x10xf32>
    %12 = linalg.generic {indexing_maps = [#map3, #map3, #map3], iterator_types = ["parallel", "parallel"]} ins(%arg0, %11 : tensor<10x10xf32>, tensor<10x10xf32>) outs(%arg2 : tensor<10x10xf32>) {
    ^bb0(%in: f32, %in_5: f32, %out: f32):
      %13 = arith.divf %in, %in_5 : f32
      linalg.yield %13 : f32
    } -> tensor<10x10xf32>
    return %12 : tensor<10x10xf32>
}
```

Lowering linalg to loops - `-convert-linalg-to-loops -cleanup -bufferize -cleanup:
```
#map = affine_map<(d0, d1) -> (d0, 0)>
#map1 = affine_map<(d0, d1) -> (d0, d1)>
module {
  func.func @entry(%arg0: memref<10x10xf32>, %arg1: memref<10x1xf32>, %arg2: memref<10x10xf32>) {
    %alloc = memref.alloc() {alignment = 64 : i64} : memref<10x10xf32>
    linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "parallel"]} ins(%arg1 : memref<10x1xf32>) outs(%alloc : memref<10x10xf32>) {
[...]

PR #587, opened 2023-06-13
[TPP] Implement `bufferizesToAllocation` method for unary and binary

chelini:
When bufferizing out-of-place unary and binary operations, `allocateTensorForShapedValue` always escaped the allocated tensor precluding bufferization from injecting proper deallocation operations. Implement a `bufferizesToAllocation` method interface that will tell us during the `bufferize` method if an OpResult will escape.

Fix: #586

### DIFF
--- a/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
+++ b/lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
@@ -46,9 +46,12 @@ static LogicalResult bufferizeUnaryOp(Operation *op, RewriterBase &rewriter,
     return failure();
   // Out-of-place bufferization.
   if (unaryOp.getInputs()[0].getType() != unaryOp.getResultType()) {
-    FailureOr<Value> alloc =
-        allocateTensorForShapedValue(rewriter, loc, unaryOp.getResult(0),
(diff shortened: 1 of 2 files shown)


lib/TPP/Dialect/Tpp/BufferizableOpInterfaceImpl.cpp
```diff
@@ -290,6 +317,16 @@ getAliasingOpResultsBinaryImpl(Operation *op, OpOperand &opOperand,
...
 }
 
+// Return true if the opResult bufferize out of place. Binary operations
+// bufferize out of place when the type of the result does not match the type of
+// any of the inputs.
+static bool bufferizesToAllocationBinaryImpl(Operation *op, OpResult opResult) {
+  auto binaryOp = cast<tpp::TppOp>(op);
+  assert(binaryOp && binaryOp.isBinary());
+  return binaryOp.getInputs()[0].getType() != opResult.getType() ||
+         binaryOp.getInputs()[1].getType() != opResult.getType();
```
adam-smnk, line 327: Shouldn't this be `&&`?

chelini: yeap.

PR #588, opened 2023-06-13
[TPP] Fix in-place bufferization

chelini:
To bufferize in place a binary operation where the lhs and rhs are the same operands, we were ignoring a conflict between the operand, but we can ignore this conflict only if the operand is dead after the current operation. If we do not ignore the conflict bufferization sees a read/write effect on the same operand and introduces copies.

### DIFF
--- a/test/Dialect/Tpp/tpp-bufferization.mlir
+++ b/test/Dialect/Tpp/tpp-bufferization.mlir
@@ -737,3 +737,45 @@ func.func @allocation_should_escape_binary(
   // CHECK: return %[[ALLOC]] : memref<10x10xf32>
   return %1 : tensor<10x10xf32>
 }
+
+// -----
+
+func.func @conflicting_in_place(%arg0: tensor<3x3xf32>) -> tensor<3x3xf32> {
+  %0 = tpp.add(%arg0: tensor<3x3xf32>, %arg0: tensor<3x3xf32>) -> tensor<3x3xf32>
+  %1 = tpp.add(%0: tensor<3x3xf32>, %arg0: tensor<3x3xf32>) -> tensor<3x3xf32>
+  return %1 : tensor<3x3xf32>
+}
+
+// CHECK-LABEL: conflicting_in_place
+// CHECK-SAME: %[[ARG0:.+]]: memref<3x3xf32>
+// CHECK: %[[ALLOC:.+]] = memref.alloc() {{.+}} : memref<3x3xf32>
+// CHECK-NEXT: memref.copy %[[ARG0]], %[[ALLOC]] : memref<3x3xf32> to memref<3x3xf32>
+// CHECK-NEXT: tpp.add ins(%[[ARG0]] : memref<3x3xf32>, %[[ALLOC]] : memref<3x3xf32>) outs(%[[ALLOC]] : memref<3x3xf32>)
+// CHECK-NEXT: tpp.add ins(%[[ALLOC]] : memref<3x3xf32>, %[[ARG0]] : memref<3x3xf32>) outs(%[[ARG0]] : memref<3x3xf32>)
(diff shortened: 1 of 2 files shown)


test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -590,3 +590,35 @@ func.func @fused_brgemm_op(%arg0: tensor<1x3x3xf32>, %arg1: tensor<1x3x3xf32>,
...
+func.func @conflict_in_place(%arg0: tensor<3x3xf32>) -> tensor<3x3xf32> {
+  %0 = tpp.add(%arg0: tensor<3x3xf32>, %arg0: tensor<3x3xf32>) -> tensor<3x3xf32>
+  %1 = tpp.add(%0: tensor<3x3xf32>, %arg0: tensor<3x3xf32>) -> tensor<3x3xf32>
+  return %1 : tensor<3x3xf32>
+}
+
+// CHECK-LABEL: conflict_in_place
+// CHECK-SAME: %[[ARG:.+]]: memref<3x3xf32>
+// CHECK: %[[ALLOC:.+]] = memref.alloc() {{.+}} : memref<3x3xf32>
+// CHECK-NEXT: memref.copy %[[ARG0]], %[[ALLOC]] : memref<3x3xf32> to memref<3x3xf32>
```
rengolin, line 605: Careful, this test defines `ARG` not `ARG0`, but you check for `ARG0` below, which was defined in the previous tests.

test/Dialect/Tpp/tpp-bufferization.mlir
```diff
@@ -590,3 +590,35 @@ func.func @fused_brgemm_op(%arg0: tensor<1x3x3xf32>, %arg1: tensor<1x3x3xf32>,
...
+
+func.func @conflict_in_place(%arg0: tensor<3x3xf32>) -> tensor<3x3xf32> {
+  %0 = tpp.add(%arg0: tensor<3x3xf32>, %arg0: tensor<3x3xf32>) -> tensor<3x3xf32>
+  %1 = tpp.add(%0: tensor<3x3xf32>, %arg0: tensor<3x3xf32>) -> tensor<3x3xf32>
+  return %1 : tensor<3x3xf32>
+}
+
+// CHECK-LABEL: conflict_in_place
+// CHECK-SAME: %[[ARG:.+]]: memref<3x3xf32>
+// CHECK: %[[ALLOC:.+]] = memref.alloc() {{.+}} : memref<3x3xf32>
```
rengolin, line 604: Is this extra copy because you use `%arg0` twice?

Can you have a test that uses a local constant to make sure there's no extra alloc?

rengolin (commented):
Code looks sensible, just have some questions on the tests, perhaps understand the wider impact of this on some benchmarks. Do we have any existing benchmark that can show a delta with this?

test/Dialect/Tpp/tpp-bufferization.mlir
chelini: fixed.

test/Dialect/Tpp/tpp-bufferization.mlir
chelini: Yes, the second ops uses %arg0, so %arg0 is not writable for the first op.

chelini:
> 

No, looks like the benchmarks do not have an in-place tpp.

adam-smnk:
> > 
> 
> No, looks like the benchmarks do not have an in-place tpp.

AFAIK, there are individual in-place tpps but not with this read-write pattern.

chelini:
> > 
> 
> No, looks like the benchmarks do not have an in-place tpp.

...(with conflicting read-write patterns).

rengolin (approved):

PR #590, opened 2023-06-14
Enable rewrite-to-brgemm to preserve parallelism at tensor level

chelini:
At tensor level `linalg::GenerateLoopNest` emits only scf.for loop even if the loops have parallel semantics. This changes enable to rewrite parallel scf.for to scf.forall in the same way we did for the tile and fuse pass. The PRs mainly move code around to make this possible exposing `populateScfForToForAllRewritePattern` as utils in `TransformUtils.h`. Simplify logic by retiring the tail mark attribute.

### DIFF
--- a/include/TPP/TransformUtils.h
+++ b/include/TPP/TransformUtils.h
@@ -78,6 +78,12 @@ bool isMatmulOp(Operation *op,
 bool hasMulAddBody(linalg::LinalgOp linalgOp,
                    SmallVectorImpl<Value> *capturedOperands = nullptr);
 
+// Rewrite scf.for to scf.forall. Assumes the loop to be parallel and
+// marked with `kLoopId`.
+constexpr const static llvm::StringLiteral kLoopId = "parallel";
+constexpr const static llvm::StringLiteral kLoopRoot = "root";
+void populateScfForToForAllRewritePattern(RewritePatternSet &patterns);
+
 } // namespace utils
 } // namespace linalgx
 } // namespace mlir
--- a/include/TPP/TransformUtils.h
+++ b/include/TPP/TransformUtils.h
@@ -80,7 +80,7 @@ bool hasMulAddBody(linalg::LinalgOp linalgOp,
 
 // Rewrite scf.for to scf.forall. Assumes the loop to be parallel and
 // marked with `kLoopId`.
-constexpr const static llvm::StringLiteral kLoopId = "parallel";
+constexpr const static llvm::StringLiteral kLoopParallel = "parallel";
 constexpr const static llvm::StringLiteral kLoopRoot = "root";
 void populateScfForToForAllRewritePattern(RewritePatternSet &patterns);
 
--- a/lib/TPP/TestForToForAllRewrite.cpp
+++ b/lib/TPP/TestForToForAllRewrite.cpp
@@ -1,0 +1,43 @@
+//===- TestForToForAllRewrite.cpp - ---------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/TransformUtils.h"
+#include "mlir/IR/FunctionInterfaces.h"
+#include "mlir/Pass/Pass.h"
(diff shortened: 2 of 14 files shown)


include/TPP/TransformUtils.h
```diff
@@ -78,6 +78,12 @@ bool isMatmulOp(Operation *op,
 bool hasMulAddBody(linalg::LinalgOp linalgOp,
                    SmallVectorImpl<Value> *capturedOperands = nullptr);
 
+// Rewrite scf.for to scf.forall. Assumes the loop to be parallel and
+// marked with `kLoopId`.
+constexpr const static llvm::StringLiteral kLoopId = "parallel";
```
rengolin, line 83: I'd name this `kLoopParallel` to make it clear throughout the code without having to look at the declaration.

lib/TPP/TestForToForAllRewrite.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "TPP/TransformUtils.h"
+#include "mlir/IR/FunctionInterfaces.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+namespace {
+// This is a test pass to check the scf.for to scf.forall rewrite.
+struct TestForToForAllRewrite
```
rengolin, line 18: I'm not sure I get the point of this test pass...

chelini: It tests the rewrite in isolation, similar to what we did for the matchers.

rengolin: The matchers do something, this one just runs greedly. Wouldn't it be the same as running `tpp-opt` with the pass name?

test/Passes/tile-and-fuse-depth.mlir
```diff
@@ -1,7 +1,7 @@
-// RUN: tpp-opt %s -tile-consumer-and-fuse-producers="tile-sizes=1,0 max-depth=0" | FileCheck -check-prefix=DEPTH0 %s
-// RUN: tpp-opt %s -tile-consumer-and-fuse-producers="tile-sizes=1,0 max-depth=1" | FileCheck -check-prefix=DEPTH1 %s
-// RUN: tpp-opt %s -tile-consumer-and-fuse-producers="tile-sizes=1,0 max-depth=2" | FileCheck -check-prefix=DEPTH2 %s
-// RUN: tpp-opt %s -tile-consumer-and-fuse-producers="tile-sizes=1,0 max-depth=3" | FileCheck -check-prefix=DEPTH3 %s
+// RUN: tpp-opt %s -tile-consumer-and-fuse-producers="tile-sizes=1,0 max-depth=0 use-for-all=false" | FileCheck -check-prefix=DEPTH0 %s
```
rengolin, line 1: Why disable it here?

rengolin:
This change doesn't seem to change any test in the default pipeline / integration tests. We're either not thorough enough there or we're not checking for the `scf.for` before in there. We should.

adam-smnk:
> This change doesn't seem to change any test in the default pipeline

Correct, the rewrite occurs during `tpp-conversion` which happens only at the memref level.
This PR addresses performance regression observed during switch to TPP on tensors.

lib/TPP/TestForToForAllRewrite.cpp
chelini: It is a pass in tpp-opt, but it is not exposed as we do with the other passes. I chose this because I am only interested in exposing the rewrite pattern that other passes like tile-and-fuse or rewrite-to-brgemm can use. I don't see this as a fully standalone pass, and we don't need to add another pass to the default pipeline. Anyway, I am open to suggestions/improvements.

chelini:
> This change doesn't seem to change any test in the default pipeline / integration tests. We're either not thorough enough there or we're not checking for the `scf.for` before in there. We should.

As @adam-smnk mentioned this PR has not effect on the current pass pipeline as rewrite-to-brgemm happens after bufferization.

lib/TPP/TestForToForAllRewrite.cpp
rengolin: I see. I'm not a fan of this approach, but I don't have a better idea right now. We should think about how we want to continue here to avoid the boilerplate this adds to both code and tests. For now, this is fine.

rengolin (approved):

PR #593, opened 2023-06-15
Fix benchmarks flops

adam-smnk:
Fixes tensor sizes and flops calculations in a benchmark kernels.

### DIFF
--- a/benchmarks/config/base/base.json
+++ b/benchmarks/config/base/base.json
@@ -64,7 +64,7 @@
       "type": "MLIR",
       "benchmark": "mlp-bf16-10layers-3584.mlir",
       "environment": {},
-      "flags": [ "-n", "100", "-run-args='-def-pack=0'" ],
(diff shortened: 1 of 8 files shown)


chelini (approved):

rengolin:
hmm
image

adam-smnk:
MLP base kernel sizes were fine but the total numbers of flops were incorrect.

adam-smnk:
image
Small drop in unpacked gemm performance as I reenabled packing. It doesn't particularly change overall test runtime but it was the only remaining test case with packing explicitly disabled. Thus, it is better to just run the default pipeline without extra options.

image
image
After the mlp flops fix, the two gemm and mlp kernels, using the same tensor sizes and both prepacked, achieve similar performance.

rengolin (approved):
Stupid copy pasta! :face_exhaling:

Issue #354, opened 2023-03-01
Lower `linalg.batch_matmul` to TPP (for MHA)


rengolin:
As exposed in #342, `linalg.batch_matmul` isn't being lowered to TPP. We probably never had the need before, but this should be a simple case.

As we ramp up the MHA priority, this will be essential to get good performance there.

@chelini @KavithaTipturMadhu @adam-smnk

chelini:
There is already a pattern: https://github.com/plaidml/tpp-mlir/blob/76e54724da4dd5e2fce5ec2b3cde357825558a14/lib/TPP/ConvertLinalgToTpp.cpp#L336 Do you know why is not converting?

rengolin:
From #342, these are the ops:
```
    %collapsed_25 = tensor.collapse_shape %100 [[0, 1], [2], [3]] : tensor<32x2x8x64xf32> into tensor<64x8x64xf32>
    %collapsed_26 = tensor.collapse_shape %102 [[0, 1], [2], [3]] : tensor<32x2x64x8xf32> into tensor<64x64x8xf32>
    %103 = tensor.empty() : tensor<64x8x8xf32>
    %104 = linalg.fill ins(%cst_1 : f32) outs(%103 : tensor<64x8x8xf32>) -> tensor<64x8x8xf32>
    %105 = linalg.batch_matmul ins(%collapsed_25, %collapsed_26 : tensor<64x8x64xf32>, tensor<64x64x8xf32>) outs(%104 : tensor<64x8x8xf32>) -> tensor<64x8x8xf32>
```

and

```
    %collapsed_32 = tensor.collapse_shape %127 [[0, 1], [2], [3]] : tensor<32x2x8x8xf32> into tensor<64x8x8xf32>
    %collapsed_33 = tensor.collapse_shape %136 [[0, 1], [2], [3]] : tensor<32x2x8x64xf32> into tensor<64x8x64xf32>
    %137 = tensor.empty() : tensor<64x8x64xf32>
    %138 = linalg.fill ins(%cst_1 : f32) outs(%137 : tensor<64x8x64xf32>) -> tensor<64x8x64xf32>
    %139 = linalg.batch_matmul ins(%collapsed_32, %collapsed_33 : tensor<64x8x8xf32>, tensor<64x8x64xf32>) outs(%138 : tensor<64x8x64xf32>) -> tensor<64x8x64xf32>
```

chelini:
[...]

PR #595, opened 2023-06-16
Improve `packVNNIMatmulOp`

chelini:
`packVNNIMatmulOp` was using hard-coded checks to pack a generic matmul operation. The "if" branch was VNNI-blocking an already blocked matmul, while the "else" branch was a "tiled" version. With a closer look at the "else" branch, it turned out that the generic is a simple brgemm operation, as a tile-blocked matmul is a brgemm. Thus remove the "else" branch and use a de-generalization pattern to get a named op. At this point, `PackVNNIBrgemm` can do the job of the "else" branch.

### DIFF
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -412,7 +412,8 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
 //===----------------------------------------------------------------------===//
 // MatmulOp (VNNI packing)
 //===----------------------------------------------------------------------===//
-// Original layout: [IB][JB][ib][jb] += [IB][KB][ib][kb] * [JB][KB][kb][jb]
+// Original layout:
+//      [IB][JB][ib][jb] += [IB][KB][ib][kb] * [JB][KB][kb][jb]
 // New      layout:
 //      [IB][JB][ib][jb] += [IB][KB][ib][kb] * [JB][KB][kb/VNNI][jb][VNNI]
 FailureOr<linalg::GenericOp>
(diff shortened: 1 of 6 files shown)


chelini:
Stumble on this when trying to make progress for: https://github.com/plaidml/tpp-mlir/issues/354

rengolin:
@KavithaTipturMadhu

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -423,6 +409,12 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
   return packMatmulOpImpl(rewriter, matmulOp, tiles);
 }
 
+//===----------------------------------------------------------------------===//
+// MatmulOp (VNNI packing)
+//===----------------------------------------------------------------------===//
+// Original layout: [IB][JB][ib][jb] += [IB][KB][ib][kb] * [JB][KB][kb][jb]
```
rengolin, line 415: Nit: can you try to align the two layouts for easier visual comparison? Maybe line break both?

rengolin (approved):
IIUC, we had two types of matmuls (one of them batch reduce) and this patch converts the first one to brgemm and thus only needs one of the patterns (rank 4-7)? If we run the brgemm conversion first, we can pack each with their own functions. I like this approach.

We're slowly converging to a combined approach and this is great, thanks!

LGTM with the small formatting nit on the comment.

chelini:
> IIUC, we had two types of matmuls (one of them batch reduce) and this patch converts the first one to brgemm and thus only needs one of the patterns (rank 4-7)? If we run the brgemm conversion first, we can pack each with their own functions. I like this approach.
> 
> We're slowly converging to a combined approach and this is great, thanks!
> 
> LGTM with the small formatting nit on the comment.

Yes, after tiling a blocked matmul is just a brgemm, so we de-generalize and apply the brgemm blocking.

PR #599, opened 2023-06-19
Default pipeline - use TPP on tensors

### LINKED ISSUE #580: Move default pipeline to TPP on tensors

adam-smnk:
Modifies the default pipeline to convert ops to TPP at the tensor instead of memref level.

The switch ensures that TPP on tensors and the custom bufferization is used by default.
The existing tests are adjusted to match the new compilation flow and to prevent memory leaks e.g., by avoiding returning buffers from helper functions.

Fixes #580


adam-smnk:
image


There is single regression on the `gemm_bf16_dp2_3x1024_omp_16_mlir` test case (grey line) on TPP-benchmark.
Initially the change was attributed to slight difference in generated IR as packing can change shape of constant buffers (weights, biases etc.) as they are represented by splat tensors. This is addressed by c550f10 which enables random initialization for this kernel. This generated IR appears to be now identical compared to the main branch.

This change addresses the performance different a bit but the regression is still present:
image

However, the same benchmark shows no performance difference in TPP-performance (blue line):
image

rengolin (approved):
Looks good, thanks!

PR #601, opened 2023-06-20
clang-format - do not error on in-place formatting

adam-smnk:
Avoids returning errors when in-place code formatting is performed.

### DIFF
--- a/scripts/ci/clang-format.sh
+++ b/scripts/ci/clang-format.sh
@@ -25,9 +25,10 @@ _clang_format() {
       if [ $INPLACE ]; then
         ${CLANG_FORMAT} -i ${FILE}
         echo "File ${FILE} updated"
+      else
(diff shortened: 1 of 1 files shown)


scripts/ci/clang-format.sh
```diff
@@ -25,9 +25,10 @@ _clang_format() {
       if [ $INPLACE ]; then
         ${CLANG_FORMAT} -i ${FILE}
         echo "File ${FILE} updated"
+      else
```
rengolin, line 28: This will make the "all clear" message confusing. How about adding a new counter, say, `CHANGED` and update when a file is reformatted, and then have a third message at the end: "${CHANGED} files were reformatted."?

adam-smnk: Good point, I like that idea

rengolin (approved):

PR #602, opened 2023-06-21
Benchmarks - randomize splats by default

### LINKED ISSUE #515: Make all of our tests / benchmarks use `splat-to-random`
We have the option and yet not one of our tests/benchmarks use it (other than the test for this feature).

We should move to use them more broadly, and more importantly, on benchmarks.

adam-smnk:
Enables splat randomization by default in the benchmarks.
Tensor random initializers now produce continuous random data upon multiple requests for the same splat data type instead of restarting the generated number sequence each time.
Also, incorrect scalar tensor initialization is fixed.

This approach aims to improve benchmarking by providing more realistic workloads containing varied inputs and model constants (weighs, biases etc.). Moreover, randomizing splats prevents unrealistic optimizations such as dimension folding caused by compile time packing.
Continuous random data generation prevent unexpected CSE and helps to measure more realistic memory consumption.

Fixes #515

### DIFF
--- a/benchmarks/harness/controller.py
+++ b/benchmarks/harness/controller.py
@@ -137,9 +137,10 @@ class BenchmarkController(object):
                                              '-e', self.args.entry,
                                              '--entry-point-result=void',
                                              '--print=0',
-                                             '--seed=123',
-                                             '--splat-to-random',
+                                             '--seed', str(self.args.seed),
                   ]
+        if self.args.splat_to_random:
+            runCmd.append('--splat-to-random')
         if self.args.run_args:
             runCmd.extend(shlex.split(self.args.run_args))
         runResult = executor.run(runCmd, irContents)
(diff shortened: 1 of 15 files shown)


benchmarks/harness/controller.py
```diff
@@ -210,6 +213,10 @@ def verifyStats(self):
                         help='Disable LSAN')
     parser.add_argument('--build', type=str, default="",
                         help='Path to the build dir')
+    parser.add_argument('--seed', type=int, default=123,
```
rengolin, line 216: This should be disabled. We want truly random behaviour to be the default if the seed isn't specified. We also don't want random behaviour to be the default via seeds.

`tpp-run` will pick `time` as the seed if any random option is chosen and `seed` isn't passed.

benchmarks/harness/controller.py
```diff
@@ -210,6 +213,10 @@ def verifyStats(self):
...
+                        help='Random seed (default: enabled)')
+    parser.add_argument('--splat-to-random', type=int, default=1,
```
rengolin, line 218: `tpp-run` has some options (`random`, `normal`, etc) and `normal` is the preferred here. Maybe this should be the same list, with default `normal`.

include/TPP/TensorInit.h
```diff
@@ -89,7 +90,7 @@ enum class TensorInitType {
 };
 
 // Unique pointer for tensor init to help with memory management
-using TensorInitPtr = std::unique_ptr<ITensorInit>;
+using TensorInitPtr = std::shared_ptr<ITensorInit>;
```
rengolin, line 93: Also change the comment

rengolin (approved):

PR #604, opened 2023-06-21
Finally fix clang-format path issue

rengolin:


hfp (approved):
Does the "../.." imply any non-local knowledge, i.e., something not done int *this* script?

hfp:
> Does the "../.." imply any non-local knowledge, i.e., something not done int _this_ script?

Ahh, got it. It's relative to `$0`.

PR #606, opened 2023-06-21
Rewrite `linalg.batch_matmul` to `linalg.matmul`

chelini:
Emit parallel loop for the batch dimension.

Work on: https://github.com/plaidml/tpp-mlir/issues/354


rengolin:
> Nice and clean approach Looks good to me

Looks good to me too, let's just make sure this is the right approach.

rengolin (approved):
libxsmm has nothing special, LGTM. Thanks!

PR #608, opened 2023-06-22
Add blocking for `linalg.batch_matmul`

chelini:
Block a `linalg.batch_matmul` in the same way we do for a `linalg.matmul`. The batch dimension is left as outermost.

### DIFF
--- a/include/TPP/Transforms.h
+++ b/include/TPP/Transforms.h
@@ -24,6 +24,7 @@ class Conv2DNchwFchwOp;
 class Conv2DNhwcHwcfOp;
 class MatmulOp;
 class BatchReduceMatmulOp;
+class BatchMatmulOp;
 } // namespace linalg
 
 namespace tpp {
@@ -54,9 +55,12 @@ FailureOr<linalg::GenericOp>
 packConv2DNhwcHwcfOp(RewriterBase &rewriter, linalg::Conv2DNhwcHwcfOp linalgOp,
                      ArrayRef<OpFoldResult> tiles);
 
-// Attempt to block a MatmulOp.
+// Attempt to block a MatmulOp or a BatchMatmulOp.
 FailureOr<linalg::GenericOp> packMatmulOp(RewriterBase &rewriter,
                                           linalg::MatmulOp linalgOp,
+                                          ArrayRef<OpFoldResult> tiles);
+FailureOr<linalg::GenericOp> packMatmulOp(RewriterBase &rewriter,
+                                          linalg::BatchMatmulOp linalgOp,
                                           ArrayRef<OpFoldResult> tiles);
 
 // Attempt to block a MatmulOp to VNNI format.
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -65,10 +65,16 @@ static Value toUnPackLayoutImpl(OpBuilder &builder, Location loc, Value input,
                                           tiles, outerDimsPerm);
 }
 
-static Value handleLayoutNC_NCnc(OpBuilder &builder, Location loc, Value input,
-                                 Value output, ArrayRef<OpFoldResult> tiles) {
+static Value handleLayoutNC_NCnc(OpBuilder &builder, linalg::LinalgOp matmulOp,
+                                 Value input, Value output,
+                                 ArrayRef<OpFoldResult> tiles) {
+  assert(isa<linalg::MatmulOp>(matmulOp) ||
+         isa<linalg::BatchMatmulOp>(matmulOp));
   assert(tiles.size() == 2 && "expect two tile sizes for NC_NCnc");
+  Location loc = matmulOp.getLoc();
   SmallVector<int64_t> innerDimPos = {0, 1};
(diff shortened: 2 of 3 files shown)


include/TPP/Transforms.h
```diff
@@ -59,6 +60,11 @@ FailureOr<linalg::GenericOp> packMatmulOp(RewriterBase &rewriter,
                                           linalg::MatmulOp linalgOp,
                                           ArrayRef<OpFoldResult> tiles);
 
+// Attempt to block a BatchMatmulOp.
```
rengolin, line 63: The name implies `Matmul`, the comment implies `BatchMatmul`.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -68,7 +68,10 @@ static Value toUnPackLayoutImpl(OpBuilder &builder, Location loc, Value input,
 static Value handleLayoutNC_NCnc(OpBuilder &builder, Location loc, Value input,
                                  Value output, ArrayRef<OpFoldResult> tiles) {
   assert(tiles.size() == 2 && "expect two tile sizes for NC_NCnc");
+  int64_t rank = input.getType().cast<RankedTensorType>().getRank();
   SmallVector<int64_t> innerDimPos = {0, 1};
+  if (rank == 3 || rank == 5)
```
rengolin, line 73: Magic numbers. What does this mean?

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -337,9 +388,13 @@ packMatmulOpImpl(RewriterBase &rewriter, linalg::MatmulOp matmulOp,
   OpFoldResult tileOnI = tiles[0];
   OpFoldResult tileOnJ = tiles[1];
   OpFoldResult tileOnK = tiles[2];
+  bool isMatmulOp = std::is_same_v<OpTy, linalg::MatmulOp>;
```
rengolin, line 391: This does not make clear what else it could be.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -324,9 +333,51 @@ mlir::linalgx::packConv2DNchwFchwOp(RewriterBase &rewriter,
   return packConvolutions(rewriter, convOp, tiles);
 }
 
+template <typename OpTy>
+static std::pair<SmallVector<AffineMap>, SmallVector<utils::IteratorType>>
+getAffineMapsAndIteratorsForBlockedMatmul(OpTy matmulOp) {
```
rengolin, line 338: This isn't just for blocked matmuls.

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: I have a static assert few lines above to make sure we are dealing with matmul or batch matmul. Any suggestion on how to improve?

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: `getAffineMapsAndIteratorsForBlockedMatmul` -> `getAffineMapsAndIteratorsForBlockedMatmulOrBatchMatmul`?

include/TPP/Transforms.h
chelini: yes, it is an overloading to share the same `PackMatmulImpl` driver. Should we call it `packMatmulAndBatchMatmulOp`?

lib/TPP/ToBlockLayoutAndBack.cpp
chelini: yes, not really great. Here we need to deal with batchMatmul. For now we have this layout only for matmul and batch matmul, the if stmt sets the right inner dims pos and outer dim perm for the batch matmul.

lib/TPP/ToBlockLayoutAndBack.cpp
```diff
@@ -65,10 +65,16 @@ static Value toUnPackLayoutImpl(OpBuilder &builder, Location loc, Value input,
...
+static Value handleLayoutNC_NCnc(OpBuilder &builder, linalg::LinalgOp matmulOp,
+                                 Value input, Value output,
+                                 ArrayRef<OpFoldResult> tiles) {
+  assert(isa<linalg::MatmulOp>(matmulOp) ||
+         isa<linalg::BatchMatmulOp>(matmulOp));
   assert(tiles.size() == 2 && "expect two tile sizes for NC_NCnc");
+  Location loc = matmulOp.getLoc();
   SmallVector<int64_t> innerDimPos = {0, 1};
+  if (isa<linalg::BatchMatmulOp>(matmulOp))
+    innerDimPos = {1, 2};
```
rengolin, line 77: The `output` check looks redundant now. Probably true on all other similar methods too.

chelini: output is used to understand if we need to pack or unpack. The check looks still valid to me.

rengolin: Ignore me, I apparently couldn't read on Friday.

rengolin (approved):

PR #607, opened 2023-06-22
bringup tests on intel gpu

hfp:
Leverage Vulkan based runtime and bring-up initial CI. This work includes changes for tpp-scripts (not tracked here).

PR #609, opened 2023-06-22
Bringup tests on intel gpu

hfp:
Leverage Vulkan based runtime and bring-up initial CI. This work includes changes for tpp-scripts (not tracked here).

### DIFF
--- a/scripts/buildkite/tpp-gpu.yml
+++ b/scripts/buildkite/tpp-gpu.yml
@@ -31,7 +31,6 @@ steps:
               'KIND=Debug COMPILER=clang LINKER=lld GPU=1 CHECK=1 \
               $${BUILDKITE_BUILD_CHECKOUT_PATH}/scripts/buildkite/build_tpp.sh'"
     env:
-      # TPP_ENV_REPLACE implies TPP_ENV_EXPORT=1
-      TPP_ENV_REPLACE: "/data/nfs_home/buildkite-slurm:/Users/buildkite"
       TPP_LAUNCH_NODE: "pcl-adl01"
       TPP_PROXY_ENABLE: "1"
+      TPP_ENV_EXPORT: "1"
--- a/scripts/buildkite/tpp-gpu.yml
+++ b/scripts/buildkite/tpp-gpu.yml
@@ -32,4 +32,5 @@ steps:
               $${BUILDKITE_BUILD_CHECKOUT_PATH}/scripts/buildkite/build_tpp.sh'"
     env:
       TPP_LAUNCH_NODE: "pcl-adl01"
+      TPP_PROXY_ENABLE: "1"
       TPP_ENV_STRICT: "1"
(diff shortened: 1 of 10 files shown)


hfp:
See also #607.

hfp:
Status: remote execution on target system works now! Remaining work is about implementing a recipe and perhaps rename or unify settings like `GPU=1` (currently CUDA specific).

hfp:
There is also some add-on/side-effect for this PR, the following works now (blank cluster environment):

```bash
source /data/nfs_home/buildkite-slurm/builds/tpp/enable-tpp
cd tpp-mlir
KIND=Debug COMPILER=clang LINKER=lld CHECK=1 GPU=vulkan CLEAN=1  scripts/buildkite/build_tpp.sh
```

This means local tests can run like in CI without the CI-environment to be present.

scripts/buildkite/tpp-gpu.yml
```diff
@@ -8,19 +8,29 @@ env:
...
+              scripts/buildkite/build_tpp.sh'"
+    env:
+      # ASAN flags required due to incompatibility with CUDA
+      # See: https://github.com/google/sanitizers/issues/629
+      ASAN_OPTIONS: "protect_shadow_gap=0:replace_intrin=0:detect_leaks=0:${ASAN_OPTIONS}"
+      TPP_CLANGVER: "16.0.4"
+      TPP_GCCVER: "9.5.0" # Use an older version to match CUDA requirements
+      TPP_CUDAVER: "11.5"
+
+  - label: "TPP-MLIR-intel"
```
rengolin, line 29: Why don't we have forced tools for Intel GPU?

scripts/ci/cmake.sh
```diff
@@ -6,119 +6,128 @@
...
+  echo "  -n: Optional link job flag, defaults to nproc"
   exit 1
 }
 
 # Cmd-line opts
-while getopts "s:b:i:m:t:c:g:l:n:SG" arg; do
-  case ${arg} in
-    s)
-      SRC_DIR=$(realpath ${OPTARG})
+while test $# -gt 0; do
```
rengolin, line 26: Why the move from `getopts`?

scripts/buildkite/tpp-gpu.yml
hfp: Not yet. We can add later. ATM, nothing is actually executed on Intel GPU, and further development (incl. source level) is needed. For example, we may need IMEX to be fiddled into our LLVM, etc. I can also enforce some tools if your prefer.

scripts/ci/cmake.sh
hfp: To support the previous command line interface like `-G` without arguments. Now, `-G` can also take `cuda` or `vulkan` (or anything else we may want later). This could not be expressed with `getopts` like checking if the "next" argument is a flag or a value.

hfp: Ok, I will bring back `getopt`. I thought people are using these scripts and wanted to provide "no change".

rengolin (approved):
Thanks!

Issue #600, opened 2023-06-19
Deprecate TPP conversion on memrefs


adam-smnk:
Once the move to TPP on tensors (#580) is complete, the existing infrastructure that supports conversion to TPP on memrefs should be deprecated. Mapping to TPP ops, and further transformations and optimizations will be taking place primarily on tensor semantics.

TODO:
- make `tpp-conversion` work only on ops with tensor semantics
- add paths to fail gracefully when called on non-tensor ops
- update tests accordingly

PR #611, opened 2023-06-22
Avoid in-place bufferization for tpp when invalid.

chelini:
We cannot bufferize in-place if the operation and the definition of the operand where we want to bufferize are not in the same repetitive region.

Work toward: https://github.com/plaidml/tpp-mlir/issues/600


chelini:
discussed offline with @adam-smnk. We will merge once we change upstream bufferization to pass the state.

adam-smnk (approved):
The helper logic we required is luckily stateless so the current workaround addresses TPP bufferization needs.
Let's switch to proper solution when upstream fix+bump is done.

PR #613, opened 2023-06-23
Deprecate tpp memref conversion

### LINKED ISSUE #600: Deprecate TPP conversion on memrefs

adam-smnk:
Constrains lowering to TPP to ops using the tensor abstraction.
Relevant tests are updated and cleaned up accordingly.

Fixes #600 

Co-authored-by: lorenzo chelini <l.chelini@icloud.com>


test/Conversion/LinalgToTpp/linalg-to-tpp-memref-unsupported.mlir
```diff
@@ -6,7 +6,7 @@
 // CHECK-SAME: %[[arg2:.*]]: memref<5x5xf32>) {
 func.func @brgemm_lowering(%arg0: memref<3x5x4xf32>, %arg1: memref<3x4x5xf32>,
                           %arg2: memref<5x5xf32>) {
-  // CHECK: tpp.brgemm ins(%[[arg0]] : memref<3x5x4xf32>, %[[arg1]] : memref<3x4x5xf32>, %[[arg2]] : memref<5x5xf32>) outs(%[[arg2]] : memref<5x5xf32>)
+  // CHECK-NOT: tpp.brgemm ins(%[[arg0]] : memref<3x5x4xf32>, %[[arg1]] : memref<3x4x5xf32>, %[[arg2]] : memref<5x5xf32>) outs(%[[arg2]] : memref<5x5xf32>)
```
rengolin, line 9: IIUC, most of the tests here (non-matching) have been moved to `linalg-to-tpp-on-tensors.mlir` and are not just redundant, but wrong. We don't match at memrefs any more not because the shape is wrong, but because the type is wrong.

I recommend to remove every test here that has been moved to the tensor file and make sure we only leave tests here (one per op) to make sure we don't match "valid shapes" to TPP on memrefs.

chelini (approved):
To me looks good.

test/Passes/mlp-all.mlir
```diff
@@ -1,107 +0,0 @@
-// RUN: tpp-opt %s -bufferize -convert-linalg-to-tpp | FileCheck %s
-
-#map0 = affine_map<(d0, d1) -> (d1)>
```
rengolin, line 3: Is there a multi-layer MLP test on tensors?

rengolin (commented):
I'm assuming the deleted test files already have alternatives on tensor. Is there one for a 4 layer mlp?

test/Passes/mlp-all.mlir
adam-smnk: Hmm looks like there isn't. This one was removed because it was a poorly written kernel and not relevant anymore.
I'll add a separate 3 layer mlp test.

adam-smnk: (They are a few in benchmarks but there should be one in `test` for clarity)

test/Conversion/LinalgToTpp/linalg-to-tpp-memref-unsupported.mlir
adam-smnk: Good idea, cleaned up the memref.
All cases apart from memref stride specific tests were moved to `linalg-to-tpp-tensor` test file.

test/Passes/mlp-all.mlir
adam-smnk: Added new mlp3-3layers test on tensors

rengolin (approved):

PR #615, opened 2023-06-26
Allow `isBlockedMatmul` to match blocked batch matmuls

chelini:
Allow affine expressions to have an (optional) extra batch dimensions that must be parallel and present in A, B and C operands.

Working on https://github.com/plaidml/tpp-mlir/issues/354


adam-smnk (approved):
lgtm

PR #616, opened 2023-06-26
Enable rank-reduction for named ops during fusion

chelini:
Enable rank-reduction for linalg named operations when fusing. Linalg can rank reduce only generic operations, thus generalize all the ops and raise back to named ops after fusion.


chelini:
Converted to draft as it requires an upstream fix.

test/Passes/pass-degeneralize.mlir
```diff
@@ -0,0 +1,61 @@
+// RUN: tpp-opt %s -linalg-generalize-named-ops -linalg-degeneralize-generic-ops -split-input-file | FileCheck %s
+
+// CHECK-LABEL: degeneralize
+func.func @degeneralize(%arg0: tensor<3x3xf32>, %arg1: tensor<3x3xf32>) -> tensor<3x3xf32> {
+  %0 = tensor.empty() : tensor<3x3xf32>
+  %cst = arith.constant 0.0 : f32
+  %1 = linalg.fill ins(%cst : f32) outs(%arg0 : tensor<3x3xf32>) -> tensor<3x3xf32>
+  // CHECK: linalg.fill
```
adam-smnk, line 8: nit: check that ins and outs are the same after the round trip

adam-smnk:
Does any test case show the rank reduction? (I just can't spot it with so many small tweaks)

rengolin (commented):
Awesome! This is exactly the direction we want upstream.

I'm happy when @adam-smnk is happy.

In the future, we may have to create some different detection for odd-shaped matmuls (N-D with reduction in odd-places), but that's just repeating the same pattern with `scf.forall` for the parallel dimensions, which should build on the code from this PR.

adam-smnk (approved):
lgtm

PR #617, opened 2023-06-26
Bump llvm

chelini:
Need for #616  an upstream patch when converting `linalg.map` to `linalg.generic` (see: https://github.com/llvm/llvm-project/commit/09d09fc335ac8255d2c3c80048922ee5c68c06c3).


rengolin (approved):
Thanks!

PR #618, opened 2023-06-27
[TPP] Add tensor builders

### LINKED ISSUE #612: Change TPP builders to accept only inputs in tensor semantics
Right now, TPP operations on tensors need to be passed an `output` tensor, like we do for memrefs, but tensors are new values and we're not using DPS for TPP, so we really need to have two versions of the builders (and check if tensor/memref with output).

Together with #600 should be the last remaining cleanups to claim full support for tensor semantics in TPP.

@chelini @adam-smnk

chelini:
Provide more ergonomics builder for tpp operations at tensor level.  At tensor abstraction there is not output value and only the output type is needed.

```cpp
// memref builder
tppOpBuilderMemRef(..., ValueRange = input(s), Value = output)

// tensor builder
tppOpBuilderTensor(..., ValueRange = input(s), Type = outputType)
```

Fix #612

### DIFF
--- a/include/TPP/Dialect/Tpp/TppOps.td
+++ b/include/TPP/Dialect/Tpp/TppOps.td
@@ -60,7 +60,8 @@ class Tpp_UnaryOp<string mnemonic, list<Trait> traits = []> :
   let skipDefaultBuilders = 1; 
 
   let builders = [
-    OpBuilder<(ins "Value":$input, "Value":$output)>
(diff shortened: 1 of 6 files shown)


rengolin (approved):
Awesome, thanks!

@KavithaTipturMadhu, you should rebase on top of this commit once merged, will help with your current work.

PR #621, opened 2023-06-29
Simplify `tensor.pack` when the source is an empty

chelini:
In upstream packing pass we avoid packing an empty but this is not always possible, for example we can have:

```mlir
%empty = tensor.empty
%other_empty = tensor.empty

%collapse = tensor.collapse(empty)
%pack = tensor.pack %collapse into %other_empty
%gen = linalg.generic ins(...) outs(%pack)

```

`FoldEmptyTensorWithReshapeOp` will fold the empty with the collapse exposing a tensor.pack where the source and destination are two empties.

### DIFF
--- a/lib/TPP/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/ToBlockLayoutAndBack.cpp
@@ -807,6 +807,20 @@ struct PropagatePackUnPack
   }
 };
 
+// TODO: (lorenzo) Upstream in `populateSimplifyTensorPack`.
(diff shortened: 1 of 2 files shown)


adam-smnk:
What does it even mean to pack an empty tensor?

chelini:
> What does it even mean to pack an empty tensor?

Nothing, indeed. But you may end up in cases where you need to remove this pattern.

adam-smnk (approved):

PR #622, opened 2023-06-29
Use matchers in `isBlockedMatmul`

chelini:


lib/TPP/TransformUtils.cpp
```diff
@@ -378,27 +378,32 @@ bool isBlockedMatmul(Operation *op) {
   if (!linalgOp)
     return false;
 
-  if (linalgOp.getNumDpsInputs() != 2 || linalgOp.getNumDpsInits() != 1)
+  using namespace mlir::tpp::structured_match;
+  AffineMap aMap, bMap, cMap;
+  // clang-format off
+  auto matcher =
```
rengolin, line 384: This should be in a `isSomething` helper to make it clear what it's matching.

chelini: Called: `maybeBlockMatmul` as we don't know yet if it will be a blocked matmul or not (depend on the loop classification).

PR #624, opened 2023-07-03
Fold reshape into eltwise consumers

chelini:
When converting an MHA block using the IREE importer, operations such as
collapse shapes and expand shapes are emitted to fit the computation to
use `linalg.matmul`. We cannot propagate packing operations through
these operations; thus, this commit attempts to mitigate the problem by
folding `collapse shape` operations with the consumer, increasing the
propagation region. This commit also adds an MHA block to benchmarks
(thanks, Adam).

Co-authored by Adam Siemieniuk <adam.siemieniuk@intel.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,9 +131,18 @@ if (NOT TPP_INSIDE_IREE)
     ${CONFIG_DIR}/fc/1024x352x512.json
     ${CONFIG_DIR}/fc/1024x512x256.json
   )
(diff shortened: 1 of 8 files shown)


chelini:
> GPU related changes are merge/rebase regressions.

Nuking the build dir worked. `TPP_GPU` was in my local build still a boolean while now is a string.

chelini:
It is not needed anymore.

PR #625, opened 2023-07-03
Improve GPU-related build flags

adam-smnk:
Ensures that CUDA dependencies are not required without explicit CMake `TPP_GPU=cuda` flag.

### DIFF
--- a/lib/TPP/GPU/CMakeLists.txt
+++ b/lib/TPP/GPU/CMakeLists.txt
@@ -14,7 +14,7 @@ add_mlir_library(TPPGPU
     MLIRGPUToNVVMTransforms
 )
 
-if (NOT TPP_GPU STREQUAL "")
+if (TPP_GPU MATCHES "cuda")
   target_compile_definitions(obj.TPPGPU
     PRIVATE
     TPP_GPU_ENABLE=1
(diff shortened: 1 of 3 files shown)


lib/TPP/GPU/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ add_mlir_library(TPPGPU
     MLIRGPUToNVVMTransforms
 )
 
-if (NOT TPP_GPU STREQUAL "")
+if (TPP_GPU MATCHES "cuda")
```
rengolin, line 17: Weird that the definition is generic `TPP_GPU_ENABLE`... Perhaps that could also be Cuda specific?

adam-smnk: I'll have to probably reorganize it later anyway but sure why not.

rengolin (approved):

hfp:
Looks good to me. CMake-wise, `if (NOT TPP_GPU STREQUAL "")` is a check for any GPU-runtime (and can be of course tested specifically against, e.g., `cuda`). Though, testing for an non-empty string in CMake looks as weird as shown (one cannot write `if (NOT TPP_GPU)`).

PR #630, opened 2023-07-05
tpp-run - replace warmup call with extra iteration

adam-smnk:
Replaces the initial warmup call with an extra timer loop iteration. This approach simplifies kernel wrapper logic and ensures no extra allocs or copies are required to handle the output of the separate warmup call.

Output printing is disabled for runs with 'n > 1' as the output buffer gets reused on each timer without any guarantees to produce stable results on every iteration. And, in general, there is no need to get both timing results and print output in a single command. This also simplifies generated wrapper code.

### DIFF
--- a/tools/tpp-run/MLIRBench.cpp
+++ b/tools/tpp-run/MLIRBench.cpp
@@ -311,7 +311,7 @@ Value MLIRBench::getTimerStats(Value accBuf) {
   // The offset is slightly inaccurate as the buffer lenght already includes
   // the warmup iters. However, the difference is negligible for the chosen
   // max iters and only a few more deltas will be skipped in the worst case
-  // does not impact the overall stats.
+  // which does not impact the overall stats.
   auto offset = getNumWarmupIters(len);
   auto size = builder.create<arith::SubIOp>(unkLoc, len, offset);
   auto stride = builder.create<arith::ConstantIndexOp>(unkLoc, 1);
--- a/tools/tpp-run/MLIRBench.cpp
+++ b/tools/tpp-run/MLIRBench.cpp
@@ -258,7 +258,9 @@ Value MLIRBench::getKernelResult(Operation *kernelCall) {
                                        : kernelCall->getOpResult(0);
 }
 
-Value MLIRBench::createTimerLoop(unsigned n) {
+Value MLIRBench::createTimerLoop(unsigned iters) {
+  int n = iters + MLIRBench::numWarmupLoops;
+
   // Allocates buffer for results
   auto count = getConstInt(builder, n, 64);
   auto memrefType = MemRefType::get({n}, builder.getF64Type());
@@ -269,14 +271,24 @@ Value MLIRBench::createTimerLoop(unsigned n) {
   builder.setInsertionPointToStart(loop.getBody());
 
   // Call the kernel, ignore output
-  callKernel();
+  auto *call = callKernel();
+  assert(call && "Failed to generate a kernel call");
 
   // Revert insertion point and return the accumulation ID
   builder.setInsertionPointAfter(loop);
   return acc;
 }
 
-Value MLIRBench::getTimerStats(Value acc) {
+Value MLIRBench::getTimerStats(Value accBuf) {
+  auto offset =
+      builder.create<arith::ConstantIndexOp>(unkLoc, MLIRBench::numWarmupLoops);
+  auto dimIdx = builder.create<arith::ConstantIndexOp>(unkLoc, 0);
+  auto len = builder.create<memref::DimOp>(unkLoc, accBuf, dimIdx);
+  auto size = builder.create<arith::SubIOp>(unkLoc, len, offset);
+  auto stride = builder.create<arith::ConstantIndexOp>(unkLoc, 1);
+  auto acc = builder.create<memref::SubViewOp>(
+      unkLoc, accBuf, ValueRange{offset}, ValueRange{size}, ValueRange{stride});
+
   auto callMean =
       builder.create<perf::MeanOp>(unkLoc, builder.getF64Type(), acc);
   auto mean = callMean.getMean();
--- a/tools/tpp-run/MLIRBench.cpp
+++ b/tools/tpp-run/MLIRBench.cpp
@@ -264,26 +264,9 @@ unsigned MLIRBench::getNumWarmupIters(unsigned iters) {
                                                 MLIRBench::maxIters));
 }
 
-Value MLIRBench::getNumWarmupIters(Value iters) {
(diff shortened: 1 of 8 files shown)


rengolin:
The warmup call is for running the JIT outside of the timer loop and avoid taking its skewed measurement into account.

We were actually discussing last week perhaps to even increase the warmup loop (outside of the main loop) to massage the caches, TLBs, etc. a bit more, especially in the small matrix OpenMP cases.

adam-smnk:
> The warmup call is for running the JIT outside of the timer loop and avoid taking its skewed measurement into account.

This behavior should be preserved by skipping the first few (`numWarmupLoops`) deltas when computing mean and stdev.
Although, it failed CI as I think it's not doing that correctly yet.

> We were actually discussing last week perhaps to even increase the warmup loop (outside of the main loop) to massage the caches, TLBs, etc. a bit more, especially in the small matrix OpenMP cases.

I see some benefits of that locally, I'll increase it here too and let's see how it behaves on CI.

rengolin:
> This behavior should be preserved by skipping the first few (`numWarmupLoops`) deltas when computing mean and stdev. Although, it failed CI as I think it's not doing that correctly yet.

I see, this could work. Perhaps not make it a `constexpr` and add a percentage of the total number of iteration?

> I see some benefits of that locally, I'll increase it here too and let's see how it behaves on CI.

Ok.

adam-smnk:
> I see, this could work. Perhaps not make it a `constexpr` and add a percentage of the total number of iteration?

Sounds good, I'll have a look around if there's any rule of thumb for that. Otherwise, maybe like 10% but no less that X would be fine.

rengolin:
> > I see, this could work. Perhaps not make it a `constexpr` and add a percentage of the total number of iteration?
> 
> Sounds good, I'll have a look around if there's any rule of thumb for that. Otherwise, maybe like 10% but no less that X would be fine.

Right, something like `max( 1, min(N*0.1, 100) )`. We don't want the minimum to be higher than 1, as for larger models, one run can be substantial already.

adam-smnk:
The answer to required number of warmup iters is as always "it depends".
I applied `max( 1, min(N / 10, 100) )` as it looks reasonable and fits our needs.

adam-smnk:
CI had some trouble succeeding with various tpp-performance runs intermixing but the newest run (build 245) passed compared to fresh main branch performance (build 244).

tools/tpp-run/MLIRBench.cpp
```diff
@@ -267,14 +294,37 @@ Value MLIRBench::createTimerLoop(unsigned n) {
...
   return acc;
 }
 
-Value MLIRBench::getTimerStats(Value acc) {
+Value MLIRBench::getTimerStats(Value accBuf) {
+  // Skip the warmup loops in stats calculation
+  auto dimIdx = builder.create<arith::ConstantIndexOp>(unkLoc, 0);
+  auto len = builder.create<memref::DimOp>(unkLoc, accBuf, dimIdx);
+
+  // The offset is slightly inaccurate as the buffer lenght already includes
```
rengolin, line 310: I'm not sure I understand this comment. Isn't the point of the offset to not include the initial runs? And if so, aren't you taking the average of the remaining runs divided by the length - offset?

tools/tpp-run/MLIRBench.cpp
```diff
@@ -267,14 +294,37 @@ Value MLIRBench::createTimerLoop(unsigned n) {
...
+  // which does not impact the overall stats.
+  auto offset = getNumWarmupIters(len);
+  auto size = builder.create<arith::SubIOp>(unkLoc, len, offset);
+  auto stride = builder.create<arith::ConstantIndexOp>(unkLoc, 1);
+  auto subview = builder.create<memref::SubViewOp>(
+      unkLoc, accBuf, ValueRange{offset}, ValueRange{size}, ValueRange{stride});
+
+  // Copy the deltas into an separate smaller buffer to ensure
+  // that perf functions do not read the warmup loop deltas
+  auto accType =
```
rengolin, line 322: Why can't you just use the subview directly?

tools/tpp-run/MLIRBench.cpp
adam-smnk: Here it refers to the fact that I apply the same formula to both `n` and `1.1 * n`. Which is technically incorrect as I should solve it always for the original `n` but this simplification works fine here.
To give an example: for `-n 100`, 10 extra warmup iters are added but when calculating how many iters should be skipped for average calculation, it will skip first 11 iters.
This is mostly because I wanted to avoid floating math or any extra info tracing within the benchmark object.

Overall, all the warmup iters are skipped and sometimes a few extra iters.

tools/tpp-run/MLIRBench.cpp
adam-smnk: `perf.mean` and `perf.stdev` functions accept unranked `<*xf32>` memrefs.
AFAIK, you can't have affine map for an unranked type so the base ptr offset is lost after cast and then incorrect values are used for stats computation.

tools/tpp-run/MLIRBench.cpp
rengolin: Why can't you just add `11` to `100`, run `111` times, then take the tail `100` measurements and then divide by `100`?

tools/tpp-run/MLIRBench.cpp
rengolin: Right, it's ok to do this copy, since it's outside of the loop and it's quite small.

tools/tpp-run/MLIRBench.cpp
adam-smnk: That's what I do. It's how I compute this addition and tail that matters:
* program launch `-n 100` - measure over 100 iters
* run this kernel - how many iters we need to execute? 100 + 0.1 * 100 = 110; run 110 iters
* compute my average - how many iters we need to skip? len(deltas) = 110, 0.1 * 110 = 11; skip first 11 deltas

It skips 1 delta too many as I don't want to bother with all the int -> float -> int conversions for better accuracy.

adam-smnk: Alright, I'll just make these methods a bit less generic, hide the num iters as internal benchmark state, and remove these runtime computations.

adam-smnk: Adapted and added tests.

adam-smnk:
`omp16` kernels still have stability issues (builds 249 vs 250 on the same commit), otherwise, `tpp-performance` looks good.

rengolin (approved):
We have a strong indication that the OMP tests instability is in the machine/OS not the way we run them. This looks good to me now, thanks!

PR #631, opened 2023-07-06
GPU Vulkan runner

adam-smnk:
Extends GPU backend with Vulkan support.

Summary:
* Adds basic GPU to Vulkan lowering pipeline.
* Executes generated IR using Vulkan runtime.
* Serializes CUDA compilation - processing multiple gpu.modules in parallel when lowering to CUDA can randomly lead to CUDA runtime errors. It is addressed by serializing compilation step and execution of CUDA tests.
* Reorganizes GPU-related builds and tests depending on the enabled TPP_GPU target.


rengolin (approved):
Awesome! Thanks!

PR #632, opened 2023-07-07
Benchmarks generic runner

adam-smnk:
Adds a generic CLI runner to the benchmarks driver.

The new runner allows benchmarks to execute a generic arbitrary command without any extra wrappers.
No postprocessing is applied on the output, the invoked command is expected ensure that the produced output is in correct format.

### DIFF
--- a/benchmarks/driver.py
+++ b/benchmarks/driver.py
@@ -305,6 +305,30 @@ class IrGeneratorRun(BaseRun):
         self.teardown()
         return True
 
+class GenericRun(BaseRun):
+    """ Generic cli runs """
+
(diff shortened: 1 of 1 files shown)


adam-smnk:
Main idea: to simplify experimentation and addition of external runners such as CUDA baseline kernels for GPU.

rengolin:
I'm ok with it, we discussed this earlier a few months back, but my main question is still the same: 

How do we check the output in a generic way so that the numbers can be actionable? 

If we don't, then what's the point?

rengolin (approved):
Nit: Add a comment making it clear that this is "here be dragons" territory.

Issue #639, opened 2023-07-14
TPP-performance tests are not running


hfp:
This issue was apparently introduced by PR #633.

hfp:
Fixed by #638.

PR #645, opened 2023-07-19
Fixed and improved #639

hfp:


hfp:
Without this PR, for instance https://buildkite.com/intel/tpp-performance does not actually run. It also worrisome for our glue-scripts to miss `set -eo pipefail` (or more elaborated setting), which means failure in script/test is not passed to CI.

adam-smnk (approved):
Let's bring this back considering CI is broken without the fixes. Also this shouldn't affect formatting so lgtm.

PR #647, opened 2023-07-20
Print format error files

rengolin:
This allows us to see at CI what's wrong and which files have reported errors wihout running clang-format locally in a special way.


adam-smnk (approved):
Tested locally - works nicely 👍

PR #648, opened 2023-07-23
Reenable BF16 tests on AX2 machines

rengolin:


test/BF16/Integration/lit.local.cfg
```diff
@@ -20,5 +20,5 @@ def has_support(feature):
     return True
 
 # AVX512 and SVE should have BF16 support
-if not has_support('avx512') and not has_support('sve'):
+if not has_support('avx512') and not has_support('avx2') and not has_support('sve'):
```
alheinecke, line 23: SVE has the same support as AVX512. in libxsmm.. not sure why it's excluded

alheinecke (changes requested):

test/BF16/Integration/lit.local.cfg
alheinecke: sse and avx should be excluded

rengolin: It's not excluded. This is a double negative. It is "supported" is there's AVX2, AVX512 or SVE.

rengolin: > sse and avx should be excluded

It is. These tests are **only** run on AVX2, AVX512 and SVE.

rengolin:
For the record, the CI machine was already running the BF16 tests, so no changes were visible on the number of "Unsupported" tests. However, on my ageing desktop I used to get 35 unsupported and now I get 21 (like the CI machine) and all tests pass.

alheinecke (approved):

alheinecke:
yes it should be fine since April, when the compiler bug was fixed

hfp:
> For the record, the CI machine was already running the BF16 tests, so no changes were visible on the number of "Unsupported" tests. However, on my ageing desktop I used to get 35 unsupported and now I get 21 (like the CI machine) and all tests pass.

( The `tpp-gpu` pipeline or the Intel based tests are running on a RaptorLake system and we should see the change there as well. )

PR #650, opened 2023-07-24
Avoid to include src/template/libxsmm_version.h and fix #566

hfp:
Addressed #566. Including src/template/libxsmm_version.h is wrong since the header file is not valid C/C++ code.

The issue seems to stem from `target_compile_definitions` like `-DLIBXSMM_DEFAULT_CONFIG` not being propagated to code using LIBXSMM (even when this definition was marked as `PUBLIC`). Apparently, `-DLIBXSMM_DEFAULT_CONFIG` previously only applied when compiling LIBXSMM's code (separate static library). The issue might be a wrong perception of how to reuse code compiled into a separate library.

### DIFF
--- a/cmake/modules/xsmm.cmake
+++ b/cmake/modules/xsmm.cmake
@@ -34,13 +34,10 @@ set(XSMM_INCLUDE_DIRS ${LIBXSMMROOT}/include)
 
 add_library(xsmm STATIC ${XSMM_SRCS})
 target_include_directories(xsmm PUBLIC ${XSMM_INCLUDE_DIRS})
-target_compile_definitions(xsmm PUBLIC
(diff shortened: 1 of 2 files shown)


hfp:
This work reproduced the issue #566 and resolved it (differently).

adam-smnk (approved):

hfp:
@nhasabni I was using your `tpp-mlir` branch or IREE to reproduce and resolve #566 correctly. This PR is now merged; perhaps you need to do something else for your IREE work? Just a head-up...

( Side effect: I can confirm, the IREE integration compiled and we can likely deploy a recipe if needed. )

nhasabni:
Thanks @hfp for the fix. Will test it in IREE soon.

Issue #555, opened 2023-05-19
Replace our vector print in `tpp-run` with upstream `printMemrefF32`


rengolin:
Right now there's only `f32` variant of print, not `bf16` so we had to use the vector lowering.

But if we upstream a `printMemrefBF16` (see #554), then we can just use that and not have to worry about the rank and dims of the returned tensors (see #176).

@adam-smnk @KavithaTipturMadhu

rengolin:
Interesting development: https://github.com/llvm/llvm-project/pull/68695

PR #654, opened 2023-07-26
Improve Vulkan pipeline

adam-smnk:
Expands Vulkan lowering pipeline with extra support passes to allow executing more linalg ops using Vulkan backend.

Changes:
* Missing SPIRV attributes required for lowering GPU to SPIRV ops are now automatically added to GPU modules.
* Missing lowering patterns are added to GPU to SPIRV conversion.
* GPU kernel prototypes and function launches are rewritten to be compliant with Vulkan ABI.
* More Vulkan tests are now enabled and working.


adam-smnk:
Missing conversion patterns have to be added directly to the upstream GPU to SPIRV conversion pass as this conversion cannot be split into multiple steps. This addition could be upstreamed later if it works well.

The SPIRV attribute adding passes are taken from IMEX and slightly adapted.

lib/TPP/GPU/GpuVulkanAbi.cpp
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // Buffers are Vulkan compatible but need to be min 1D and max 3D shaped.
+  // Scalars cannot be passed directly to a Vulkan kernels.
+  // Wrap them into Vulkan compatible buffers.
+  auto funcType = funcOp.getFunctionType();
+  SmallVector<Type> inputs;
+  for (Type input : funcType.getInputs()) {
+    if (auto memrefType = input.dyn_cast<MemRefType>())
+      inputs.push_back(FlattenMemrefType(memrefType));
```
rengolin, line 96: This seems to be used only in conjunction with `getVulkanTypeWrapper`, and could simply be called from `getVulkanTypeWrapper` to simplify this loop.

rengolin (commented):
Quick review of the source, will look at tests later

include/TPP/Passes.td
```diff
@@ -446,4 +446,67 @@ def DefaultPipeline : Pass<"default-pipeline", "ModuleOp"> {
   ];
 }
 
+def GPUToSPIRV : Pass<"gpu-to-spirv", "ModuleOp"> {
```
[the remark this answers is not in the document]
rengolin: My guess is that different downstream groups wanted to do it differently downstream and no one would accept the other's contribution. I guess we'll see once we try to upstream.

test/GPU/CUDA/Integration/tpp-brgemm.mlir
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  %out = memref.alloc() : memref<2x8x32x32xf32>
+  %tw = gpu.wait async
+  %tOut = gpu.memcpy async [%tw] %out, %arg2 : memref<2x8x32x32xf32>, memref<2x8x32x32xf32>
+  gpu.wait [%tOut]
+
+  %d1 = arith.constant -1.0 : f32
+  %zeroCst = arith.constant 0 : index
+  %v0 = vector.transfer_read %out[%zeroCst, %zeroCst, %zeroCst, %zeroCst], %d1 : memref<2x8x32x32xf32>, vector<32xf32>
```
rengolin, line 26: Why are we printing here rather than `-print`?

test/GPU/Vulkan/Integration/matmul-vulkan.mlir
```diff
@@ -0,0 +1,41 @@
...
+  spirv.target_env = #spirv.target_env<
+    #spirv.vce<v1.0, [Shader], [SPV_KHR_storage_buffer_storage_class]>, #spirv.resource_limits<>>
+} {
+  func.func @entry(%arg0: memref<8x8xf32>, %arg1: memref<8x8xf32>, %arg2: memref<8x8xf32>) {
+    %c8 = arith.constant 8 : index
+    %c0 = arith.constant 0 : index
+    %c1 = arith.constant 1 : index
+    gpu.launch_func  @entry_kernel::@entry_kernel blocks in (%c8, %c8, %c1) threads in (%c1, %c1, %c1) args(%arg0 : memref<8x8xf32>, %arg1 : memref<8x8xf32>, %arg2 : memref<8x8xf32>, %c0 : index, %c8 : index, %c1 : index)
+    %cast = memref.cast %arg2 : memref<8x8xf32> to memref<*xf32>
+    call @printMemrefF32(%cast) : (memref<*xf32>) -> ()
```
rengolin, line 17: #555 is pertinent here, for another PR of course

test/GPU/Vulkan/Integration/matmul-vulkan.mlir
```diff
@@ -0,0 +1,41 @@
...
+    return
+  }
+
+  gpu.module @entry_kernel {
+    gpu.func @entry_kernel(%arg0: memref<8x8xf32>, %arg1: memref<8x8xf32>, %arg2: memref<8x8xf32>, %arg3: index, %arg4: index, %arg5: index)
+    kernel attributes { spirv.entry_point_abi = #spirv.entry_point_abi<workgroup_size = [1, 1, 1]> } {
```
rengolin, line 23: If this is comparable to the CUDA matmul example, than this is way more complex. Can we get a similar kernel from CUDA and lower it all the way down to SPIRV via Vulkan and get this IR?

test/GPU/CUDA/Integration/tpp-brgemm.mlir
adam-smnk: Well I could `-print` in a few other tests but I just kept `printMemref` from some upstream test cases I later adapted.
Imo, it doesn't really matter.

test/GPU/Vulkan/Integration/matmul-vulkan.mlir
adam-smnk: It's not meant to be comparable. This test's goal is to have lower level IR at SPIRV(ish) level to test that for loop lowering works properly as previously it was not possible. It's done at the same level as `addf-vulkan.mlir`.

Comparable examples are covered by `linalg-matmul.mlir`, `tpp-gemm.mlir`, and `tpp-brgemm.mlir` tests present in both `CUDA` and `Vulkan`.

test/GPU/CUDA/Integration/tpp-brgemm.mlir
rengolin: Right, #555 to the rescue, for now, looks good.

rengolin (approved):
Right, this has too much Vulkan specific logic for my comfort and we should look into upstreaming all of it, but for now, it should be fine. We really need to move to something like OpenCL soon.

PR #655, opened 2023-07-27
Tppfication of pack pass 

KavithaTipturMadhu:


test/Passes/tpp-identity.mlir
```diff
@@ -0,0 +1,21 @@
+//RUN: tpp-opt %s --default-tpp-passes | tpp-run -e main  -entry-point-result=void | FileCheck %s
```
adam-smnk, line 1: Just `tpp-run %s` should be enough

test/Passes/tpp-identity.mlir
adam-smnk, line 1: Please move the test to test/Integration dir

test/Passes/DefaultPipeline/tpp-mapping.mlir
```diff
@@ -216,19 +212,16 @@ func.func @propagate_pack_unpack(%arg0: tensor<128x512xf32>, %arg1: tensor<512x2
 // CHECK: scf.for
 // CHECK:   scf.for
 // CHECK:     tensor.extract_slice{{[^:]+}}: tensor<128x512xf32> to tensor<32x32xf32>
-// CHECK:     linalg.transpose
 // CHECK:     tensor.insert_slice{{[^:]+}}: tensor<32x32xf32> into tensor<4x16x32x32xf32>
 // Generalized pack of the second input
 // CHECK: scf.for
 // CHECK:   scf.for
 // CHECK:     tensor.extract_slice{{[^:]+}}: tensor<512x256xf32> to tensor<32x32xf32>
-// CHECK:     linalg.transpose
```
rengolin, line 225: The second input has `outer_dims_perm`, so perhaps there's some additional logic needed here?

test/Passes/tpp-identity.mlir
```diff
@@ -0,0 +1,21 @@
...
+  %0 = arith.constant dense<[[0.0, 1.0, 2.0, 3.0],
+                               [4.0, 5.0, 6.0, 7.0],
+                               [8.0, 9.0, 10.0, 11.0],
+                               [12.0, 13.0, 14.0, 15.0]]> : tensor<4x4xf32>
+  %1 = tensor.empty():tensor<2x2x2x2xf32>
+  %c0 = arith.constant 0 : index
+  %2 = call @myfunc(%0,%1) : (tensor<4x4xf32>, tensor<2x2x2x2xf32>) -> tensor<2x2x2x2xf32>
+  %d1 = arith.constant -1.0 : f32
+  %v0 = vector.transfer_read %1[%c0, %c0, %c0, %c0], %d1 : tensor<2x2x2x2xf32>, vector<2x2x2x2xf32>
+  // CHECK: ( ( ( ( 0, 1 ), ( 4, 5 ) ), ( ( 2, 3 ), ( 6, 7 ) ) ), ( ( ( 8, 9 ), ( 12, 13 ) ), ( ( 10, 11 ), ( 14, 15 ) ) ) )
```
rengolin, line 18: You can avoid the print by comparing to another constant.

Would also be good to get some more variations on the tests.

test/Passes/tpp-identity.mlir
```diff
@@ -0,0 +1,63 @@
+// RUN:  tpp-opt %s  --default-tpp-passes | tpp-run -e entry  -entry-point-result=void | FileCheck %s
+
+func.func @pack1(%in: tensor<4x4xf32>, %out: tensor<2x2x2x2xf32>) ->  tensor<2x2x2x2xf32> {
+  %1 = tensor.pack %in inner_dims_pos = [0, 1] inner_tiles = [2,2] into %out : tensor<4x4xf32> -> tensor<2x2x2x2xf32>
+  return %1 : tensor<2x2x2x2xf32>
+}
+
+func.func @pack2(%0: tensor<1x2x2x4xf32>, %1:  tensor<1x2x2x2x2xf32>)-> tensor<1x2x2x2x2xf32>{
+ %2 = tensor.pack %0  outer_dims_perm = [0, 3, 1, 2] inner_dims_pos = [3] inner_tiles = [2] into %1 : tensor<1x2x2x4xf32> -> tensor<1x2x2x2x2xf32>
```
rengolin, line 9: Would be good to have these tests checking IR here and the execution in `Integration`, so that we can make sure that it's not only doing what we want, but how we want.

test/Passes/tpp-identity.mlir
```diff
@@ -0,0 +1,63 @@
...
+  %2 = call @pack1(%0,%1) : (tensor<4x4xf32>, tensor<2x2x2x2xf32>) -> tensor<2x2x2x2xf32>
+  %d1 = arith.constant -1.0 : f32
+  %v0 = vector.transfer_read %1[%c0, %c0, %c0, %c0], %d1 : tensor<2x2x2x2xf32>, vector<2x2x2x2xf32>
+  vector.print %v0 : vector<2x2x2x2xf32> 
+  // CHECK: ( ( ( ( 0, 1 ), ( 4, 5 ) ), ( ( 2, 3 ), ( 6, 7 ) ) ), ( ( ( 8, 9 ), ( 12, 13 ) ), ( ( 10, 11 ), ( 14, 15 ) ) ) ) 
+  
+  %3 = arith.constant dense <[[[[0.0, 1.0, 2.0, 3.0],  [4.0, 5.0, 6.0, 7.0]],
+                                [[8.0, 9.0, 10.0, 11.0],[12.0, 13.0, 14.0, 15.0]]]]>: tensor<1x2x2x4xf32>
+  %4 = tensor.empty():tensor<1x2x2x2x2xf32>
+  %5 = call @pack2(%3, %4):(tensor<1x2x2x4xf32>,  tensor<1x2x2x2x2xf32>)->(tensor<1x2x2x2x2xf32>)
```
rengolin, line 35: It is a bit confusing to have to look up to the `pack` function to see what pack is being done, here. I think it's easier to just have the pack operation here and not a function call.

lib/TPP/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ add_mlir_library(MLIRTPP
     ConvertCheckToLoops.cpp
     ConvertPerfToLoops.cpp
     ConvertPerfToFunc.cpp
-
+    ConvertTppToIdentity.cpp
```
rengolin, line 44: This isn't converting "TPP to Identity", it's converting `tensor.pack` and `tensor.unpack` to TPP identity. I'd name it `ConvertPackToXSMM.cpp` because we'll soon not be using TPP ops and will convert straight to an XSMM unary (with different flags for transpose and vnni).

lib/TPP/ConvertTppToIdentity.cpp
```diff
@@ -0,0 +1,219 @@
...
+      SmallVector<Value> ubs;
+      SmallVector<Value> steps;
+
+      for (int i = 0; i < numLoops; i++) {
+        lbs.push_back(zero);
+        steps.push_back(one);
+      }
+
+      for (int i = 0; i < numLoops; i++) {
+
```
rengolin, line 90: extra space

lib/TPP/ConvertTppToIdentity.cpp
```diff
@@ -0,0 +1,219 @@
...
+                offsets.push_back(localIvs[i]);
+              }
+            }
+            SmallVector<OpFoldResult> strides;
+            for (int i = 0; i < numLoops; i++) {
+              strides.push_back(rewriter.getIndexAttr(1));
+            }
+
+            SmallVector<OpFoldResult> sizes;
+
```
rengolin, line 139: extra space

lib/TPP/ConvertTppToIdentity.cpp
```diff
@@ -0,0 +1,219 @@
...
+          rewriter, packOp.getLoc(), lbs, ubs, steps, reduc,
+          [&reduc, &pos, bodyBuilder, &packOp,
+           &numLoops](OpBuilder &rewriter, Location loc, ValueRange localIvs,
+                      ValueRange iterArgs) -> scf::ValueVector {
+            reduc.assign(iterArgs.begin(), iterArgs.end());
+
+            SmallVector<OpFoldResult> offsets;
+            SmallVector<MatchResult> haveMatched = innerDimMatchesIndex(packOp);
+            for (int i = 0; i < numLoops; i++) {
+
```
rengolin, line 121: extra space

lib/TPP/ConvertTppToIdentity.cpp
```diff
@@ -0,0 +1,219 @@
...
+    : public ConvertTppToIdentityBase<ConvertTppToIdentity> {
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    populateTppToIdentityPatterns(patterns);
+    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
+    return;
+  }
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<scf::SCFDialect>();
+    registry.insert<tpp::TppDialect>();
```
rengolin, line 210: We're not using TPP identity any more, so I guess we can drop the requirement.

For transpose copies we can keep them as `linalg.transpose` and lower straight to XSMM.

For VNNI work we'll have to work out how we support it in `linalg` or if we lower everything to XSMM at a high level (and make it work on `tensor`?).

include/TPP/Passes.td
```diff
@@ -446,4 +446,12 @@ def DefaultPipeline : Pass<"default-pipeline", "ModuleOp"> {
   ];
 }
 
+def ConvertTppToIdentity: Pass<"tpp-to-identity", "func::FuncOp"> {
+  let summary = "Lower TPP transpose to tpp identity";
+  let description = [{
+    Convert tpp transpose operations to tpp identity loops.
```
rengolin, line 452: This is not what the pass is doing. The pass is just optimizing the packing algorithm, and most of it can actually go upstream later, because even `memref.copy` can be faster (on a line/tile) than one-by-one exchange.

We're also not replacing all of the `pack`s, so this isn't a conversion pass, but an optimization pass.

I'd call this pass "pack-optimization" and have its description as something like: "Optimize certain `pack` and `unpack` patterns (block-transpose) into faster implementation".

We can have a separate one to lower to XSMM that we keep local.

lib/TPP/ConvertTppToIdentity.cpp
alheinecke: For the now the goal is to have the two packs needed for VNNI lowered into a single `scf.for` loop nest with 2 memref copy (one inner at the length of VNNI packing into a small temp buffer, one outer which copies the temp buffer into the final destination buffer). This will allow for fusion of packing with GEMM, what we need (hopefully) and much later in the lowering pass we can match the 2 `scf.for` + outer memref copy into on `XSMM` call for norm->vnni. Lowering to XSMM at high-level could kill fusion in tiled formulation.

include/TPP/Passes.td
alheinecke: Yepp, see above for now there should be a pass that lower `memref.copy` to `XSMM` identity and `XSMM` norm->vnni. The one for `XSMM` identity already exists and it will need massaging to support norm->vnni.

rengolin: Awesome! Ok, so we don't need TPP identity at all even now. Let's drop the requirement on the pass.

This will make it a lot easier to upstream this pass later on.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
+//===ConvertPackOptimization.cpp --------------------------------*----- C++-*-===//
```
chelini, line 1: please fix top-level comment.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+////===----------------------------------------------------------------------===//
+//
+//
+#include "TPP/BuilderUtils.h"
+#include "TPP/Dialect/Tpp/TppOps.h"
+#include "TPP/Dialect/Tpp/TppUtils.h"
+#include "TPP/Passes.h"
+#include "TPP/TransformUtils.h"
+#include "TPP/Transforms.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
```
chelini, line 16: please have a look at the include. Most of them are not needed (i.e., Bufferization).

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include "llvm/ADT/TypeSwitch.h"
+using namespace mlir;
+using namespace mlir::tpp;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+typedef struct MatchResult {
```
chelini, line 37: drop typedef.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+  bool haveMatch;
+  int index;
+} MatchResult;
+
+SmallVector<MatchResult> innerDimMatchesIndex(tensor::PackOp packOp) {
```
chelini, line 42: I am a bit lost here. What are you trying to do?

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+struct ConvertPackOptimizationOp : public OpRewritePattern<tensor::PackOp> {
+  using OpRewritePattern<tensor::PackOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(tensor::PackOp packOp,
+                                PatternRewriter &rewriter) const override {
+    if (packOp.getStaticInnerTiles().size() <= 0) {
+      return failure();
+    }
+    auto shape = packOp.getSource().getType().getShape();
```
chelini, line 76: you can also write: `packOp.getSourceType().getShape()`.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    auto zero = getConstIndex(rewriter, 0);
+    auto one = getConstIndex(rewriter, 1);
+
+    SmallVector<Value> lbs;
+    SmallVector<Value> ubs;
+    SmallVector<Value> steps;
+
+    for (int i = 0; i < numLoops; i++) {
```
chelini, line 95: you can avoid the loop here: `SmallVector<Value> lbs(numLoops, zero)`

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+      lbs.push_back(zero);
+      steps.push_back(one);
+    }
+
+    for (int i = 0; i < numLoops; i++) {
+
+      if (haveMatched[i].haveMatch) {
+        ubs.push_back(getConstIndex(
+            rewriter,
+            shape[i] / packOp.getStaticInnerTiles()[haveMatched[i].index]));
```
chelini, line 105: this is not correct in case the sizes or the shape are values.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+      } else {
+        ubs.push_back(getConstIndex(rewriter, shape[i]));
+      }
+    }
+    Value pos;
+
+    SmallVector<Value> reduc = {
```
chelini, line 113: what do you want to do here?

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+        packOp.getDest(),
+    };
+
+    auto bodyBuilder = [&](OpBuilder &builder, Location, Value iv,
```
chelini, line 117: can you pass nullptr?

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+          }
+
+          SmallVector<OpFoldResult> insertSliceStrides(
+              packOp.getDestRank(), rewriter.getIndexAttr(1));
+          auto insertSliceOp = rewriter.create<tensor::InsertSliceOp>(
+              loc, tensorExtract.getResult(), iterArgs[0], insertSliceOffsets,
+              insertSliceSizes, insertSliceStrides);
+          return {insertSliceOp};
+        });
+    rewriter.replaceAllUsesWith(packOp.getResult(),
```
chelini, line 195: why not replaceOp?

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+                  getConstIndex(
+                      rewriter,
+                      packOp.getStaticInnerTiles()[haveMatched[i].index]));
+              offsets.push_back(muliOp);
+            } else {
+              offsets.push_back(localIvs[i]);
+            }
+          }
+          SmallVector<OpFoldResult> strides;
+          for (int i = 0; i < numLoops; i++) {
```
chelini, line 143: avoid loops pls.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+                  packOp.getSource().getType().cast<RankedTensorType>(),
+                  offsets, sizes, strides);
+
+          auto tensorExtract = rewriter.create<tensor::ExtractSliceOp>(
+              loc, tensorExtractType.cast<RankedTensorType>(),
+              packOp.getSource(), offsets, sizes, strides);
+
+          SmallVector<OpFoldResult> insertSliceOffsets;
+          for (int i = 0; i < numLoops; i++) {
+            int indirection = i;
```
chelini, line 170: what is `indirection`?

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+     patterns.add<ConvertPackOptimizationOp>(patterns.getContext());
+  // clang-format on
+}
+
+struct ConvertPackOptimization : public ConvertPackOptimizationBase<ConvertPackOptimization> {
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    populatePackOptimizationPatterns(patterns);
+    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
+    return;
```
chelini, line 212: no need for return.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,223 @@
...
+  }
+  void getDependentDialects(DialectRegistry &registry) const override {
```
chelini, line 214: let's move it in the .td file, pls.

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: indirection refers to the index used to index based on outer_dims_perm. Since it is an indirect access through local induction variable, I have named it so.

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: Added a line of comment in code. Lemme know if that helps

include/TPP/Passes.h
```diff
@@ -147,6 +147,7 @@ std::unique_ptr<OperationPass<ModuleOp>>
 createDefaultPipelinePass(StringRef gpuBackend = "");
 
 // Testing passes.
+std::unique_ptr<OperationPass<func::FuncOp>> createConvertPackOptimization();
```
chelini, line 150: this is not a testing pass. Please move this before the comment.

include/TPP/Passes.td
```diff
@@ -241,6 +241,15 @@ def RewriteBatchMatmulToMatmul : Pass<"rewrite-batch-matmul-to-matmul",
   let dependentDialects = ["scf::SCFDialect", "linalg::LinalgDialect"];
 }
 
+def ConvertPackOptimization : Pass<"pack-optimization", "func::FuncOp"> {
+  let summary = "Optimize certain pack patterns into faster implementation";
+  let description =
+      [{Optimize certain pack patterns into faster implementation.}];
+  let constructor = "mlir::tpp::createConvertPackOptimization()";
+
+  let dependentDialects = ["scf::SCFDialect"];
```
chelini, line 250: it probably depends on tensor dialect too.

include/TPP/Passes.td
```diff
@@ -241,6 +241,15 @@ def RewriteBatchMatmulToMatmul : Pass<"rewrite-batch-matmul-to-matmul",
   let dependentDialects = ["scf::SCFDialect", "linalg::LinalgDialect"];
 }
 
+def ConvertPackOptimization : Pass<"pack-optimization", "func::FuncOp"> {
+  let summary = "Optimize certain pack patterns into faster implementation";
+  let description =
+      [{Optimize certain pack patterns into faster implementation.}];
```
chelini, line 247: can you specify which types or patterns are you optimizing for? "certain" is a bit too generic.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,202 @@
+////===ConvertPackOptimization.cpp ----------------------------*---- C++-*-===//
+////
```
chelini, line 2: this is not yet fixed.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+namespace {
+
+struct MatchResult {
+  bool haveMatch;
+  int index;
+};
+
+/*
+ * Returns a list of indices that match the inner dimensions' position
```
chelini, line 32: I still don't understand, sorry. The inner_dims_pos are already the indices that get tiled. You seems to return the same indices.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,202 @@
...
+                                PatternRewriter &rewriter) const override {
+    if (packOp.getStaticInnerTiles().size() <= 0)
+      return failure();
+    auto shape = packOp.getSourceType().getShape();
+    int numLoops = shape.size();
+
+    SmallVector<MatchResult> haveMatched = innerDimMatchesIndex(packOp);
+    for (int i = 0; i < numLoops; i++) {
+      if (haveMatched[i].haveMatch) {
+        if (shape[i] < packOp.getStaticInnerTiles()[haveMatched[i].index]) {
```
chelini, line 73: why do you check this?

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,202 @@
...
+        ubs.push_back(getConstIndex(
+            rewriter,
+            shape[i] / packOp.getStaticInnerTiles()[haveMatched[i].index]));
+
+      } else {
+        ubs.push_back(getConstIndex(rewriter, shape[i]));
+      }
+    }
+
+    SmallVector<Value> reduc = {
```
chelini, line 99: You don't need this `reduc`. I think you can pass directly `packOp.getDest()`.

lib/TPP/ConvertPackOptimization.cpp
```diff
@@ -0,0 +1,202 @@
...
+        packOp.getDest(),
+    };
+
+    auto loopNest = mlir::scf::buildLoopNest(
+        rewriter, packOp.getLoc(), lbs, ubs, steps, reduc,
+        [&reduc, &packOp, &numLoops](OpBuilder &rewriter, Location loc,
+                                     ValueRange localIvs,
+                                     ValueRange iterArgs) -> scf::ValueVector {
+          reduc.assign(iterArgs.begin(), iterArgs.end());
```
chelini, line 108: dead code?

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: I don't return whatever is in inner_dims_pos, but the tiling dimension for every inner_dims_pos index that matches the index of the tensor shape.

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: This is in case the returned index is -1.

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: removed.

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: %pack1 = tensor.pack %input_tensor_bcast4 inner_dims_pos = [3, 2] inner_tiles = [8, 32] into %packed_tensor1 : tensor<1x1x32x8xf32> -> tensor<1x1x1x1x8x32xf32>
In this case, inner_tiles size is 32 whereas corresponding shape is 8.

chelini: Here you are using a tile of 8 for the position in the input tensor at index 3 that is 8, similarly you use a 32 tile factor for the position in the input tensor at index 2 that is 32. The condition above in my opinion is not correct.

test/Integration/tpp-pack.mlir
```diff
@@ -0,0 +1,63 @@
+// RUN:  tpp-opt %s  --default-tpp-passes | tpp-run -e entry  -entry-point-result=void | FileCheck %s
```
rengolin, line 1: No need for `tpp-opt` here, just `tpp-run`.

lib/TPP/ConvertPackOptimization.cpp
KavithaTipturMadhu: Replaced with monotonically increasing dimensions check

test/Integration/tpp-pack.mlir
KavithaTipturMadhu: Not having tpp-opt throws an error : Entry point entry not found as a func.func
Did you mean get rid of the flag -default-tpp-passes?

rengolin: You should be able to run this with just `tpp-run`. If you drop the `-default-tpp-passes` from `tpp-opt` then it does nothing.

rengolin:
@chelini can you check the CHECK lines on the final test to make sure they're what you expected?

KavithaTipturMadhu:
> @chelini can you check the CHECK lines on the final test to make sure they're what you expected?

I am adding a few more tests, you can review them after, I will tag you once I am done updating the tests @chelini.

KavithaTipturMadhu:
@chelini test cases are good for review now

chelini:
> @chelini test cases are good for review now

Tests are ok.

KavithaTipturMadhu:
@chelini @rengolin @adam-smnk any more comments?

PR #656, opened 2023-07-27
GPU benchmarks - CUDA kernels

adam-smnk:
Adds benchmark kernels for CUDA performance evaluation.

The old `benchmarks/ref` reference implementation infrastructure is brought back and adapted to GPU testing needs.
A naive CUDA kernel and CUBLAS are added as baselines for MLIR performance evaluation.


adam-smnk:
Initial results from a local run:
```
Benchmark: cpu_gemm
fp32_1024_base_loops        :     5.323 +-     0.114 gflops
fp32_1024_base_tpp          :   139.346 +-     3.031 gflops
fp32_1024_unp_base_tpp      :     3.201 +-     0.043 gflops

Benchmark: gpu_gemm
fp32_1024_cuda_ref          :    98.575 +-     5.994 gflops
fp32_1024_cublas_ref        :  4432.781 +-   387.365 gflops
fp32_1024_base_mlir         :    11.537 +-     0.720 gflops
fp32_1024_tpp_mlir          :    25.961 +-     0.125 gflops
fp32_1024_manual_kernel_mlir:   101.767 +-     0.833 gflops
fp32_1024_unp_base_mlir     :     5.187 +-     0.110 gflops
```

`fp32_1024_manual_kernel_mlir` is manual recreation of baseline CUDA kernel using MLIR GPU dialect. Performance of the two kernels is similar as expected.
Current default GPU pipeline doesn't parallelize well for GPU yet. Thus, other kernels have poor performance.

Some next steps:
* add the GPU benchmarks to existing GPU CI jobs
* add Vulkan benchmarks - after #654 
* start iterating on linalg -> GPU lowering pipeline to at least match performance of the manual kernel

CMakeLists.txt
```diff
@@ -152,6 +153,27 @@ if (NOT TPP_INSIDE_IREE)
...
+      set(BENCH_CUDA_DEPS
+        bench_cuda_matmul
+      )
+    endif()
+
+    string(JOIN ',' BENCH_GPU_CFGS_STR ${BENCH_CUDA_CFGS})
+    string(JOIN ' ' BENCH_GPU_DEPS_STR ${BENCH_CUDA_DEPS})
+
+    # Run a small set of benchmarks with small iterations to test the benchmarks and run locally on small machines
+    add_custom_target(quick-gpu-bench ${BENCHMARK_DIR}/driver.py -v --build ${PROJECT_BINARY_DIR} -n 10
```
rengolin, line 171: Guard this target with `BENCH_GPU_CFGS_STR` not being empty. Right now, non-cuda targets will make it empty

benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(ref)
```
rengolin, line 1: This should go into `tools` not `benchmarks`

benchmarks/config/GPU/cuda.json
```diff
@@ -0,0 +1,71 @@
+[
+  {
+  "cpu_gemm": {
```
rengolin, line 3: Unnecessary to run CPU as a baseline for GPU, we can't really compare them 1:1 and we already have them in other runs, even quick ones.

benchmarks/harness/controller.py
```diff
@@ -278,6 +283,13 @@ def verifyStats(self):
     if args.disable_lsan:
         asan_options.append("detect_leaks=0")
 
+    # GPU tests require extra ASAN flags due to incompatibility with CUDA
+    # See: https://github.com/google/sanitizers/issues/629
+    if args.gpu is not None:
+        asan_options.extend(
+            ["protect_shadow_gap=0", "replace_intrin=0", "detect_leaks=0"]
```
rengolin, line 290: You're adding these here, but also in the JSON file. We need to pick a place and stick with it.

benchmarks/mlir/GPU/gemm-fp32-1024-base.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: tpp-run %s -n 10 \
+// RUN:  -e entry -entry-point-result=void
+
+// BENCH_TOTAL_FLOPS: 536870912
+
+#map = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
+module {
+  func.func @entry(%arg0: tensor<8x32x32x32xf32>, %arg1: tensor<32x32x32x32xf32>, %arg2: tensor<8x32x32x32xf32>) -> tensor<8x32x32x32xf32> {
```
rengolin, line 10: Why is this packed?

benchmarks/mlir/GPU/gemm-fp32-1024-unpacked-base.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: tpp-run %s -n 10 \
+// RUN:  -e entry -entry-point-result=void
+
+// BENCH_TOTAL_FLOPS: 536870912
+
+#map = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+module {
+  func.func @entry(%arg0: tensor<256x1024xf32>, %arg1: tensor<1024x1024xf32>, %arg2: tensor<256x1024xf32>) -> tensor<256x1024xf32> {
```
rengolin, line 10: I'd call this one `base` and the other one `packed-base` or something.

benchmarks/ref/include/Bench.h
```diff
@@ -0,0 +1,91 @@
...
+
+  // Warms up the kernel and populates the output (inside the args vector).
+  void warmup() { kernel.runRef(args); }
+
+  // Runs the benchmark `iter` times and take timings
+  void run() {
+    // Easier to optimize / inline a branchless loop
+    for (size_t i = 0; i < iter; i++) {
+      perf.startTimer();
+      kernel.runRef(args);
```
rengolin, line 68: In a GPU, will this also time dispatch? If so, can you bundle the dispatch of multiple kernels in one go?

benchmarks/ref/include/Bench.h
```diff
@@ -0,0 +1,91 @@
...
+
+// Benchmark: runs a kernel multiple times, takes timings, return avg.
+//
+// Supports warm-up (JIT compilers, cache flush, etc), tensor initialization,
+// performance tracking, output comparison, etc.
+//
+// Usage:
+//  * Implement a kernel that implements the KernelInterface
+//  * Define a variable as Benchmark<MyKernel>
+template <class Kernel, class T> class Benchmark {
```
rengolin, line 36: We had this before, but we moved to IR based timer, right?

This should be good for both CPU and GPU.

benchmarks/ref/include/Tensor.h
```diff
@@ -0,0 +1,184 @@
...
+#include <vector>
+
+namespace {
+using ListArg = std::vector<unsigned>;
+using InitArg = std::initializer_list<unsigned>;
+} // namespace
+
+// Tensor: represents a tensor in memory, basically a pointer to data plus
+// rank and dimension.
+template <typename T> class Tensor {
```
rengolin, line 28: Do we need to differentiate between CPU tensor and GPU tensor?

For example, if we want to "read" a result tensor on the GPU, we'll need to copy it back or something. 

Or is that supposed to be done by the kernels themselves? And if so, how can we avoid timing that?

benchmarks/harness/controller.py
adam-smnk: My view is that JSON allows you to add these (and other flags) more on per need basis to individual tests.
However, these sanitizers have to be disabled otherwise you can't run CUDA kernels at all. Thus, they are here in controller as default options for GPU runs.

CMakeLists.txt
adam-smnk: Correct, it will just throw an error that it misses cfg file. Later we'll also have vulkan so there'll be always sth active.

benchmarks/ref/include/Bench.h
adam-smnk: That's for my CUDA baseline implementations.

benchmarks/ref/include/Tensor.h
adam-smnk: Currently, we need a bit due to memory management. That's why there's `CudaTensor` which provides GPU wrapper over other Tensor types.

Otherwise, good questions but I'll worry about them if I ever need these features. For now I don't expect to expand much on the side of manual CUDA kernels.

benchmarks/ref/include/Bench.h
adam-smnk: It does include dispatch which I think is in line with MLIR measurements i.e., timing loop around kernel call that invokes `gpu.launch_func`.

Multiple kernels could be bundled or queued through streams but I haven't added this logic.

scripts/buildkite/tpp-gpu.yml
```diff
@@ -47,3 +47,19 @@ steps:
...
+    command: "${SRUN} --partition=nv-v100,nv-a100 --time=0:30:00 -- \
+              'KIND=Release COMPILER=clang LINKER=lld GPU=cuda \
+              scripts/buildkite/benchmark_gpu.sh'"
+    env:
+      # ASAN flags required due to incompatibility with CUDA
+      # See: https://github.com/google/sanitizers/issues/629
+      ASAN_OPTIONS: "protect_shadow_gap=0:replace_intrin=0:detect_leaks=0:${ASAN_OPTIONS}"
+      TPP_CLANGVER: "16.0.4"
+      TPP_GCCVER: "9.5.0" # Use an older version to match CUDA requirements
+      TPP_CUDAVER: "11.5"
```
hfp, line 61: This shall be changed to `12.2`.

tools/bench-ref/GPU/cuda/MatmulCUDA.cu
```diff
@@ -0,0 +1,37 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include <math.h>
+
+#include <cuda_runtime.h>
+
+__global__ void matmulKernel(float *A, float *B, float *C, int m, int n,
```
hfp, line 13: I guess const-ness does not really matter, but `A` and `B` can both be `const`.

tools/bench-ref/GPU/cuda/MatmulCUDA.cu
```diff
@@ -0,0 +1,37 @@
...
+
+  float sum = 0;
+  for (int i = 0; i < k; ++i) {
+    sum += A[row * k + i] * B[i * n + col];
+  }
+
+  C[row * n + col] += sum;
+}
+
+void matmulCUDA(float *A, float *B, float *C, int m, int n, int k) {
```
hfp, line 29: Const...

tools/bench-ref/GPU/cuda/MatmulCUDA.h
```diff
@@ -0,0 +1,11 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+void matmulCUDA(float *A, float *B, float *C, int m, int n, int k);
```
hfp, line 11: Const...

tools/bench-ref/GPU/cuda/MatmulRef.cpp
```diff
@@ -0,0 +1,195 @@
...
+  void runCUDA(float *A, float *B, float *C, int m, int n, int k) {
+    matmulCUDA(A, B, C, m, n, k);
+    cudaError_t syncStatus = cudaDeviceSynchronize();
+    if (syncStatus != cudaSuccess) {
+      std::cerr << "cudaDeviceSynchronize error : cuda code=" << syncStatus
+                << " - " << cudaGetErrorString(syncStatus) << "\n";
+    }
+  }
+
+  void runCUBLAS(float *A, float *B, float *C, int m, int n, int k) {
```
hfp, line 65: Const...

hfp (approved):
I made a few inline comments regarding const-ness. I also had a look at the Yaml files, etc. Looks good to me!

benchmarks/ref/include/Tensor.h
adam-smnk: About data movement outside vs inside kernel.
It's a bit of a design choice but we'll have to see what others do when we want to compare any numbers. For now, both in the reference CUDA implementation and MLIR kernels the args are copied to GPU outside of the timing loop.

In the future, it might be beneficial to get more exact numbers using some Nvidia profiler.

adam-smnk:
@rengolin any other comments?

rengolin (approved):
No, looks good as a first approach. Let's iterate in tree from now on.

PR #663, opened 2023-08-07
Move `createRewriteBatchMatmulToMatmulPass` 

chelini:
There is no reason to keep `createRewriteBatchMatmulToMatmulPass` as pass in `createTppConversionPass`. It is not related to tpp, it is a linalg to linalg pass.

Rename:
- linalg-to-loops.mlir
- linalg-tensor.mlir

More descriptive names.

Regression:

In `pass-rewrite-batch-matmul-to-matmul.mlir` the fully dynamic case is not mapped anymore to a linalg.matmul. ForAll tiling insert an affine min that prevents rank reduction. Unlcear what this is the case.


test/Passes/pass-rewrite-batch-matmul-to-matmul.mlir
```diff
@@ -74,16 +70,18 @@ func.func @batch_matmul_rewrite(%arg0: tensor<?x?x?xf32>,
...
-// CHECK: %{{.+}} = linalg.matmul ins(%[[SLICE]], %[[SLICE1]] : tensor<?x?xf32>, tensor<?x?xf32>) 
-// CHECK-SAME:  outs(%[[SLICE2]] : tensor<?x?xf32>) -> tensor<?x?xf32>
+// CHECK: %[[MIN:.+]] = affine.min #[[MAP]](%[[ARG5]])[%[[DIM0]]]
+// CHECK: %[[SLICE:.+]] = tensor.extract_slice %[[ARG0]][%[[ARG5]], 0, 0] [%[[MIN]], %[[DIM1]], %[[DIM2]]] [1, 1, 1] 
+// CHECK-SAME:  : tensor<?x?x?xf32> to tensor<?x?x?xf32>
+// CHECK: %[[SLICE1:.+]] = tensor.extract_slice %[[ARG1]][%[[ARG5]], 0, 0] [%[[MIN]], %[[DIM2]], %[[DIM3]]] [1, 1, 1] 
+// CHECK-SAME:  : tensor<?x?x?xf32> to tensor<?x?x?xf32>
+// CHECK: %[[SLICE2:.+]] = tensor.extract_slice %[[ARG6]][%[[ARG5]], 0, 0] [%[[MIN]], %[[DIM1]], %[[DIM3]]] [1, 1, 1] 
+// CHECK-SAME:  : tensor<?x?x?xf32> to tensor<?x?x?xf32>
+// CHECK: %{{.+}} = linalg.batch_matmul ins(%[[SLICE]], %[[SLICE1]] : tensor<?x?x?xf32>, tensor<?x?x?xf32>)
```
rengolin, line 86: Is this doing the original `batch_matmul` N times for a `tensor<Nx?x?>`, in parallel?

chelini: Sorry got stuck chasing a bug. It executes a batch_matmul with batch = 1 for N times. It seems that a canonicalization is missing to remove the affine.min and rank-reduce the tensor.

rengolin: Could it be because you pass as a separate argument and the compiler cannot prove that `%batch == dim(%arg0, 0)`? 

This kernel looks fragile, I would not write it that way in the first place.

Perhaps if you take the dim directly form the inputs, it can prove that the `min` is redundant and collapse the `matmul`.

chelini: The affine min should not be there `affine_map<(d0)[s0] -> (-d0 + s0, 1)>` is always 1, as d0 goes from 0 to DIM0 - 1 and you subtract DIM0. Upstream tiling using scf.forall may miss some canonicalization for affine min, but I don't know which one :)

rengolin: It seems `canonicalize` does work on min/max:
https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Affine/canonicalize.mlir#L1104

Have you tried adding it as a following pass in the RUN command line?

test/Passes/DefaultPipeline/default-tpp-passes.mlir
```diff
@@ -206,3 +206,15 @@ func.func @mlp(%arg0: tensor<128x256xf32>, %arg1: tensor<256x512xf32>,
...
   return %3 : tensor<128x512xf32>
 }
+
+// -----
+
+// CHECK-LABEL: batch_matmul_rewrite
+func.func @batch_matmul_rewrite(%arg0: tensor<512x32x64xf32>, %arg1: tensor<512x64x32xf32>) -> tensor<512x32x32xf32> {
+  %0 = tensor.empty() : tensor<512x32x32xf32>
+  // CHECK: scf.parallel
+  // CHECK: xsmm_brgemm_invoke
```
rengolin, line 216: Shouldn't this be `xsmm.gemm` given this pass?

chelini: The GEMM get's packed and mapped to a BRGEMM. The sequence here is: linalg.batch_matmul -> linalg.matmul -> generic (blocked matmul) -> tpp.brgemm

test/Passes/pass-rewrite-batch-matmul-to-matmul.mlir
chelini: yes, canonicalize in this case does not help.

rengolin: Can we make canonicalise work on this upstream then update this? I'm ok with a TODO for now if we create an issue and work on it soon.

rengolin (approved):

PR #664, opened 2023-08-07
Add pass to decompose aggregated operations

chelini:
Needed for softmax. Remove module from the mlp test in `tpp.mlir`.

### DIFF
--- a/lib/TPP/DecomposeAggregatedOps.cpp
+++ b/lib/TPP/DecomposeAggregatedOps.cpp
@@ -1,0 +1,53 @@
+//===- DecomposeAggregatedOps.cpp --------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+struct DecomposeAggregateOpsImpl : public OpRewritePattern<linalg::SoftmaxOp> {
+  using OpRewritePattern<linalg::SoftmaxOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::SoftmaxOp softmaxOp,
+                                PatternRewriter &rewriter) const override {
+    auto decomposableOp =
+        cast<linalg::AggregatedOpInterface>(softmaxOp.getOperation());
+    if (!decomposableOp)
+      return failure();
+    FailureOr<SmallVector<Value>> maybeNewResult =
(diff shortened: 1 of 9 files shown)


chelini:
Need #665

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -416,6 +417,12 @@ struct DefaultTppPasses : public DefaultTppPassesBase<DefaultTppPasses>,
       pm.addNestedPass<func::FuncOp>(createTppConversionPass());
       pm.addNestedPass<func::FuncOp>(createCleanupPass());
 
+      // Decompose Aggregated operations. These ops currently do not
+      // bufferize. Once this is possible we can move this pass after
+      // bufferization.
+      pm.addNestedPass<func::FuncOp>(createDecomposeAggregatedOpsPass());
```
adam-smnk, line 423: Maybe this could be done before TPP conversion in case any of the individual ops match.
But not sure if we really care atm.

lib/TPP/DecomposeAggregatedOps.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+struct DecomposeAggregateOpsImpl : public OpRewritePattern<linalg::SoftmaxOp> {
+  using OpRewritePattern<linalg::SoftmaxOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::SoftmaxOp softmaxOp,
+                                PatternRewriter &rewriter) const override {
+    auto decomposableOp =
+        cast<linalg::AggregatedOpInterface>(softmaxOp.getOperation());
+    if (!decomposableOp)
+      return failure();
```
adam-smnk, line 29: `dyn_cast` if it can fail or no check needed otherwise.

adam-smnk (approved):

rengolin (commented):
Why isn't this going upstream directly?

lib/TPP/DefaultTppPasses.cpp
chelini: moved. Let's see, once we have e2e tests we can reshuffle if needed.

xurui1995:
hi, I found this PR when I was trying to decompose `linalg.softmax`, may I ask do you know if there is a similar pass in the mlir repo, or do you plan to upstream this pass to mlir?

adam-smnk:
@xurui1995 It's just a simple wrapper over upstream Linalg decomposition utility. While one could add some similar generic pass wrapper upstream, I'd say it brings little value as you usually want more customized approach that control which operations and when you decompose.
Overall, I'd suggest you create your own small wrapper over the decomposition API.
But for testing purposes you can also find a similar test pass (`-test-linalg-decompose-ops`) upstream here:
https://github.com/llvm/llvm-project/blob/b7730a23efb222944b732bbdb3a7b965b7bffd98/mlir/test/lib/Dialect/Linalg/TestLinalgDecomposeOps.cpp#L21

xurui1995:
> @xurui1995 It's just a simple wrapper over upstream Linalg decomposition utility. While one could add some similar generic pass wrapper upstream, I'd say it brings little value as you usually want more customized approach that control which operations and when you decompose. Overall, I'd suggest you create your own small wrapper over the decomposition API. But for testing purposes you can also find a similar test pass (`-test-linalg-decompose-ops`) upstream here: https://github.com/llvm/llvm-project/blob/b7730a23efb222944b732bbdb3a7b965b7bffd98/mlir/test/lib/Dialect/Linalg/TestLinalgDecomposeOps.cpp#L21

thanks!

PR #666, opened 2023-08-10
Retire the rewrite to BRGEMM pass

chelini:
Fusion now is general enough and the tiling strategy used in tile and
fuse already exposes BRGEMMs. Thus, `rewrite-to-brgemm` becomes a simple
trivial rewrite. Adjust fusion to tile if the working group consists of
a single operation, we need to do this as we benchmarks and map 
single operations (for single op we tile only). 
Previously, fusion was assuming a working group
of at least two operations.

Fixes: https://github.com/plaidml/tpp-mlir/issues/392


lib/TPP/LinalgDeGeneralize.cpp
```diff
@@ -152,6 +156,56 @@ struct FillOpDeGeneralizationPattern
   }
 };
 
+// From linalg.generic to TPP brgemm (VNNI).
+struct TppBrgemmDeGeneralizationPattern
+    : public OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  bool isTppVnniOp(linalg::GenericOp linalgOp) const {
```
rengolin, line 164: This looks like it could be in a more generic place?

lib/TPP/LinalgDeGeneralize.cpp
```diff
@@ -152,6 +156,56 @@ struct FillOpDeGeneralizationPattern
...
+            .output(MatchOne(0), HasMap(EqualsTo(mapList[2])))
+            .region(MatchOne(0), WithOpChain<arith::MulFOp, arith::AddFOp>(
+                                     /*captures=*/nullptr));
+    return matmulMatcher.match(linalgOp);
+  }
+
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (!isTppVnniOp(linalgOp))
+      return failure();
```
rengolin, line 199: Can't the non-VNNI variant be like this as well?

rengolin (commented):
Nice!

lib/TPP/LinalgDeGeneralize.cpp
chelini: I moved in TppUtils.h. However I think we should pull out all the `isXXXOps` from TppUtils as we are going to use them also when converting linalg to xsmm, and including `TppUtils.h` looks borken.

lib/TPP/LinalgDeGeneralize.cpp
chelini: the non-VNNI variant is `BatchReduceOpDeGeneralizationPattern`. I had to keep `mlir::linalgx::rewriteToBRGemmOp` as it is used in transform. I have a patch to clean this up, if you want I can queue it here.

PR #667, opened 2023-08-11
Cleanup `getDataType` in XSMM and tile and fuse (NFC)

chelini:
- `getDataType` is now an utils. Need as utils for later patches.

- Use .td to specify dialects deps for tile and fuse.

### DIFF
--- a/include/TPP/Dialect/Xsmm/XsmmUtils.h
+++ b/include/TPP/Dialect/Xsmm/XsmmUtils.h
@@ -1,0 +1,27 @@
+//===- XsmmUtils.h - --------------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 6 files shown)


include/TPP/Passes.td
```diff
@@ -218,6 +218,8 @@ def TileConsumerAndFuseProducers : Pass<"tile-consumer-and-fuse-producers",
            "true", "Fuse from the last fusable consumer of the current target">,
     Option<"useForAll", "use-for-all", "bool", "true", "Use parallel forAll">
   ];
+  let dependentDialects = ["linalg::LinalgDialect", "scf::SCFDialect",
```
KavithaTipturMadhu, line 221: Is this an unrelated change?

KavithaTipturMadhu (approved):

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -0,0 +1,27 @@
...
+#include "TPP/Dialect/Xsmm/XsmmEnum.h"
+
+namespace mlir {
+class Type;
+class RewriterBase;
+
+namespace xsmm {
+namespace utils {
+
+xsmm::DataTypeAttr getXsmmDataType(RewriterBase &rewriter, Type type);
```
rengolin, line 21: Nit: If this is already in the `xsmm` namespace, does it really need to be called `getXsmmDataType`?

rengolin (approved):

PR #668, opened 2023-08-11
Linalg to XSMM: UnaryOps conversion

chelini:


lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -0,0 +1,338 @@
...
+struct ConvertFillOpToUnaryZero : public OpRewritePattern<linalg::FillOp> {
+  using OpRewritePattern<linalg::FillOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::FillOp fillOp,
+                                PatternRewriter &rewriter) const override {
+    if (!fillOp.hasBufferSemantics() || fillOp.hasDynamicShape() ||
+        !hasEqualTypes(fillOp)) {
+      return failure();
+    }
+    auto input = fillOp.getDpsInputOperands()[0];
```
rengolin, line 128: Can we not use the matchers?

test/Conversion/LinalgToXsmm/linalg-to-unary.mlir
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// CHECK-LABEL: transpose_op
+// CHECK-NOT: xsmm.unary transpose
+// CHECK: linalg.transpose
+
+// -----
+
+#map = affine_map<(d0, d1) -> (d0, d1)>
+func.func @relu(%arg0: memref<4x3xf32>) {
```
rengolin, line 78: There's also `linalg.max`

lib/TPP/ConvertLinalgToXsmm.cpp
chelini: Yes, we could use the matchers but it turned out to be more verbose here.

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+  int64_t ldi;
+  int64_t ldo;
+};
+} // namespace
+
+// Check if the strides associated with `operand` are valid strides
+// for XSMM: Strides must be statically known.
+static FailureOr<SmallVector<int64_t>> verifyStrides(Type operandType) {
+  assert(!isa<RankedTensorType>(operandType));
```
adam-smnk, line 49:
```suggestion
  assert(!isa<TensorType>(operandType));
```

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -0,0 +1,338 @@
...
+}
+
+static FailureOr<SmallVector<int64_t>> verifyStrides(OpOperand *operand) {
+  return verifyStrides(operand->get().getType());
+}
+
+// Return true if all the operand have the same type. No implicit conversion in
+// the linalgOp.
+static bool hasEqualTypes(linalg::LinalgOp linalgOp) {
+  // assert(linalgOp.getNumInitOperands > 0);
```
adam-smnk, line 79: needed?

lib/TPP/ConvertLinalgToXsmm.cpp
rengolin: The point of the matchers is not to reduce verbosity, but to make it clear what we're matching and be composable, so we can cascade them in a clear way. Going back to manually matching things makes it unclear and confusing.

lib/TPP/ConvertLinalgToXsmm.cpp
chelini: We should not allow unrankedTensor.

lib/TPP/ConvertLinalgToXsmm.cpp
rengolin: Nice matcher structure!

test/Conversion/LinalgToXsmm/linalg-to-unary.mlir
chelini: Since it is not strictly needed, I prefer tracking this as an issue and returning to it later (see: https://github.com/plaidml/tpp-mlir/issues/671).

chelini:
@rengolin any concern?

rengolin (approved):
Nope, LGTM, thanks!

Issue #669, opened 2023-08-14
Update CUDA driver on the CI.


chelini:
This is needed for using newer LLVM version (see: #665 )
We use 11.5 but it does not have CUtensorMap. I found CUtensorMap in newer version (12.x).

hfp:
This means upgrading all nvGPU nodes to CUDA 12.x (latest CUDA without limiting the general case). The latest version we have installed is v11.5. This must be done lockstep (cluster).

hfp:
Completed per #670.

PR #665, opened 2023-08-07
Bump LLVM

chelini:
- Upstream changes for AggregateOp interface

- Better bufferization for eltwise

PR #670, opened 2023-08-15
Issue #669: updated CUDA API to v12.2

hfp:


rengolin (changes requested):
This will change the LLVM version without fixing the delta and will break the build.

If the cluster is ready to take on the 12.2 change, then this should be done in #665 directly.

hfp:
On the other hand, basic tests like building `tpp-mlir` are failing with the new LLVM version hence I think #665 must go first.

adam-smnk:
It seems that after changing the CUDA version, the sanitizer kicks in when it should be disabled for the GPU tests.
`tpp-gpu` is optional so #665 could be merged to move things forward and I can investigate this CI failure separately.

rengolin:
> On the other hand, basic tests like building `tpp-mlir` are failing with the new LLVM version hence I think #665 must go first.

Not first, together. This change in #665 would probably fix the test, which is what we want. We don't need two different PRs.

hfp:
> > On the other hand, basic tests like building `tpp-mlir` are failing with the new LLVM version hence I think #665 must go first.
> 
> Not first, together. This change in #665 would probably fix the test, which is what we want. We don't need two different PRs.

Never mind. I can pull @chelini 's draft-PR into this PR.

hfp:
> It seems that after changing the CUDA version, the sanitizer kicks in when it should be disabled for the GPU tests. `tpp-gpu` is optional so #665 could be merged to move things forward and I can investigate this CI failure separately.

@adam-smnk Now, only the ASAN issue seems to be leftover.

adam-smnk:
CUDA not cooperating with asan is nothing new. But it seems that sth is even more broken in the newest drivers with CUDA 12+ as mentioned [here](https://github.com/google/sanitizers/issues/1630).
For now, simply disabling sanitizers in CUDA CI might be the only way forward.

rengolin (approved):

PR #672, opened 2023-08-17
Linalg to XSMM: AddOp conversion

chelini:


lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -260,6 +301,101 @@ struct ConvertGenericToUnaryRelu : public OpRewritePattern<linalg::GenericOp> {
...
+        genericOp.getMatchingIndexingMap(rhsOperand), /*operandIdx=*/1);
+    if (failed(broadCastFlagRhs))
+      return failure();
+
+    auto flagLhs =
+        xsmm::BinaryFlagsAttr::get(rewriter.getContext(), *broadCastFlagLhs);
+    auto flagRhs =
+        xsmm::BinaryFlagsAttr::get(rewriter.getContext(), *broadCastFlagRhs);
+
+    // Spaghetti code to handle 'NONE' as it conflicts with other flags; we
```
adam-smnk, line 379: Where does this conflict come from?

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
adam-smnk, line 1: Please add a test case with no broadcast

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -260,6 +301,101 @@ struct ConvertGenericToUnaryRelu : public OpRewritePattern<linalg::GenericOp> {
...
+    // best solution is to get rid of it.
+    ArrayAttr flags;
+    if (flagLhs.getValue() != xsmm::BinaryFlags::NONE &&
+        flagRhs.getValue() != xsmm::BinaryFlags::NONE) {
+      flags = rewriter.getArrayAttr({flagLhs, flagRhs});
+    } else if (flagLhs.getValue() != xsmm::BinaryFlags::NONE) {
+      flags = rewriter.getArrayAttr({flagLhs});
+    } else if (flagRhs.getValue() != xsmm::BinaryFlags::NONE) {
+      flags = rewriter.getArrayAttr({flagRhs});
+    }
```
adam-smnk, line 390: Doesn't this need one more default case to add NONE?

lib/TPP/ConvertLinalgToXsmm.cpp
chelini: the verifier. If we have "NONE" in the flags no other flags is allowed.

lib/TPP/ConvertLinalgToXsmm.cpp
chelini: yes..

PR #674, opened 2023-08-17
Enable GPU threading through tiling

adam-smnk:
Adds tiling to the GPU pipeline to allow kernel outlining to create more optimal kernel split into blocks and threads.

### DIFF
--- a/include/TPP/Passes.h
+++ b/include/TPP/Passes.h
@@ -110,7 +110,10 @@ createPackConv2DNchwFchwPass(ArrayRef<int64_t> blockingFactors = {});
 std::unique_ptr<OperationPass<func::FuncOp>>
 createPackConv2DNhwcHwcfPass(ArrayRef<int64_t> blockingFactors = {});
 std::unique_ptr<OperationPass<func::FuncOp>>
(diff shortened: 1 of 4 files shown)


chelini (approved):
By default we should already tile matmul by i = 32, j = 32, but I am fine exposing the options if it is easier.

adam-smnk:
Improved `GPU/gemm-fp32-1024-base.mlir` kernel performance from 26 gflops ([link](https://buildkite.com/intel/tpp-gpu/builds/353#018a0372-d34a-4be2-adba-7b47e03f03a4)) to 65.9 gflops ([link](https://buildkite.com/intel/tpp-gpu/builds/354#018a03f1-1e2f-4829-b8d6-95314bdb3af5)) as now the kernel is now split not only into blocks but also threads.
The base performance is still below naive CUDA ref performance due to suboptimal kernel operations. In case of matmul threads perform too many memory loads and stores.

adam-smnk:
> By default we should already tile matmul by i = 32, j = 32, but I am fine exposing the options if it is easier.

We might want to tile with 16x16 for Intel GPUs plus I felt it's better to have it stated explicitly in GPU pipeline.

PR #679, opened 2023-08-18
Enable linalg-to-xsmm in tpp-run

chelini:
Enable `linalg-to-xsmm` in tpp.run and fix mapping issue for binary and unary.
For unary and binary operations we need to use the captured operand and not
the operands of the linalg generic.

### DIFF
--- a/lib/TPP/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/ConvertLinalgToXsmm.cpp
@@ -88,7 +88,8 @@ static FailureOr<UnaryInfo> getUnaryInfo(Value input, Value output) {
 
 // Replace `linalgOp` with a unary dispatch plus invoke.
 static void replaceOpWithUnary(RewriterBase &rewriter,
-                               linalg::LinalgOp linalgOp, UnaryInfo unaryInfo,
+                               linalg::LinalgOp linalgOp,
+                               ArrayRef<Value> operands, UnaryInfo unaryInfo,
                                ArrayAttr flags, xsmm::UnaryKindAttr kind) {
   Location loc = linalgOp.getLoc();
   IntegerType integer64 = IntegerType::get(rewriter.getContext(), 64);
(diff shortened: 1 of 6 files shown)


test/Integration/tpp-run-xsmm-path.mlir
```diff
@@ -0,0 +1,44 @@
...
+  %added = call @add(%fill, %fill_1) : (tensor<2x2xf32>, tensor<2x2xf32>) -> tensor<2x2xf32>
+  %c0 = arith.constant 0 : index
+  %d1 = arith.constant -1.0 : f32
+ 
+  // CHECK: ( ( 3, 3 ), ( 3, 3 ) ) 
+  %v1 = vector.transfer_read %added[%c0, %c0], %d1 : tensor<2x2xf32>, vector<2x2xf32>
+  vector.print %v1 : vector<2x2xf32>
+  
+  bufferization.dealloc_tensor %arg0 : tensor<2x2xf32>
+  bufferization.dealloc_tensor %arg1 : tensor<2x2xf32>
```
adam-smnk, line 41: Is this manual memory management really needed?

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -101,17 +102,16 @@ static void replaceOpWithUnary(RewriterBase &rewriter,
       loc, integer64, kind, dims, flags, dtype);
   SmallVector<Value> invokeOperands;
   invokeOperands.push_back(dispatched);
-  invokeOperands.append(linalgOp->getOperands().begin(),
-                        linalgOp->getOperands().end());
+  invokeOperands.append(operands.begin(), operands.end());
```
adam-smnk, line 105: Does any test capture this change?

chelini: yes, relu_5 and add_9. The add has 2 operands and the xsmm lowering has 3 operands. Same for the relu 1 operand and 2 operands in xsmm.

test/Integration/tpp-run-xsmm-path.mlir
chelini: I saw this pattern in other tests. I can remove it

adam-smnk: It's only for weird edge cases in bufferization when there is (or at least used to be) issue with lifetimes.

test/Integration/tpp-run-xsmm-path.mlir
```diff
@@ -0,0 +1,41 @@
...
+      ^bb0(%in: f32, %in_1: f32, %out: f32):
+        %0 = arith.addf %in, %in_1 : f32
+        linalg.yield %0 : f32
+  } -> tensor<2x2xf32>
+  return %0 : tensor<2x2xf32>
+}
+
+func.func @entry() {
+  %cst_one = arith.constant 1.0 : f32
+  %arg0 = bufferization.alloc_tensor() : tensor<2x2xf32>
```
adam-smnk, line 25: nit: use empty instead?

adam-smnk (approved):

PR #680, opened 2023-08-21
CI: remove using "clxap" and "clxaptrb" partitions

hfp:


hfp:
This is NFC but required to accommodate lab refresh.

adam-smnk (approved):

PR #681, opened 2023-08-21
Linalg to GPU pass

adam-smnk:
Adds a custom lowering from linalg to set of operations more suitable for GPU computations.

This new pass aims to be the central place for generation of GPU optimized kernels from linalg ops. This can range from GPU-centric rewriting into lower-level ops to calls to specialized compute libraries.
The GPU pipeline applies the custom lowering first and then falls back to the standard lowering for any remaining ops.


adam-smnk:
Current GPU bench numbers from CI:
image

The improved `linalg.matmul` lowering increased the `base` kernel's performance to the level of `cuda` and `manual_kernel`.
Which means that the default GPU pipeline finally matched the naive hand implementation.

Future steps (PRs) will look into FC and MLP to check how the performance holds up with multiple computations.

hfp:
How much control do we have about GPU specific memory hierarchy? This question aims at utilizing specific kinds of memory on GPUs like "local store" or "shared memory"; terminology depends on the actual high-level programming model. I am aware, we may not talk about "high-level programming", but it is of interest if we can express utilizing properties that deviate (from a programming a CPU core).

I think we need control of what's done at [work-group level](https://downloads.ti.com/mctools/esd/docs/opencl/execution/terminology.html). Anyway, my comment(s) should not block this PR in any way (as it is just about "going forward").

If we assume, GPUs to be less general than CPUs (not necessarily true), we want to execute/test and lower "GPU-specific" kernels on CPU "just for the sake of comparison. We have multiple ways of doing so: (1) via the execution-backend like Vulkan based CPU device or whatever kind of "runner", e.g, mlir cpu runner, or OpenCL based CPU-device, (2) neutralizing specialties in the incoming IR/attributes and "simply" running it (on CPU).

It could be a starting point to run all "GPU specific" tests on CPU as well (or make this possible).

adam-smnk:
> How much control do we have about GPU specific memory hierarchy?

Good question, I'll look into it.

hfp:
> > How much control do we have about GPU specific memory hierarchy?
> 
> Good question, I'll look into it.

Best proof would be running your naive CUDA/GPU kernel on CPU. Naive in this context means we are still an order of magnitude away from what's perceived to be common sense. I really wonder if we can elide GPU-specifics and run on CPU just for comparison.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -0,0 +1,149 @@
...
+                                   .getType()
+                                   .cast<ShapedType>()
+                                   .getShape();
+    ArrayRef<int64_t> shapeB =
+        matmulOp.getInputs()[1].getType().cast<ShapedType>().getShape();
+    ArrayRef<int64_t> shapeA =
+        matmulOp.getInputs()[0].getType().cast<ShapedType>().getShape();
+    if (shapeB.size() == 3) {
+      return rewriter.notifyMatchFailure(matmulOp,
+                                         "Packed BF16 loops unsupported");
```
rengolin, line 55: Not all 3D matmul ops are VNNI. This needs a better matcher to make sure you get what you want here and give a reasonable error message if not.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -0,0 +1,149 @@
...
+    SmallVector<Value> ubs = {i, j, k};
+    // Lbs.
+    Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+    SmallVector<Value> lbs = {zero, zero, zero};
+    // Step.
+    Value one = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+    SmallVector<Value> steps = {one, one, one};
+    SmallVector<Value> ivs;
+
+    // Create parallel loops over the outer dimensions.
```
rengolin, line 71: This looks like packing, but for GPUs... We really need to use the same packing for all, and lower to GPUs after packing.

adam-smnk: Currently there is no packing involved in GPU lowering.
Linalg operation are only tiled to split them into GPU blocks operating on tiles which represent groups of threads per blocks.
Then this rewrite lowers linalg on tiles to elementwise operations such that they can be executed by individual threads.

Essentially, this rewrites `linalg.matmul` to a bit better kernel than the default `convert-linalg-to-parallel-loops` provides with respect to GPU.

lib/TPP/GPU/LinalgToGpu.cpp
adam-smnk: This check shouldn't be here at all as I go through named linalg op. To be removed.

lib/TPP/GPU/LinalgToGpu.cpp
rengolin: Hm, I see. This will play with the packing, then, we need these two to talk to each other later on.

adam-smnk: Depending on packing I'd expect need to generate different element-wise kernels. There will definitely be some need to coordinate the two but at the moment I have no examples. Although hopefully this last step will be replaced by some tile-based GPU kernel library like we do with libxsmm.

hfp (approved):
LGTM.

The following comment is not necessarily meant to be addressed by this PR. Looking at `LinalgToGpu.cpp` and corresponding CMake update, I wonder if the translation unit is a misnamer like having a subfolder `LinalgToGpu` with, e.g., `MatmulOp.cpp` inside. Given many named ops, it may blow-up the translation unit or readability with future/additional ops.

adam-smnk:
@hfp Fair point, I might need to reorganized some of these passes. It might make sense to have a bit larger pass like `linalg-to-gpu-kernel`.

PR #685, opened 2023-08-23
Enable named op conversion to xsmm for matmul and brgemm

chelini:
Use xsmm gemm instead of brgemm when batch is 1.

### DIFF
--- a/lib/TPP/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/ConvertLinalgToXsmm.cpp
@@ -437,30 +437,47 @@ static void replaceOpWithBrgemm(RewriterBase &rewriter,
   int64_t strideA = brgemmInfo.strideA;
   int64_t strideB = brgemmInfo.strideB;
 
-  DenseI64ArrayAttr dims = DenseI64ArrayAttr::get(
-      rewriter.getContext(),
-      ArrayRef<int64_t>{loops[m], loops[n], loops[k], lda, ldb, ldc, strideA,
-                        strideB});
+  bool hasBatch = batch != std::numeric_limits<unsigned>::max();
+
   auto dtype = xsmm::utils::getDataType(
       rewriter, linalgOp.getDpsInitOperands()[0]->get().getType());
   IntegerType integer64 = IntegerType::get(rewriter.getContext(), 64);
   Location loc = linalgOp.getLoc();
-  auto flags = rewriter.getArrayAttr(
-      xsmm::GemmFlagsAttr::get(rewriter.getContext(), xsmm::GemmFlags::NONE));
-  Value dispatched = rewriter.create<xsmm::BrgemmDispatchOp>(
-      loc, integer64, dims, flags, dtype);
 
-  unsigned batchVal = 1;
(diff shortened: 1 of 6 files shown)


lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -437,30 +437,47 @@ static void replaceOpWithBrgemm(RewriterBase &rewriter,
   int64_t strideA = brgemmInfo.strideA;
   int64_t strideB = brgemmInfo.strideB;
 
-  DenseI64ArrayAttr dims = DenseI64ArrayAttr::get(
-      rewriter.getContext(),
-      ArrayRef<int64_t>{loops[m], loops[n], loops[k], lda, ldb, ldc, strideA,
-                        strideB});
+  bool hasBatch = batch != std::numeric_limits<unsigned>::max();
```
rengolin, line 440:
```suggestion
  bool hasBatch = (batch != std::numeric_limits<unsigned>::max());
```

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -437,30 +437,47 @@ static void replaceOpWithBrgemm(RewriterBase &rewriter,
...
-  Value batchDim = rewriter.create<arith::ConstantOp>(
-      loc, integer64, rewriter.getIntegerAttr(integer64, batchVal));
-  SmallVector<Value> invokeOperands;
-  invokeOperands.push_back(dispatched);
-  invokeOperands.append(linalgOp->getOperands().begin(),
-                        linalgOp->getOperands().end());
-  invokeOperands.push_back(batchDim);
-  rewriter.replaceOpWithNewOp<xsmm::BrgemmOp>(linalgOp, dtype, invokeOperands);
+
+  if (hasBatch) {
```
rengolin, line 447: These two codes can be merged into one by selecting the flags' strides and adding the `batchVal` inside an `if (hasBatch)` and leaving the rest identical.

chelini: I moved the flags out. Note that the dispatch calls are not the same, and in the brgemm we have an extra parameter to pass, the batch.

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -798,12 +811,46 @@ void ConvertLinalgToXsmm::runOnOperation() {
...
+  using OpRewritePattern<linalg::MatmulOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::MatmulOp matmulOp,
+                                PatternRewriter &rewriter) const override {
+    auto gemmInfo = isMappableToBrgemm(matmulOp);
+    if (failed(gemmInfo)) {
+      llvm::errs() << "Failed\n";
+      return failure();
+    }
+    replaceOpWithBrgemm(rewriter, matmulOp, *gemmInfo);
```
adam-smnk, line 826: nit: this function name should be updated as it's not only brgemm anymore

adam-smnk (approved):

PR #686, opened 2023-08-23
Unpack optimization

KavithaTipturMadhu:


lib/TPP/ConvertPackUnpackOptimization.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+using namespace mlir;
+using namespace mlir::tpp;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+template <typename Op, bool IsPack>
```
rengolin, line 25: No need for the second argument. You can create a `constexpr bool` with a match on the `Op`.

lib/TPP/ConvertPackUnpackOptimization.cpp
```diff
@@ -0,0 +1,216 @@
...
+         &tiledDims](OpBuilder &rewriter, Location loc, ValueRange localIvs,
+                     ValueRange iterArgs) -> scf::ValueVector {
+          SmallVector<OpFoldResult> extractSliceOffsets;
+          SmallVector<OpFoldResult> extractSliceStrides;
+          SmallVector<OpFoldResult> extractSliceSizes;
+          SmallVector<OpFoldResult> insertSliceOffsets;
+          SmallVector<OpFoldResult> insertSliceStrides;
+          SmallVector<OpFoldResult> insertSliceSizes;
+
+          if (IsPack) {
```
rengolin, line 84: I can't see the difference between `IsPack` and not, but it doesn't look big. Can you extract the differences and have a single block of code for both?

lib/TPP/ConvertPackUnpackOptimization.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+          return {insertSliceOp};
+        });
+    rewriter.replaceOp(packOp, loopNest.loops[0].getResults()[0]);
+    return success();
+  }
+};
+
+void populatePackUnpackOptimizationPatterns(RewritePatternSet &patterns) {
+  // clang-format off
```
rengolin, line 193: there's no need for clang-format to be off here, this is a long line

lib/TPP/ConvertPackUnpackOptimization.cpp
KavithaTipturMadhu: Problem is offsets, strides and sizes are interchanged for extract and insert slice operations in case of pack and unpack operations. So it's a little tricky to refactor this into common code of some kind. @alheinecke 
is of the opinion that we can refactor it after merging when we get to it while working on vnni pack and unpack.

lib/TPP/ConvertPackUnpackOptimization.cpp
```diff
@@ -0,0 +1,189 @@
...
+         &IsPack](OpBuilder &rewriter, Location loc, ValueRange localIvs,
+                  ValueRange iterArgs) -> scf::ValueVector {
+          SmallVector<OpFoldResult> extractSliceOffsets;
+          SmallVector<OpFoldResult> extractSliceStrides;
+          SmallVector<OpFoldResult> extractSliceSizes;
+          SmallVector<OpFoldResult> insertSliceOffsets;
+          SmallVector<OpFoldResult> insertSliceStrides;
+          SmallVector<OpFoldResult> insertSliceSizes;
+
+          auto setArgs = [&numLoops, &tiledDims, &rewriter, &localIvs, &packOp,
```
rengolin, line 86: This looks dangerous. Not only the lambda is overriding the same variables from the main function, but it's doing so in the opposite meaning for one case and not the other. People are going to get confused over this code.

I suggest you rename it to something like `sourceSlice*` and `destSlice*`. Also, please add a comment before creating the lambda to understand its construction.

rengolin (approved):

PR #687, opened 2023-08-24
GPU kernel eltwise operation fusion

adam-smnk:
Extends linalg to GPU lowering with a simple fusion strategy to allow efficient GPU kernel generation by combining matmul computation with following element wise operations.
Adds a fully connected kernel to GPU benchmarks.

The fusion is added for both scalar and WMMA lowering.

### DIFF
--- a/lib/TPP/GPU/LinalgToGpu.cpp
+++ b/lib/TPP/GPU/LinalgToGpu.cpp
@@ -94,6 +94,191 @@ static bool supportsMMACompute(linalg::LinalgOp linalgOp) {
          cType.getElementType().isF16() && m == 16 && n == 16 && k == 16;
 }
 
+Operation *mmaFusion(linalg::LinalgOp rootOp, linalg::LinalgOp consumer,
+                     gpu::SubgroupMmaStoreMatrixOp rootStoreOp,
+                     ValueRange storeIndices, PatternRewriter &rewriter) {
+  Location loc = rootOp.getLoc();
+
+  auto rootOutput = rootOp.getDpsInitOperands()[0]->get();
+  auto outputType = rootOutput.getType().cast<ShapedType>();
+  gpu::MMAMatrixType mmaOutputType = gpu::MMAMatrixType::get(
+      outputType.getShape(), outputType.getElementType(), "COp");
+  auto leadingDim = rootStoreOp.getLeadDimension();
+
+  Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+
+  // Insert fused eltwise ops before the store and later replace the store
+  // with a new result.
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(rootStoreOp);
+
+  Operation *newStore = nullptr;
+  SmallVector<Value> operands;
+  if (structured_match::utils::isTwoDAddOp(consumer, &operands)) {
+    // Get the value to be added - load the tile first.
+    // Must be a buffer of the same type - scalar broadcast is not supported.
+    auto addValue = operands[0] != rootOutput ? operands[0] : operands[1];
+    if (addValue.getType() != rootOutput.getType())
+      return nullptr;
+    // Fuse the add into the matmul body.
+    addValue = rewriter
+                   .create<gpu::SubgroupMmaLoadMatrixOp>(
+                       loc, mmaOutputType, addValue, ValueRange{zero, zero},
+                       leadingDim,
+                       /*transpose=*/UnitAttr())
+                   .getRes();
+    auto eltwiseAttr = isa<FloatType>(outputType.getElementType())
+                           ? gpu::MMAElementwiseOp::ADDF
+                           : gpu::MMAElementwiseOp::ADDI;
+    auto addRes =
+        rewriter
+            .create<gpu::SubgroupMmaElementwiseOp>(
+                loc, mmaOutputType, ValueRange{rootStoreOp.getSrc(), addValue},
+                eltwiseAttr)
+            .getRes();
+    // Store the new result.
+    newStore = rewriter.replaceOpWithNewOp<gpu::SubgroupMmaStoreMatrixOp>(
+        rootStoreOp, addRes, rootStoreOp.getDstMemref(), ValueRange{zero, zero},
+        leadingDim,
+        /*transpose=*/UnitAttr());
+  } else if (structured_match::utils::isTwoDReluOp(consumer, &operands)) {
+    // Must be a floating point type.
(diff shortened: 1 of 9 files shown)


adam-smnk:
image

FC benchmark kernel (`cuda_fc`) achieves the same gflops as the base matmul only kernel (`cuda_gemm`) as expected.

chelini (commented):
High-level question, why cannot we use the fusion we have? My fear is that we are going to create two parallel codegens.

adam-smnk:
> High-level question, why cannot we use the fusion we have? My fear is that we are going to create two parallel codegens.

I depend on the current `tile-and-fuse` we have to gather ops within the same loop.
This fusion is specifically for generating GPU kernels at the memref level where the flow is:
- lower matmul (root) op into a GPU-friendly loops
- pull following eltwise consumer into the root loop

Lowering first to loops and then doing parallel loop fusion does not work ideally plus it would need a separate cleanup to remove unnecessary store/load pairs after loop fusion.
I find it much easier to generate better GPU kernels straight from linalg level.

adam-smnk:
> So do you actually use our tile-and-fuse? I cannot see from the diff. What you describe is what tile and fuse does. You get a matmul you tile using the tiling interface and then fuse into the loop nest.

I use it in the overall GPU pipeline to slice linalg into device.
What I do here essentially I need is `lower-and-fuse` at memref level. Perhaps I could better reuse some linalg interfaces.
Of course, ideally I would want to preserve fusion information from tensor level through bufferization and skip need for any kind of fusion analysis at memref. But that is a missing piece atm.

rengolin:
> So do you actually use our tile-and-fuse? I cannot see from the diff. What you describe is what tile and fuse does. You get a matmul you tile using the tiling interface and then fuse into the loop nest.

Not all fusion can happen at Linalg, because it cannot represent non-perfectly nested loops. Eventually, some fusion will have to be done in loops. 

The GPU case is probably a bit orthogonal too, since that's about "what should we bundle into one call to reduce target dispatch" which also crosses linalg calls.

adam-smnk:
image

Local run of the two new GPU benchmarks.
MLP is slower as the measurement includes copy of the constants into device memory.

adam-smnk:
After investigating some possible alternative fusion strategies, it does not appear that there would be any advantage going with a grouping like a region fusion op or a separate outlining step.

So, the strategy remains to perform a simple fusion on memrefs. The main assumption is that the parallel loop body should have simple and nicely fusable layout. Appropriate constraints are placed on the consumer op selection before fusion is performed.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -93,6 +94,200 @@ static bool supportsMMACompute(linalg::LinalgOp linalgOp) {
          cType.getElementType().isF16() && m == 16 && n == 16 && k == 16;
 }
 
+// Fuse a consumer using WMMA operations.
+// Returns updated store op or nullptr if the fusion fails.
+static Operation *mmaFusion(linalg::LinalgOp rootOp, linalg::LinalgOp consumer,
```
chelini, line 99: failureOr ?

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -93,6 +94,200 @@ static bool supportsMMACompute(linalg::LinalgOp linalgOp) {
...
+  // with a new result.
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(rootStoreOp);
+
+  Operation *newStore = nullptr;
+  SmallVector<Value> operands;
+  if (structured_match::utils::isTwoDAddOp(consumer, &operands)) {
+    // Get the value to be added - load the tile first.
+    // Must be a buffer of the same type - scalar broadcast is not supported.
+    auto addValue = operands[0] != rootOutput ? operands[0] : operands[1];
```
chelini, line 123: nit: `(operands[0] != rootOutput)` ...

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -93,6 +94,200 @@ static bool supportsMMACompute(linalg::LinalgOp linalgOp) {
...
+                eltwiseAttr)
+            .getRes();
+    // Store the new result.
+    newStore = rewriter.replaceOpWithNewOp<gpu::SubgroupMmaStoreMatrixOp>(
+        rootStoreOp, addRes, rootStoreOp.getDstMemref(), ValueRange{zero, zero},
+        leadingDim,
+        /*transpose=*/UnitAttr());
+  } else if (structured_match::utils::isTwoDReluOp(consumer, &operands)) {
+    // Must be a floating point type.
+    auto floatType = dyn_cast<FloatType>(outputType.getElementType());
```
chelini, line 149: why is this not checked for the add? Can we simply constraint the matcher to match only float ops?

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -93,6 +94,200 @@ static bool supportsMMACompute(linalg::LinalgOp linalgOp) {
...
+                                       Operation *rootStoreOp,
+                                       ValueRange storeIndices,
+                                       PatternRewriter &rewriter) {
+  auto parentOp = rootOp->getParentOp();
+  auto rootOutput = rootOp.getDpsInits()[0];
+
+  // Traverse other ops within the same region and collect consumers.
+  SmallVector<linalg::LinalgOp> consumers;
+  Operation *nextOp = rootOp;
+  while ((nextOp = nextOp->getNextNode())) {
```
chelini, line 248: Is this safe?, example:
```
matmul (%A, %B) (%C)
func (%A, %B, %C) // can write/read to all the operands
relu (%C)
```
If you fuse and the relu and the matml, you can get wrong results.

lib/TPP/GPU/LinalgToGpu.cpp
adam-smnk: I ran into some weird lifetime issues when passing around FailureOr so I just used simpler std::optional.

lib/TPP/GPU/LinalgToGpu.cpp
adam-smnk: Constrained to floats only in general.

lib/TPP/GPU/LinalgToGpu.cpp
adam-smnk: There are various checks for that.
In general, I collect only linalgs and stop walking over ops any other operation.
Anyway, added a test case like your example.

chelini (approved):

PR #690, opened 2023-08-24
Retire `createConvertMemRefToTppPass`

chelini:
It is better to convert memref to xsmm directly without passing via TPP operations. This simplifies the pipeline (see `tensor-to-unary.mlir`) and avoid TPP dependency.


lib/TPP/ConvertMemRefToXsmm.cpp
```diff
@@ -0,0 +1,61 @@
+//===- LinalgMemRefToXsmm.cpp ------------------------------------*- C++-*-===//
```
adam-smnk, line 1: nit: file name

PR #691, opened 2023-08-25
Add pack/unpack benchmarks

chelini:


rengolin:
> Can you actually also add two hard-coded benchmarks for large matrices (>512):

Ignore me, we already have those tests and they're already showing improvements!

benchmarks/config/base/base.json
```diff
@@ -113,4 +113,31 @@
       "extensions": [ "avx2" ]
     }
   }}
+  {
+    "pack": {
+      "gemm_operand_a_512x1024": {
```
rengolin, line 118: Call them by their types `fp32` / `bf16`, since soon we'll have both types of tests.

rengolin (approved):

PR #693, opened 2023-08-29
Conda for more reproducible environment

hfp:
* Added Conda environment (requirements.yml) and Python environment (requirements.txt).
* CMakeLists.txt: include Sanitizer module as late as possible (append)
  - Error messages w/o fix are misleading (cannot find OpenMP libraries, etc).
  - Interacts badly with PThreads and upfront packages.
* Uncovered issues (somewhat unrelated)
  - build_tpp.sh: check actual value of supplied keywords (more scripts suffer from it).
  - lint-python.sh: check if Python linter works and avoid printing an empty variable.


hfp:
This is an initial PR for Conda support. It's complemented by changes in `tpp-scripts`.

Goals:
* Keep current operation based on cluster environment (Conda disabled by default).
* Mixed operation possible, i.e., Conda and Cluster environment.
* Conda can be enabled without additional PR.

Regarding mixed operation, this is necessary until all bits are rebuilt or Conda based like/including LLVM, and all GPU runtimes. Next steps will focus on GPU environment.

requirements.yml
```diff
@@ -0,0 +1,13 @@
+channels:
+  - conda-forge
+dependencies:
+  - python
+  - clang=16
+  - clang-format
+  - compiler-rt_linux-64
```
rengolin, line 7: why do we need compiler-rt?

requirements.yml
```diff
@@ -0,0 +1,13 @@
+channels:
+  - conda-forge
+dependencies:
+  - python
+  - clang=16
+  - clang-format
```
rengolin, line 6: clang-format also needs to be 16

requirements.yml
```diff
@@ -0,0 +1,13 @@
...
+  - compiler-rt_linux-64
+  - gcc_linux-64=12
+  - gxx_linux-64=12
+  - llvm-openmp
+  - cmake
+  - ninja
+  - lld
```
rengolin, line 13: should we also fix lld to 16?

scripts/buildkite/build_tpp.sh
```diff
@@ -33,10 +33,10 @@ fi
 if [ "${COMPILER}" == "clang" ]; then
   GCC_COMPAT_OPTION="-g ${GCC_TOOLCHAIN}"
 fi
-if [ "${SANITIZERS}" ]; then
+if [ "${SANITIZERS}" ] && [ "0" != "${SANITIZERS}" ]; then
```
rengolin, line 36: I'd just make:
```
if [ "1" == "${SANITIZERS}" ]; then
```

requirements.yml
hfp: Mainly because it contains the sanitizer libraries.

requirements.yml
hfp: Indeed, overlooked it. Btw, it seems to resolve to v16 because clang is v16.

requirements.yml
hfp: I think I ran into a problem with this, at least in the beginning. But let me retry.

requirements.yml
hfp: Btw, it seems impossible to find out a package given a file name (that supposedly belong to it).

scripts/buildkite/build_tpp.sh
hfp: Yeah, then SANITIZER=2 would not match although unequal zero. Anyway, the current pattern is used elsewhere also.

rengolin: Who would ever use `=2`? I'm trying to simplify the pattern elsewhere too.

hfp: Completed.

rengolin (approved):

PR #695, opened 2023-08-30
MHA benchmarks

chelini:
Add MHA benchmarks, do not expect good performance. The xsmm flow is not enable yet and more work is needed in main to map the kernels to xsmm.


benchmarks/mlir/fp32-out-softmax-times-value.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: tpp-run %s -n 10 \
+// RUN:  -e entry -entry-point-result=void
+
+// BENCH_TOTAL_FLOPS: 67108864
```
adam-smnk, line 4: Are the flops here and in the `query-times-key` correct?
It looks like the two kernels have different argument sizes.

chelini: Looks correct to me:
`softmax-times-value`: 64 * 32 * 8 * 64 * 32 (reduction) * 2 = 67108864
`query-times-key`: 64 * 8 * 32 * 32 * 64 (reduction) * 2 = 67108864

adam-smnk (approved):

PR #697, opened 2023-08-30
Run fusion for a fixed-point number of iterations

chelini:
When running more complex benchmarks (i.e., MHA), fusion opportunities can be exposed by running the pass multiple times. Currently, the pass runs three times as this is needed for MHA; we may want more sophisticated heuristics in the future. Retire also the option `start-from-last-fusable-consumer` as this is always profitable. Drop `tile-and-fuse-with-cse.mlir` as the allocation in the loops is needed because it iterates over a tile of 32x32.


lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -704,72 +771,27 @@ struct TileConsumerAndFuseProducers
...
+        // `populateFoldUnitExtentDimsViaSlicesPatterns` works only with
+        // linalg.generic we need to generalize first using
+        // `populateLinalgNamedOpsGeneralizationPatterns`.
+        linalg::ControlDropUnitDims options;
+        options.rankReductionStrategy = linalg::ControlDropUnitDims::
+            RankReductionStrategy::ExtractInsertSlice;
+        linalg::populateFoldUnitExtentDimsPatterns(patterns, options);
+        tensor::populateMergeConsecutiveInsertExtractSlicePatterns(patterns);
+        linalg::populateLinalgNamedOpsGeneralizationPatterns(patterns);
+        (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
```
adam-smnk, line 792: Is it needed to run on every fusion iteration?

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -704,72 +771,27 @@ struct TileConsumerAndFuseProducers
...
-    // Compute the tile sizes for each contraction operations or
-    // use the default one.
-    llvm::DenseMap<Operation *, SmallVector<OpFoldResult>> defaultTiles;
-    for (auto contractionOp : linalgContractionOperations) {
-      auto tiles = getDefaultTileSizes(contractionOp, this->tileSizes);
-      if (failed(tiles)) {
-        LLVM_DEBUG(llvm::dbgs() << "Failed to compute default tile sizes for: "
-                                << contractionOp << "\n");
-        return;
+    int fixedPoint = 3;
```
adam-smnk, line 774: Any test covers/showcases this new approach?

lib/TPP/TileConsumerAndFuseProducers.cpp
chelini: yes, you need to rank reduce to expose opportunities.

chelini: perhaps we could pull out `populateLinalgNamedOpsGeneralizationPatterns`.

chelini: (ignore last comment, we need it).

lib/TPP/TileConsumerAndFuseProducers.cpp
chelini: I can extract a smaller test case from MHA

lib/TPP/TileConsumerAndFuseProducers.cpp
```diff
@@ -704,72 +771,27 @@ struct TileConsumerAndFuseProducers
...
-                            defaultTiles, fusedOps, this->maxDepth);
-        LLVM_DEBUG(llvm::dbgs() << "\n\n");
-        if (succeeded(fuseAndTileResult)) {
-          rewriter.replaceOp(
-              linalgOp,
-              (*fuseAndTileResult).replacements[linalgOp->getResults()[0]]);
-        }
-      }
-    }
+    } while (fixedPoint--);
```
adam-smnk, line 794: Now it runs 4 iterations.
Either `--fixedPoint` or `while {}` loop for clarity.

test/Passes/pass-tile-and-fuse-mha.mlir
```diff
@@ -38,6 +37,8 @@ func.func @mha_tensorflow(%arg1: tensor<64x32x8x64xf32>,
   %0 = tensor.empty() : tensor<64x32x8x64xf32>
   
   // CHECK: %{{.+}} = scf.forall (%{{.+}}) in (%[[C64]])
+  // Should be an scf.forall
```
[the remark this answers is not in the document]
chelini: need to investigate.

lib/TPP/TileConsumerAndFuseProducers.cpp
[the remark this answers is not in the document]
chelini: ok

lib/TPP/TileConsumerAndFuseProducers.cpp
chelini: true, my bad.

adam-smnk (approved):

PR #698, opened 2023-08-31
Remove unused option (NFC)

chelini:


rengolin (approved):

rengolin:
@chelini make sure you install the Python linter locally to catch that kind of error before submitting the PR.

PR #699, opened 2023-08-31
Enable linalg-to-xsmm in MHA benchmarks

chelini:


adam-smnk (approved):
Looks good overall. Let's see if benchmark results change.

rengolin:
> > Looks good overall. Let's see if benchmark results change.
> 
> MHA kernel perf improved 👍

Considerably!! :smiley:

PR #700, opened 2023-09-01
Tensor init F16

adam-smnk:
Adds tensor initialization for 'f16' type.
Useful for testing GPU kernels.


chelini (approved):
Looks good to me.

PR #702, opened 2023-09-05
Descend linalgToXsmm decision into TPP passes

rengolin:
This allows us to combine the existing TPP passes to the Linalg-to-XSMM path without the need to re-write all the pipeline for both. The decision to lower to TPP at Tensor level or not is accompanied by the decision to lower either TPP or Linalg to XSMM, respectively.


rengolin:
Ran benchmarks locally, no changes against `main`.

adam-smnk (approved):

chelini (approved):

PR #703, opened 2023-09-05
Fold flags 

chelini:
I see 5% perf impr. on MHA.

### DIFF
--- a/lib/TPP/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/ConvertLinalgToXsmm.cpp
@@ -15,6 +15,7 @@
 #include "TPP/TransformUtils.h"
 #include "TPP/Transforms.h"
 #include "TPP/ValueUtils.h"
+#include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Linalg/Utils/Utils.h"
@@ -35,6 +36,10 @@ namespace {
 
 struct ConvertLinalgToXsmm
     : public ConvertLinalgToXsmmBase<ConvertLinalgToXsmm> {
+  void runOnOperation() override;
+};
+
+struct FoldXsmmFlags : public FoldXsmmFlagsBase<FoldXsmmFlags> {
   void runOnOperation() override;
 };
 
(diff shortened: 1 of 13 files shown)


chelini:
Forgot to add the folding pass to the default pipeline. Marking as draft for now.

rengolin:
PS: no need to mark "do not fold", you'll be the one merging anyway. :)

adam-smnk:
Any impact on other benchmarks compared to main?

chelini:
> Any impact on other benchmarks compared to main?

I see 3% to %5 impr. on MHA locally.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -256,18 +256,15 @@ struct TppMappingPass : public TppMappingBase<TppMappingPass>,
     pm.addPass(createConstantFoldPackPass());
     pm.addPass(createSimplifyAndCanonicalizePackPass());
 
-    // Looks like we want to agressively remove tensor.empty before fusion.
-    // See: `test/Passes/tile-and-fuse-with-cse.mlir`.
-    pm.addPass(createCleanupPass());
     pm.addPass(createTileConsumerAndFuseProducersPass());
-    pm.addPass(createCleanupPass());
+    pm.addPass(createCanonicalizerPass());
```
adam-smnk, line 260: I get the idea but it seems a bit brittle.
The full default pipeline runs cleanup in a few more places before it gets to XSMM lowering.
This will probably break after #702

lib/TPP/ConvertLinalgToXsmm.cpp
```diff
@@ -720,6 +725,81 @@ makeMinorDimensionsInnerMost(RewriterBase &rewriter, linalg::GenericOp linalgOp,
   return linalgOp;
 }
 
+template <typename OpTy>
+static LogicalResult foldFlags(RewriterBase &rewriter, OpTy gemmDispatchOp) {
+  SmallVector<Operation *> users(gemmDispatchOp.getResults().getUsers());
+  if (users.size() != 1)
+    return failure();
```
adam-smnk, line 732: Wouldn't it be more versatile to start from invokes, look at their operands, and update their dispatch if they happen to be zero initialized?

lib/TPP/DefaultTppPasses.cpp
chelini: To me running CSE as a cleanup is not correct, what are the advantages?. The alternative solution is to do a pass that canonicalize the IR before and after fusion, basically undoing CSE.

lib/TPP/ConvertLinalgToXsmm.cpp
chelini: If we use pattern rewriter, the root op needs to be either updated in-place, replaced, or erased (https://mlir.llvm.org/docs/PatternRewriter/#restrictions); since I don't touch the invoke, this would
not be the correct usage of pattern rewriter.

lib/TPP/DefaultTppPasses.cpp
chelini: I see CSE as an optimization not a cleanup.

adam-smnk: Not entirely sure how these two concepts are connected. Tensor semantics define that the two forms are equivalent and the "less is more" is usually more applicable 😉 

While I agree there might be little benefit for us in this case in applying CSE, I think it's a bit unrealistic to expect upstream passes or 10 other passes in between to never run CSE.

chelini: To which concepts are you referring? I agree that the IR looks equivalent, but I can't entirely agree with how we use CSE. We consider it as a cleanup pass for no good reason. In the worst case,  if the IR is not in the canonical form, we will not fold the zero. Keep in mind that the canonical form is only needed by the folding pass; you can run CSE before fusion. However, if you run CSE after fusion, you may not fold the zero.

chelini: The bottom line is:

1. I am more concerned about having CSE as a cleanup pass than having CSE in the default pipeline. I did not see any advantages to it being a cleanup pass. "It just happens to be there".

2. Assuming IR in a given shape is fine before running a pass, as we are building a pipeline. Either by carefully looking at the passes we inject in the default pipeline or some more canonical pass, I don't care.

adam-smnk: I wouldn't call intricate chain of operations exactly canonical.
I don't mind assumption on IR within a single pass or worst case some one-shot bundle of passes. But between fusion and folding there are so many independent transformation that it is a large assumption that might brake 3 new passes in the future when CSE is beneficial for sth else.

The cleanup as it currently stands is there to promote (force) some degree of modularity in our pipeline. I would consider CSE fundamental enough that it is large requirement to assume its absence.

Anyway, is lack of CSE required for folding only because the pass cannot detect the zero fill properly? Then the pass's analysis is too simple.
Or does it create some other issues?

lib/TPP/ConvertLinalgToXsmm.cpp
adam-smnk: That's fair but then we could use a different tool.
Anyway, it's just a general thought. If it is fine as is, then no changes needed.

lib/TPP/DefaultTppPasses.cpp
chelini: It is not an intricate chain of operation but simply:
```
%0 = tensor.empty()
%1 = linalg.fill ins(...) outs(%0)
%2 = some contraction ins(...) outs(%1)
```
I miss what kind of modularity CSE adds to the pipeline. The fact that we copy and paste the cleanup in multiple stages in the pipeline? It should only be there if it is helpful, and I did not find any advantages. We could run a pass before the folding that attempts to canonicalize the IR as above, but it is difficult as it is a memref level already.

Putting this aside, if you run CSE after fusion and assuming you have:

```
%cst = arith.constant 0.0 : f32

%0 = tensor.empty() : tensor<3x3xf32>
%1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<3x3xf32>)
%2 = linalg.contraction ins(...) outs(%1)

// CSE: replace a possible empty + fill with %1.
%3 = linalg.contraction ins(...) outs(%1)
```
It will bufferize as:

```
%alloc = memref.alloc() : memref<3x3xf32>
%1 = xsmm.zero (%alloc)
%2 = xsmm.gemm (..., %alloc)

%alloc_1 =  memref.alloc() : memref<3x3xf32>
memref.copy (%alloc, %alloc_1)
%3 = xsmm.gemm (..., %alloc_1)
```
You cannot fold the zero in the second operation. You can extend the pattern to handle this case too but at memref level everything is more complicated.

adam-smnk: It is a chain that has to survive all the way from tensors through multiple passes and bufferization.

CSE in itself doesn't add to modularity, of course. But the current approach assumes we run our pipeline in stages (tensor-level, bufferization, memref-level, down to LLVM). So the current assumption is that cleanup is reasonably expected to happen at least in between these major steps. IIRC, this was somewhat motivated by pass construction in IREE.

Perhaps this assumption could be revised. However, if a pass at tensor-level dictates everything after it then we're essentially going back to a monolithic pipeline. And I currently don't see benefit offsetting this cost.

rengolin: > Perhaps this assumption could be revised. However, if a pass at tensor-level dictates everything after it then we're essentially going back to a monolithic pipeline. And I currently don't see benefit offsetting cost of going back to that design.

This is a good point. We can't assume CSE (or any other cleanup and canonicalizaion) won't be run before our passes unless we fully control our pipeline and we don't really want to do that. 

In fact, we want those passes to be upstream, which means they'll be composed on a variety of frameworks' pipelines and we need to keep it robust in the presence of canonical IR.

Alternatively, we could run a canonicalization of our own, in the XSMM side, to replace:
```mlir
  %0 = memref.alloc()
  %1 = xsmm.zero %0
  ...
  %5 = memref.alloc()
  %6 = memref.copy ins(%0) outs(%5)
```

with:
```mlir
  %0 = memref.alloc()
  %1 = xsmm.zero %0
  ...
  %5 = memref.alloc()
  %6 = xsmm.zero %5
```

Then we get the pattern we want. This is "always" beneficial, because a copy from a zero buffer to another buffer is always equal or worse than setting the memory to zero directly.

adam-smnk: Well unless it is only down to bufferization again. If it bufferizes better without CSE before it then maybe we could reduce/eliminate its usage.
But if CSE at memref still breaks this pass then it is, imo, too brittle.

chelini: I am not proposing to "avoid the CSE" nor the pass "breaks" if the IR is not in the right shape. While I agree that each pass should be as much as possible "standalone," I disagree with us not being flexible on changing the default pipeline just because a pass was already there or because it is *considered* necessary. There is no technical value in running CSE at multiple points as we do now, nor in adding complexity to the folding pass now because we want to keep CSE there. If CSE is not needed why running it?

rengolin: Oh, I totally agree with you. I don't think the CSE is _necessary_ or that we should treat *our* pipeline as fixed or "important".

But because the `memref`/`xsmm` passes run at the end of any pipeline (ours or someone else's), we have little control on what _they_ pass. If we _require_ them to _not_ be canonical, then we have a high risk of not having any effect.

On those situations, if you treat your input as "whatever would come after multiple passes of canonicalization steps", then it's more robust, because best case it's already in the right shape, and worst case we canonicalize ourselves.

This is why _our_ pipeline has a lot of canonicalizations. This way we can always treat the IR at a every particular level as the "ground truth".

PR #708, opened 2023-09-06
Remove IREE build logic

rengolin:
As we move to integrate with other frameworks and since our prototype with IREE did not yield a standalone plugin architecture, we should remove the IREE-specific logic from our build files.

If we look again to integrate our work with IREE (or any other framework at that level), we should look into building shared libraries and linking via run time plugin calls instead of source code and LLVM dependencies, which clearly do not scale.


nhasabni (approved):
LGTM. Minor point - do we still need `TwoDBiasReluOp` in `MatcherUtils.cpp`?

rengolin:
> LGTM. Minor point - do we still need `TwoDBiasReluOp` in `MatcherUtils.cpp`?

Technically no, but we can leave it as is for now as an example on how to do this, as we'll need later.

PR #709, opened 2023-09-06
CUDA enable WMMA

adam-smnk:
Adds an opt-in support for WMMA operations support to the CUDA pipeline.

WMMA requires higher than default compute capability.


adam-smnk:
Vulkan support will follow next as it requires separate pipeline tweaks and hardware capability detection.

adam-smnk:
Passes locally and on [GPU CI](https://buildkite.com/intel/tpp-gpu/builds/416)

hfp (commented):
I wonder if we want to detect HW capabilities at runtime of the compiler, or if this already allows flags to force certain code (without that HW being present)?

hfp (approved):
LGTM

PR #710, opened 2023-09-07
SPIR-V enable WMMA

adam-smnk:
Adds support for WMMA operations to the SPIR-V pipeline.

WMMA requires underlying Vulkan device to support `CooperativeMatrixNV` device feature.
Tests are guarded behind a naive device filter as querying device properties is not accessible directly through `vulkaninfo`.
The extra SPIR-V capabilities can be applied blindly as they do not effect code generation or execution of the existing kernels as long as no `wmma` ops are present in lowered IR.


adam-smnk:
Passes tpp-gpu [test](https://buildkite.com/intel/tpp-gpu/builds/427).

rengolin (approved):

adam-smnk:
Tweaked 'gemm-wmma' tests to use the new '-print' f16 support + one more CI run after rebase.

PR #712, opened 2023-09-11
Bench - print more data types

adam-smnk:
Expands tpp-run '-print' utility to support more data types.


rengolin:
How does this relate to #555? Could we fix these two birds?

rengolin (approved):

PR #713, opened 2023-09-11
Add MAH with sequence length 1024

chelini:
GPT-2 uses 1024 as sequence lenght, see: ByteTransformer: A High-Performance Transformer Boosted for Variable-Length Inputs


chelini:
Note: perf are not great as we end up with large M and N dimension for the contractions. I am improving fusion to deal with it.

adam-smnk (approved):
Benchmark results seem fine

PR #715, opened 2023-09-12
Cleanup CUDA tests

adam-smnk:
Removes tests relying on `gpu.host_register` as the op is not universally available.


adam-smnk:
Passes tpp-gpu [test](https://buildkite.com/intel/tpp-gpu/builds/428).

rengolin (approved):

PR #716, opened 2023-09-12
Basic OneDNN integration

chelini:
Basic integration with OneDNN. The idea is to use `convert-linalg-to-std` pass and then provide a tiny runtime.


runtime/OneDnnl/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_mlir_library(tpp_dnnl_runner_utils
+  SHARED
+  OneDnnlRunnerUtils.cpp
+
```
rengolin, line 4: Could we add the perf runtime here too, for later?

CMakeLists.txt
```diff
@@ -49,6 +49,8 @@ include(xsmm)
 message (STATUS "LIBXSMM Include dir: ${XSMM_INCLUDE_DIRS}")
 include(xsmm-dnn)
 message (STATUS "LIBXSMM-DNN Include dir: ${XSMM_DNN_INCLUDE_DIRS}")
+include(one-dnn)
+message (STATUS "LIBXSMM-DNN Include dir: ${DNNL_INCLUDE_DIR}")
```
rengolin, line 53: Copy pasta 😁

rengolin (approved):
This is awesome! Really simple entry point. A couple of nits, but LGTM as the first commit. We can iterate on the API of the runtime later. Thanks!

alheinecke (commented):
Naming the thing is now called oneDNN (instead of DNNL), si I suggest all OneDnnl -> OneDnn at some point. 

For Simple test dnnl_sgemm is fine.

In Future we should focus on inner product and/or matmul apis:
https://oneapi-src.github.io/oneDNN/group_dnnl_api_inner_product.html
https://oneapi-src.github.io/oneDNN/group_dnnl_api_matmul.html

chelini:
> Naming the thing is now called oneDNN (instead of DNNL), si I suggest all OneDnnl -> OneDnn at some point.
> 
> For Simple test dnnl_sgemm is fine.
> 
> In Future we should focus on inner product and/or matmul apis: https://oneapi-src.github.io/oneDNN/group_dnnl_api_inner_product.html https://oneapi-src.github.io/oneDNN/group_dnnl_api_matmul.html

Ok thanks, will do another pass tomorrow.

rengolin:
```
tpp-mlir/runtime/OneDnnl/OneDnnlRunnerUtils.h:4:10: fatal error: 'dnnl.h' file not found
--
  | #include "dnnl.h"
  | ^~~~~~~~
```

Missing include path somewhere

chelini:
heads up here: build time double if we merge one-dnn.

PR #718, opened 2023-09-13
Improve `tpp-run` CMake

chelini:
`tpp_xsmm_runner_utils` and `tpp_dnnl_runner_utils` are shared library and target_link_options is enough, there is not need to add them in LIBS.

### DIFF
--- a/tools/tpp-run/CMakeLists.txt
+++ b/tools/tpp-run/CMakeLists.txt
@@ -15,8 +15,6 @@ set(LIBS
         MLIRSupport
         MLIROptLib
         MLIRTPP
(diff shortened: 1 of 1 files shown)


tools/tpp-run/CMakeLists.txt
```diff
@@ -64,7 +60,6 @@ target_link_options(tpp-run PRIVATE
   -lmlir_async_runtime
   ${TPP_GPU_LINK_FLAGS}
   -lomp
-  -Wl,--as-needed
```
rengolin, line 67: This was there to close the `no-as-needed` above. It is necessary if something adds more deps after this.

Issue #719, opened 2023-09-14
Linalg to XSMM lowering not detecting broadcast as XSMM


rengolin:
Reproduction:
```
./build/bin/tpp-opt ./test/Passes/DefaultPipeline/default-tpp-passes.mlir -default-tpp-passes="linalg-to-xsmm" -split-input-file
```

Expected behaviour (`@mlp` function):
```
    // Broadcast before GEMM lowered
    %0 = call @xsmm_unary_dispatch(%c1_i64, %c1_i64, %c128_i64, %c512_i64, %c512_i64, %c512_i64, %c4_i64) : (i64, i64, i64, i64, i64, i64, i64) -> i64
    %intptr = memref.extract_aligned_pointer_as_index %arg2 : memref<512xf32> -> index
    %1 = arith.index_cast %intptr : index to i64
    %2 = llvm.inttoptr %1 : i64 to !llvm.ptr<f32>
    %intptr_0 = memref.extract_aligned_pointer_as_index %alloc : memref<128x512xf32> -> index
    %3 = arith.index_cast %intptr_0 : index to i64
    %4 = llvm.inttoptr %3 : i64 to !llvm.ptr<f32>
    call @xsmm_unary_invoke(%c1_i64, %0, %2, %c0, %4, %c0) : (i64, i64, !llvm.ptr<f32>, index, !llvm.ptr<f32>, index) -> ()
 ```

Actual behaviour (`@mlp` function):
```
    // Broadcast before GEMM untouched
    linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "parallel"]} ins(%arg2 : memref<512xf32>) outs(%alloc : memref<128x512xf32>) {
    ^bb0(%in: f32, %out: f32):
      linalg.yield %in : f32
    }
```

Everything else works in that test when moving to the new lowering.

@chelini

rengolin:
Looking at the actual test, this is not an MLP and it makes no sense.

When changing it to an actual MLP, I get the following (expected) behaviour, with the broadcast still not expanded (but in the right place).

From:
```mlir
#map0 = affine_map<(d0, d1) -> (d1)>
#map1 = affine_map<(d0, d1) -> (d0, d1)>
#map2 = affine_map<(d0, d1, d2) -> (d0, d2)>
#map3 = affine_map<(d0, d1, d2) -> (d2, d1)>
#map4 = affine_map<(d0, d1, d2) -> (d0, d1)>

[...]

PR #720, opened 2023-09-18
Add identity conversion to linalg-to-xsmm

chelini:
Fixes: #719

### DIFF
--- a/lib/TPP/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/ConvertLinalgToXsmm.cpp
@@ -266,6 +266,37 @@ struct ConvertGenericToUnaryRelu : public OpRewritePattern<linalg::GenericOp> {
   }
 };
 
+// Convert linalg.generic to xsmm unary identity op.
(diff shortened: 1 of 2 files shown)


rengolin:
I just realised that the original pattern was broken, but when I fix it to a real MLP, I get the proper lowering. I'm wondering why your new pass was needed for those similar cases.

chelini:
> I just realised that the original pattern was broken, but when I fix it to a real MLP, I get the proper lowering. I'm wondering why your new pass was needed for those similar cases.

I think because in the real MLP we don't use bias broadcast. In the IR you posted the identity comes from a fill operation (linalg.fill). There is no conversion for linalg.generic at the moment.

rengolin:
> I just realised that the original pattern was broken, but when I fix it to a real MLP, I get the proper lowering. I'm wondering why your new pass was needed for those similar cases.

Arg! Ignore me. I had merged your patch before. Without it, the identity doesn't match. :D

All good, thanks!

rengolin (approved):

PR #723, opened 2023-09-18
Generalize OneDNN call to sgemm

chelini:


test/Passes/pass-convert-linalg-to-func.mlir
```diff
@@ -0,0 +1,98 @@
...
+// CHECK: %[[ARG1_PTR_CAST:.+]] = arith.index_cast %[[ARG1_PTR]] : index to i64
+// CHECK: %[[ARG1_LLVM_PTR:.+]] = llvm.inttoptr %[[ARG1_PTR_CAST]] : i64 to !llvm.ptr<f32>
+// CHECK: %[[ARG2_PTR:.+]] = memref.extract_aligned_pointer_as_index %[[ARG2]] : memref<3x5xf32> -> index
+// CHECK: %[[ARG2_PTR_CAST:.+]] = arith.index_cast %[[ARG2_PTR]] : index to i64
+// CHECK: %[[ARG2_LLVM_PTR:.+]] = llvm.inttoptr %[[ARG2_PTR_CAST]] : i64 to !llvm.ptr<f32>
+// CHECK: call @linalg_matmul_blas(%[[C3]], %[[C5]], %[[C4]], %[[ARG0_LLVM_PTR]], %[[C0]], %[[C3]], %[[ARG1_LLVM_PTR]], %[[C0]], %[[C4]], %[[ARG2_LLVM_PTR]], %[[C0]], %[[C3]])
+
+// -----
+
+func.func @static_matmul_double(%arg0: memref<3x4xf64>, %arg1: memref<4x5xf64>, %arg2: memref<3x5xf64>) {
```
rengolin, line 90: OneDNN is good for very large kernels. Would be nice to have some examples from `matmul` into `sgemm` calls for large (> 2k x 2k) matrices. We can add that as a small benchmark on a later PR just to see that it's good perf, but if you can just run locally and check, that'd be ok for now.

chelini: added.

PR #724, opened 2023-09-19
Linalg to WMMA

adam-smnk:
Extends linalg to GPU lowering with `linalg.brgemm` lowering and WMMA operations.

The new rewrite patterns place suitable matmul-like operations on GPU wmma cores such as Nvidia tensor cores.
The suite of WMMA tests is extended to validate correctness of the new lowering.


adam-smnk:
Current locally run GPU benchmark numbers:
image

Use of tensor cores gives better performance as expected. Note that the kernel does not measure packing/unpacking overhead.
Proper packing placement (host vs device) and, if needed, support (GPU buffer allocation) still have to be addressed.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
 
 namespace {
 
+// Return true if the operation can be represented with WMMA compute.
+bool supportsMMACompute(linalg::LinalgOp linalgOp) {
+  if (!(isa_and_nonnull<linalg::MatmulOp>(linalgOp) ||
```
chelini, line 35: nit braces.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
 
 namespace {
 
+// Return true if the operation can be represented with WMMA compute.
+bool supportsMMACompute(linalg::LinalgOp linalgOp) {
```
chelini, line 34: static bool..

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
...
+  ArrayRef<int64_t> shapeC = cType.getShape();
+  int64_t m = shapeC[0];
+  int64_t n = shapeC[1];
+  // Matrix A might be 2D (gemm) or 3D (brgemm) but the last dimension will
+  // always be reduction.
+  int64_t k = shapeA.back();
+
+  // For now, only M-N-K F16[16] x F16[16] x F16[16] WMMA variant is supported.
+  // TODO: add more WMMA combinations.
+  return aType.getElementType().isF16() && bType.getElementType().isF16() &&
```
chelini, line 56: I think the check for static shape is missing.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
...
+  auto bType =
+      linalgOp.getDpsInputOperands()[1]->get().getType().cast<ShapedType>();
+  auto cType =
+      linalgOp.getDpsInitOperands()[0]->get().getType().cast<ShapedType>();
+
+  ArrayRef<int64_t> shapeA = aType.getShape();
+  ArrayRef<int64_t> shapeC = cType.getShape();
+  int64_t m = shapeC[0];
+  int64_t n = shapeC[1];
+  // Matrix A might be 2D (gemm) or 3D (brgemm) but the last dimension will
```
chelini, line 50: nit: I would say "tensor" not matrix as matrix is 2d.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
...
+      typeA.getShape().take_back(2), typeA.getElementType(), "AOp");
+  gpu::MMAMatrixType mmaTypeB = gpu::MMAMatrixType::get(
+      typeB.getShape().take_back(2), typeB.getElementType(), "BOp");
+  gpu::MMAMatrixType mmaTypeC =
+      gpu::MMAMatrixType::get(typeC.getShape(), typeC.getElementType(), "COp");
+
+  // Matrix A might be 2D (gemm) or 3D (brgemm) but the last dimension will
+  // always be reduction.
+  auto ldb = rewriter.getIndexAttr(typeA.getShape().back());
+  auto lda = rewriter.getIndexAttr(typeC.getShape()[0]);
```
chelini, line 86: Are you at tensor level? You need to enforce identity layout otherwise. Looking at shape for the leading dimension is not correct when you have strided memref.
`memref<64x32xf32, strides[512, 1]...>` the leading dim here is 512 not 32.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
...
+  int64_t k = shapeA.back();
+
+  // For now, only M-N-K F16[16] x F16[16] x F16[16] WMMA variant is supported.
+  // TODO: add more WMMA combinations.
+  return aType.getElementType().isF16() && bType.getElementType().isF16() &&
+         cType.getElementType().isF16() && m == 16 && n == 16 && k == 16;
+}
+
+// Create WMMA instructions out of matmul-like operation.
+void gemmToGpuMMA(linalg::LinalgOp linalgOp, PatternRewriter &rewriter) {
```
chelini, line 61: static void..

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -30,108 +30,301 @@ using namespace mlir::tpp;
...
+    rewriter.setInsertionPointAfter(batchLoop);
+  }
+
+  // Write back the total sum to the output buffer.
+  rewriter.create<gpu::SubgroupMmaStoreMatrixOp>(
+      loc, result, matC, ValueRange{zero, zero}, ldc, /*transpose=*/UnitAttr());
+}
+
+// Create loops out of matmul-like operation.
+void gemmToGpuLoops(linalg::LinalgOp linalgOp, PatternRewriter &rewriter) {
```
chelini, line 159: static

lib/TPP/GPU/LinalgToGpu.cpp
adam-smnk: Good point, this can be handled.
Added proper strides support.

test/GPU/linalg-to-gpu-wmma.mlir
```diff
@@ -0,0 +1,141 @@
...
+  linalg.matmul ins(%arg0, %arg1 : memref<16x16xf16>, memref<16x16xf16>)
+                outs(%arg2 : memref<16x16xf16>)
+  return
+}
+
+// CHECK-LABEL: func.func @matmul(
+// CHECK-SAME:  %[[A:.+]]: memref<16x16xf16>, %[[B:.+]]: memref<16x16xf16>, %[[C:.+]]: memref<16x16xf16>
+// CHECK-DAG:     %[[subgroup_size:.+]] = arith.constant 32 : index
+// CHECK-DAG:     %[[one:.+]] = arith.constant 1 : index
+// CHECK:         scf.parallel {{.*}}to (%[[one]], %[[one]])
```
chelini, line 15: Can you please clarify why we create an outer loops with a single iteration? Is it not canonicalized away?

adam-smnk:
Removed strided test from Vulkan due to current limitations:
- no subview support in spirv and non-contiguous subview cannot be folded - will need to be addressed in the future
- Vulkan lowering does not support multiple kernel launches - shouldn't be an issue once we move to a different runtime

test/GPU/linalg-to-gpu-wmma.mlir
adam-smnk: Currently the kernel outlining i.e., creating GPU kernels and function launch, is modeled using nested parallel loops - outer loop models blocks, inner loop models threads.

In case of larger workloads, tiling provides me with the outer loop as GPU blocks model number of matmul blocks. Also, normally running standard eltwise kernel with only blocks and a thread/subgroup size of 1 is fine.
However, to use WMMA operation, kernel must execute as a wrap/subgroup of 32 threads. So, I need this dummy wrapper to enforce kernel outlining with 32 threads specifically for WMMA in case of small (test) workloads like this one.

Because I can run linalg to parallel loops conversion back to back with GPU parallel loop mapping (and kernel outlining), it is a good enough "hack" for now.

But indeed, it will get canonicalized away which a bit of limitation on how to properly model kernel with `blocks (1, 1, 1) threads(X, Y, Z)`. Addressing that would require better kernel mapping/outlining which is a problem for the future.

chelini (approved):

PR #725, opened 2023-09-21
Allow disabling of OneDNN and OpenMP

rengolin:
Makes sure libraries are not build/included in tools and tests not run if the libraries are disabled.


rengolin:
Tested with both, without either and with neither, on my machine. All good. Right number of tests disabled.

chelini (approved):
Looks good to me.

PR #726, opened 2023-09-21
Disable code format as dependency for check-all

rengolin:
Still keeps the targets and scripts to help formatting, but the version issue in clang-format is non-trivial, so disable from the default target for CI builds.

Policy changes to "format only the part of the code you change", similar to LLVM upstream. Then you can use any format tool that makes the code look "ok".


hfp (approved):
LGTM

PR #732, opened 2023-10-03
[CI] New cluster build

rengolin:
This simplifies the scripts for the new cluster. So far, only the tpp-mlir build on Clang. No GCC, no Cuda, no Vulkan, no benchmarks, no log reports, nothing.

Those will come later, but we need a pre-commit CI to approve pull requests, so I'm drawing the line here.


adam-smnk (approved):
Looks fine, let's get CI spinning once again.
Other CI jobs will be reenabled in different PRs.

PR #736, opened 2023-10-05
GPU data transfer

adam-smnk:
Adds a new pass that inserts data copies around GPU function launches when they refer to buffers that reside on the host e.g., memref.alloc or host's global variables.

This improves GPU code generation capabilities by enabling execution of IR that starts at the tensor level. After bufferization and kernel outlining, kernel might access buffers that still reside on the host. This is addressed by copying host data into the device.
This pass targets GPUs or runtimes that do not support unified memory e.g., our current CUDA runner setup.


adam-smnk:
Performing full one-shot data flow analysis between host and GPU ended up being a bit too complex for now. So, the current approach is to examine individual kernel calls and move data back and forth between host and GPU when necessary.

This can be later improved by adding a few cleanup patterns that fold unnecessary copies and even replace original `memref.allocs` with `gpu.allocs`.

Once we focus more exclusively on the Intel GPU, it should be possible to just unified shared memory. But for now it will enable some more GPU testing.

include/TPP/Passes.td
```diff
@@ -569,4 +569,16 @@ def LinalgToGpu : Pass<"linalg-to-gpu", "func::FuncOp"> {
   ];
 }
 
+def GpuDataTransfer : Pass<"gpu-data-transfer", "func::FuncOp"> {
+  let summary = "Transfer data to and from GPU.";
+  let description = [{
+    Make host data required by GPU kernels accessible by the device.
+    It might involve data copies and/or movement.
+  }];
+  let constructor = "mlir::tpp::createGpuDataTransferPass()";
```
chelini, line 578: nit: this could be generated by tableGen, see: https://github.com/chelini/tpp-sandbox/commit/732f47e1ebc1ef904658674c9295c13a5d93fe47

lib/TPP/GPU/GpuDataTransfer.cpp
```diff
@@ -0,0 +1,245 @@
...
+#define GEN_PASS_CLASSES
+#include "TPP/Passes.h.inc"
+
+namespace {
+
+// Returns the source operation of the given value (data) e.g., host allocation,
+// if it should be moved to the device.
+// Nullptr is returned (nothing to do), if the source is unknown and/or
+// the buffer is believed to already live on the device.
+static Operation *getDeviceTransferableBuffer(Value val) {
```
chelini, line 32: can you return std::optional or failure<Operation*>?

lib/TPP/GPU/GpuDataTransfer.cpp
```diff
@@ -0,0 +1,245 @@
...
+  Operation *op = val.getDefiningOp();
+  if (!op)
+    return nullptr;
+
+  // Host-space allocation or a global variable can be easily transfered to
+  // a device.
+  if (isa<memref::AllocOp, memref::GetGlobalOp>(op))
+    return op;
+
+  // Follow through mmeref alias to get to the real source.
```
chelini, line 50: nit: memref.

lib/TPP/GPU/GpuDataTransfer.cpp
```diff
@@ -0,0 +1,245 @@
...
+  rewriter.setInsertionPoint(block.getTerminator());
+  rewriter.create<gpu::DeallocOp>(loc, std::nullopt, gpuAlloc.getMemref());
+
+  return gpuBuffer;
+}
+
+// Move host data used by GPU kernel calls to the device.
+struct TransferDataToGpu : public OpRewritePattern<gpu::LaunchFuncOp> {
+  TransferDataToGpu(MLIRContext *context, PatternBenefit benefit = 1)
+      : OpRewritePattern<gpu::LaunchFuncOp>(context, benefit) {}
```
chelini, line 163: nit: `using OpRewritePattern<gpu::LaunchFuncOp>::OpRewritePattern` seems sufficient.

lib/TPP/GPU/GpuDataTransfer.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// Transfer host allocated data between the host and a device.
+// Optionally, copies data back to the host after device computations.
+// Returns device-allocated buffer with the moved data.
+static FailureOr<Value> transferMemref(RewriterBase &rewriter,
+                                       gpu::LaunchFuncOp launchFuncOp,
+                                       Value operand, Value hostBuffer,
+                                       bool copyDataBack = true) {
+  // A memref buffer is expected.
```
chelini, line 118: I would assert here.

lib/TPP/GPU/GpuDataTransfer.cpp
```diff
@@ -0,0 +1,245 @@
...
+  // a function call, a function arguments, device allocation etc.
+  return nullptr;
+}
+
+// Matches the kernel operand and device buffer types such that data can be
+// transfered correctly from the original host buffer.
+// Returns failure if the two types cannot be connected.
+static LogicalResult matchTransferBuffers(RewriterBase &rewriter,
+                                          Value kernelOperand,
+                                          Value &gpuBuffer) {
```
chelini, line 65: let's not pass Value by reference.

lib/TPP/GPU/GpuDataTransfer.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include "mlir/Conversion/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/GPU/IR/GPUDialect.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/Dialect.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
+
+using namespace mlir;
```
chelini, line 20: nit: probably you don't want to import mlir::tpp namespace.

adam-smnk: Good point, removed.

lib/TPP/GPU/GpuDataTransfer.cpp
adam-smnk: Added assert for quicker feedback while debugging.

include/TPP/Passes.td
adam-smnk: Seems not bad, converted to TableGen.

rengolin (approved):

test/GPU/CUDA/Integration/linalg-mlp.mlir
```diff
@@ -0,0 +1,42 @@
+// RUN: ASAN_OPTIONS=protect_shadow_gap=0:replace_intrin=0:detect_leaks=0:${ASAN_OPTIONS} \
```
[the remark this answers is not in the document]
adam-smnk: Added check and description.

Issue #734, opened 2023-10-04
Investigate non scalability of OpenMP benchmarks in MLIR compiler


rengolin:
As exposed in #733, the MLIR OMP benchmarks don't scale at all, as if all of them are single-threaded. Must be some mistake on the flags side.

rengolin:
The test `test/Integration/openmp/tpp-run-parallel.mlir` checks that OMP calls are made to an `scf.parallel`, and that is still working (test pass, I looked at the output, looks fine).

However, when running the same `tpp-run` command on the benchmarks, OMP calls are not made. Here's a reproducer:
```
$ tpp-run -def-parallel -e entry -entry-point-result=void -print-mlir=late benchmarks/mlir/mlp-fp32-3layers-1024.mlir
```

Expecting `omp.parallel`, `omp.wsloop`, `omp.yield`, `omp.terminator` but got `scf.forall`.

This was still working before #730, so I'm guessing this is related to upstream's change in `scf.parallel` to `scf.forall`.

@KavithaTipturMadhu, perhaps we need to recognize `scf.forall` for our OMP change?

rengolin:
Hm, not so simple. The pass is [upstream](https://github.com/llvm/llvm-project/blob/ca003ee06d0eac7e8facc179181298a05e4d03ed/mlir/lib/Conversion/SCFToOpenMP/SCFToOpenMP.cpp#L358) and only works on `scf.parallel`.

Perhaps we should just not convert to `scf.forall` when using OpenMP? Or contribute to a rewriter for `scf.forall` upstream?

chelini:
We could add back the pass `createConvertForAllToParallelOpPass` or contribute the pattern upstream.

adam-smnk:
Looks like it's the same issue as with kernel outlining that it only expects `scf.parallel`.
We should add the conversion pass back for now.
Long term, upstream changes would be ideal.

chelini:
mm, upstream seems ok: https://github.com/llvm/llvm-project/commit/87568ff3efb216e35161bf201897b2c587ffc1c4

chelini:
I can take care of it tomorrow.

rengolin:
> mm, upstream seems ok: [llvm/llvm-project@87568ff](https://github.com/llvm/llvm-project/commit/87568ff3efb216e35161bf201897b2c587ffc1c4)

Ah, I missed that one! Ok, perhaps we just need to add this pass before the OpenMP one and it'd work again. Thanks!

PR #738, opened 2023-10-09
Fix scf.forall to scf.parallel conversion

chelini:
Upstream converts `scf.forall` to `scf.parallel`, which is, in turn, lowered to an `scf.for` and then to structured control flow in `convert-scf-to-cf`. This lowering is atomic, thus we need to materialize `scf.parallel` before `convert-scf-to-cf` if we want to map to `omp.parallel`. The ideal solution would be to decouple the lowering from `scf.forall` to `scf.parallel` from upstream: `convert-scf-to-cf`.

fixes: #734


test/Integration/openmp/tpp-run-parallel.mlir
```diff
@@ -1,6 +1,8 @@
-// RUN: tpp-run %s -def-parallel=0 -e entry -entry-point-result=void -print-mlir=late -print 2>&1 | \
+// RUN: tpp-opt %s -default-tpp-passes | \
```
rengolin, line 1: Why do you need to pipe `tpp-opt`?

rengolin:
Some weird crashes on the tests, did you not see this on your machine?

test/Integration/openmp/tpp-run-parallel.mlir
chelini: We need to make sure we get from scf.forall to scf.parallel. Previously we were only lowering: `scf.parallel` using tpp-run this did not catch the regression.

chelini:
> Some weird crashes on the tests, did you not see this on your machine?

nope.

test/Integration/openmp/tpp-run-parallel.mlir
rengolin: That's what I want to understand. both `tpp-opt` and `tpp-run` need to converge to the same IR, especially when both are running the default TPP passes. If they don't converge, then there is a pipeline problem.

chelini: there was not need for `tpp-opt`.

rengolin (approved):

PR #739, opened 2023-10-09
Add GCC/Clang release builds

rengolin:
Fix type range, hard-code gcc to 12.3 in the cluster.


adam-smnk (approved):
Changes and builds look good

PR #740, opened 2023-10-10
Re-work pack lowering

chelini:
Rewriting `tensor.pack` to loops was not composable, preventing further optimizations like fusion. A better approach is to materialize loops only during tiling and fusion. The pack is then generalized to an expanded shape and a transpose. Further optimizations are added to map the transpose and the expand shape to xsmm vnni calls.


adam-smnk (commented):
Should there be any improvement in any benchmarks?

lib/TPP/LowerPacksAndUnpacks.cpp
```diff
@@ -0,0 +1,254 @@
...
+    if (tileSizes.empty())
+      continue;
+    auto tilingResult = tileOp(
+        rewriter, cast<TilingInterface>(packOp.getOperation()), tileSizes);
+    if (failed(tilingResult))
+      continue;
+    rewriter.replaceOp(packOp, tilingResult->replacements);
+  }
+
+  // Tile unpacks.
```
adam-smnk, line 130: Is unpack fusion coming in this or separate PR?

adam-smnk (commented):
Looks good after the first pass

benchmarks/config/base/base.json
```diff
@@ -46,6 +46,13 @@
       "flags": [ "-n", "100" ],
       "extensions": [ "(avx2|asimd)" ]
     },
+    "fp32_3x512_mlir": {
```
rengolin, line 49: What do these add over the existing 1024 case?

benchmarks/config/base/base.json
```diff
@@ -60,6 +67,13 @@
       "flags": [ "-n", "100" ],
       "extensions": [ "avx512.*" ]
     },
+    "f32_flat_3x1024_unp_mlir": {
```
rengolin, line 70: We really need to be moving off hand-crafted MLIR files and use `mlir-gen` for everything.

lib/TPP/LowerPacksAndUnpacks.cpp
```diff
@@ -0,0 +1,254 @@
...
+  }
+  return tilingResult;
+}
+
+// Fuse producer and consumer pack. Standalone packs are tiled into 2d tiles.
+static void fuseOrTilePacks(RewriterBase &rewriter, FunctionOpInterface func) {
+  SmallVector<tensor::PackOp> consumerPackOps;
+  SmallVector<tensor::PackOp> otherPacks;
+  SmallVector<tensor::UnPackOp> unPacks;
+  func->walk<WalkOrder::PostOrder>([&](tensor::PackOp consumerPackOp) {
```
rengolin, line 77: Nit, it took me a while to figure that `consumer` here means "of another pack", not just any other op. I'd call it `chained` instead, to make it clearer.

lib/TPP/LowerPacksAndUnpacks.cpp
```diff
@@ -0,0 +1,254 @@
...
+    if (producerPackOp)
+      consumerPackOps.push_back(consumerPackOp);
+    else
+      otherPacks.push_back(consumerPackOp);
+  });
+  func->walk<WalkOrder::PostOrder>(
+      [&](tensor::UnPackOp unpackOp) { unPacks.push_back(unpackOp); });
+
+  // Tile and fuse.
+  for (auto consumerPackOp : consumerPackOps) {
```
rengolin, line 90: Are you assuming here that the chained packs (consumers) are handled before the non-chained (producer) ones?

If you have a 3-chain, wouldn't the second be in the list before the third, breaking that rule?

lib/TPP/LowerPacksAndUnpacks.cpp
```diff
@@ -0,0 +1,254 @@
...
+            cast<tensor::ExtractSliceOp>(tiledPack.getSource().getDefiningOp()),
+            forLoops);
+    if (!fusedProducer)
+      continue;
+    rewriter.replaceOp(consumerPackOp, tilingResult->replacements);
+  }
+
+  // Tile packs.
+  for (auto packOp : otherPacks) {
+    SmallVector<int64_t> tileSizes = getTileSizes(packOp);
```
rengolin, line 120: Shouldn't you be using the same tiles as the consumer? There doesn't seem to be any guarantee to this here or in `getTileSizes`.

chelini: Here I don't have any consumer. All the packs are standalone.

test/Passes/pass-pack-lowering.mlir
adam-smnk, line 1: It'd be nice to have more fusion tests also for more complex chains.

benchmarks/config/base/base.json
chelini: removed.

benchmarks/config/base/base.json
rengolin: No, just move it up and keep a consistent naming. I'll refactor this in one single PR later.

chelini: moved up, the name is consistent with the other unpack we have, no?

rengolin: one nit `fp32`, otherwise ok.

adam-smnk (approved):

PR #741, opened 2023-10-10
Delete `simple-copy.mlir` (NFC)

chelini:
The linalg.generic gets canonicalized away so we are not testing much.


adam-smnk (approved):
Functionality is covered by other tests

PR #742, opened 2023-10-12
Simplify flops calculation, apply for all kernel types

rengolin:
All tests pass, they check FLOPS calculations, so we know the new style is correct for all tested cases.

Now it also applies MLP FLOP calculations.

### DIFF
--- a/tools/mlir-gen/MLIRGen.cpp
+++ b/tools/mlir-gen/MLIRGen.cpp
@@ -72,7 +72,7 @@ MLIRGenerator::MLIRGenerator(StringRef kernelStr, unsigned miniBatch,
                              unsigned typeWidth, int seed, bool enableSoftmax,
                              bool biasAcc, int vnniBlockingFactor)
     : builder(&context), loc(builder.getUnknownLoc()), miniBatch(miniBatch),
-      seed(seed), enableSoftmax(enableSoftmax), biasAcc(biasAcc),
+      seed(seed), flops(0), enableSoftmax(enableSoftmax), biasAcc(biasAcc),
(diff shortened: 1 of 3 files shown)


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -378,6 +351,16 @@ Value MLIRGenerator::lowerMatmul(MatMulArgs args) {
               })
           .getResult(0);
 
+  // Matmul flops = 2 * M * N * K = 2 * prod(inputDims) * N (weightShape[0])
+  int64_t mkFlops = 1;
+  for (int i = 0, max = inputShape.getRank(); i < max; i++)
+    mkFlops *= inputShape.getDimSize(i);
+  // Tiled: N = NB * n = weightShape[0] + weightShape[3]
+  int64_t nFlops = weightShape.getDimSize(0);
```
adam-smnk, line 359: This grabs K dim if there's no tiling.

rengolin:
> I like the idea but it gets a bit convoluted with complex tiling and non eltwise operations.

Thats why we do it by operation and not at the end. Once we compose everything, flops will always be correct.

Also, it's not super important to get the absolute perfection. Ball park is enough to give us what we want and this gives the exact same numbers as before for all tested cases.

tools/mlir-gen/MLIRGen.cpp
rengolin: ouch, ok. Not enough tests...

adam-smnk (approved):
Looks good - benchmark numbers are unchanged.
Thanks for adding extra test cases 👍

PR #745, opened 2023-10-17
Bump LLVM + CI LLVM build scripts

adam-smnk:
Bumps LLVM version and adds a CI script for building LLVM directly from TPP-MLIR.

The changes move the LLVM CI job to use TPP-MLIR repository to contain the whole build logic in one repository.
The overall LLVM build flow remains as before:
- check for if specified LLVM version is available before building TPP-MLIR
- if no LLVM available, schedule LLVM CI job
- recheck LLVM availability to avoid double builds
- proceed to build, run check-all, and finally install LLVM

LLVM CI job now checks out TPP-MLIR and fetches LLVM codebase separately.
After build and checks are complete, LLVM is still installed locally in the CI agent home directory.


scripts/buildkite/build_llvm.sh
```diff
@@ -0,0 +1,111 @@
...
+echo "--- LLVM"
+LLVM_VERSION=${BUILDKITE_COMMIT}
+if [ ! "${LLVM_VERSION}" ]; then
+  echo "Unknown LLVM_VERSION version"
+  exit 1
+fi
+echo "LLVM version: ${LLVM_VERSION}"
+
+LLVM_INSTALL_DIR=${LLVMROOT}/${LLVM_VERSION}
+if [ ! -d ${LLVM_INSTALL_DIR} ]; then
```
rengolin, line 27: You don't need the `if` check, `mkdir -p` ignores if the dir already exists.

rengolin:
If the buildkite job is bound to the `tpp-mlir` repo, it'll checkout the `tpp-mlir` code, not `llvm`. So, the LLVM src won't be in the same place or even checked out. You'll need to use the shallow clone or download a tar ball to fetch the LLVM code. You can also cache and check if the tar ball already exists (in a shared NFS dir) and just unpack it over the same place again.

scripts/buildkite/tpp-llvm.yml
```diff
@@ -0,0 +1,11 @@
+env:
+  NPROCS_LIMIT_LINK: "8"
+
+steps:
+  - label: "TPP-LLVM"
+    concurrency: 1
+    concurrency_group: "tpp-llvm"
+    command: "scripts/buildkite/check_llvm.sh || \
+              ${SRUN} --partition=spr --time=1:00:00 -- \
```
chelini, line 9: Do we give back the partition to slurm or we keep it for 1h? I think 30 minutes are enough.

chelini (approved):
Generally looks good to me.

scripts/buildkite/tpp-llvm.yml
adam-smnk: It is returned when the job is done. The 1h is just the upper limit.
I kept the same higher time allocation as it was in the old tpp-llvm setup.

adam-smnk: I'll lower it to 30min to be consistent with others, should be enough.

rengolin:
Thank you! This was the last hard-coded CI we had. Glad we finally moved on.

PR #747, opened 2023-10-18
[mlir-gen] Refactor all kernels into a single generator

rengolin:
Avoid duplicating the logic across `matmul`, `fc` and `mlp` into a single generator that enables/disables features based on command line flags.

Also enables multi-layer for `fc` and `matmul` that previously wasn't possible.

These are the new choices:
 * Kernel = `layer` (operands are arguments), `model` (constants)
 * Enable/disable bias/relu/softmax

The matching is:
 * `matmul` = kernel=layer
 * `fc` = kernel=layer + bias + relu
 * `mlp` = kernel=model + bias + relu

There should be a softmax at the end of MLP, but we're not doing it yet.

All tests pass. All performance benchmarks remained idential and kept their performance profiles.

Changes in tests are due to real (and necessary) output changes.

### DIFF
--- a/tools/mlir-gen/MLIRGen.cpp
+++ b/tools/mlir-gen/MLIRGen.cpp
@@ -45,35 +45,16 @@ void parseStringList(StringRef str, SmallVector<int64_t> &list) {
   }
 }
 
-SmallVector<int64_t> getMatMulResultShape(ShapedType lhs, ShapedType rhs) {
-  SmallVector<int64_t> shape;
-  assert(lhs.getRank() == rhs.getRank() && "Matmul types must have same rank");
-  // M x K x N -> M x N
-  assert(lhs.getDimSize(1) == rhs.getDimSize(0) &&
-         "Incompatible matmul shapes");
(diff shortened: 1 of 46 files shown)


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -655,3 +619,11 @@ int MLIRGenerator::getRand() {
...
   return temp;
 }
+
+Value MLIRGenerator::getZeroInitTensor(TensorType type) {
+  auto zero = getConstFloat(builder, 0.0, dataType.getIntOrFloatBitWidth());
+  Value tensor =
+      builder.create<tensor::EmptyOp>(loc, type, ValueRange{}).getResult();
+  tensor = builder.create<linalg::FillOp>(loc, zero, tensor).getResult(0);
+  return tensor;
+}
```
adam-smnk, line 629: nit: newline

adam-smnk (approved):
Great improvement, thanks 👍
Benchmark numbers look correct.

PR #748, opened 2023-10-18
Fix leading dimension for VNNI2

chelini:
We need to get ldo from the output memref, it is not the same as the ldi. In addition the ldo must be scaled by the vnni factor.


adam-smnk (commented):
Can we add a simple integration test that validates it too?

adam-smnk (approved):
Thanks, looks good 👍

Issue #581, opened 2023-06-09
`xsmm.fused_brgemm` lacks input validation


adam-smnk:
The snippet taken from `xsmm-quarternary-bf16.mlir` test:
```
func.func @entry(%arg0: memref<64x4x4xbf16>, %arg1: memref<64x2x4x2xbf16>, %arg2: memref<4xbf16>, %arg3: memref<4x4xbf16>) {
  %c16_i64 = arith.constant 16 : i64
  %func = xsmm.fused_brgemm.dispatch [4, 4, 4, 4, 4, 4][add, relu]
    flags = (vnni_b) binary_flags = (bcast_col_in0) unary_flags = (none) data_type = bf16
  xsmm.fused_brgemm(data_type = bf16, %func, %arg0, %arg1, %arg2, %arg3, %c16_i64) : (i64, memref<64x4x4xbf16>, memref<64x2x4x2xbf16>, memref<4xbf16>, memref<4x4xbf16>, i64) -> ()
  return
}
```

The output of `xsmm.fused_brgemm` operation is stored in `%arg2` which is a 1D tensor instead of 2D.
Probably XSMM ops validation is too relaxed.

rengolin:
Test has changed since and no longer exhibits this problem. Not saying the verify is fixed, but we're planning to change the dialect, so no point in looking at this now.

PR #752, opened 2023-10-20
Xsmm fusion of brgemm+add+relu

KavithaTipturMadhu:


test/Passes/pass-combine-tpp.mlir
adam-smnk, line 1: Could it be converted to XSMM combine test?

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+ArrayAttr getBinaryFlags(RewriterBase &rewriter, xsmm::BinaryOp addOp) {
+  auto binaryInputType = addOp.getOperand(2).getType().cast<MemRefType>();
+  auto outputType = addOp.getOperand(3).getType().cast<MemRefType>();
+  return rewriter.getArrayAttr(xsmm::BinaryFlagsAttr::get(
+      rewriter.getContext(),
+      xsmm::utils::getBinaryBCast(binaryInputType, outputType,
+                                  /*operandNumber=*/0)));
+}
+
+xsmm::BinaryKindAttr getBinaryKind(RewriterBase &rewriter) {
```
adam-smnk, line 85: These helper functions return hardcoded attributes.
I'd give them more specific names to reflect that e.g., `getAddBinaryKind`.

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+  LogicalResult matchAndRewrite(xsmm::BrgemmOp brgemmOp,
+                                PatternRewriter &rewriter) const override {
+    auto brgemmResultOperand = brgemmOp.getOperand(3);
+
+    for (auto use = brgemmResultOperand.user_begin();
+         use != brgemmResultOperand.user_end(); use++) {
+      auto user = *use;
+      if (isa<xsmm::BinaryOp>(user) &&
+          (dyn_cast<xsmm::BinaryOp>(user)).getCallee() ==
+              mlir::xsmm::BinaryKind::ADD) {
```
adam-smnk, line 116: It might be safer to break fusion on unknown users.
Or maybe it is handled well, just make sure to add some test cases to cover e.g., a copy in between XSMM calls.

include/TPP/Passes.h
```diff
@@ -100,7 +100,6 @@ std::unique_ptr<OperationPass<ModuleOp>>
 createTransformDialectInterpreterPass();
 std::unique_ptr<OperationPass<func::FuncOp>> createConvertPerfToLoopsPass();
 std::unique_ptr<OperationPass<ModuleOp>> createConvertPerfToFuncPass();
-std::unique_ptr<OperationPass<func::FuncOp>> createCombineTppPass();
```
rengolin, line 103: Please don't remove it in this PR, since it's not used in the "linalg-to-xsmm" pass anyway, it won't interfere with your work.

include/TPP/Passes.td
```diff
@@ -256,6 +247,16 @@ def RewriteBatchMatmulToMatmul : Pass<"rewrite-batch-matmul-to-matmul",
   let dependentDialects = ["scf::SCFDialect", "linalg::LinalgDialect"];
 }
 
+def CombineXsmmOpPass : Pass<"combine-xsmm-op-optimization", "func::FuncOp"> {
+  let summary = "Fuse gemm-relu-bias into a bigger xsmm op";
```
rengolin, line 251: I'd be more generic here. We want to add all sorts of XSMM fusions in this pass. For now we only have those.

lib/TPP/CombineXsmmPass.cpp
rengolin: Honestly, this looks like a silly wrapper, totally unnecessary. Best to remove and use the call altogether.

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+                                  /*operandNumber=*/0)));
+}
+
+xsmm::BinaryKindAttr getBinaryKind(RewriterBase &rewriter) {
+  return xsmm::BinaryKindAttr::get(rewriter.getContext(),
+                                   xsmm::BinaryKind::ADD);
+}
+
+xsmm::UnaryKindAttr getUnaryKind(RewriterBase &rewriter) {
+  return xsmm::UnaryKindAttr::get(rewriter.getContext(), xsmm::UnaryKind::RELU);
```
rengolin, line 91: Same here.

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+        auto isFusedAdd = dispatchOp.getKind() == xsmm::BinaryKind::ADD;
+        auto binaryFlags = dispatchOp.getFlags();
+        if (isFusedAdd &&
+            (binaryFlags.size() != 1 ||
+             binaryFlags[0].cast<mlir::xsmm::BinaryFlagsAttr>().getValue() !=
+                 mlir::xsmm::BinaryFlags::BCAST_COL_IN_0)) {
+          return failure();
+        }
+        auto addResultOperand = ((xsmm::BinaryOp)user).getOperand(3);
+        for (auto relu = addResultOperand.user_begin();
```
rengolin, line 129: Do not have loops inside loops. This won't scale for anything that is not an ADD followed by a RELU.

As we discussed earlier, have an initial scan for binary/unary, put them in a list/struct/local vars, and then handle the generic case from there.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -246,7 +246,6 @@ struct TppConversionPass : public TppConversionBase<TppConversionPass>,
 
     // Convert all higher level dialects to TPP.
     pm.addPass(createConvertLinalgToTppPass());
-    pm.addPass(createCombineTppPass());
```
rengolin, line 249: Please do not remove this pass from the pipeline in this PR. Let's do that once your PR is merged, as a clean up, while removing the TPP dialect.

test/Passes/DefaultPipeline/tpp-conversion.mlir
```diff
@@ -63,6 +63,7 @@ func.func @linalg_dialect_expect_fused_brgemm(
 // CHECK-LABEL: linalg_dialect_expect_fused_brgemm
 // CHECK-SAME:  %[[ARG0:.+]]: tensor<1x3x2xf32>, %[[ARG1:.+]]: tensor<1x2x3xf32>, 
 // CHECK-SAME:  %[[ARG2:.+]]: tensor<3x3xf32>, %[[ARG3:.+]]: tensor<1x3xf32>
-// CHECK: %{{.+}} = tpp.fused_brgemm [unary = relu, binary = add] 
-// CHECK-SAME:  (%[[ARG0]] : tensor<1x3x2xf32>, %[[ARG1]] : tensor<1x2x3xf32>, %[[ARG2]] : tensor<3x3xf32>, %[[ARG3]] : tensor<1x3xf32>) -> (tensor<3x3xf32>)
-
+// CHECK: %{{.+}} = tpp.brgemm
```
rengolin, line 66: Nope, still needs to match a fused_brgemm, until we remove the TPP dialect. Let's not revert behaviour, even if we'll remove it soon.

rengolin (commented):
One benchmark was affected: `bf16_3x1024_flat_mlir`, from 500 GF to 390 GF.

I see no tests to show that your work can match XSMM sequences into fused ones. We also need tests to *not* match cases that we don't want.

include/TPP/Passes.td
```diff
@@ -256,6 +247,16 @@ def RewriteBatchMatmulToMatmul : Pass<"rewrite-batch-matmul-to-matmul",
...
+  let description =
+      [{Fuse gemm-relu-bias into a bigger xsmm op }];
+  let constructor = "mlir::tpp::createCombineXsmmOpPass()";
```
chelini, line 254: nit: from now on each new pass should not have `let constructor` this is generated by tablegen. You can have a look here on how to retire it: https://github.com/chelini/tpp-sandbox/commits/fix-vnni2

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+using namespace mlir;
+using namespace mlir::tpp;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Dialect/Xsmm/XsmmUtils.h"
+#include "TPP/Passes.h.inc"
+#include "TPP/VNNIUtils.h"
+#include <iostream>
```
chelini, line 26: drop `iostream` pls.

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+//===----------------------------------------------------------------------===//
+//
+
+#include "TPP/BuilderUtils.h"
+#include "TPP/Dialect/Xsmm/XsmmOps.h"
+#include "TPP/Dialect/Xsmm/XsmmUtils.h"
+#include "TPP/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
```
chelini, line 16: why you include tensor?

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+using namespace mlir::tpp;
+
+#define GEN_PASS_CLASSES
+#include "TPP/Dialect/Xsmm/XsmmUtils.h"
+#include "TPP/Passes.h.inc"
+#include "TPP/VNNIUtils.h"
+#include <iostream>
+namespace {
+
+static FailureOr<DenseI64ArrayAttr>
```
chelini, line 29: is this copied from `ConvertToXsmm.cpp`?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -78,6 +78,123 @@ FailureOr<UnaryInfo> getUnaryInfo(Value input, Value output) {
   return unaryInfo;
 }
 
+// Examples:
+// If lower=[c], higher=[a, b, c], [c] reshaped into [1, 1, c].
+// If lower=[b, c], higher=[a, b, c], [b, c] reshaped into [1, b, c].
+// If lower=[a], higher=[a, a], [a] reshaped into [1, a].
+// If lower=[a], target=[a, b, a], [a] reshaped into [1, 1, a].
+// If lower=[], target=[a, b, c], [] reshaped into [1, 1, 1].
+void computeBcastShapeInput(ArrayRef<int64_t> higherRankShape,
```
chelini, line 87: this methods will soon disappear. We used them for TPPs but now that we are moving to linalg we can use affine maps. It is unclear also why you need to recompute this informations. Each XSMM operations as a flag specifying the broadcast for each of its operand.

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,194 @@
...
+  return rewriter.getArrayAttr(gemmFlag);
+}
+
+struct CombineXsmmOp : public OpRewritePattern<xsmm::BrgemmOp> {
+  using OpRewritePattern<xsmm::BrgemmOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(xsmm::BrgemmOp brgemmOp,
+                                PatternRewriter &rewriter) const override {
+    auto brgemmResultOperand = brgemmOp.getOperand(3);
+
+    for (auto use = brgemmResultOperand.user_begin();
```
chelini, line 111: Is this actually safe? We need to be more strict with the users, what if you have:
```
xsmm.add(B, A)
xsmm.sub(C, A)
xsmm.brgemm(A ..)
```
this should not fuse as the sub writes into A. Can you please add some tests along this line?

chelini:
@KavithaTipturMadhu do you need help in landing this?

lib/TPP/CombineXsmmPass.cpp
KavithaTipturMadhu: will address this in a test case in the next commit

KavithaTipturMadhu:
> One benchmark was affected: `bf16_3x1024_flat_mlir`, from 500 GF to 390 GF.
> 
> I see no tests to show that your work can match XSMM sequences into fused ones. We also need tests to _not_ match cases that we don't want.

This should be fixed in the current commit

KavithaTipturMadhu:
> One benchmark was affected: `bf16_3x1024_flat_mlir`, from 500 GF to 390 GF.
> 
> I see no tests to show that your work can match XSMM sequences into fused ones. We also need tests to _not_ match cases that we don't want.

I have added positive test case for now, will add negative ones in a subsequent commit

include/TPP/Passes.h
KavithaTipturMadhu: reverted

lib/TPP/CombineXsmmPass.cpp
KavithaTipturMadhu: will address in a subsequent commit

lib/TPP/DefaultTppPasses.cpp
KavithaTipturMadhu: reverted

test/Passes/DefaultPipeline/tpp-conversion.mlir
KavithaTipturMadhu: reverted

include/TPP/Passes.td
KavithaTipturMadhu: fixed

lib/TPP/CombineXsmmPass.cpp
KavithaTipturMadhu: removed

lib/TPP/CombineXsmmPass.cpp
KavithaTipturMadhu: renamed

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: removed

lib/TPP/CombineXsmmPass.cpp
KavithaTipturMadhu: will address in a a subsequent commit

rengolin: This commit should be safe enough to not handle untested cases.

lib/TPP/CombineXsmmPass.cpp
rengolin: If there's anything untested, it must return failure and not transform.

lib/TPP/CombineXsmmPass.cpp
rengolin: No, this will create technical debt that will be hard to remove. We can't go on adding things that will be removed later.

lib/TPP/DefaultTppPasses.cpp
rengolin: Still seems to be here in the last diff

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,28 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
   return binaryInfo;
 }
 
+FailureOr<int64_t> getLeadingDim(Type type, size_t pos) {
+  // Not shaped type, the leading dimension is the single scalar.
+  if (!isa<ShapedType>(type))
```
rengolin, line 153: This is unsafe. You check for `ShapedType` but cast to `MemrefType`.

A much simpler pattern is:
```c++
  auto memref = dyn_cast<MemRefType>(type);
  if (!memref)
    return 1;
  // Continue assuming it is a memref
```

test/Passes/DefaultPipeline/tpp-conversion.mlir
rengolin: test still shows `brgemm`, not fused. did you push the commit that reverts it?

rengolin (commented):
We need a test with broadcast on add, and one that uses allocs inside the inner loop for the add/relu, even if we don't transform it. We must prove that we fail gracefully, not crash the compiler.

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+namespace {
+
+static FailureOr<DenseI64ArrayAttr>
+getSizesAndLeadingDimForBrgemmOp(RewriterBase &rewriter, xsmm::BrgemmOp opTy) {
+
+  auto memrefC = opTy.getOperand(3).getType();
+  auto memrefA = opTy.getOperand(1).getType();
+  auto memrefB = opTy.getOperand(2).getType();
+
+  int64_t m = memrefC.cast<ShapedType>().getShape()[0];
```
rengolin, line 38: This is unsafe. There's no guarantee that the type is indeed `ShapedType` and the cast will succeed. If it does not, a segfault will occur. Use the `isa` or `dyn_cast` patterns to make sure you're dealing with the right type.

Further unsafe for use of `MemRefType` casts below without checking it's a valid cast. Since XSMM only works on memrefs, just use `MemRefType` everywhere.

lib/TPP/CombineXsmmPass.cpp
rengolin: The pattern here is in memrefs, so you need the producer of the result operand's user list. This can be an `alloc`, `alloca`, `region_op`, `subview`, etc.

As @chelini said, we need to make sure the `brgemm` is the first operation on the operand (other than `zero`) and the other "known" ops follow. We need to make sure the graph is consistent, too.

```mlir
  %a = memref.subview ...
  %b = memref.subview ...
  %c = memref.subview ...

  // optional
  xsmm.unary zero (%c)

  // brgemm, %c needs to be outs
  xsmm.brgemm (%a, %b) -> (%c)

  // binary in-place (%bias can be a broadcast, cannot be %a or %b)
  xsmm.binary add (%c, %bias) -> (%c)

  // binary out-of-place (%bias can be a broadcast, cannot be %a or %b)
  %add = memref.alloc
  xsmm.binary add (%c, %bias) -> (%add)

  // unary in-place
  xsmm.unary relu (%c) -> (%c)

  // unary out-of-place
  %relu = memref.alloc
  xsmm.unary relu (%c) -> (%relu)

  // User or %c / %add / %relu
  return ...
```

lib/TPP/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+      auto user = *use;
+      if (isa<xsmm::BinaryOp>(user) &&
+          (dyn_cast<xsmm::BinaryOp>(user)).getCallee() ==
+              mlir::xsmm::BinaryKind::ADD) {
+        xsmm::BinaryOp binaryOp = ((xsmm::BinaryOp)user);
+        auto dispatchOp = dyn_cast<xsmm::BinaryDispatchOp>(
+            binaryOp.getOperand(0).getDefiningOp());
+        auto isFusedAdd = dispatchOp.getKind() == xsmm::BinaryKind::ADD;
+        auto binaryFlags = dispatchOp.getFlags();
+        if (isFusedAdd &&
```
rengolin, line 99: Could this be a matcher? @chelini

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -151,6 +91,103 @@ getSizesAndLeadingDimsForGemmLikeOp(RewriterBase &rewriter, OpTy opTy) {
...
+  }
+}
+
+// Given the operand type and the output type return the broadcast
+// to use in the XSMM call.
+static xsmm::BinaryFlags getBinaryBCast(MemRefType operandType,
+                                        MemRefType outputType,
+                                        size_t operandNumber) {
+
+  enum class BCastType { NONE = 0, SCALAR, ROW, COL };
```
rengolin, line 134: Just use `xsmm::BinaryFlags`?

lib/TPP/ConvertTppToXsmm.cpp
```diff
@@ -151,6 +91,103 @@ getSizesAndLeadingDimsForGemmLikeOp(RewriterBase &rewriter, OpTy opTy) {
...
+  auto shapeOutput = outputType.getShape();
+  auto shapeOperand = operandType.getShape();
+  assert(shapeOutput.size() >= shapeOperand.size() &&
+         "Output rank must be >= operand rank");
+  SmallVector<int64_t> bOperandShape;
+  computeBcastShapeInput(shapeOutput, shapeOperand, bOperandShape);
+  assert(shapeOutput.size() == bOperandShape.size());
+  assert(shapeOutput.size() == 2);
+
+  auto getBCastEnum = [](BCastType bCastType,
```
rengolin, line 145: I get it that this is code move, but lets avoid lambdas whenever possible. They're horrible to debug.

The alternative here is to swap the order of the `if/else` pattern below adding the results to local variables and then have the switch on that local variable.

Very little code will actually change here.

lib/TPP/CombineXsmmPass.cpp
chelini: No, current matchers allow you to check the properties of a single op; here, you need something to detect a DAG. MLIR has some matchers for this: https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/Matchers.h

lib/TPP/CombineXsmmPass.cpp
rengolin: For the record, initially we should not worry about `allocs`. We can clean these up on an earlier pass, or worry about them after we create the first implementation. Right now, let's assume all is in-place.

lib/TPP/ConvertTppToXsmm.cpp
rengolin: As discussed, this code move is unnecessary and will be reverted.

lib/TPP/ConvertTppToXsmm.cpp
rengolin: As discussed, this code move is unnecessary and will be reverted.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -309,6 +309,7 @@ struct TppLowering : public tpp::impl::TppLoweringBase<TppLowering>,
       } else {
         // Tpp to Xsmm conversion patterns.
         pm.addPass(createConvertTppToXsmm());
+	pm.addPass(createCombineXsmmOpPass());
```
rengolin, line 312: No, this pass needs to go in the `linalgToXsmm` path above, not the TPP path.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,30 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
...
```
rengolin, line 153: You don't need `!isa<ShapedType>(type)` since you already have `!isa<MemRefType>(type)`.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,181 @@
...
+#define GEN_PASS_DEF_COMBINEXSMMOPPASS
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+namespace {
+
+static FailureOr<DenseI64ArrayAttr>
+getSizesAndLeadingDimForBrgemmOp(RewriterBase &rewriter, xsmm::BrgemmOp opTy) {
+
+  auto memrefC = opTy.getOperand(3).getType();
```
rengolin, line 33: nit: keep order 1, 2, 3.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,181 @@
...
+}
+
+struct CombineXsmmOp : public OpRewritePattern<xsmm::BrgemmOp> {
+  using OpRewritePattern<xsmm::BrgemmOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(xsmm::BrgemmOp brgemmOp,
+                                PatternRewriter &rewriter) const override {
+    auto use = brgemmOp.getOperand(3).getDefiningOp();
+    xsmm::BinaryOp binaryOp;
+    xsmm::UnaryOp unaryOp;
+    for (auto user : use->getUsers()) {
```
rengolin, line 93: This is insufficient.

If the pattern is `GEMM + RELU + ADD`, this will match and it will fuse wrong. Not that there's any reason for the pattern to be that, but we need to write safe matchers.

There are the following cases:
* GEMM + ADD + RELU: match all three
* GEMM + ADD: match binary only
* GEM + RELU: match unary only

Everything else, for now, is not a match.

Variations we need to detect and reject:
* GEMM + RELU + ADD
* GEMM + ??? + ADD + RELU
* GEMM + ADD + ??? + RELU

Variations we don't care but still need to match:
* GEMM + ADD + RELU + ??? -> FUSED_GEMM + ???
* Zero + GEMM + ADD + RELU -> Zero + FUSED_GEMM
* Later:  Zero + FUSED_GEMM -> FUSED_GEMM(Beta=0)

There are two ways of doing this without graph structured matchers:
1. Force the order in a particular way for the users (`getUser[1].isa<Add>()` etc). This is inflexible and leads to too many if/else blocks.
2. Loop through all users (like you do), but verify that you have the right number of ops in the right order. You can match expected ops, even if you don't handle them (ex. Zero, for now). But you have to continue on everh match, and you have to stop once you already have matched all you can.

To do (2) above, you need to:
* Check for Zero, and ignore for now, `continue`.
* Check for Add, verify structure/flags/operands, save and `continue`
* Check for Relu, verify structure/flags/operands, save and `continue`
* If you have a unary (which is meant to be the last op), you must stop looking.

If you don't continue on each match, you'll be trying to match everything every time. If you don't stop at the unary, you'll try to fuse following ops that don't fuse.

In the end, you need to add the fused op and remove all matches above. Because this is in `memref`, you don't need to reconstruct the chain between the last non-match op from before and the first non-match op from after, since they'll all be operating (reading/writing) directly to the `memref` (alloc, subview).

test/Passes/pass-combine-xsmm.mlir
```diff
@@ -0,0 +1,147 @@
+//RUN: tpp-opt %s -combine-xsmm-op-optimization -split-input-file | FileCheck %s
+
+  func.func @entry1(%arg0: memref<8x32x32x32xf32>, %arg1: memref<32x32x32x32xf32>, %arg2: memref<1024xf32>, %arg3: memref<8x32x32x32xf32>, %arg4: memref<32x32x32x32xf32>, %arg5: memref<1024xf32>, %arg6: memref<8x32x32x32xf32>, %arg7: memref<32x32x32x32xf32>, %arg8: memref<1024xf32>, %arg9: memref<8x32x32x32xf32>) {
```
rengolin, line 3: The tests are confusing. Please add a comment to what they're testing and name them appropriately.

You need to test all the variations I mention above for both matches and non-matches.

test/Passes/DefaultPipeline/tpp-conversion.mlir
```diff
@@ -63,6 +63,5 @@ func.func @linalg_dialect_expect_fused_brgemm(
 // CHECK-LABEL: linalg_dialect_expect_fused_brgemm
 // CHECK-SAME:  %[[ARG0:.+]]: tensor<1x3x2xf32>, %[[ARG1:.+]]: tensor<1x2x3xf32>, 
 // CHECK-SAME:  %[[ARG2:.+]]: tensor<3x3xf32>, %[[ARG3:.+]]: tensor<1x3xf32>
-// CHECK: %{{.+}} = tpp.fused_brgemm [unary = relu, binary = add] 
+// CHECK: %{{.+}} = tpp.fused_brgemm [unary = relu, binary = add]
```
rengolin, line 66: unnecessary whitespace change

test/Passes/pass-combine-xsmm.mlir
rengolin: Looking again, I do not understand these tests at all. 

The `col0` vs. `col1` tests only change the flag, not the arguments. Aren't they supposed to mark which argument (first or second) is the one being broadcasted?

There are multiple variations of tests with more than one flags but only the first is ever checked. `entry4` has one flag but I don't understand what is the difference or why it should fail. The shape is wrong for `row` broadcast, but that's not what the check is doing, it's only checking for `row` bcast.

You need tests for { `col0`, `col1`, `row0`, `row1` } and set the appropriate shape (`<1024xf32>` and `<256x1xf32>`) and match all of those. Having `<1024xf32>` and `bcast_row_in0` with `<256x1024xf3>` should be a verification error (see #779 and #581).

`entry5` is one case where it should not match, repeating the `relu` is another and needs a test.

rengolin:
@KavithaTipturMadhu this is the code that I expected to see on the matcher side:
https://github.com/rengolin/tpp-mlir/tree/xsmm_fusion

It's super verbose, splatted out into one big function, but it has the basic structure: match loop, flags validation, rewrite.

The checks in the match loop are repeated but necessary. We need to make sure we only match one. The loop goes through each operation multiple times, once for each operand (use), so we need to be careful and only fail when it's actually another op, not just the same op on another argument.

It does not check for everything. What's missing is:
* Check that the broadcast argument follows the flag (arg0, arg1, col/row)
* Check that the "other" input to the ADD isn't an input to the BRGEMM
* Check that the uses are actually in the output of the binary/unary (input won't match correctly)

I'm sure there's more, but those are the basic ones we must have, and we must have tests for each of those corner cases.

Going through the code changes led to many segfaults on the compiler and the generated libxsmm kernels because the calls aren't correct (ex. row bcast flag when the shape row tensor or arg0 bcast when arg1 is smaller), which also needs to be checked (first point above).

So we need as many "positive" tests (actual matches) as "negative" ones (shouldn't match), and verify that the kernel call is indeed correct.

I hope the code I wrote today can help you move forward from the base match+check problem to the more advanced phase of understanding the shape of the IR, but that starts by writing all tests and seeing how the changes behave across them. 

Feel free to ping me if there's anything you don't understand in the code or the tests.

rengolin:
With the minimal modifications for this PR to detect fusion on the benchmarks (from my branch), I get the error:

```
loc("<stdin>":21:10): error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: xsmm.fused_brgemm.dispatch
tpp-run: /home/rengolin/devel/intel/tpp-mlir/tools/tpp-run/tpp-run.cpp:199: std::unique_ptr<llvm::Module> lowerToLLVMIR(Operation *, llvm::LLVMContext &): Assertion `llvmModule' failed.
```

This means `xsmm.fused_brgemm.dispatch` does not get converted into function calls, while the invoke part is. Please add an execution test to make sure that gets done correctly.

rengolin:
@KavithaTipturMadhu, I have factored all helper functions into `XsmmHelper.cpp` and made enough changes to support all the cases we can possibly have. The code is _untested_, so it may have bugs, but the intention and the mechanics should be correct.

The code is now in my branch: https://github.com/rengolin/tpp-mlir/tree/xsmm_fusion

Please create tests for all of the options that we should and should not match and run through that code, you should get the right results or fine tune my matchers.

We should also add integration tests to make sure that the xsmm calls you are generating have the correct format and don't crash execution.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -282,6 +284,129 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+  }
+  // We don't know how to fuse more than two tail ops after the BRGEMM
+  if (chain.size() > 3)
+    return failure();
+  // List is in reverse order, put the brgemm at the top
+  std::reverse(chain.begin(), chain.end());
+  // If we haven't found a BRGEMM, this are not the droids we're looking for
+  assert(isa<xsmm::BrgemmOp>(chain[0]) && "First op must be brgemm");
+
+  // New, we're sure we have a chain, but not yet if it has the right types
```
rengolin, line 324:
```suggestion
  // Now, we're sure we have a chain, but not yet if it has the right types
```

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -282,6 +284,129 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+  return fusedMatch;
+}
+
+FailureOr<int64_t> getLeadingDim(Type type, size_t pos) {
+  // Not shaped type, the leading dimension is the single scalar.
+  if (!isa<ShapedType>(type))
+    return 1;
+  if (!isa<MemRefType>(type))
+    return 1;
+  MemRefType memref = type.cast<MemRefType>();
```
rengolin, line 392:
```suggestion
  // Not shaped type, the leading dimension is the single scalar.
  auto memref = dyn_cast<MemRefType>(type);
  if (!memref)
    return 1;
```

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+getSizesAndLeadingDimForBrgemmOp(RewriterBase &rewriter, xsmm::BrgemmOp opTy) {
+
+  auto memrefC = opTy.getOperand(3).getType();
+  auto memrefA = opTy.getOperand(1).getType();
+  auto memrefB = opTy.getOperand(2).getType();
+
+  int64_t m, n, k;
+  if (!isa<ShapedType>(memrefC) || !isa<ShapedType>(memrefA)) {
+    return failure();
+  }
```
rengolin, line 39:
```suggestion
  if (!isa<MemRefType>(memrefC) || !isa<MemRefType>(memrefA)) {
    return failure();
  }
  int64_t m, n, k;
```

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+} // namespace tpp
+} // namespace mlir
+namespace {
+
+static FailureOr<DenseI64ArrayAttr>
+getSizesAndLeadingDimForBrgemmOp(RewriterBase &rewriter, xsmm::BrgemmOp opTy) {
+
+  auto memrefC = opTy.getOperand(3).getType();
+  auto memrefA = opTy.getOperand(1).getType();
+  auto memrefB = opTy.getOperand(2).getType();
```
rengolin, line 34:
```suggestion
  auto memrefA = opTy.getOperand(1).getType();
  auto memrefB = opTy.getOperand(2).getType();
  auto memrefC = opTy.getOperand(3).getType();
```

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,250 @@
+//RUN: tpp-opt --combine-xsmm-op-optimization %s --split-input-file | FileCheck %s
```
rengolin, line 1:
```suggestion
//RUN: tpp-opt -verify-xsmm-calls -combine-xsmm-op-optimization -verify-xsmm-calls %s --split-input-file | FileCheck %s
```

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,250 @@
...
+// CHECK-LABEL: func.func @entry(
+// CHECK: %[[ARG0:.*]]: memref<256x128xf32>) -> memref<256x512xf32> {
+// CHECK: xsmm.fused_brgemm.dispatch
+// CHECK: xsmm.fused_brgemm
+
+memref.global "private" constant @__constant_4x32x32xf32 : memref<4x32x32xf32> = dense<1.000000e+00> {alignment = 128 : i64}
+memref.global "private" constant @__constant_8x32x32xf32 :  memref<8x32x32xf32> = dense<1.000000e+00> {alignment = 128 : i64}
+memref.global "private" constant @__constant_32xf32:  memref<32xf32, strided<[32], offset:?>> = dense<1.000000e+00> {alignment = 128 : i64}
+
+//bcast_col_in0 flag set on binary add
```
rengolin, line 12: Instead of comment + generic function name, change the function name to what the comment says, then it's east to map what CHECK lines apply to what functionality, not just what split test.

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,250 @@
+//RUN: tpp-opt --combine-xsmm-op-optimization %s --split-input-file | FileCheck %s
+
+// CHECK-LABEL: func.func @entry(
+// CHECK: %[[ARG0:.*]]: memref<256x128xf32>) -> memref<256x512xf32> {
+// CHECK: xsmm.fused_brgemm.dispatch
```
rengolin, line 5: Please also show the args and flags in the CHECK lines

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,250 @@
...
+  %4 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 32, 1024, 1024] flags = (beta_0) data_type = f32
+  %5 = xsmm.binary.dispatch add [32, 32, 32, 32, 32] flags = (none) data_type = f32
+  %6 = xsmm.unary.dispatch relu [32, 32, 32, 32] flags = (none) data_type = f32
+  %alloc_1 = memref.alloc() {alignment = 64 : i64} : memref<256x512xf32>
+  scf.parallel (%arg3, %arg2) = (%c0, %c0) to (%c8, %c8) step (%c1, %c1) {
+        %subview = memref.subview %alloc_0[%arg3, %arg2, 0, 0] [1, 1, 32, 32] [1, 1, 1, 1] : memref<8x8x32x32xf32> to memref<32x32xf32, strided<[32, 1], offset: ?>>
+        %subview_2 = memref.subview %alloc[%arg3, 0, 0, 0] [1, 4, 32, 32] [1, 1, 1, 1] : memref<8x4x32x32xf32> to memref<4x32x32xf32, strided<[1024, 32, 1], offset: ?>>
+        xsmm.brgemm(data_type = f32, %4, %subview_2, %0, %subview, %c4_i64) : (i64, memref<4x32x32xf32, strided<[1024, 32, 1], offset: ?>>, memref<4x32x32xf32>, memref<32x32xf32, strided<[32, 1], offset: ?>>, i64) -> ()
+        xsmm.binary add(data_type = f32, %5, %subview, %2,  %subview) : (i64, memref<32x32xf32, strided<[32,1], offset: ?>>,  memref<32xf32, strided<[32], offset:?>>, memref<32x32xf32, strided<[32, 1], offset: ?>>) -> ()
```
rengolin, line 119: This is broken IR. The flag says `none` but the shape is 1D. This test cannot guarantee behaviour.

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,250 @@
...
+  %c8_i64 = arith.constant 8 : i64
+  %cst = arith.constant 0.000000e+00 : bf16
+  %0 = memref.get_global @__constant_4x16x32x2xbf16 : memref<4x16x32x2xbf16>
+  %1 = memref.get_global @__constant_8x16x32x2xbf16 : memref<8x16x32x2xbf16>
+  %2 = memref.get_global @__constant_32xbf16 : memref<32xbf16, strided<[32], offset:?>>
+  %alloc = memref.alloc() {alignment = 64 : i64} : memref<8x4x32x32xbf16>
+  %3 = xsmm.unary.dispatch identity [32, 32, 128, 32] flags = (none) data_type = bf16
+  %alloc_0 = memref.alloc() {alignment = 64 : i64} : memref<8x8x32x32xbf16>
+  %4 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 32, 1024, 1024] flags = (vnni_b, beta_0) data_type = bf16
+  %5 = xsmm.binary.dispatch add [32, 32, 32, 32, 32] flags = (none) data_type = bf16
```
rengolin, line 153: This test is broken. The flag says `none` but the shape is 1D. It should NOT be allowed to pass.

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,250 @@
...
```
rengolin, line 237: This test is broken. The flag says none but the shape is 1D. It should NOT be allowed to pass.

rengolin:
You need to add this to the default pipeline and enable integration tests, to make sure we're not only generating good IR but also lowering the calls correctly to XSMM. This should have a small impact on all benchmarks.

KavithaTipturMadhu:
@rengolin Updated the PR, we're good for another round of review.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+  // - strideA and strideB that represent the stride between different GEMM
+  // in BRGEMM.
+  int64_t strideA = lda * m;
+  int64_t strideB = ldb * k;
+  return DenseI64ArrayAttr::get(
+      rewriter.getContext(),
+      ArrayRef<int64_t>{m, n, k, lda, ldb, ldc, strideA, strideB});
+}
+
+static ArrayAttr getBrgemmFlags(RewriterBase &rewriter, xsmm::BrgemmOp opTy) {
```
chelini, line 76: Why not having this method in TableGen in the op?

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+  auto memrefB = opTy.getOperand(2).getType().cast<MemRefType>();
+  xsmm::GemmFlagsAttr gemmFlag =
```
chelini, line 78: What if we have BETA_0, do we loose it? This method does not looks correct to me.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+          ? xsmm::GemmFlagsAttr::get(rewriter.getContext(),
+                                     xsmm::GemmFlags::VNNI_B)
+          : xsmm::GemmFlagsAttr::get(rewriter.getContext(),
+                                     xsmm::GemmFlags::NONE);
+  return rewriter.getArrayAttr(gemmFlag);
+}
+
+struct CombineXsmmOp : public OpRewritePattern<xsmm::BrgemmOp> {
+  using OpRewritePattern<xsmm::BrgemmOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(xsmm::BrgemmOp brgemmOp,
```
chelini, line 90: nit new line pls.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+        dtype);
+
+    Value batchDim = rewriter.create<arith::ConstantOp>(
+        loc, integer64, rewriter.getIntegerAttr(integer64, batchSize));
+    SmallVector<Value, 6> invokeOperands;
+    invokeOperands.push_back(dispatched);
+    auto opItr = brgemmOp->getOperands().begin();
+    std::advance(opItr, 1);
+    invokeOperands.append(opItr, brgemmOp->getOperands().end());
+    // Drop the aliasing output operand.
```
chelini, line 150: sure about this? It was specific to tpp to linalg conversion.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+        rewriter.getArrayAttr(
+            xsmm::BinaryFlagsAttr::get(rewriter.getContext(), *binaryFlags)),
+        dtype);
+
+    Value batchDim = rewriter.create<arith::ConstantOp>(
+        loc, integer64, rewriter.getIntegerAttr(integer64, batchSize));
+    SmallVector<Value, 6> invokeOperands;
+    invokeOperands.push_back(dispatched);
+    auto opItr = brgemmOp->getOperands().begin();
+    std::advance(opItr, 1);
```
chelini, line 148: can you please add a comment here? Are you skipping the dispatch?

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+    auto dtype =
+        xsmm::utils::getDataType(rewriter, brgemmOp.getOperand(1).getType());
+    IntegerType integer64 = IntegerType::get(rewriter.getContext(), 64);
+
+    Location loc = brgemmOp.getLoc();
+    auto dims = getSizesAndLeadingDimForBrgemmOp(rewriter, brgemmOp);
+    auto memrefB = brgemmOp.getOperand(2);
+    int64_t batchSize = memrefB.getType().cast<ShapedType>().getShape()[0];
+
+    OpBuilder::InsertionGuard guard(rewriter);
```
chelini, line 130: why are you changing the insertion point?

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+    brgemmOp.getOperand(0).getDefiningOp()->erase();
+    fusedMatch.binaryOp->erase();
+    fusedMatch.binaryOp->getOperand(0).getDefiningOp()->erase();
+    fusedMatch.unaryOp->erase();
+    fusedMatch.unaryOp->getOperand(0).getDefiningOp()->erase();
+    return success();
+  }
+};
+
+void populateCombinePatterns(RewritePatternSet &patterns) {
```
chelini, line 167: Unless we plan to expose this we can drop for now.

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,247 @@
+//RUN: tpp-opt -verify-xsmm-calls  --combine-xsmm-op-optimization  -verify-xsmm-calls %s --split-input-file | FileCheck %s
```
chelini, line 1: General comments on the tests: They are really difficult to follow. Can we simplify them?

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,247 @@
...
+
+// CHECK-LABEL: func.func @bcast_col_in0_on_binary_add(
```
chelini, line 3: Please move the check lines after the IR.

chelini (commented):
Left some comments. I think it still requires some more iterations, but we are getting there.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+namespace {
+
+static FailureOr<DenseI64ArrayAttr>
+getSizesAndLeadingDimForBrgemmOp(RewriterBase &rewriter, xsmm::BrgemmOp opTy) {
+
+  auto memrefA = opTy.getOperand(1).getType();
+  auto memrefB = opTy.getOperand(2).getType();
+  auto memrefC = opTy.getOperand(3).getType();
+
+  if (!isa<ShapedType>(memrefC) || !isa<ShapedType>(memrefA)) {
```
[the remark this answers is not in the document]
chelini: There are other stmts below please take care of them too.

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -38,6 +39,21 @@ struct BinaryInfo {
...
+/// should have already been converted to flags. All stray allocations
+/// should have already been converted to in-place reuse. Init zero
+/// should have already been converted to Beta=0.
+struct FusedMatch {
+  // This is the BRGEMM op
+  BrgemmOp brgemmOp;
+  // This is the (optional) binary op that follows the GEMM
+  BinaryOp binaryOp;
+  BinaryKind binaryKind;
+  // This is the (optional) binary op that follows the GEMM/Binary
```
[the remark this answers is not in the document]
KavithaTipturMadhu: It's brgemm followed by binary add followed by unary relu that gets fused in one op.

lib/TPP/Transforms/CombineXsmmPass.cpp
rengolin: Ah, yes! @alheinecke @egeor does libxsmm support fused brgemm with BETA=0?

lib/TPP/Transforms/CombineXsmmPass.cpp
rengolin: Because the matcher starts at the BRGEMM location but in case there's an ADD, the subview for the bias is created _after_ the BRGEMM. If we insert the fused BRGEMM _before_ the bias subview, we create an IR problem.

This insertion point is to make sure we place the fused BRGEMM at the binary op level (because the unary doesn't need any extra subviews).

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,179 @@
...
+    fusedMatch.unaryOp->getOperand(0).getDefiningOp()->erase();
+    return success();
+  }
+};
+
+void populateCombinePatterns(RewritePatternSet &patterns) {
+  patterns.add<CombineXsmmOp>(patterns.getContext());
+}
+
+struct CombineXsmmOpPass
```
rengolin, line 171: This must be added to the default pipeline and we must have Pass tests with `-default-tpp-passes` and Integration tests with `tpp-run` showing that the lowering is correct.

Earlier, I had crashes when the invoke was trying to call a null pointer, meaning the dispatch was never called (DCE'd?).

lib/TPP/Transforms/CombineXsmmPass.cpp
KavithaTipturMadhu: Indeed, will add a comment

lib/TPP/Transforms/CombineXsmmPass.cpp
KavithaTipturMadhu: This will be done in subsequent iterations, I tried moving it but kept encountering compilation errors.

lib/TPP/Transforms/CombineXsmmPass.cpp
KavithaTipturMadhu: Updated the comment

lib/TPP/Transforms/CombineXsmmPass.cpp
KavithaTipturMadhu: We expose the pass to tpp-opt so keeping it for now.

test/Passes/xsmm-combine.mlir
KavithaTipturMadhu: We're getting the IR from mlir-gen, hence simplifying the IR isn't trivial.

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,254 @@
...
+	xsmm.unary relu(data_type = f32, %6, %subview, %subview) : (i64, memref<32x32xf32, strided<[32, 1], offset: ?>>, memref<32x32xf32, strided<[32, 1], offset: ?>>) -> ()
+	scf.yield
+  }
+  return %alloc_1 : memref<256x512xf32>
+ }
+
+// CHECK-LABEL: func.func @bcast_col_in0_on_binary_add(
+// CHECK: %[[ARG0:.*]]: memref<256x128xf32>) -> memref<256x512xf32> {
+// CHECK: %[[dispatch:.*]] = xsmm.fused_brgemm.dispatch [32, 32, 32, 32, 32, 32, 1024, 1024][add,relu]  flags = (none)  binary_flags = (bcast_col_in0)  unary_flags = (none) data_type = f32
+// CHECK: xsmm.fused_brgemm
```
rengolin, line 40: Add an CHECK line for %2 and make sure it's in the right order on the invoke:
```mlir
  CHECK: %[[BIAS:.*]] = memref.get_global {{.*}} memref<32xf32>
  ...
  CHECK: xsmm.fused_brgemm(data_type = f32, [[dispatch]], {{.*}}, {{,.*}}, {{.*}}, [[BIAS]]) 
```

Same below with `col_in1` etc to make sure you got the right bias to fuse.

test/Passes/xsmm-combine.mlir
```diff
@@ -0,0 +1,254 @@
...
+	xsmm.binary add(data_type = f32, %5,  %subview, %2, %subview) : (i64, memref<32x32xf32, strided<[32,1], offset: ?>>,  memref<32xf32, strided<[32], offset:?>>, memref<32x32xf32, strided<[32, 1], offset: ?>>) -> ()
+	xsmm.unary relu(data_type = f32, %6, %subview, %subview) : (i64, memref<32x32xf32, strided<[32, 1], offset: ?>>, memref<32x32xf32, strided<[32, 1], offset: ?>>) -> ()
+	scf.yield
+  }
+  return %alloc_1 : memref<256x512xf32>
+ }
+
+// CHECK-LABEL: func.func @bcast_col_in1_on_binary_add(
+// CHECK: %[[ARG0:.*]]: memref<256x128xf32>) -> memref<256x512xf32> {
+// CHECK-NOT: %[[dispatch:.*]] = xsmm.fused_brgemm.dispatch [32, 32, 32, 32, 32, 32, 1024, 1024][add,relu]  flags = (none)  binary_flags = (bcast_col_in1)  unary_flags = (none) data_type = f32
```
rengolin, line 81: IIUC, the `fused_brgemm` flag should _always_ be `bcast_col_in0` even if the `add` flag was `in1`. 

Same comment below on BF16.

@alheinecke

lib/TPP/Transforms/CombineXsmmPass.cpp
KavithaTipturMadhu: Sorry I missed this comment, there were no tests that were crashing when I ran cmake locally, I don't see any tests in test/Integration with fusion pattern which require updates. Is there something I am missing?

test/Passes/xsmm-combine.mlir
KavithaTipturMadhu: This is debatable, shouldn't the fusion happen along in1 dimension technically? Isn't the limitation due to libxsmm's current behaviour?

rengolin: I meant as an API restriction (both `in0` and `in1` would be converted to `in0` because there is no `0` or `1` in the fused, only one binary arg), but it seems the fusion generator can cope with either: https://github.com/libxsmm/libxsmm/blob/b520352e4be31ef71d4f24684c898b945f1c1a58/src/generator_gemm_common.c#L1684.

Once we have an integration test that shows both cases work (dispatch and invoke and print result), we can rest assured this actually works.

lib/TPP/Transforms/CombineXsmmPass.cpp
rengolin: My interpretation of libxsmm's code is that the `LIBXSMM_GEMM_FLAG_BETA_0` is handled on all tile/packed/fused generators (mostly on store sequences), so it's independent of which op was originally called.

The libxsmm-dnn example for DNN models seems to corroborate it: https://github.com/libxsmm/libxsmm-dnn/blob/2075b99f38af245a81293de1d0b35a5224de93c7/src/libxsmm_dnn_fc.c#L289

Anyway, if we don't detect BETA=0 we won't do any fusion in real world code because both TF and PT examples get BETA=0 through out code. You should pass the flag and make sure the execution is correct in the integration test, as below.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -0,0 +1,193 @@
...
+    if (unaryFlags != mlir::xsmm::UnaryFlags::BCAST_SCALAR &&
+        unaryFlags != mlir::xsmm::UnaryFlags::NONE)
+      return failure();
+
+    // TODO: Support more than just COL_0 BCAST
+    auto binaryFlags =
+        xsmm::utils::getBinaryFlags(fusedMatch.binaryOp.getOperand(1).getType(),
+                                    fusedMatch.binaryOp.getOperand(3).getType(),
+                                    mlir::xsmm::utils::OperandPos::LHS);
+    if (binaryFlags != mlir::xsmm::BinaryFlags::BCAST_COL_IN_0)
```
rengolin, line 126: Checked with @alheinecke, the behavior I describe above will change soon, so we should just have a simple conversion in the lowering:
```c++
  int binaryArg = 0;
  switch (binaryFlags) {
    case BinaryFlags::BCAST_COL_IN_0:
      binaryArg = 2;
      break;
    case BinaryFlags::BCAST_COL_IN_1:
      binaryArg = 3;
      binaryFlags = BinaryFlags::BCAST_COL_IN_0;
      break;
    default:
      return failure();
  }
```

Then lower down, you replace:
```c++
  invokeOperands.push_back(fusedMatch.binaryOp->getOperand(2));
```

with:
```c++
  invokeOperands.push_back(fusedMatch.binaryOp->getOperand(binaryArg));
```

KavithaTipturMadhu:
@rengolin I have added an integration test and updated the matching algorithm a little bit too. Please take a look.

lib/TPP/Transforms/CombineXsmmPass.cpp
rengolin: You are now ignoring BETA=0, which means it will not apply to any current integration test or benchmark. In addition, your integration test does not check the output of the execution, which means we don't know if it's doing the right thing in addition to lowering correctly.

You also seem to have removed the `xsmm.zero` from the IR tests below, which means now you have adapted your tests to the code and not the other way around. This way, we'll never know if your code can match the IR we actually see in the wild.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -317,8 +318,10 @@ FailureOr<FusedMatch> getFusedBrgemmSequenceFromProducer(Operation *op) {
   // We don't know how to fuse more than two tail ops after the BRGEMM
   if (chain.size() > 3)
     return failure();
-  // List is in reverse order, put the brgemm at the top
-  std::reverse(chain.begin(), chain.end());
+  if (!isa<xsmm::BrgemmOp>(chain[0]))
```
rengolin, line 321: I don't understand this change. In which case the BRGEMM ends up as the first op from the loop above? Is this now coming in program order?

lib/TPP/Transforms/CombineXsmmPass.cpp
KavithaTipturMadhu: I thought beta=0 was a case which was no longer supposed to be fused, therefore I am not sure how to make it applicable to an integration test other than build a case where I initialize beta to 0 and not fuse. As for the end to end case, I will update the integration test to check results. 

I am not sure what you mean by xsmm.zero, do you mean beta assigned to xsmm.zero?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Yes, this was in the integration test where the uses were traversed with brgemm first.

lib/TPP/Transforms/CombineXsmmPass.cpp
rengolin: > I thought beta=0 was a case which was no longer supposed to be fused

I thought we could (as per my now deprecated comment above), but later I realised (and commented above) that all BRGEMMs we have on tests and benchmarks have BETA=0, so we cannot ignore it.

> I am not sure what you mean by xsmm.zero, do you mean beta assigned to xsmm.zero?

Sorry, I meant `xsmm.brgemm.dispatch ... flags = (beta_0)` to `flags = (none)` in https://github.com/plaidml/tpp-mlir/pull/752/commits/e2a3597d066c410d2f0e52214404181638652c50.

rengolin:
The pass is still not fusing the input from `mlir-gen`... Can you create a test that is like your integration test but:

`// RUN: mlir-gen | tpp-run ... -print-mlir=mid -print 2>&1 | FileCheck`

And then CHECK for both `xsmm.fused_brgemm` and the result values?

KavithaTipturMadhu:
> // RUN: mlir-gen | tpp-run ... -print-mlir=mid -print 2>&1 | FileCheck

I have added a test case that uses mlir-gen instead of my own integration test and that seems to fuse two layers generated by mlir-gen just fine.

rengolin (approved):

PR #756, opened 2023-10-24
CI LLVM GPU build

adam-smnk:
Adds CI logic to build LLVM with GPU support and reenable TPP GPU testing jobs.
 
Dependency management is simplified by creating separate builds of both LLVM and TPP-MLIR with extra features (CUDA, Vulkan) enabled depending on the test or benchmark coverage. This ensures that extra environment dependencies are not required for every build e.g., no runtime linkage to libcuda.so.1 when no CUDA related logic is needed by a CI job.
CI agents now install different LLVM builds for the same LLVM version (commit SHA).

Minor tweaks are added to the CI scripts to ensure better error propagation.
'GPU-Intel-Vulkan' job is disabled due to missing GPU device in the Vulkan runtime.

### DIFF
--- a/scripts/buildkite/build_tpp.sh
+++ b/scripts/buildkite/build_tpp.sh
@@ -53,22 +53,7 @@ fi
 
 if [ "${GPU}" ]; then
   GPU_OPTION="-G ${GPU}"
-fi
-
-# Env CUDA setup
-if [[ ${GPU,,} =~ "cuda" ]]; then
-  echo "Setting up CUDA environment"
-  echo "Hard-coding CUDA-compatible GCC version (12.3)"
-  source /swtools/gcc/gcc-12.3.0/gcc_vars.sh
-  source /swtools/cuda/latest/cuda_vars.sh
-  check_program nvcc
-fi
-
-# Env Vulkan setup
-if [[ ${GPU,,} =~ "vulkan" ]]; then
-  echo "Setting up Vulkan environment"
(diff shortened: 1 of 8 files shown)


adam-smnk:
Successful LLVM builds: [CI link](https://buildkite.com/intel/tpp-llvm/builds/300)
Two extra LLVM build variants are added: `-cuda` and `-vulkan`. All LLVM builds are based on the same `llvm_version.txt` from TPP-MLIR. They are build, checked, and installed separately for the CI agent under different installation directories (see buildkite scripts e.g., `check_llvm.sh`, for the naming scheme).

Successful tpp-gpu tests: [CI link](https://buildkite.com/intel/tpp-gpu/builds/442)
GPU CI tests and benchmarks are passing once again. PVC vulkan job is disable due to missing PVC device in Vulkan runtime. But the Vulkan build is still validated on an Nvidia node.
CUDA benchmarks numbers looks correctly.

scripts/buildkite/build_tpp.sh
```diff
@@ -51,6 +55,22 @@ if [ "${GPU}" ]; then
   GPU_OPTION="-G ${GPU}"
 fi
 
+# Env CUDA setup
+if [[ ${GPU,,} =~ "cuda" ]]; then
```
chelini, line 59: would it make sense to have this in a top level .sh file?

chelini (approved):
Other than we suggested looks good.

PR #757, opened 2023-10-25
Use linalg-to-xsmm as default

chelini:
Move the default pipeline to use linalg to XSMM conversion, thus bringing us closer to deprecating TPPs.


adam-smnk (approved):
Thanks, looks good.
No regressions observed in the performance numbers. Tested locally too, everything seems fine.

PR #759, opened 2023-11-01
Fix default argument values after #757

adam-smnk:
Sets linalg-to-xsmm flag values in pass declarations to be true by default.


chelini:
I am retiring all of these. It is confusing and constructors are generated by TabGen.

adam-smnk:
> I am retiring all of these. It is confusing and constructors are generated by TabGen.

Cool, this will simplify things 👍

PR #761, opened 2023-11-02
Do not pick default in `DefaultTppPasses` (NFC)

chelini:
Localize pick for default blocking factors in the blocking passes rather than in the default pipeline. The default pipeline should not take any decision on defaults; passes should pick defaults.


rengolin (approved):
Indeed! This also makes it easier to move to a cost model based compiler. Thanks!

adam-smnk (approved):

PR #765, opened 2023-11-03
Add debug scripts for compiler passes

rengolin:
This is a simple set of scripts that run mlir-gen/tpp-opt, dumps the IR between passes, splits the output for each pass into files, looks at each one when the IR changes and runs the diff tool of your choice to show these differences.

For now, mlir-gen is only emitting the standard odd-shaped MLP layer over the standard TPP passes, but future changes to the scripts can add arguments to each tool.

### DIFF
--- a/scripts/debug/split.py
+++ b/scripts/debug/split.py
@@ -22,7 +22,7 @@ if __name__ == '__main__':
     ext=os.path.splitext(args.filename)[1]
     counter = 0
     filename = f"%03d{ext}" % counter
-    outfile = open(filename, "a")
+    outfile = open(filename, "w")
     with open(args.filename, "r") as file:
         for line in file:
             # New files
--- a/scripts/debug/split.py
+++ b/scripts/debug/split.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+
+argParser = argparse.ArgumentParser()
+argParser.add_argument('filename')
+
+def syntax():
+    print("Use -h for options\n")
+    exit(1)
+
+if __name__ == '__main__':
+    args = argParser.parse_args()
+    if args.filename is None:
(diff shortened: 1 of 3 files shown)


adam-smnk:
When I run `debug_all_passes.sh` it finishes with:
```
Traceback (most recent call last):
  File "/home/asiemien/plaidml-tpp-mlir/scripts/debug/diff.py", line 77, in <module>
    if sameFile(curr, next):
  File "/home/asiemien/plaidml-tpp-mlir/scripts/debug/diff.py", line 32, in sameFile
    rhs = readFile(next)
  File "/home/asiemien/plaidml-tpp-mlir/scripts/debug/diff.py", line 21, in readFile
    with open(file, "r") as fp:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

scripts/debug/split.py
```diff
@@ -0,0 +1,36 @@
...
+
+    if not os.path.isfile(args.filename):
+        print(f"{args.filename} is not a file\n")
+        syntax()
+
+    ext=os.path.splitext(args.filename)[1]
+    counter = 0
+    filename = f"%03d{ext}" % counter
+    outfile = open(filename, "a")
+    with open(args.filename, "r") as file:
```
adam-smnk, line 26: nit: could be combined into one `with` block to avoid explicit close

adam-smnk (approved):
Minus the mentioned error, looks nice 👍

scripts/debug/split.py
rengolin: It cannot, because the outfile is closed/opened inside the loop, it's not the same outfile, it's ALL outfiles.

adam-smnk: Ah right, somehow missed that 👍

PR #766, opened 2023-11-06
Use 64 as blocking factors for `omp` benchmarks

chelini:
In DNN we use 32, while in the IR generated by `mlir-gen` we use 64.

### DIFF
--- a/benchmarks/config/omp/dnn-bf16.json
+++ b/benchmarks/config/omp/dnn-bf16.json
@@ -5,28 +5,28 @@
       "type": "XSMM-DNN",
       "benchmark": "xsmm_dnn_mlp",
       "environment": { "OMP_NUM_THREADS": "1" },
-      "flags": [ "100", "256", "0", "F", "32", "32", "32", "1", "1024", "1024", "1024", "1024" ],
+      "flags": [ "100", "256", "0", "F", "64", "64", "64", "1", "1024", "1024", "1024", "1024" ],
       "extensions": []
     },
     "gemm_bf16_omp_4_dnn": {
       "type": "XSMM-DNN",
       "benchmark": "xsmm_dnn_mlp",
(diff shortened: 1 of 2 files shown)


rengolin:
My thinking was the other way round. Now that we can pass the tile sizes through `tpp-opt`, we should make them all `64`, because that's also what we use in libxsmm-dnn.

adam-smnk (commented):
Most kernels use 64 instead of 32. Maybe we should move the DNN to 64 instead?
Could you also align/update bf16 benchmarks?

chelini:
`omp/dnn-fp32.json` uses 32 as blocking factors. For me 64 or 32 is the same as long as we try to compare apple-with-apple. Should I use 64?

adam-smnk:
> `omp/dnn-fp32.json` (aka libxsmm-dnn) uses 32 as blocking factors. For me 64 or 32 is the same as long as we try to compare apple-with-apple. Should I use 64?

I think so, let's switch to 64 to align better our basic benchmarks same as `base.json` uses 64.
No need to change `config/matmul` benchmarks as they use values from parlooper.

rengolin:
Also, make sure to pass `tile-sizes=64,64,64` to `tpp-opt` on the non-pre-packed `base` benchmarks.

And to rename the PR. :)

chelini:
> > Also, make sure to pass `tile-sizes=64,64,64` to `tpp-opt` on the non-pre-packed `base` benchmarks.
> 
> Hm, it seems `default-tpp-passes` does not have that option to propagate. Can be on a separate PR, then.

Why not changing the default instead of passing cl options?

rengolin:
> Why not changing the default instead of passing cl options?

We could. But we still need a way to pass those options and right now we don't have good tests for that.

I don't mind if this PR has that change, or if we wait for another PR to have the pass options, but we need the latter anyway.

adam-smnk (approved):
Changes and benchmarks numbers look fine 👍

rengolin (approved):

PR #767, opened 2023-11-06
Allow fusion of fill with multiple users

chelini:
If an operation has multiple users, not all in the same fusion domain, fusion does not take the operation into account. Allowing operations with multiple users in various fusion domains introduces recomputation. Introduce an exception for FillOp. This allows us to bring fillOps operation in the fusion domain even after CSE and later mapping the operation to XSMM calls. Without this change we end up with scalar fills or copy.

### DIFF
--- a/lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
@@ -371,6 +371,12 @@ static llvm::SmallDenseSet<Operation *> collectFusableProducers(
     processingQueue.pop();
     for (OpOperand &operand : currentOp->getOpOperands()) {
       Operation *producer = operand.get().getDefiningOp();
+      if (producer && isa<linalg::FillOp>(producer)) {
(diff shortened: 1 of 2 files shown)


chelini:
this PR improves performance in the `omp` benchmarks.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -492,7 +498,8 @@ static FailureOr<scf::SCFTileAndFuseResult> fuseWithEltwise(
     Operation *candidateOp = getUntiledProducerFromSliceSource(
         &candidateSliceOp->getOpOperand(0), forLoops);
     if (!candidateOp || worklist.count(candidateOp) == 0 ||
-        alreadyFusedOps.count(candidateOp))
+        (alreadyFusedOps.count(candidateOp) &&
+         !isa<linalg::FillOp>(candidateOp)))
       continue;
```
adam-smnk, line 503: nit: braces

adam-smnk (approved):

Issue #770, opened 2023-11-07
tpp-run crashes on TorchDynamo generated MLP


nhasabni:
I was trying to run TorchDynamo MLP model (via its MLIR) with TPP-run for benchmarking. But I am getting a crash with `tpp-run`. Actual MLIR attached below.

**tpp-mlir commit** (Nov 2)
`5bba2907effe2d34fbb6c6af50f5eff8cb7e1ce9`

**Command and output**
```
$ tpp-run mlir/linear.mlir -e forward -entry-point-result=void
tpp-run: /nfs_home/nhasabni/other/tensor_compiler/nhasabni_tpp-sandbox/llvm-project/mlir/include/mlir/IR/OpDefinition.h:890: mlir::Block *mlir::OpTrait::SingleBlock<mlir::scf::ForallOp>::getBody(unsigned int) [ConcreteType = mlir::scf::ForallOp]: Assertion `!region.empty() && "unexpected empty region"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
 #0 0x000000000067af6c llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /data/nfs_home/nhasabni/other/tensor_compiler/nhasabni_tpp-sandbox/llvm-project/llvm/lib/Support/Unix/Signals.inc:723:13
 #1 0x0000000000678e50 llvm::sys::RunSignalHandlers() /data/nfs_home/nhasabni/other/tensor_compiler/nhasabni_tpp-sandbox/llvm-project/llvm/lib/Support/Signals.cpp:106:18
 #2 0x000000000067b7ff SignalHandler(int) /data/nfs_home/nhasabni/other/tensor_compiler/nhasabni_tpp-sandbox/llvm-project/llvm/lib/Support/Unix/Signals.inc:0:3
 #3 0x0000150348854df0 __restore_rt (/usr/lib64/libc.so.6+0x54df0)
 #4 0x00001503488a154c __pthread_kill_implementation (/usr/lib64/libc.so.6+0xa154c)
 #5 0x0000150348854d46 gsignal (/usr/lib64/libc.so.6+0x54d46)
 #6 0x00001503488287f3 abort (/usr/lib64/libc.so.6+0x287f3)
 #7 0x000015034882871b _nl_load_domain.cold (/usr/lib64/libc.so.6+0x2871b)
[...]

PR #771, opened 2023-11-07
Do not fold unpack when for has multiple uses

chelini:
We cannot take the body of the for if it has multiple uses. The for ends up with an empty body and the verifier fails.
Fix: #770


rengolin (approved):

nhasabni:
Thanks @chelini!

PR #772, opened 2023-11-07
Add linalg->xsmm conversion for sub operation

chelini:


lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -323,66 +323,90 @@ getBroadCastBinaryFlagFromMap(AffineMap map, unsigned operandIdx) {
...
+
+  auto flagLhs =
+      xsmm::BinaryFlagsAttr::get(rewriter.getContext(), *broadCastFlagLhs);
+  auto flagRhs =
+      xsmm::BinaryFlagsAttr::get(rewriter.getContext(), *broadCastFlagRhs);
+
+  // Spaghetti code to handle 'NONE' as it conflicts with other flags; we
+  // cannot add it if at least the RHS or the LHS is not 'NONE'. Maybe the
+  // best solution is to get rid of it.
+  ArrayAttr flags;
```
rengolin, line 360: For a different PR...

Perhaps this could go to a utility function somewhere else. Looks fairly independent.

Though `NONE` is just `0`, you can `if (!flagsLhs.getValue())`. Or you could make the attribute optional and not create it in the first place (for NONE)?

Also, you could just create a small vector and create the array attr from the vector?
```c++
  SmallVector flags;
  if (flagsLhs.getValue())
    flags.push_back(flagsLhs.getValue());
  if (flagsRhs.getValue())
    flags.push_back(flagsRhs.getValue());
  ArrayAttr flagsAttr = rewriter.getArrayAttr(flags);
```

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
```diff
@@ -261,3 +261,25 @@ func.func @add_10(%arg0: memref<256x1024xf32>, %arg1: memref<256xf32>) {
 // C_HECK: %[[DIS:.+]] = xsmm.binary.dispatch add [256, 1024, 1024, 1, 1024]
 // C_HECK-SAME:  flags = (bcast_col_in1) data_type = f32
 // C_HECK: xsmm.binary add(data_type = f32, %[[DIS]], %[[ARG0]], %[[ARG1]], %[[ARG0]])
+
+// -----
+
+#map = affine_map<(d0, d1) -> (d0, d1)>
+
+func.func @sub1(%arg0: memref<256x1024xf32>, %arg1: memref<256x1024xf32>) {
```
rengolin, line 269: Can you add examples for broadcast (scalar/col/row)?

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
chelini: `getValue()` return the attribute value and it is not convertible to zero (you still need to have `!= xsmm::BinaryFlags::NONE`). Switch the rest to use a SmallVector

chelini: The NONE business is a know annoying problem. I should have made it optional but at the moment is not (you cannot have an empty flags vector).

rengolin: No worries, it can be another PR.

rengolin (approved):

PR #774, opened 2023-11-07
[Bench] Move benchmarks to show apples-to-apples comparison

rengolin:
This change modifies the `base` benchmarks to show an actual comparison between libxsmm-dnn and our compiler on fair comparisons. It also splits the pack and MHA benchmarks to separate configuration files so they don't affect the baseline.

Assumptions below...

Packing behaviour:
 * libxsmm-dnn packs at 64x64x64 and uses VNNI (dp2 and dp4)
 * mlir-gen can generate similar MLIR off the bat, so the compiler tile and fuses optimally, generating the same results

BF16 comparison:
 * libxsmm-dnn changes behaviour on x86 and Arm
 * mlir-gen has dp2/dp4 variants that only run on x86/arm resp.

GEMM/MLP "Models":
 * "const" means weights and biases are constant, like inference
 * "args" means weights and biases are not constant, like training
 * All models are unpacked and the compiler needs to pack them
 * Packing is done at argument (const) and weight/bias (args)

Expectations:
 * Pre-packed benchmarks need to be as fast as XSMM-DNN
   * Any inconsistency needs to be addressed to achieve parity
 * Unpacked benchmarks will be off by a few percent
   * But still faster than unpacked tiled benchmarks
   * And much faster than unpacked untiled benchmarks

TODO:
 * Add options to the compiler to avoid packing to create benchmarks for unpacked tiling and fuse to compare as "baseline" for traditional compilers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,14 +67,21 @@ add_subdirectory(test)
 set(BENCHMARK_DIR "${PROJECT_SOURCE_DIR}/benchmarks")
 set(CONFIG_DIR "${BENCHMARK_DIR}/config")
 
+# Run baseline benchmarks with default iterations to track simple performance
+set(BENCH_CFGS
+  ${CONFIG_DIR}/base/base.json
+  ${CONFIG_DIR}/base/pack.json
+  ${CONFIG_DIR}/base/mha.json
+)
+string(JOIN ',' BENCH_CFGS_STR ${BENCH_CFGS})
 # Run a small set of benchmarks with small iterations to test the benchmarks and run locally on small machines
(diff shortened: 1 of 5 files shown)


rengolin:
Current status:
image

rengolin:
Take aways:
1. FP32 GEMM 94% & MLP 92% needs understanding what's different. There should be no difference.
2. BF16 is still broken. Even 70% is too low for pre-packed, the rest is just broken. Needs a serious look into it.
3. Packing costs 3.19% for GEMMs and 2.25% for MLPs over pre-packed. That's really good.
4. Const (inference) is understandably better than Args (training), especially with packing (a lot more to pack).
5. BF16 packing is still under development, hopefully will improve soon.

Though, none of this is related to this PR. Here, we just renamed and moved the benchmarks around for clarity.

chelini (approved):
thanks.

PR #777, opened 2023-11-08
Enable folding beta 0 in XSMM flags

chelini:
This PR allows us to generate the same micro-kernel as DNN, which has
been checked manually for the `prepacked_targets.` However, there is
still some performance gap with DNN. Some observations in regard:
- There are no allocations in DNN. Everything is pre-allocated. We do
  have allocations in the kernels. For example, `gemm_fp32_mlir` ends up
  with two mallocs.
- Allocations have different alignments (64 for us while 2097152 for
  DNN).

### DIFF
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -59,7 +59,7 @@ def Xsmm_UnaryOp : Xsmm_Op<"unary"> {
 
   let extraClassDeclaration = [{
     bool hasScalarInput() {
-      // skip the function pointer. The operand
+      // Skip the function pointer. The operand
       // is at position 1.
       Type operand = getInputs()[1].getType();
       if (!operand.isa<ShapedType>())
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -117,6 +117,10 @@ def Xsmm_FusedBrgemmOp : Xsmm_Op<"fused_brgemm"> {
     `(` `data_type` `=` $data_type `,` $inputs `)`
     attr-dict `:` functional-type($inputs, results)
   }];
+
+  let extraClassDeclaration = [{
+    Value getOutput() { return getInputs()[3]; }
+  }];
 }
 
 //===----------------------------------------------------------------------===//
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -83,7 +83,7 @@ def Xsmm_GemmOp : Xsmm_Op<"gemm"> {
   }];
 
   let extraClassDeclaration = [{
-    Value getOutput() { return getInputs().back(); }
+    Value getOutput() { return getInputs()[3]; }
   }];
 }
 
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -81,6 +81,10 @@ def Xsmm_GemmOp : Xsmm_Op<"gemm"> {
     `(` `data_type` `=` $data_type `,` $inputs `)`
     attr-dict `:` functional-type($inputs, results)
   }];
+
+  let extraClassDeclaration = [{
+    Value getOutput() { return getInputs().back(); }
+  }];
 }
 
 //===----------------------------------------------------------------------===//
@@ -94,6 +98,10 @@ def Xsmm_BrgemmOp : Xsmm_Op<"brgemm"> {
   let assemblyFormat = [{
     `(` `data_type` `=` $data_type `,` $inputs `)`
     attr-dict `:` functional-type($inputs, results)
+  }];
+
+  let extraClassDeclaration = [{
+    Value getOutput() { return getInputs()[3]; }
   }];
 }
 
--- a/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
@@ -894,6 +894,7 @@ static void fuseZeroWithGemmOrBrgemm(RewriterBase &rewriter,
                                      xsmm::UnaryOp rootOp) {
   LLVM_DEBUG(llvm::dbgs() << "[fuseZeroWithGemmOrBrgemm] Candidate op: "
                           << rootOp << "\n");
+  // 1. Check if we have a gemm zero initialized by rootOp.
   auto gemmLikeOp = getZeroInitGemmLikeOp(rootOp);
   if (!gemmLikeOp)
     return;
--- a/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
@@ -814,8 +814,9 @@ static void fuseZeroWithGemmOrBrgemm(RewriterBase &rewriter,
   LLVM_DEBUG(llvm::dbgs() << "[fuseZeroWithGemmOrBrgemm] Candidate op: "
                           << rootOp << "\n");
   // 1. Prove the alloc to be zero. Linear walk of the bb
-  // and make sure there are only pure operations between
-  // the zero op and the gemm.
+  // and make sure there are only side-effect free operations between
+  // the zero op and the gemm. Bail out if any operations take
+  // a subview from `dest`.
   Value dest = rootOp.getInputs().back();
   DenseSet<Operation *> destUsers(dest.getUsers().begin(),
                                   dest.getUsers().end());
@@ -832,10 +833,18 @@ static void fuseZeroWithGemmOrBrgemm(RewriterBase &rewriter,
   assert(blck && "must be a valid ptr");
   auto it = blck->begin();
   auto itEnd = blck->end();
-  while (it != itEnd && &*it != rootOp.getOperation())
+  while (it != itEnd && &*it != rootOp.getOperation()) {
+    // View may introduce aliasing.
+    if (auto view = dyn_cast<ViewLikeOpInterface>(&*it)) {
+      if (view.getViewSource() == dest)
+        return;
+    }
     it++;
+  }
+
   if (it == itEnd)
     return;
+
   while (++it != itEnd) {
     // Skip operations that do not touch `dest`.
     if (!destUsers.count(&*it))
--- a/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
@@ -782,12 +782,16 @@ void ConvertLinalgToXsmm::runOnOperation() {
     return signalPassFailure();
 }
 
+// Set the beta flags of a gemm dispatch to zero by cloning and updating the
+// clone.
 template <typename XsmmDisTy, typename XsmmTy>
 static void updateGemmOpFlags(RewriterBase &rewriter, XsmmDisTy gemmDispatchOp,
                               XsmmTy gemmOp) {
-  static_assert(llvm::is_one_of<XsmmDisTy, xsmm::GemmDispatchOp,
-                                xsmm::BrgemmDispatchOp>::value);
-  static_assert(llvm::is_one_of<XsmmTy, xsmm::GemmOp, xsmm::BrgemmOp>::value);
+  static_assert(
+      llvm::is_one_of<XsmmDisTy, xsmm::GemmDispatchOp, xsmm::BrgemmDispatchOp,
+                      xsmm::FusedBrgemmDispatchOp>::value);
+  static_assert(llvm::is_one_of<XsmmTy, xsmm::GemmOp, xsmm::BrgemmOp,
+                                xsmm::FusedBrgemmOp>::value);
 
   OpBuilder::InsertionGuard guard(rewriter);
   rewriter.setInsertionPoint(gemmDispatchOp);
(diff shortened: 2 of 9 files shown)


lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -808,6 +814,100 @@ void ConvertLinalgToXsmm::runOnOperation() {
...
+                                                xsmm::GemmFlags::BETA_0));
+    gemmDispatchOp.setFlagsAttr(rewriter.getArrayAttr(newFlags));
+  });
+}
+
+static void fuseZeroWithGemmOrBrgemm(RewriterBase &rewriter,
+                                     xsmm::UnaryOp rootOp) {
+  LLVM_DEBUG(llvm::dbgs() << "[fuseZeroWithGemmOrBrgemm] Candidate op: "
+                          << rootOp << "\n");
+  // 1. Prove the alloc to be zero. Linear walk of the bb
```
adam-smnk, line 848: There's no check that the same buffer is zeroed out. For example:
```
func.func @zero_test(%arg0: memref<1x32x32xf32>, %arg1: memref<1x32x32xf32>, %arg2: memref<32x32xf32>) {
  %c1_i64 = arith.constant 1 : i64
  %alloc = memref.alloc() {alignment = 64 : i64} : memref<32x32xf32>
  %0 = xsmm.unary.dispatch zero [32, 512, 1, 512] flags = (bcast_scalar) data_type = f32
  xsmm.unary zero(data_type = f32, %0, %cst, %alloc) : (i64, f32, memref<32x32xf32>) -> ()
  %1 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 32, 32, 1] flags = (none) data_type = f32
  // Sets incorrectly flags = (beta_0) when zero acts on %alloc not %arg2
  xsmm.brgemm(data_type = f32, %1, %arg0, %arg1, %arg2, %c1_i64) : (i64, memref<1x32x32xf32>, memref<1x32x32xf32>, memref<32x32xf32>, i64) -> ()
  return
}
```

Not saying we need to handle weird patterns but just avoid wrong codegen.

adam-smnk (commented):
Approach looks valid just needs to be more conservative.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -866,14 +869,23 @@ static void fuseZeroWithGemmOrBrgemm(RewriterBase &rewriter,
   if (it == itEnd)
     return;
   while (++it != itEnd) {
-    if (mlir::isPure(&*it))
+    // Skip pure operations and operations that do not
+    // touch `dest`.
```
adam-smnk, line 873: Not sure if that's safe in case of aliasing.

chelini: if we have a subview on `dest` we fails.

chelini: Do you have in mind a specific pattern?

adam-smnk: > if we have a subview on dest we fails.

Why would we, it is pure afaik and touches dest.

For example:
```
func.func @zero_test(%arg0: memref<1x32x32xf32>, %arg1: memref<1x32x32xf32>, %arg2: memref<32x32xf32>) {
  %c1_i64 = arith.constant 1 : i64
  %alloc = memref.alloc() {alignment = 64 : i64} : memref<32x32xf32>
  %0 = xsmm.unary.dispatch zero [32, 512, 1, 512] flags = (bcast_scalar) data_type = f32
  xsmm.unary zero(data_type = f32, %0, %cst, %alloc) : (i64, f32, memref<32x32xf32>) -> ()
  %sub = memref.subview %alloc[0, 0] [16, 16] [1, 1] : memref<32x32xf32> to memref<16x16xf32, strided<[32, 1]>>
  call @test(%sub) : (memref<16x16xf32, strided<[32, 1]>>) -> ()
  // Inserts beta_0 while we cannot prove it is zero anymore.
  %1 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 32, 32, 1] flags = (none) data_type = f32
  xsmm.brgemm(data_type = f32, %1, %arg0, %arg1, %alloc, %c1_i64) : (i64, memref<1x32x32xf32>, memref<1x32x32xf32>, memref<32x32xf32>, i64) -> ()
  return
}
func.func private @test(%arg0 : memref<16x16xf32, strided<[32, 1]>>)
```

adam-smnk: This still is fails if `memref.subview` is created before xsmm zero op.

chelini: Not sure if I get it. You can zero a subview, as long as you don't do something with it before the GEMM.

test/Integration/xsmm-strided-brgemm3.mlir
```diff
@@ -24,12 +25,12 @@ func.func @matmul_static(%A : !A_tensor_t, %B : !B_tensor_t, %C : !C_tensor_t) {
...
   %fill = linalg.fill ins(%cst_fill : f32) outs(%empty: tensor<2x2x8x2xf32>) -> tensor<2x2x8x2xf32>
 
-  // IR: %[[C2:.+]] = arith.constant 2 : i64
+  // IR-DAG: %[[C2:.+]] = arith.constant 2 : i64
   // IR-DAG: %[[C1:.+]] = arith.constant 1 : i64
   // IR-DAG: %[[C8:.+]] = arith.constant 8 : i64
   // IR-DAG: %[[C4:.+]] = arith.constant 4 : i64
   // IR-DAG: %[[C0:.+]] = arith.constant 0 : i64
-  // IR: xsmm_gemm_dispatch(%[[C1]], %[[C8]], %[[C2]], %[[C4]], %[[C8]], %[[C2]], %[[C2]], %[[C0]]) 
+  // IR: xsmm_gemm_dispatch(%[[C1]], %[[C8]], %[[C2]], %[[C4]], %[[C8]], %[[C2]], %[[C2]], %[[C4]])
```
rengolin, line 33: I'd also add here a check for NOT zero

test/Passes/DefaultPipeline/linalg-to-xsmm.mlir
```diff
@@ -30,3 +30,34 @@ func.func @fill_op_i32(%arg0: memref<3x3xi32>) {
...
+  return %mul : tensor<3x3xf32>
+}
+
+// CHECK-LABEL: gemm_with_zero
+// CHECK-SAME: %[[ARG0:.+]]: memref<3x3xf32>, %[[ARG1:.+]]: memref<3x3xf32>
+// CHECK-DAG: %[[C0:.+]] = arith.constant 0 : index
+// CHECK-DAG: %[[C1:.+]] = arith.constant 1 : i64
+// CHECK-DAG: %[[C3:.+]] = arith.constant 3 : i64
+// CHECK-DAG: %[[C4:.+]] = arith.constant 4 : i64
+// CHECK: %[[ALLOC:.+]] = memref.alloc() {alignment = 64 : i64} : memref<3x3xf32>
```
rengolin, line 52: // CHECK-NOT: zero

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
adam-smnk:
```
  %alloc = memref.alloc() {alignment = 64 : i64} : memref<32x32xf32>
  %sub = memref.subview %alloc[0, 0] [16, 16] [1, 1] : memref<32x32xf32> to memref<16x16xf32, strided<[32, 1]>>
  %0 = xsmm.unary.dispatch zero [32, 512, 1, 512] flags = (bcast_scalar) data_type = f32
  xsmm.unary zero(data_type = f32, %0, %cst, %alloc) : (i64, f32, memref<32x32xf32>) -> ()
  call @test(%sub) : (memref<16x16xf32, strided<[32, 1]>>) -> ()
  %1 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 32, 32, 1] flags = (none) data_type = f32
```

In this pattern, folding shouldn't kick in as `%alloc` is modified through an alias.

rengolin:
I don't see a big difference in the benchmarks, are you sure this pass is triggered on the benchmarks?

chelini:
> > I don't see a big difference in the benchmarks, are you sure this pass is triggered on the benchmarks?
> 
> Yes, it does. I checked the generated LIBXSMM kernel.

at least in the ` prepacked_targets`.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -808,6 +814,120 @@ void ConvertLinalgToXsmm::runOnOperation() {
     return signalPassFailure();
 }
 
+template <typename OpTy>
+static void updateGemmOpFlags(RewriterBase &rewriter, OpTy gemmDispatchOp) {
+  static_assert(llvm::is_one_of<OpTy, xsmm::GemmDispatchOp,
+                                xsmm::BrgemmDispatchOp>::value);
+  if (!gemmDispatchOp->hasOneUse())
```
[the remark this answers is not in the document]
rengolin: There is one right after LICM. If we have layers of the same size, it does remove the later `dispatch` calls.

Reproduce with:
```
mlir-gen --layers=1024,1024,1024,1024 | tpp-opt --default-tpp-passes
```

and see the third `scf.parallel` reusing the dispatch from the previous layer (`%5` and `%6`).

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -808,6 +814,120 @@ void ConvertLinalgToXsmm::runOnOperation() {
...
+                          << rootOp << "\n");
+  // 1. Prove the alloc to be zero. Linear walk of the bb
+  // and make sure there are only pure operations between
+  // the zero op and the gemm.
+  Value dest = rootOp.getInputs().back();
+  DenseSet<Operation *> destUsers(dest.getUsers().begin(),
+                                  dest.getUsers().end());
+
+  Block *blck = nullptr;
+  if (auto bbArg = dyn_cast<BlockArgument>(dest)) {
```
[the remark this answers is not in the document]
rengolin: So, the function is called `fuseZeroWithGemmOrBrgemm`, and the walk is about making sure they're side-effect free zeroes, not fusing. The actual logic is below (line 913). There's a lot to unpack in the walk, what each match means, why some continue others return, etc.

I'd try to find common patterns, create list holders for the matches and return if none were found, etc.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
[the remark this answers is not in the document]
rengolin: > If we allow CSE we need to start cloning the operations.

No, just run before CSE or LICM would do. Then different `dispatch`es won't get commoned up.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -808,6 +814,129 @@ void ConvertLinalgToXsmm::runOnOperation() {
...
+void FoldXsmmFlags::runOnOperation() {
+  SmallVector<xsmm::UnaryOp> producers;
+  getOperation()->walk([&](xsmm::UnaryOp unaryOp) {
+    auto kind = unaryOp.getCallee();
+    if (kind == xsmm::UnaryKind::ZERO)
+      producers.push_back(unaryOp);
+  });
+
+  IRRewriter rewriter(&getContext());
+  for (xsmm::UnaryOp producer : producers)
```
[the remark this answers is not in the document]
chelini: Done.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -808,6 +814,120 @@ void ConvertLinalgToXsmm::runOnOperation() {
...
+  }
+  if (it == itEnd)
+    return;
+
+  LLVM_DEBUG(llvm::dbgs() << "[fuseZeroWithGemmOrBrgemm] Candidate op OK: "
+                          << rootOp << "\n");
+
+  // 2. Update flags.
+  assert(isa<xsmm::GemmOp>(*it) || isa<xsmm::BrgemmOp>(*it));
+  if (auto gemmOp = dyn_cast<xsmm::GemmOp>(*it)) {
```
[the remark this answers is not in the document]
chelini: Added case for fusedBrgemm.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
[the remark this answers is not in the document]
chelini: I tried to pull it out. Let me know what you think.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
chelini: thanks, fixed.

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -81,6 +81,10 @@ def Xsmm_GemmOp : Xsmm_Op<"gemm"> {
     `(` `data_type` `=` $data_type `,` $inputs `)`
     attr-dict `:` functional-type($inputs, results)
   }];
+
+  let extraClassDeclaration = [{
+    Value getOutput() { return getInputs().back(); }
```
rengolin, line 86: Nit: to be in sync with the others, maybe use `getInputs()[3]` too?

rengolin (approved):
Small nit, but LGTM, thanks!

chelini:
@adam-smnk any more comments?

adam-smnk (approved):
Looks good, thanks 👍

Issue #773, opened 2023-11-07
Detect cmpf and select as Relu


nhasabni:
Below is a typical pattern coming from PyTorch-dynamo based MLP model. Currently, `cmpf` and `select` are not detected as `Relu` op.

```
    %4 = linalg.matmul ins(%arg4, %1 : tensor<256x1024xf32>, tensor<1024x1024xf32>) outs(%3 : tensor<256x1024xf32>) -> tensor<256x1024xf32>
   
    %5 = linalg.generic {indexing_maps = [#map2, #map, #map], iterator_types = ["parallel", "parallel"]} ins(%arg1, %4 : tensor<1024xf32>, tensor<256x1024xf32>) outs(%2 : tensor<256x1024xf32>) {
    ^bb0(%in: f32, %in_1: f32, %out: f32):
      %22 = arith.addf %in, %in_1 : f32
      linalg.yield %22 : f32
    } -> tensor<256x1024xf32>
    
    %6 = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} ins(%5 : tensor<256x1024xf32>) outs(%2 : tensor<256x1024xf32>) {
    ^bb0(%in: f32, %out: f32):
      %22 = arith.cmpf ugt, %in, %cst : f32
      %23 = arith.select %22, %in, %cst : f32
      linalg.yield %23 : f32
    } -> tensor<256x1024xf32>
```

PR #778, opened 2023-11-08
Adding support for detecting cmp + select as Relu

nhasabni:
Adding unit tests as well and they pass.

Addresses #773.

### DIFF
--- a/lib/TPP/IR/MatcherUtils.cpp
+++ b/lib/TPP/IR/MatcherUtils.cpp
@@ -146,21 +146,21 @@ static bool hasReluBody(Operation *op, SmallVectorImpl<Value> *captured) {
   };
 
   // Multiple patterns map to Relu.
-  if (isa<arith::MaximumFOp>(innerOp)) {
+  if (auto maxfOp = dyn_cast<arith::MaximumFOp>(innerOp)) {
     // Pattern 1 - arith.maximumf(tensor, const 0)
     if (yieldOp->getOperand(0).getDefiningOp() != innerOp)
       return false;
-    auto maxfOp = cast<arith::MaximumFOp>(innerOp);
     Value maxfLhs = maxfOp.getLhs();
     Value maxfRhs = maxfOp.getRhs();
 
     return (getOperand(maxfLhs, maxfRhs) || getOperand(maxfRhs, maxfLhs));
-  } else if (isa<arith::CmpFOp>(innerOp)) {
+  } else if (auto cmpfOp = dyn_cast<arith::CmpFOp>(innerOp)) {
     // Pattern 2 - arith.cmpf, arith.select
-    // %22 = arith.cmpf ugt, %in, %cst : f32
-    // %23 = arith.select %22, %in, %cst : f32
-    // linalg.yield %23 : f32
-
+    //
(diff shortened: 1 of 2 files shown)


rengolin (commented):
Great set of tests! You couldn't find a way to use the  op chain method?

lib/TPP/IR/MatcherUtils.cpp
```diff
@@ -151,7 +144,82 @@ static bool hasReluBody(Operation *op, SmallVectorImpl<Value> *captured) {
     }
     return false;
   };
-  return (getOperand(maxfLhs, maxfRhs) || getOperand(maxfRhs, maxfLhs));
+
+  // Multiple patterns map to Relu.
+  if (isa<arith::MaximumFOp>(innerOp)) {
```
chelini, line 149: I would change as: `if (auto maxfOp = dyn_cast<arith::MaximumFOp>(innerOp))`. This allows you yo avoid the cast in line 153.

lib/TPP/IR/MatcherUtils.cpp
```diff
@@ -151,7 +144,82 @@ static bool hasReluBody(Operation *op, SmallVectorImpl<Value> *captured) {
...
+    // Pattern 1 - arith.maximumf(tensor, const 0)
+    if (yieldOp->getOperand(0).getDefiningOp() != innerOp)
+      return false;
+    auto maxfOp = cast<arith::MaximumFOp>(innerOp);
+    Value maxfLhs = maxfOp.getLhs();
+    Value maxfRhs = maxfOp.getRhs();
+
+    return (getOperand(maxfLhs, maxfRhs) || getOperand(maxfRhs, maxfLhs));
+  } else if (isa<arith::CmpFOp>(innerOp)) {
```
chelini, line 158: same here.

lib/TPP/IR/MatcherUtils.cpp
```diff
@@ -151,7 +144,82 @@ static bool hasReluBody(Operation *op, SmallVectorImpl<Value> *captured) {
...
+    // Pattern 2 - arith.cmpf, arith.select
+    // %22 = arith.cmpf ugt, %in, %cst : f32
```
chelini, line 160: I would drop this IR.

chelini (commented):
thanks, only minor comments for now.

lib/TPP/IR/MatcherUtils.cpp
nhasabni: I was keeping it as a doc for easier code understanding.

nhasabni: Reworded it to make it more abstract.

nhasabni:
Thanks @chelini @rengolin  for comments. Please take a look after the change now.

rengolin:
> Great set of tests! You couldn't find a way to use the op chain method?

Discussed offline with @nhasabni, we'll defer the move to structured matchers because the patterns here are not that simple and we may need some new support in the matchers themselves. For now, we'll carry on with the bloated function and refactor later.

rengolin (approved):
Sorry it took so long, it fell out of my radar. LGTM, thanks!

PR #781, opened 2023-11-13
Add linalg->xsmm conversion for mul operation

chelini:


lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -929,11 +946,12 @@ struct ConvertGenericToVnniBrgemm : public OpRewritePattern<linalg::GenericOp> {
...
 void mlir::tpp::populateLinalgToXsmmPatterns(RewritePatternSet &patterns) {
-  patterns.add<ConvertFillOpToUnaryZero, ConvertTransposeOpToUnaryTranspose,
-               ConvertGenericToUnaryRelu, ConvertGenericToBinaryAdd,
-               ConvertGenericToBinarySub, ConvertGenericToBrgemm,
-               ConvertBatchReduceMatmulToBatchReduceMatmul,
-               ConvertMatmulToMatmul, ConvertGenericToUnaryIdentity,
-               ConvertVnniPacking, ConvertGenericToVnniBrgemm>(
-      patterns.getContext());
+  patterns
+      .add<ConvertFillOpToUnaryZero, ConvertTransposeOpToUnaryTranspose,
```
rengolin, line 950: This is getting bloated. Can we have a single rewrite with all matchers inside? At least one for unary, one for binary and one for gemms.

lib/TPP/IR/MatcherUtils.cpp
```diff
@@ -113,6 +113,10 @@ bool isTwoDSubOp(linalg::LinalgOp linalgOp, SmallVectorImpl<Value> *operands) {
   return isTwoDEltWiseOpOfTypeTy<arith::SubFOp>(linalgOp, operands);
 }
 
+bool isTwoDMulOp(linalg::LinalgOp linalgOp, SmallVectorImpl<Value> *operands) {
+  return isTwoDEltWiseOpOfTypeTy<arith::MulFOp>(linalgOp, operands);
```
rengolin, line 117: Honestly, these helpers don't really help anything. Can we just move the inner call to the outer users?

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
```diff
@@ -377,3 +377,119 @@ func.func @sub_bcast_row_1(%arg0: memref<256x1024xf32>, %arg1: memref<256xf32>)
...
+      linalg.yield %6 : f32
+  }
+  return
+}
+
+// CHECK-LABEL: mul_bcast_row_1
+// CHECK-SAME: %[[ARG0:.+]]: memref<256x1024xf32>, %[[ARG1:.+]]: memref<256xf32>
+// CHECK: linalg.generic
+// CHECK-NOT: xsmm.binary sub
+// We need to lift this restriction:
```
rengolin, line 494: This looks like a bad broadcast to me, not something we want to cover. Both `<1024>` and `<256x1>` are covered above.

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
```diff
@@ -377,3 +377,119 @@ func.func @sub_bcast_row_1(%arg0: memref<256x1024xf32>, %arg1: memref<256xf32>)
...
+      ^bb0(%in: f32, %in_4: f32, %out: f32):
+        %19 = arith.mulf %in, %in_4 : f32
+        linalg.yield %19 : f32
+  }
+  return 
+}
+
+// CHECK-LABEL: mul_bcast_row
+// CHECK-SAME: %[[ARG0:.+]]: memref<256x1024xf32>, %[[ARG1:.+]]: memref<256x1xf32>, %[[ARG2:.+]]: memref<256x1024xf32>
+// CHECK: %[[DIS:.+]] = xsmm.binary.dispatch mul [256, 1024, 1024, 1, 1024] flags = (bcast_row_in1) data_type = f32
```
rengolin, line 469: Do we support broadcast on the first argument? If so, would be nice to have tests. If not, would be nice to have negative tests.

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
chelini: Why? It is perfectly defined by the affine map. If I am not mistaken we can end up in this situation if we fold unit dims in linalg.

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
chelini: Yes, added test for sub/mul/add.

test/Conversion/LinalgToXsmm/linalg-to-binary.mlir
rengolin: Ah, good point! Ignore me.

lib/TPP/IR/MatcherUtils.cpp
rengolin: Fair enough, not important for this PR.

rengolin (approved):

Issue #783, opened 2023-11-13
Remaining Issues for MLP performance on par with libxsmm-dnn


rengolin:
These are the known issues to reach libxsmm-dnn performance on "pre-packed layer" MLPs:
- [ ] Single allocation on page boundary (see #777)
- [ ] Beta=Zero (see #777)
- [ ] XSMM fusion (see #752)
- [ ] Pass loop order to libxsmm brgemm calls

In theory, if we get all of those in, it should reach parity. If more is discovered, please add to the list. Let's only close this issue when we reach parity on the base MLP benchmarks we have for pre-packed MLPs.

@chelini @alheinecke

rengolin:
Beta=0 is done, benchmark IR is affected, but we got <1% performance change from that, probably within noise. We didn't expect a huge change, so not a big deal.

PR #784, opened 2023-11-16
Duplicate fill on contractions

chelini:
Duplicate fill operations when the use is a contraction and we can fold the fill in the contraction later on in the pipeline using: `fold-xsmm-flags`.  Duplication avoids introducing `memref.copies` by bufferization. Example,

```mlir
%0 = tensor.empty()
%1 = linalg.fill ins(...) outs(%0) // fill with zeros.
%2 = linalg.matmul ins(...) outs(%1)
%3 = linalg.matmul ins(...) outs(%1)
```
Without this PR it bufferizes as:

```mlir
%0 = memref.alloc()
%1 = memref.alloc()
linalg.fill ins(...) outs(%0) // fill with zeros.
memref.copy %0 into %1
linalg.matmul ins(...) outs(%0)
linalg.matmul ins(...) outs(%1)
```

With this PR the IR looks like:

```mlir
// no copies and fills folded as beta = 0.
%0 = memref.alloc()
%1 = memref.alloc()
xsmm.matmul ins(...) outs(%0) // beta = 0
xsmm.matmul ins(...) outs(%1) // beta = 0
```

The PR has minor performance impact, the only notable improvement is for `fp32_mha_tensorflow_seq_len_32`. The IR looks cleaner too with 1 less allocation and all the beta flags properly folded. `fp32_mha_tensorflow_seq_len_1024` does not improve because dimensionality allows fusion to distribute the fill, see: https://github.com/plaidml/tpp-mlir/commit/b1167fea3bfcfbb89a46b8f4fa01169020de4d92.

This PR is part of #783

### DIFF
--- a/lib/TPP/Transforms/Bufferize.cpp
+++ b/lib/TPP/Transforms/Bufferize.cpp
@@ -9,6 +9,9 @@
 #include "TPP/Passes.h"
 #include "TPP/Transforms/Transforms.h"
 
+#include "TPP/Transforms/Utils/TransformUtils.h"
+#include "TPP/Transforms/Utils/ValueUtils.h"
+#include "mlir/Analysis/SliceAnalysis.h"
 #include "mlir/Conversion/BufferizationToMemRef/BufferizationToMemRef.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
@@ -42,6 +45,8 @@ namespace mlir {
 namespace tpp {
 #define GEN_PASS_DEF_BUFFERIZE
 #include "TPP/Passes.h.inc"
+#define GEN_PASS_DEF_DUPLICATEFILL
+#include "TPP/Passes.h.inc"
 } // namespace tpp
 } // namespace mlir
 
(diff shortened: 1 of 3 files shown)


lib/TPP/Transforms/Bufferize.cpp
```diff
@@ -72,12 +77,51 @@ struct Bufferize : public tpp::impl::BufferizeBase<Bufferize> {
...
+      return WalkResult::advance();
+    SetVector<Operation *> forwardSlice;
+    getForwardSlice(fillVal, &forwardSlice);
+    for (size_t idx = /*Skip first user. Use the current fill*/ 1;
+         idx < forwardSlice.size(); idx++)
+      if (auto linalgOp = dyn_cast<linalg::LinalgOp>(forwardSlice[idx]))
+        if (succeeded(linalgx::utils::isContraction(linalgOp))) {
+          assert(linalgOp.getNumDpsInits() == 1);
+          Value outLinalg = linalgOp.getDpsInits()[0];
+          if (outLinalg == fillVal) {
```
rengolin, line 103: Control nesting by advancing on negative matches.

lib/TPP/Transforms/Bufferize.cpp
```diff
@@ -72,12 +77,51 @@ struct Bufferize : public tpp::impl::BufferizeBase<Bufferize> {
...
+            rewriter.setInsertionPoint(linalgOp);
+            Operation *clonedOp = rewriter.clone(*fillOp.getOperation());
+            rewriter.replaceUsesWithIf(fillOp->getResults(),
```
rengolin, line 106: Don't you also have to check if each user is single user?

rengolin (commented):
I can't remember what were the examples we discussed about external dependencies to following ops, but we need to make sure that the copy isn't being used by other non contraction/matmul operations, so probably need an artificial test to make sure we check for number of users of each following fill/contraction.

lib/TPP/Transforms/Bufferize.cpp
chelini: User of the cloned op? The idea is to clone the operation and replace the uses of the fillOp with the clonedOp only for the current linalg contraction operation.

rengolin:
Something like:
```mlir
%buf = tensor.empty()
%fill = linalg.fill ins(%buf)

%0 = linalg.matmul inst(...) outs(%fill)
%1 = linalg.matmul inst(...) outs(%fill)
%add = linalg.add ins(...) outs(%fill)
%2 = linalg.matmul inst(...) outs(%fill)
%sub = linalg.sub ins(...) outs(%fill) 
%3 = linalg.matmul inst(...) outs(%fill)
```
My question is: without this pass, does that code bufferizes to `alloc+copy` for **every** user? Or does it try to look at the lifetime of `%add` and `%sub` or something else?

If the answer is YES, then adding new `empty+fill` patterns is equal or better than the original, because it will "zero" all matmuls (for xsmm) but not the others (add, sub) which they wouldn't anyway. More importantly, if the element-wise ops operate on the matmul result (`%1` chains into `%add` and there are no more users of `%1`), can we not add a `empty+fill` patter to the tail element-wise?

If the answer is NO, then what care must we take to not break bufferization "smartness" when it comes to non-matmul ops (ex. in-place reuse due to liveness analysis)? This is mostly important when the operations do not chain together (ie. they do not kill the previous SSA values).

lib/TPP/Transforms/Bufferize.cpp
rengolin: Something like:
```mlir
%buf = tensor.empty()
%fill = linalg.fill ins(%buf)

%0 = linalg.matmul inst(...) outs(%fill)
%1 = linalg.matmul inst(...) outs(%fill)
%add = linalg.add ins(...) outs(%fill)
%2 = linalg.matmul inst(...) outs(%fill)
%sub = linalg.sub ins(...) outs(%fill) 
%3 = linalg.matmul inst(...) outs(%fill)
```
My question is: without this pass, does that code bufferizes to `alloc+copy` for **every** user? Or does it try to look at the lifetime of `%add` and `%sub` or something else?

If the answer is YES, then adding new `empty+fill` patterns is equal or better than the original, because it will "zero" all matmuls (for xsmm) but not the others (add, sub) which they wouldn't anyway. More importantly, if the element-wise ops operate on the matmul result (`%1` chains into `%add` and there are no more users of `%1`), can we not add a `empty+fill` patter to the tail element-wise?

If the answer is NO, then what care must we take to not break bufferization "smartness" when it comes to non-matmul ops (ex. in-place reuse due to liveness analysis)? This is mostly important when the operations do not chain together (ie. they do not kill the previous SSA values).

test/Passes/pass-duplicate-fill.mlir
```diff
@@ -0,0 +1,203 @@
...
+
+// -----
+
+// CHECK-LABEL: matmuls_1
+func.func @matmuls_1(%arg0: tensor<32x32xf32>, %arg1: tensor<32x32xf32>) -> tensor<32x32xf32> {
+  // BUFF-COUNT-3: memref.alloc
+  // BUFF-COUNT-1: memref.copy
+  // BUFFNOTDUP-COUNT-3: memref.alloc
+  // BUFFNOTDUP-COUNT-1: memref.copy
+  // CHECK-COUNT-2: linalg.fill
```
rengolin, line 192: There should be a copy, too, right? For the `add`?

chelini: yes, there is 1 memref.copy

rengolin: then just add a `CHECK-COUNT-1: memref.copy` to make it clear and it LGTM. Thanks!

rengolin (approved):

test/Passes/pass-duplicate-fill.mlir
chelini: the CHECK line does not bufferize, it only applies the transformation at the tensor level. Copy and allocations are checked with BUFFXX

PR #785, opened 2023-11-16
Fail compilation if `OpTy` is not a dispatch in `verifyInputs` (NFC)

chelini:


lib/TPP/Dialect/Xsmm/XsmmOps.cpp
```diff
@@ -277,6 +277,11 @@ static LogicalResult verifyGemmFlags(ArrayAttr flags, DataType dataType,
 
 template <typename OpTy>
 static LogicalResult verifyInputs(OpTy op, size_t expected) {
+  static_assert(llvm::is_one_of<OpTy, xsmm::UnaryDispatchOp,
```
rengolin, line 280: nit: name it accordingly, too? `verifyDispatchInputs`?

rengolin (approved):

PR #786, opened 2023-11-17
Adding PyTorch-Dynamo models as benchmarks

### LINKED ISSUE #776: PyTorch GEMM/MLP into benchmarks
Now that we can generate models with PyTorch into MLIR linalg and can consume them in our compiler, we need to add them to the benchmark list. Once we fix #773, we'll hopefully get the same performance as the ones from `mlir-gen`.

To avoid compatibility issues and creating too many scripts, we should just have two models in https://github.com/plaidml/mlir-generator to create similar benchmarks as we have ({GEMM only / MLP }, 3-layers, 256 batch, 1024 layer size, no softmax, {FP32 / BF16}).

The 4 new models should be added to the "gemm_models" and "mlp_models" (as in #774) two each (32, 16).

This should conclude the task that connects PyTorch (Dynamo) to our compiler. Of course, not all models work, and the connection is flaky (manual steps), but automation of that particular task and adding new models and making them work will be handled elsewhere.

nhasabni:
Added {FP32, BF16}, {MLP, GEMM} MLIR from PyTorch-Dynamo with 256 batch size and 3 layers of 1024.

Also added benchmarks-pt as a target for ninja.

Fixes #776


chelini:
Thank you @nhasabni, really nice to see this.

rengolin:
You also need to add it to the CI scripts (`scripts/buildkite/benchmarks.sh`). Add them to a new category after the `base` category.

rengolin:
The 4 new MLIR files are 21MB each. It's not massive, but 99% of it is pre-trained weights. For the sake of benchmarking, we can convert that to a `dense<1.0>` which takes a lot less space and the benchmark harness converts those weights back to a normal distribution before timing it.

Just make sure to convert different weights to incremental values of `dense` so that they don't get commoned up in the canonicalizer. Ex:
```mlir
  %w0 = arith.constant dense<1.0> : ...
  %b0 = arith.constant dense<1.1> : ...
  %w1 = arith.constant dense<1.2> : ...
  %b1 = arith.constant dense<1.3> : ...
  %w2 = arith.constant dense<1.4> : ...
  %b3 = arith.constant dense<1.5> : ...
  ...
```

nhasabni:
> The 4 new MLIR files are 21MB each. It's not massive, but 99% of it is pre-trained weights. For the sake of benchmarking, we can convert that to a `dense<1.0>` which takes a lot less space and the benchmark harness converts those weights back to a normal distribution before timing it.
> 
> Just make sure to convert different weights to incremental values of `dense` so that they don't get commoned up in the canonicalizer. Ex:
> 
> ```mlir
>   %w0 = arith.constant dense<1.0> : ...
>   %b0 = arith.constant dense<1.1> : ...
>   %w1 = arith.constant dense<1.2> : ...
>   %b1 = arith.constant dense<1.3> : ...
>   %w2 = arith.constant dense<1.4> : ...
>   %b3 = arith.constant dense<1.5> : ...
>   ...
> ```

Right. I was not aware that the benchmark harness would replace dense consts before benchmarking. Good to know. Will make the change.

alheinecke:
we should also make sure, that in the final version such things:
`%3 = linalg.fill ins(%cst_2 : bf16) outs(%2 : tensor<256x1024xbf16>) -> tensor<256x1024xbf16>`
are removed as we don't want to benchmark it. IMHO, when following @rengolin  suggestion that should be taken of (I think, just pointing it out here to double check).

nhasabni:
> we should also make sure, that in the final version such things: `%3 = linalg.fill ins(%cst_2 : bf16) outs(%2 : tensor<256x1024xbf16>) -> tensor<256x1024xbf16>` are removed as we don't want to benchmark it. IMHO, when following @rengolin suggestion that should be taken of (I think, just pointing it out here to double check).

These patterns are coming from converting PyTorch models via torch-mlir. Do we want to modify MLIR generated via this process?

rengolin:
> These patterns are coming from converting PyTorch models via torch-mlir. Do we want to modify MLIR generated via this process?

Not by hand, no. We need a "faithful" representation of what PyTorch generates, to make sure our compiler can handle everything. But we may need a pass (or use an upstream pass) to get rid of bad patterns.

scripts/buildkite/benchmark.sh
```diff
@@ -10,9 +10,10 @@ SCRIPT_DIR=$(realpath "$(dirname "$0")/..")
 source "${SCRIPT_DIR}/ci/common.sh"
 
 die_syntax() {
-  echo "Syntax: $0 [-b] [-o] [-m] [-f]"
+  echo "Syntax: $0 [-b] [-p] [-o] [-m] [-f]"
```
rengolin, line 13: Ah, one final missing thing: add `-p` to the option in `tpp-benchmark.yml`.

rengolin:
> we should also make sure, that in the final version such things: `%3 = linalg.fill ins(%cst_2 : bf16) outs(%2 : tensor<256x1024xbf16>) -> tensor<256x1024xbf16>` are removed as we don't want to benchmark it. IMHO, when following @rengolin suggestion that should be taken of (I think, just pointing it out here to double check).

`%cst_2` here is `0.00000` and that's an `xsmm.zero` before the `xsmm.brgemm`, which should be folded into "beta=0" brgemm by #777 and propagated to all layers by #784.

nhasabni:
> > These patterns are coming from converting PyTorch models via torch-mlir. Do we want to modify MLIR generated via this process?
> 
> Not by hand, no. We need a "faithful" representation of what PyTorch generates, to make sure our compiler can handle everything. But we may need a pass (or use an upstream pass) to get rid of bad patterns.

So to confirm and conclude, @rengolin I do not need to take any action here, right?

rengolin:
> So to confirm and conclude, @rengolin I do not need to take any action here, right?

Correct! LGTM, thanks!

GEMM and MLP are is 96% of our hand-crafted versions, which is a good start. We can investigate how to get the remaining 4% in a separate PR.

rengolin (approved):

PR #787, opened 2023-11-17
Change `perf` dialect to use `iter_args` correctly and fix dependency problem

### LINKED ISSUE #277: Missing value dependency between `perf.bench` `args` and `perf.yield`
`perf.bench` op does not capture correctly the relation between the op's `args` and their reuse within the region.
The `args` should model the same data relation and reuse as `scf.for` op's `iter_args`.

However, currently this link is missing and it allows folders to freely remove operations that use `args` inputs.
As an example, the following test fails:
```
// RUN: tpp-opt %s -canonicalize -convert-perf-to-loops | FileCheck %s

// CHECK: @bench_canonicalize({{.*}}, %[[k:.*]]: i64)
func.func @bench_canonicalize(%n: i64, %k: i64) -> (i64) {
  // CHECK-DAG: %[[output:.*]] = arith.constant 0 : i64
  // CHECK: %[[deltas:.*]] = memref.alloc
  %size = arith.index_cast %n : i64 to index
  %deltas = memref.alloc(%size) : memref<?xf64>
  %output = arith.constant 0 : i64

  // CHECK: %[[res:.*]] = scf.for %[[i:.*]] = {{.*}}iter_args(%[[iarg0:.*]] = %[[output]]) -> (i64) {
  // CHECK:   %[[timer:.*]] = perf.start_timer
  // CHECK:   %[[sum:.*]] = arith.addi %[[iarg0]], %[[k]]
  // CHECK:   %[[delta:.*]] = perf.stop_timer(%[[timer]] {{.*}})
  // CHECK:   memref.store %[[delta]], %[[deltas]][%[[i]]]
  // CHECK:   scf.yield %[[sum]]
  // CHECK: }
  %res = perf.bench (%n, %deltas : memref<?xf64>) args(%output : i64) {
    %sum = arith.addi %output, %k : i64
    perf.yield %sum : i64
  } -> i64

  memref.dealloc %deltas : memref<?xf64>

  // CHECK: return %[[res]]
  return %res : i64
}
```

Just by running `tpp-opt -canonicalize`, the `arith.addi` get folded away as it fits `a + 0` pattern.
The missing dependency between the initial `%output` and the fact that its value gets updated through `perf.yield` allows this folding to happen.

The goal of this issue is to somehow prevent this erroneous folding from happening.

rengolin:
This makes correct usage of `iter_args` by creating actual loop bound values and using them inside the loop body doesn't get removed away during canonicalization.

I've changed the existing tests to stress this feature to make sure we don't regress.

Fixes #277

### DIFF
--- a/lib/TPP/Dialect/Perf/PerfOps.cpp
+++ b/lib/TPP/Dialect/Perf/PerfOps.cpp
@@ -170,29 +170,28 @@ ParseResult BenchOp::parse(OpAsmParser &parser, OperationState &result) {
   locs.clear();
 
   // Parse iter_args, if any
-  if (succeeded(parser.parseOptionalKeyword("iter_args"))) {
-    if (parser.parseLParen())
+  SmallVector<OpAsmParser::Argument, 4> regionArgs;
+  bool hasIterArgs = succeeded(parser.parseOptionalKeyword("iter_args"));
+  if (hasIterArgs) {
+    if (parser.parseAssignmentList(regionArgs, operands) ||
+        parser.parseArrowTypeList(result.types))
       return failure();
+  }
 
-    locs.push_back(parser.getCurrentLocation());
-    if (parser.parseOperandList(operands))
-      return failure();
-    if (parser.parseColon())
-      return failure();
+  if (regionArgs.size() != result.types.size() + 1)
+    return parser.emitError(parser.getNameLoc(),
+        "mismatch in number of loop-carried values and defined values");
 
-    // Avoid using parseTypeList because it crashes on GCC
-    if (parser.parseCommaSeparatedList(parseType))
-      return failure();
-    if (parser.parseRParen())
-      return failure();
-
-    // Validate operands
-    if (!operands.size())
-      return parser.emitError(parser.getCurrentLocation(), "expect arguments");
-    if (operands.size() != types.size())
-      return parser.emitError(locs[0], "expect same number of types as args");
-    if (parser.resolveOperands(operands, types, locs[0], result.operands))
-      return failure();
+    // Resolve input operands.
+  if (hasIterArgs) {
+    for (auto argOperandType :
+         llvm::zip(llvm::drop_begin(regionArgs), operands, result.types)) {
+      Type type = std::get<2>(argOperandType);
+      std::get<0>(argOperandType).type = type;
+      if (parser.resolveOperand(std::get<1>(argOperandType), type,
+                                result.operands))
+        return failure();
+    }
   }
 
   // Parse region
--- a/lib/TPP/Dialect/Perf/PerfOps.cpp
+++ b/lib/TPP/Dialect/Perf/PerfOps.cpp
@@ -74,6 +74,7 @@ YieldOp BenchOp::getYieldOp() {
 }
 
 void BenchOp::print(OpAsmPrinter &printer) {
+  // Print base args
   printer << "(";
   printer << getNumIters();
   printer << ", ";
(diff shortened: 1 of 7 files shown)


adam-smnk (approved):
Looks clean - thanks 👍

lib/TPP/Dialect/Perf/PerfOps.cpp
```diff
@@ -89,16 +90,22 @@ void BenchOp::print(OpAsmPrinter &printer) {
...
-    printer << getIterArgs();
-    printer << " : ";
-    printer << getIterArgs().getTypes();
-    printer << ")";
+    auto blockArgs = getRegion().getArguments();
+    auto initArgs = ValueRange{getIterArgs()};
+    assert(blockArgs.size() == initArgs.size() &&
+           "expected same length of arguments and initializers");
+    printer << " iter_args(";
+    llvm::interleaveComma(llvm::zip(blockArgs, initArgs), printer, [&](auto it) {
```
chelini, line 101: llvm::zip_equal?

lib/TPP/Dialect/Perf/PerfOps.cpp
```diff
@@ -170,48 +173,40 @@ ParseResult BenchOp::parse(OpAsmParser &parser, OperationState &result) {
...
+    return parser.emitError(
+        parser.getNameLoc(),
+        "mismatch in number of loop-carried values and defined values");
+
+  // Resolve input operands.
+  if (hasIterArgs) {
+    for (auto argOperandType : llvm::zip(regionArgs, operands, result.types)) {
+      Type type = std::get<2>(argOperandType);
+      std::get<0>(argOperandType).type = type;
+      if (parser.resolveOperand(std::get<1>(argOperandType), type,
```
chelini, line 194: I would add brace here.

test/Dialect/Perf/perf-invalid.mlir
```diff
@@ -5,7 +5,7 @@ func.func @perf_no_outs(%n: i64) {
   %deltas = memref.alloc(%size) : memref<?xf64>
   %out = arith.constant 0 : i64
 
-  // expected-error @below {{'perf.bench' op failed to verify that result types match types of iter_args}}
+  // expected-error @below {{cannot name an operation with no results}}
```
chelini, line 8: Why are we testing this? This is MLIR stuff, no?

lib/TPP/Dialect/Perf/PerfOps.cpp
```diff
@@ -170,48 +173,40 @@ ParseResult BenchOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  }
 
-    // Validate operands
-    if (!operands.size())
-      return parser.emitError(parser.getCurrentLocation(), "expect arguments");
-    if (operands.size() != types.size())
-      return parser.emitError(locs[0], "expect same number of types as args");
-    if (parser.resolveOperands(operands, types, locs[0], result.operands))
-      return failure();
+  if (regionArgs.size() != result.types.size())
```
chelini, line 184: brace here too.

test/Dialect/Perf/perf-invalid.mlir
rengolin: Not entirely. The error comes from the basic verifier, yes, but it means we're not generating the `perf.bench` correctly. Before, this was an error in the `perf` verifier, but that error was actually misplaced. This is the right place to put it.

I can remove the test, but I think it works well to encode expectation, even if the error is somewhere else. Doesn't cost us much either.

PR #794, opened 2023-11-20
Clean-up definitions in `XsmmOps.td` (NFC)

chelini:


adam-smnk (approved):
Looks like 2D and 4D defs are unused anyway.
Should be fine to remove.

PR #800, opened 2023-11-22
[NFC] No need for disabling lsan anymore on benchmarks

### LINKED ISSUE #379: MLIRBench - improve kernel result deallocation
Current MLIRBench kernel wrapped generation examines entry function signature and if it explicitly returns a memref, a deallocation op is generate to prevent memory leaks.

This approach does not always work when the input kernel is at the tensor level. Often returned tensor is automatically converted into a destination-passing memref (last argument is used instead of return) and then no additional memory cleanup is necessary. However, this bufferization behavior is not guaranteed and sometimes the kernel can still return a memref buffer. With current approach, this leads to memory leaks.

This fix this issue, MLIRBench should examine kernel IR after lowering (at least after the default pipeline) to determine whether result buffer deallocation is necessary.

rengolin:
With bufferization under control, more stable kernel generation and ownership based deallocation passes, we don't have memory leaks any more on benchmarks, as shown by the total lack of usage of the -disable-lsan flag.

Fixes #379

### DIFF
--- a/benchmarks/config/base/base.json
+++ b/benchmarks/config/base/base.json
@@ -26,14 +26,14 @@
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=model --float-width=16 --batch=256 --layers=1024,1024,1024,1024 --tiles=64,64,64 --vnni=2" ],
       "environment": {},
-      "flags": [ "-n", "100", "--disable-lsan" ],
+      "flags": [ "-n", "100" ],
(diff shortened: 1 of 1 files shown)


adam-smnk (approved):
Should be fine to remove.
Good to go if CI passes.

rengolin:
CI builds in release mode, without sanitizers, but I also ran this locally on a debug build with sanitizers and did not see any leaks.

PR #802, opened 2023-11-22
[Debug] Add suport for arbitrary MLIR file

rengolin:
This helps debug any MLIR file, not only the ones generated by mlir-gen, for example the PyTorch and other files we have in the benchmark suite.


adam-smnk (approved):
Works nicely 👍

nhasabni (approved):
LGTM

PR #804, opened 2023-11-23
[XSMM][Bench] Bump XSMM and change benchmarks to match

rengolin:
This is related to libxsmm-dnn PR:
https://github.com/libxsmm/libxsmm-dnn/pull/27

List of changes:
 * Changing fusion type to 5 (from 3) on MLP XSMM-DNN kernels
 * Adding a buildkite option for increasing the number of iterations
 * Increasing the timeout to 1h to accomodate longer (weekend) runs
 * Fix OMP benchmark names, move to (2, 4, 8, 16) threads
 * Change OMP tile sizes to 32x32 instead of 64x64


benchmarks/config/base/base.json
```diff
@@ -40,7 +40,7 @@
       "type": "XSMM-DNN",
       "benchmark": "xsmm_dnn_mlp",
       "environment": {},
-      "flags": [ "100", "256", "3", "F", "64", "64", "64", "0", "1024", "1024", "1024", "1024" ],
+      "flags": [ "100", "256", "5", "F", "64", "64", "64", "0", "1024", "1024", "1024", "1024" ],
```
adam-smnk, line 43: Why not use 32x32 to align with the rest?

rengolin: Yeah, it would also make more sense when comparing against the "model" ones, since the compiler still doesn't have a way to propagate tile sizes from `tpp-opt` yet.

alheinecke (approved):
LGTM

PR #807, opened 2023-11-24
[Perf] Move timer out of the loop

rengolin:
This is a big change to the perf dialect: we're moving the timer start/stop to outside of the loop and just dividing it by the number of iterations, instead of calculating mean and stdev for each timer start/stop inside the loop.

Benchmarks on the cluster have shown identical results to all previous machines we tested on, so there's no actual different on the run time.

But the code is significantly better:
 * No need to implement mean/stdev logic
 * No need to allocate results tensor, insert/extract
 * Higher precision with low-precision timers
 * Significantly simpler bench.perf logic and lowering

The main changes are:
 * Perf dialect to remove unused ops and simplify bench
 * tpp-run to produce the right warmup, bench loop
 * benchmark driver to not print stdev on results

Some related changes:
 * Reinforced tpp-run to not print void functions
 * Warmup loop of 1% num_iter between 2 and 50


adam-smnk:
Side note, added returns to a few failing CUDA tests.

Some comments after testing:
1) Lack of stdev seems like a small regression. On one hand, timing outside of the perf loop can be beneficial for short running benchmarks. On the other hand, we switched to a single loop (opposed to warmup loop then perf loop) as collecting measurements all at once and discarding warmup timestamps afterwards seemed to reduce stdev in the past.
Now we cannot evaluate that but indeed current benchmarks numbers appear to be the same vs main branch.
Overall, not a blocker for me but just some thoughts.
1) New bench loop creation breaks CUDA benchmarking. Bufferization injects copy to a non-GPU memory buffer in the warmup loop and passes it to the GPU kernel which crashes at runtime.
This might need to be addressed at the warmup loop creation.
1) (minor) An extra unnecessary call is inserted before warmup loop in benchmarks:
```
call @_entry // unnecessary
for warmups {
  call @_entry
}
for bench {
  call @_entry
}
```

rengolin:
> Kernel calls expect to get valid GPU allocated buffers as arguments.

Reverted to old behaviour for running / printing.

adam-smnk (approved):
Looks and works good
Thanks 👍

PR #811, opened 2023-11-28
Expose `getUnaryFlags` and `getBinaryFlags` as xsmm utils (NFC)

chelini:
Needed in #810 for better XSMM verification.

### DIFF
--- a/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
+++ b/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
@@ -148,6 +148,140 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
   return binaryInfo;
 }
 
+// Examples:
+// If lower=[c], higher=[a, b, c], [c] reshaped into [1, 1, c].
+// If lower=[b, c], higher=[a, b, c], [b, c] reshaped into [1, b, c].
+// If lower=[a], higher=[a, a], [a] reshaped into [1, a].
+// If lower=[a], target=[a, b, a], [a] reshaped into [1, 1, a].
+// If lower=[], target=[a, b, c], [] reshaped into [1, 1, 1].
+static void
+computeBcastShapeInput(ArrayRef<int64_t> higherRankShape,
+                       ArrayRef<int64_t> lowerRankShape,
+                       SmallVectorImpl<int64_t> &reshapeOutputShape) {
+  // Initialize new shapes with [1] * higherRank.
+  int64_t higherRank = higherRankShape.size();
+  int64_t lowerRank = lowerRankShape.size();
+
+  reshapeOutputShape.assign(higherRank, 1);
+
+  int64_t higherRankDim;
+  int64_t lowerRankDim;
+
+  for (int64_t i = higherRank - 1, j = lowerRank - 1; i >= 0 && j >= 0;
+       i--, j--) {
+    higherRankDim = higherRankShape[i];
+    lowerRankDim = lowerRankShape[j];
+
+    if (lowerRankDim == 1 && higherRankDim > 1)
+      reshapeOutputShape[i] = 1;
+    else if ((lowerRankDim > 1 && higherRankDim == 1) ||
+             (lowerRankDim == higherRankDim))
+      reshapeOutputShape[i] = lowerRankDim;
+    else if (higherRankDim != lowerRankDim)
+      assert(false && "bCast semantics for identity op broken");
+  }
+}
+
+// Return true if `type` is a scalar or a 'scalar' memref (i.e., memref<f32> or
+// memref<1x1xf32>).
+static bool isScalarType(Type type) {
+  if (!type.isa<ShapedType>())
+    return true;
+  ShapedType shapedType = type.cast<ShapedType>();
+  ArrayRef<int64_t> shapeInput = shapedType.getShape();
+  auto isOne = [](int64_t val) { return val == 1; };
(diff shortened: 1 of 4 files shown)


lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,134 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
...
+      reshapeOutputShape[i] = lowerRankDim;
+    else if (higherRankDim != lowerRankDim)
+      assert(false && "bCast semantics for identity op broken");
+  }
+}
+
+FailureOr<UnaryFlags> getUnaryFlags(Type inputType, Type outputType) {
+  // There are multiple ways to define a scalar.  f32, memref<1x1xf32> or
+  // memref<f32>. Handle f32, and memref<1x1xf32>. memref<f32> is not allowed
+  // in tpp at the moment.
```
rengolin, line 188: I can't remember if this was a hard restriction of just we didn't care enough. AFAICS, `memref<1xf32>`, `memref<1x1xf32>`, etc are identical to `memref<f32>` in memory representation (not logically). XSMM shouldn't care about the logical semantics, right?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,134 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
...
+
+  ArrayRef<int64_t> shapeOutput = outputType.cast<ShapedType>().getShape();
+  assert(shapeOutput.size() >= shapeInput.size() &&
+         "output rank must be >= input rank");
+  SmallVector<int64_t> bShapeInput;
+  computeBcastShapeInput(shapeOutput, shapeInput, bShapeInput);
+  assert(shapeOutput.size() == bShapeInput.size());
+  shapeInput = bShapeInput;
+
+  if (shapeInput[1] == 1 && shapeOutput[1] > 1)
```
rengolin, line 204: Do we need to check if the rank is 2?

Is `2 x 1 x 2 -> 2 x 2 x 2` a row broadcast?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,134 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
...
+    return xsmm::UnaryFlags::BCAST_ROW;
+
+  if (shapeInput[0] == 1 && shapeOutput[0] > 1)
+    return xsmm::UnaryFlags::BCAST_COL;
+
+  if (shapeInput[0] == shapeOutput[0] && shapeInput[1] == shapeOutput[1])
```
rengolin, line 210: Can't remember if you can do `(shapeInput == shapeOutput)`.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,134 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
...
+    return xsmm::UnaryFlags::NONE;
+  return failure();
+}
+
+FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
```
rengolin, line 215: This requires the caller to validate if the "other" operand has the same shape as the output. I'd add all three operands and make that check here (even as a static call on the same unit).

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -148,6 +148,134 @@ FailureOr<BinaryInfo> getBinaryInfo(Value lhs, BinaryFlags lhsFlag, Value rhs,
...
+    assert(false && "unrechable");
+  };
+
+  // Multiple way to define a scalar. Check if the memref
+  // is a scalar here.
+  auto isOne = [](int64_t val) { return val == 1; };
+  if (llvm::all_of(bOperandShape, isOne))
+    return getBCastEnum(BCastType::SCALAR, operandNumber);
+
+  if (bOperandShape[1] == 1 && shapeOutput[1] > 1)
```
rengolin, line 270: Same question about rank as in the unary case.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
chelini: We used to pass memref to the runtime and we had to match the memref ABI, I can relax this checks now that we pass pointers.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
chelini: Yes, we can.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
chelini: We can have different flags per operand as far as I understood from LIBXSMM. Basically, the flags are computed per operand with respect to the output. Does this sound wrong to you?

rengolin: Good point. In theory we could have `(<10x1xf32>, <20xf32>) -> <10x20xf32>`, which would have both row/col broadcasts on different operands.

PR #816, opened 2023-12-01
Remove trailing whitespaces from *.mlir files (NFC)

chelini:


rengolin:
There is no clang-format support for MLIR:
https://discourse.llvm.org/t/clang-format-or-some-other-auto-format-for-mlir-files/75258

nhasabni:
Oh, okay, got it. Thanks

PR #820, opened 2023-12-13
Bump llvm

chelini:


lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
```diff
@@ -933,18 +947,25 @@ struct FoldUnPackIntoInsertSlice : public OpRewritePattern<tensor::UnPackOp> {
...
     auto forallOp = cast<scf::ForallOp>(loop);
-    if (!forallOp->hasOneUse())
+    if (!forallOp->hasOneUse() || forallOp->getNumResults() != 1)
       return failure();
     OpBuilder::InsertionGuard g(rewriter);
     rewriter.setInsertionPoint(forallOp);
 
     // Create a new scf.forall operation, updating its output.
-    Value unPackOpOutput = unPackOp.getDest();
+    assert(forallOp->getNumResults() == 1);
```
rengolin, line 956: This assert will never trigger

test/Passes/pack-unpack-propagation.mlir
```diff
@@ -478,7 +478,7 @@ func.func @matmul_with_relu_and_bias(%arg0: tensor<256x512xf32>, %arg1: tensor<5
 // CHECK-SAME:  iterator_types = ["parallel", "parallel", "parallel", "parallel"]
 // CHECK-SAME:  outs(%[[MATMUL]]
 // CHECK: %[[UNPACK:.+]] = tensor.unpack %[[RELU]] inner_dims_pos = [0, 1]
-// CHECK-SAME:  inner_tiles = [32, 32] into %[[BCAST]] : tensor<8x32x32x32xf32> -> tensor<256x1024xf32>
+// CHECK-SAME:  inner_tiles = [32, 32] into %[[ARG2]] : tensor<8x32x32x32xf32> -> tensor<256x1024xf32>
```
rengolin, line 481: This is still tensor semantics, which means `%0` can become dead and be removed... doesn't seem right.

chelini: This comes from an upstream changes. To me looks fine: https://github.com/llvm/llvm-project/commit/14e7846d6e2c5c99c52ba3882e59bfb021a5f0fa

test/Passes/fold-pack-and-constant.mlir
```diff
@@ -13,21 +13,6 @@ func.func @expect_to_fold_cst() ->  tensor<8x2x1x1x32x32xi64> {
 
 // -----
 
-func.func @expect_not_to_fold_cst(%dest : tensor<8x2x1x1x32x32xi64>) -> tensor<8x2x1x1x32x32xi64> {
```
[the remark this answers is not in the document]
chelini: We did not fold a splat tensor if the destination was not a tensor.empty; this is because it is unclear if we want to make a constant "writable" after bufferization by copying it into a new buffer, upstream folds in all cases: https://github.com/llvm/llvm-project/commit/d0da3d8393939790bb1a6b3b5a36daeeef000c9b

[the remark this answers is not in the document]
rengolin: Weird, to me this example should always constant pack... There's no empty here to argue about... Well, now it does.

adam-smnk (approved):
Passes both CPU and GPU tests

PR #823, opened 2023-12-15
Improve `packVNNIMatmulOp`

chelini:
Checking the rank is not the right way. A better approach is to check the affine map and bail out if the map is already in VNNI layout.

### DIFF
--- a/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
@@ -401,6 +401,41 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter,
   return packMatmulOpImpl<linalg::BatchMatmulOp>(rewriter, matmulOp, tiles);
 }
 
+// Check if the affine map associated to `operand` is already in VNNI format.
+static bool isInVnniLayout(OpOperand &operand, linalg::GenericOp linalgOp,
+                           int64_t blockingFactor) {
+  assert(operand.getOwner() == linalgOp);
+  AffineMap map = linalgOp.getMatchingIndexingMap(&operand);
+  ArrayRef<AffineExpr> results = map.getResults();
+  SmallVector<utils::IteratorType> iteratorTypes =
+      linalgOp.getIteratorTypesArray();
+
+  AffineExpr vnniDim = results.back();
+  auto dimExpr = dyn_cast<AffineDimExpr>(vnniDim);
+  if (!dimExpr ||
+      iteratorTypes[dimExpr.getPosition()] != utils::IteratorType::reduction) {
+    return false;
+  }
+
+  for (auto result : results) {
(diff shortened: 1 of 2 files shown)


lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
```diff
@@ -401,6 +401,41 @@ mlir::linalgx::packMatmulOp(RewriterBase &rewriter,
...
+  AffineExpr vnniDim = results.back();
+  auto dimExpr = dyn_cast<AffineDimExpr>(vnniDim);
+  if (!dimExpr ||
+      iteratorTypes[dimExpr.getPosition()] != utils::IteratorType::reduction) {
+    return false;
+  }
+
+  for (auto result : results) {
+    if (auto blockeDim = dyn_cast<AffineBinaryOpExpr>(result)) {
+      if (blockeDim.getKind() == AffineExprKind::FloorDiv) {
```
rengolin, line 422:
```suggestion
    auto blockeDim = dyn_cast<AffineBinaryOpExpr>(result);
    if (!blockedDim)
      continue;
    if (blockeDim.getKind() != AffineExprKind::FloorDiv)
      continue;
```

test/BF16/brgemm-vnni.mlir
```diff
@@ -1,14 +1,120 @@
...
+// CHECK-LABEL: brgemm
+// CHECK-SAME:  %[[ARG0:.+]]: tensor<32x4x4xbf16>, %[[ARG1:.+]]: tensor<32x4x4xbf16>, 
+// CHECK-SAME:  %[[ARG2:.+]]: tensor<4x4xbf16>
 // CHECK: %[[EMPTY:.+]] = tensor.empty() : tensor<32x2x4x2xbf16>
 // CHECK: %[[PACK:.+]] = tensor.pack %[[ARG1]]
-// CHECK-SAME:  inner_dims_pos = [1] inner_tiles = [2] into %[[EMPTY]] : tensor<32x4x4xbf16> -> tensor<32x2x4x2xbf16>
-// CHECK: tpp.brgemm (%[[ARG0]] : tensor<32x4x4xbf16>, %[[PACK]] : tensor<32x2x4x2xbf16>, %[[ARG2]] : tensor<4x4xbf16>) -> (tensor<4x4xbf16>)
+// CHECK-SAME:  inner_dims_pos = [1] inner_tiles = [2] into %[[EMPTY]] 
+// CHECK-SAME:  : tensor<32x4x4xbf16> -> tensor<32x2x4x2xbf16>
+// CHECK: tpp.brgemm (%[[ARG0]] : tensor<32x4x4xbf16>, %[[PACK]] : tensor<32x2x4x2xbf16>,
```
rengolin, line 17: We really should not be adding more tpp dialect tests...

chelini: if you wait for another patch all of these will disappear.

chelini: "all of these" = tpp.brgemm

PR #826, opened 2023-12-18
Do not interchange in `packVNNIMatmulOp`

chelini:
- Packing introduces an interchange on the packed dimension; there is no
  obvious reason for doing this, and it breaks the "canonical" form. If we
  need to introduce an interchnage, we need to do this explicitly and not
  while packing.

- Make sure we can handle possible interchanges when mapping to BRGEMM.
  See changes in `isBrgemmVnniOp`.

### DIFF
--- a/include/TPP/Transforms/Utils/VNNIUtils.h
+++ b/include/TPP/Transforms/Utils/VNNIUtils.h
@@ -1,4 +1,4 @@
-//===- VNNIUtils.cpp ---------------------------------------------*- C++-*-===//
+//===- VNNIUtils.h -----------------------------------------------*- C++-*-===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -9,12 +9,20 @@
 #ifndef TPP_TRANSFORMS_UTILS_VNNIUTILS_H
 #define TPP_TRANSFORMS_UTILS_VNNIUTILS_H
 
+#include "mlir/Support/LogicalResult.h"
 #include <cstdint>
 #include <optional>
 
 namespace mlir {
 class Type;
 class MemRefType;
+class OpOperand;
+class AffineDimExpr;
+class AffineMap;
+
+namespace linalg {
+class GenericOp;
+} // namespace linalg
 
 namespace vnni {
 namespace utils {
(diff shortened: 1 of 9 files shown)


adam-smnk:
Any impact on benchmarks?

chelini:
> Any impact on benchmarks?

Currently, tests are timed out as we need to match what MLIR-gen generates. During packing, we moved the VNNI dimension. We pack as [p, p, r, r, p, p, r] instead of [p, p, r, p, p, r, r]. MLIR-gen copied the behavior, @KavithaTipturMadhu . Do you remember why we pack this way?

Possible solutions:
1. Adjust mlir-gen
2. Improve detection `isVnniBrgemmOp` to cover also possible interchange happening during packing

KavithaTipturMadhu:
> > Any impact on benchmarks?
> 
> Currently, tests are timed out as we need to match what MLIR-gen generates. During packing, we interchange the VNNI dimension. We pack as [p, p, r, r, p, p, r] instead of [p, p, r, p, p, r, r]. MLIR-gen copied the behavior, @KavithaTipturMadhu . Do you remember why we pack this way?
> 
> Possible solutions:
> 
> 1. Adjust mlir-gen
> 2. Improve detection `isBrgemmVnniOp` to cover also possible interchange happening during packing. The current form: https://github.com/plaidml/tpp-mlir/blob/dc64f73abb679b1f6d6333acf9764d4d8e898ec6/lib/TPP/Dialect/Tpp/TppUtils.cpp#L30
>     expects a specific iterator order.
> 
> We should generally not care about possible interchanges, so extending the matcher should be done anyway. But I feel we should not interchange during packing if there are no good reasons for doing that.

I don't recall exactly why I chose to pack as [p, p, r, r, p, p, r], sorry. It should've been [p, p, r, p, p, r, r] as you pointed out.

chelini:
> > > Any impact on benchmarks?
> > 
> > 
> > Currently, tests are timed out as we need to match what MLIR-gen generates. During packing, we interchange the VNNI dimension. We pack as [p, p, r, r, p, p, r] instead of [p, p, r, p, p, r, r]. MLIR-gen copied the behavior, @KavithaTipturMadhu . Do you remember why we pack this way?
> > Possible solutions:
> > 
> > 1. Adjust mlir-gen
> > 2. Improve detection `isBrgemmVnniOp` to cover also possible interchange happening during packing. The current form: https://github.com/plaidml/tpp-mlir/blob/dc64f73abb679b1f6d6333acf9764d4d8e898ec6/lib/TPP/Dialect/Tpp/TppUtils.cpp#L30
> >    
> >    expects a specific iterator order.
> > 
> > We should generally not care about possible interchanges, so extending the matcher should be done anyway. But I feel we should not interchange during packing if there are no good reasons for doing that.
> 
> I don't recall exactly why I chose to pack as [p, p, r, r, p, p, r], sorry. It should've been [p, p, r, p, p, r, r] as you pointed out.

Ok thanks, np. This should be ok now.

Issue #806, opened 2023-11-24
BF16 BRGEMM is not matched on mlir-gen input when using prepacked weights and tilesize and tensor size are identical


alheinecke:
How to reproduce:
`./mlir-gen --kernel=model --float-width=16 --batch=256 --layers=64,64,64,64 --tiles=64,64,64 --vnni=2 > gemm_bf16.mlir`
In this case the weight is in VNNI layout:
`%cst = arith.constant dense<1.000000e+00> : tensor<1x1x32x64x2xbf16>`
However, when running `./tpp-opt -default-tpp-passes gemm_bf16.mlir` the `linalg.generic` is not matched to BRGEMM with count 1 or a regular GEMM.

This issue only exists for BF16, when using it on a similar case for FP32
`./mlir-gen --kernel=model --float-width=32 --batch=256 --layers=64,64,64,64 --tiles=64,64,64 > gemm_fp32.mlir`
everything works as expected. 

@chelini

PR #837, opened 2024-01-03
Allow conversion for VNNI layout matmul in linalg to xsmm

chelini:
Matmuls are BRGEMMs without a batch dimension, this is true also in the VNNI layout. This PR made some changes to `isBrgemmVnniOp` and `ConvertGenericToVnniBrgemm` to allow matmuls lowering in VNNI layout.

Fix: #806

### DIFF
--- a/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
+++ b/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
@@ -1054,13 +1054,19 @@ struct ConvertVnniPacking : public OpRewritePattern<linalg::TransposeOp> {
   }
 };
 
-struct ConvertGenericToVnniBrgemm : public OpRewritePattern<linalg::GenericOp> {
+// Converts linalg.generic with the following layout:
+// [i][j] = [i][k] [k/VNNI][j][VNNI] -> xsmm.matmul
+// [i][j] = [b][i][k] [b][k/VNNI][j][VNNI] -> xsmm.brgemm
+struct ConvertGenericToVnniMatmulLikeOp
+    : public OpRewritePattern<linalg::GenericOp> {
   using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
 
   LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
                                 PatternRewriter &rewriter) const override {
+    bool hasBatch = false;
     if (!genericOp.hasBufferSemantics() ||
-        !tpp::utils::isBrgemmVnniOp(genericOp, /*captures=*/nullptr)) {
+        !tpp::utils::isBrgemmVnniOp(genericOp, hasBatch,
+                                    /*captures=*/nullptr)) {
       return failure();
     }
     Value bufferA = genericOp.getDpsInputs()[0];
@@ -1069,8 +1075,13 @@ struct ConvertGenericToVnniBrgemm : public OpRewritePattern<linalg::GenericOp> {
 
     int64_t m = bufferC.getType().cast<ShapedType>().getShape()[0];
     int64_t n = bufferC.getType().cast<ShapedType>().getShape()[1];
-    int64_t k = bufferA.getType().cast<ShapedType>().getShape()[2];
-    int64_t batch = bufferA.getType().cast<ShapedType>().getShape()[0];
+    int64_t kPos = 1;
+    if (hasBatch)
+      kPos++;
+    int64_t k = bufferA.getType().cast<ShapedType>().getShape()[kPos];
+    int64_t batch = std::numeric_limits<int64_t>::max();
+    if (hasBatch)
+      batch = bufferA.getType().cast<ShapedType>().getShape()[0];
 
     auto stridesOnLhs = utils::getStaticStrides(bufferA);
     auto stridesOnRhs = utils::getStaticStrides(bufferB);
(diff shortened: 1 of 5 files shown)


lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -1069,8 +1075,13 @@ struct ConvertGenericToVnniBrgemm : public OpRewritePattern<linalg::GenericOp> {
 
     int64_t m = bufferC.getType().cast<ShapedType>().getShape()[0];
     int64_t n = bufferC.getType().cast<ShapedType>().getShape()[1];
-    int64_t k = bufferA.getType().cast<ShapedType>().getShape()[2];
-    int64_t batch = bufferA.getType().cast<ShapedType>().getShape()[0];
+    int64_t kPos = 1;
+    if (hasBatch)
+      kPos++;
+    int64_t k = bufferA.getType().cast<ShapedType>().getShape()[kPos];
+    int64_t batch = std::numeric_limits<int64_t>::max();
```
rengolin, line 1082: I don't understand why `batch = MAX_INT` is "no batch". The code that uses this makes a check and doesn't use it so it could very well be just zero and we wouldn't even need the extra flag.

chelini: Indeed we set "batch" to MAX_INT if the detected operation does not have the batch dimension (hasBatch = false). We pass the batch dimension to https://github.com/plaidml/tpp-mlir/blob/99af3d548226ad5ff0082fd6ff8359ef1b8494c5/lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp#L443. Do you see a way to avoid the flag or the batch value?

chelini: We could also set batch to zero and not max_int.

rengolin: Yup, my point above. :)

chelini: Maybe you cannot see from the diff but `batch` is actually used and passed to `replaceOpWithGemmLikeOp` later on.

rengolin: I did see it and it's precisely `replaceOpWithGemmLikeOp` that can change, since `batch` inside it is only used if `hasBatch` is true, and that is set to true in the line you mentioned. This can all be avoided if `batch` is zero and then you don't need the flag inside `replaceOpWithGemmLikeOp`, just do `if (batch) { ... }`.

chelini: https://github.com/plaidml/tpp-mlir/pull/837/commits/27299c73e7e9ea7476c5a339b0d81b08b2ff78a6

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -575,9 +573,8 @@ static FailureOr<BrgemmInfo> checkAccess(linalg::LinalgOp linalgOp, unsigned m,
   }
 
   auto loops = linalgOp.computeStaticLoopSizes();
-  int64_t batchVal = (batch != std::numeric_limits<unsigned>::max())
-                         ? loops[batch]
-                         : std::numeric_limits<int64_t>::max();
+  int64_t batchVal =
+      (batchPos != std::numeric_limits<unsigned>::max()) ? loops[batchPos] : 0;
```
rengolin, line 577: Why would `batchPos` be `MAX_INT` here? You can easily replace the existing callers to change to zero and not have to worry about int `max` any more anywhere.

chelini: We cannot use 0 for batchPos because it represent the position of the batch dimension in the domain of the generic.

rengolin: The thing is, you're encoding a "not-a-number" value type with an "inf" value and that is confusing. Can we perhaps use `std::optional` here?

chelini: We could, but this would mean treating the batch in a special way different from the other dimensions like m, n, and k and encoding the special case in methods like: `getPosInCodomain.`

rengolin: Fair, let's postpone this discussion.

rengolin (approved):

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
chelini: Let me merge this and create a PR on this, sounds good?

chelini: if we don't like the changes we can discard the PR.

rengolin: Sounds good, thanks!

PR #838, opened 2024-01-04
Use std::optional for batch position

chelini:
Instead of encoding the batch position using the maximum unsigned number, use std::optional to improve readability. Follow-up for #837.


rengolin (approved):
LGTM, thanks!

PR #848, opened 2024-01-08
Step 14 retire tpp

chelini:


include/TPP/IR/StructuredOpMatcher.h
```diff
@@ -95,11 +95,31 @@ struct HasMap {
 };
 
 // Callble object to verify if `map` is a projected permutation map.
+// We require the dimensions to be in sorted order this avoid filtering
```
chelini, line 98: `isProjectedPermutation` is not enough for our use case as we need to make sure the dimensions are broadcastable. For example, we should disallow a transpose add operation with map:
affine_map<(d0, d1) -> (d1, d0)> on input and affine_map<(d0, d1) -> (d0, d1)> on output.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -233,14 +230,14 @@ struct TppMapping : public tpp::impl::TppMappingBase<TppMapping>,
 };
 
 // Convert all matching operations to TPP.
+// TODO: We can remove it now. It has a single pass inside.
```
chelini, line 233: I will address this TODO in a later PR if is ok for you.

test/Models/mlp-3layers.mlir
chelini, line 1: I drop all the models, they are super annoying to have in the repo. Let me know what you think.

lib/TPP/DefaultTppPasses.cpp
adam-smnk: Fair enough, we can cleanup pipeline and other passes after.

test/Models/mlp-3layers.mlir
chelini: @rengolin what do you want to do with the Models?

test/Passes/DefaultPipeline/tpp.mlir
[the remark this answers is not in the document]
chelini: https://github.com/plaidml/tpp-mlir/pull/848/commits/c6bab53a197b4ce67808469d04aa541f7417208f

adam-smnk (approved):

PR #855, opened 2024-01-09
Move dialect registration in TD file for `LinalgLowering` (NFC)

chelini:


chelini:
Trivial NFC I will merge, sorry for the noise.

PR #860, opened 2024-01-11
Update fusion after LLVM bump

chelini:
Previously, only building blocks were exposed, and fusion users were supposed to assemble them. Most heavy lifting is shifted into `tileConsumerAndFuseProducerGreedilyUsingSCFForOp,` which tiles a consumer and fuses it with producers. Note that the method relies on a callback to decide what to fuse; what to fuse is the tricky part, which for us is computed in `collectFusableProducers`.

LLVM commit: https://github.com/llvm/llvm-project/commit/4435ced94998c00a6589c3500822015b6341c9e3


adam-smnk (approved):
Great simplification 👍

PR #869, opened 2024-01-17
Trim includes in `DefaultPipeline.cpp` (NFC)

chelini:


chelini:
It is broken, trim too much.

adam-smnk (approved):

PR #871, opened 2024-01-23
Linalg to WMMA lowering rework

adam-smnk:
Reworks linalg lowering to WMMA ops to support computations on memory tiles larger than hardware WMMA tiles.

The new lowering effectively applies warp (or subgroup) tiling on 2D tiled linalg GEMM-like operations. Additionally, the reduction dimension tiling is performed at the same time to ensure that the computation fits within limits of hardware resources.
The elementwise consumers are fused with GEMM when possible.

The warp tiling is done at the time of linalg lowering as the parameter decisions, such as tile sizes, are driven by the compute workload. Mapping first to individual WMMA operations and splitting them later into multiple ops operating on smaller sub-tiles would be more complex due to need for use-chain analysis and extra validation.


adam-smnk:
Lowering GEMM through gpu dialect WMMA ops acts as a high level GPU representation between pure computation (linalg) and a target-specific operations of Nvidia (nvgpu/nvvm - already supported) and Intel (XeGPU - to be added).
That is WMMA representation brings GEMM operation closer to GPU realm - initial data movement, subgroup/warp cooperative operations - without vendor-specific implementation details.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -69,137 +71,213 @@ createGpuBlocksWrapper(Operation *op, ArrayRef<int64_t> blockDims,
...
+
+  // Choose WMMA tile sizes.
+  // The computation dimensions must fit into the tiles.
+  // Reduction dimension tile size has to be compatible
+  // with the hardware sizes.
+  //
+  // TODO: Add more possible tile sizes and choose optimal one.
+  int wmmaTileM = 16;
+  int wmmaTileN = 16;
+  int wmmaTileK = 16;
```
adam-smnk, line 137: TODO: move this to a pass option provided by a user

It will be more flexible and provide space for easier experimentation in further lowering.

adam-smnk:
WMMA to XeGPU lowering has been added to `linalg-to-xegpu` branch.

lib/TPP/GPU/GpuPipeline.cpp
```diff
@@ -70,6 +70,31 @@ GpuType parseGpuOption(StringRef gpuStr) {
...
+  GpuOptions options;
+
+  switch (gpuType) {
+  case GpuType::Cuda: {
+    options.triple = "nvptx64-nvidia-cuda";
+    options.chip = "sm_70";
+    options.features = "+ptx60";
+    break;
+  }
+  case GpuType::Vulkan: {
```
rengolin, line 89: We'll change this to `SYCL` or some other non-Vulkan thing, right?

lib/TPP/GPU/GpuPipeline.cpp
```diff
@@ -127,22 +153,22 @@ struct GpuPipeline : public tpp::impl::GpuPipelineBase<GpuPipeline>,
     pm.addPass(createConvertForAllToParallelOp());
     pm.addNestedPass<func::FuncOp>(createCleanup());
 
+    // Default warp tile sizes.
+    SmallVector<int64_t> warpTileSizes = {16, 16, 16};
```
rengolin, line 157: Don't leave default values spread around the code. Either make it default option in the function signature or add a command line option with a default value.

lib/TPP/GPU/LinalgToGpu.cpp
```diff
@@ -84,122 +90,171 @@ static bool supportsMMACompute(linalg::LinalgOp linalgOp) {
...
+
+  // Collect new results after fusion.
+  SmallVector<Value> fusedRes;
 
-  std::optional<Operation *> newStore = std::nullopt;
   SmallVector<Value> operands;
   if (structured_match::utils::isTwoDAddOp(consumer, &operands)) {
     // Get the value to be added - load the tile first.
     // Must be a buffer of the same type - scalar broadcast is not supported.
+    // TODO: Consider adding support for eltwise with broadcast.
```
rengolin, line 160:
```suggestion
    // TODO: Add support for eltwise with broadcast.
```

rengolin (commented):
From a high-level POV, this looks like it should work, but it may not be super generic. It's ok to be like that initially, especially while XeGPU isn't upstream yet, but we need to make sure whatever we do can go upstream ASAP.

The state we need to aim for is a generic lowering for GPUs that works on various dialects (nv, amd, xe) and doesn't break common targets. Later on we add cost models for each specific device to know how to lower different patterns to different targets.

lib/TPP/GPU/GpuPipeline.cpp
adam-smnk: Yes, I intend to completely replace Vulkan with SYCL.

lib/TPP/GPU/GpuPipeline.cpp
adam-smnk: Added CLI options as three separate flags as `cl::opt` does not accept SmallVector.

rengolin: Isn't there a version that accepts a string list like "16,16,16", then we split?

adam-smnk: Found a way to do that through `llvm::cl` 👍

Issue #704, opened 2023-09-05
tpp-run does not support ml_program dialect


nhasabni:
Tried running `torch-mlir` exported ResNet in `linalg-on-tensor` via `tpp-run` and found a crash. `tpp-opt` works fine though.

*Commands* (Install `torch-mlir` using `pip`)

```
$ python examples/torchscript_resnet18_all_output_types.py
$ tpp-opt rn18.mlir -o rn18.mlir.opt
$ tpp-run rn18.mlir.opt -e forward -entry-point-result=void
```

*Error*

```
$ ./tpp-run -e forward -entry-point-result=void rn18.mlir.opt
loc("rn18.mlir.opt":9:3): error: cannot be converted to LLVM IR: missing `LLVMTranslationDialectInterface` registration for dialect for op: ml_program.global
tpp-run: /nfs_home/nhasabni/other/tensor_compiler/nhasabni_tpp-sandbox/tools/tpp-run/tpp-run.cpp:199: std::unique_ptr<llvm::Module> lowerToLLVMIR(mlir::Operation *, llvm::LLVMContext &): Assertion `llvmModule' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
```

adam-smnk:
`ml_program` seems to be a dead-end in the upstream MLIR.

The basic dialect ops are defined (that's why `tpp-opt` is fine with it), however, there are no conversion passes or any further integration. The dialect seems like a stub for frontend conversion but not much more.
IREE has custom lowering passes for `ml_program` (see `iree/compiler/MHLO/MHLOToLinalgOnTensors.cpp`) but I see nothing relevant available upstream.

Looking at this rn18 example from torch-mlir, it seems like the one `ml_program.global` variable isn't used anywhere. So, I hope we can get away with some minor IR cleanup are run the rest as is.

@nhasabni do you think that's feasible or lack of `ml_program` lowering might become a blocker in near future?

rengolin:
Could this be an upstream pass to simple bufferize `ml_program.global` to `memref.global`?

rengolin:
I see a few ways to work on this:
1. Try to upstream some conversion/bufferization pass to `ml_program`. This is quick and should be uncontroversial, unless people are already preparing to kill that dialect.
[...]

PR #879, opened 2024-02-04
Bump llvm

chelini:
Things that need to be understood before merge:
1. in `dce.mlir`, it looks like the constant is not eliminated anymore; why?
2. MemAlloc effect for perfBench, is correct? Or does buffer deallocation bail out on operation marked with `AutomaticAllocationScope` trait?


adam-smnk:
Added minor tweaks to fix GPU tests.
A few tests now fail due to upstream changes in deallocation pass. We'll see how to handle that later but it probably needs to be fixed upstream.

> 1. in dce.mlir, it looks like the constant is not eliminated anymore; why?

MLProgram has upstream bufferization now so, the original problem from #704 disappeared. And it looks that `symbol-dce` does not remove `llvm.mlir.global` ops. But it is fine to remove this check.

adam-smnk:
> 2. MemAlloc effect for perfBench, is correct? Or does buffer deallocation bail out on operation marked with AutomaticAllocationScope trait?

Added `AutomaticAllocationScopeResource` to `MemAlloc` to capture the automatic scope dependency.
Also added memref read, write, and recursive effects to capture side effects on `iter_args`.

include/TPP/Dialect/Perf/PerfOps.td
```diff
@@ -98,6 +98,8 @@ def Perf_StopTimerOp : Perf_Op<"stop_timer", []> {
 
 def Perf_BenchOp : Perf_Op<"bench",
     [AutomaticAllocationScope, SingleBlockImplicitTerminator<"perf::YieldOp">,
+     MemoryEffects<[MemWrite, MemRead, MemAlloc<AutomaticAllocationScopeResource>]>,
```
rengolin, line 101: We don't allocate memory any more, why do we need this effect?

rengolin (commented):
> Added minor tweaks to fix GPU tests.
> A few tests now fail due to upstream changes in deallocation pass. We'll see how to handle that later but it probably needs to be fixed upstream.

Is this related to Vulkan (so won't happen in the SYCL runner), or is it related to how we have our tests?

adam-smnk:
> Is this related to Vulkan (so won't happen in the SYCL runner), or is it related to how we have our tests?

I run deallocation pass as a part of SPIR-V pipeline as unified memory can be easily used with the SPIR-V runtimes.
However, in these cases `gpu.launch_func` does not implement 'MemoryEffectOpInterface' which causes the new stricter deallocation pass to fail. It is a more general issue.

include/TPP/Dialect/Perf/PerfOps.td
adam-smnk: After digging a bit further, good point.
`RecursiveMemoryEffects` should be enough to "derive" memory effect from nested ops.

rengolin (approved):

PR #883, opened 2024-02-15
Allow tiling dimensions equal to tile size

adam-smnk:
Adds a new option to tile-and-fuse pass to expose control over dimension to tile size factor (ratio) when selecting eligible operations for tiling. The chosen factor requirement must be fulfilled for all dimensions.

This change allows tiling when dimensions are equal to their corresponding tile size or restricts tiling to larger workloads.
For example, tile factor of 1 creates more opportunities in kernel outlining for wide and tall workloads e.g., memref<128x1024> into 128x128 tiles.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -153,7 +153,9 @@ def TileConsumerAndFuseProducers : Pass<"tile-consumer-and-fuse-producers",
            "Get producers till maxDepth">,
     Option<"numIters", "num-iters", "int64_t", "3",
            "Run fusion for the given number of iterations">,
-    Option<"useForAll", "use-for-all", "bool", "true", "Use parallel forAll">
+    Option<"useForAll", "use-for-all", "bool", "true", "Use parallel forAll">,
+    Option<"minTileFactor", "min-tile-factor", "int64_t", "2",
+           "Minimum factor between dimension size and a tile size">
   ];
   let dependentDialects = ["linalg::LinalgDialect", "scf::SCFDialect",
                            "tensor::TensorDialect"];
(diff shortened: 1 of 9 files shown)


adam-smnk:
2~4% regression on some MHA benchmarks due to more fine-grained tiling after relaxing heuristic.
No impact on simple gemm and mlp benchmarks.

The main motivation for the change is improving GPU outlining where tile-and-fuse will be used to map workloads into block tiles and thread subtiles. Long term we might need to have more flexible way to manage tiling heuristic.

rengolin:
> The main motivation for the change is improving GPU outlining where tile-and-fuse will be used to map workloads into block tiles and thread subtiles. Long term we might need to have more flexible way to manage tiling heuristic.

You mean, like this? 😄 
https://discourse.llvm.org/t/rfc-target-description-and-cost-model-in-mlir/76990

adam-smnk:
> > The main motivation for the change is improving GPU outlining where tile-and-fuse will be used to map workloads into block tiles and thread subtiles. Long term we might need to have more flexible way to manage tiling heuristic.
> 
> You mean, like this? 😄 https://discourse.llvm.org/t/rfc-target-description-and-cost-model-in-mlir/76990

This will definitely help 🔥
In tile and fuse there are internally quite a few decisions made, some of that could be more exposed perhaps through optional user callbacks.

For now the performance penalty is small and I think it is worth greater flexibility.

rengolin:
> This will definitely help 🔥
> In tile and fuse there are internally quite a few decisions made, some of that could be more exposed perhaps through optional user callbacks.

@nhasabni is working on that right now, we'll try to push that upstream, but maybe test in tpp-mlir first, not sure yet.

rengolin (commented):
Small comment in test, otherwise, LGTM.

adam-smnk:
Reverted changes to the default tiling validation. Instead, added it as an option with the default value corresponding to the previous behavior.

TODO: add tiling tests for the options and test tiling for GPU.

rengolin (approved):
LGTM, thanks!

@nhasabni this is a good candidate to the cost model!

adam-smnk:
Added missing test. Default tiling behavior is unchanged, thus, no more regressions in benchmarks.

PR #885, opened 2024-02-16
Parallel Grid loop creation for gemms

KavithaTipturMadhu:


rengolin:
Awesome! I'm seeing 38% improvement on `mlir-gen` BF16 16-threads (smaller improvements with less threads), and 22% for our PyTorch models BF16 16-threads, and no reductions in single-threaded benchmarks. I'll look at the code soon, but at least now we can run some benchmarks in parallel.

include/TPP/Dialect/Transform/SCF/Dialect/Passes.h
```diff
@@ -0,0 +1,72 @@
+//===- Passes.h - Pass Entrypoints ------------------------------*- C++ -*-===//
```
rengolin, line 1: What's the point of this header? Don't we already have those passes in the pipeline?

lib/TPP/DefaultPipeline.cpp
```diff
@@ -46,6 +46,12 @@ llvm::cl::opt<bool>
                 llvm::cl::desc("Default pipeline - enable parallel execution"),
                 llvm::cl::init(false));
 
+// Control grid parallelism sizes.
+llvm::cl::opt<std::string>
+    parallelTaskGrid("parallel-task-grid",
+                     llvm::cl::desc("Grid-sizes for parallel tasks"),
+                     llvm::cl::init("2,8"));
```
rengolin, line 53: What would it be the right arguments for the OMP benchmarks we have?

Can you add them to the BF16 benchmarks to see if it improves from the current state?

Here's my calculations on speed up from baseline:
| threads | speedup |
| -- | -- |
| 2 | 10% |
| 4 | 20% |
| 8 | 22% |
| 16 | 38% | 

At the very least, having them in the benchmark gives us "documentation" of what the expected values are, so we use it in the compiler once it starts doing that automatically.

lib/TPP/DefaultPipeline.cpp
```diff
@@ -74,6 +80,18 @@ PrintStage parsePrintStage(StringRef stage) {
       .Default(PrintStage::None);
 }
 
+std::vector<int64_t> parseParallelTaskGrid(std::string parallelTaskGrid) {
```
rengolin, line 83: Don't do this, use `llvm::cl::list<int64_t>` in the cmd-line option.

https://github.com/plaidml/tpp-mlir/blob/00dbee7c39d7576d2b8aa86624fd98aef4bd4d8d/lib/TPP/GPU/GpuPipeline.cpp#L49

lib/TPP/Dialect/Transform/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 add_mlir_dialect_library(TPPLinalgXTransformOps
     LinalgXTransformOps.cpp
     IteratorCollapsing.cpp
+    SCFParallelLoopTiling.cpp
```
rengolin, line 4: Why is this in the transform dialect? We're not using the transform dialect here, just passes.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -411,6 +411,7 @@ FailureOr<int64_t> getLeadingDim(Type type, size_t pos) {
   return strides[pos];
 }
 
+
```
rengolin, line 414: spurious change, please revert this file

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -146,14 +102,17 @@ struct CombineXsmmOp : public OpRewritePattern<xsmm::BrgemmOp> {
     IntegerType integer64 = IntegerType::get(rewriter.getContext(), 64);
 
     Location loc = brgemmOp.getLoc();
-    auto dims = getSizesAndLeadingDimForBrgemmOp(rewriter, brgemmOp);
+    auto dims = DenseI64ArrayAttr::get(
```
rengolin, line 105: The old code was doing validation of the input. Are we sure that the brgemm dispatch will always be valid for the fused brgemm?

test/Passes/pass-parallel-tile.mlir
```diff
@@ -0,0 +1,47 @@
+//RUN: tpp-opt %s  --scf-parallel-loop-tiling-pass -split-input-file | FileCheck %s
+
+  func.func @entry() {
```
rengolin, line 3: You really need more tests. At least one GEMM and one MLP layer checking the right loops are being created.

I'd also create smaller tests with different pass tile options and make sure the right split is done.

alheinecke:
BTW: I cannot build this PR:

/opt/rh/gcc-toolset-13/root/usr/lib/gcc/x86_64-redhat-linux/13/../../../../bin/ld: lib/libTPPPipeline.a(DefaultPipeline.cpp.o): in function `(anonymous namespace)::DefaultPipeline::constructPipeline()':
/../lib/TPP/DefaultPipeline.cpp:159: undefined reference to `mlir::tpp::createSCFParallelLoopTiling(mlir::tpp::SCFParallelLoopTilingOptions const&)'
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)

lib/TPP/Transforms/CombineXsmmPass.cpp
alheinecke: The old was broken (it recomputed and values changed parameters, e.g. stride), it's unclear why it was there and being used. In this case we just need to copy from BRGEMM to fusedBRGEMM. So this patch is fix.

alheinecke:
I could make it compile, after making it a default TPP pass
https://github.com/alheinecke/tpp-mlir/commit/f97604c758323a6bd9285804bbfb7ddfbaf0d547

lib/TPP/DefaultPipeline.cpp
alheinecke: 2,8 is for the 16 thread case and the sizes (256x1024). When we do a 32x32 blocking, we get a grid of 8x32 32x32 blocks for the 256x1024 result matrix. 2x8 as parallel task grid means we stamp out 2x8 blocks (and assign them to 1 cores) from the 8x32 task grid. That means the original matrix is now optimal computed in a 4x4 grid (minimal communication).

Having said this, we have to try for all sizes and thread counts which is "good", e.g. I just did this for 8 threads and got best performance for 4,8 , for 4 threads it's 8,8 and for 2 threads it's 8,16

... which is also the chosen libxsmm-dnn config :-) and it makes sense as we minimize weight matrix movement with these settings.

Perhaps we want to change the parameter to #workers in each dimensions instead of tiles/worker in each dimension... However, this we can also fix in a future PR.

include/TPP/Dialect/Transform/SCF/Dialect/Passes.h
alheinecke: see https://github.com/alheinecke/tpp-mlir/commit/f97604c758323a6bd9285804bbfb7ddfbaf0d547 . I had issues compiling the PR, not sure why CI was able to do it, I got linker errors for tpp-run....

I guess this is only there for an include and a define. I moved things around in https://github.com/alheinecke/tpp-mlir/commit/f97604c758323a6bd9285804bbfb7ddfbaf0d547

rengolin:
> I could make it compile, after making it a default TPP pass [alheinecke@f97604c](https://github.com/alheinecke/tpp-mlir/commit/f97604c758323a6bd9285804bbfb7ddfbaf0d547)

This looks like the right fix. Adding it to the transform dialect is the wrong direction.

I'm not sure why the CI can build it and you can't. Perhaps a clean build would works? Regardless, your patch is the right direction anyway.

lib/TPP/DefaultPipeline.cpp
rengolin: > Having said this, we have to try for all sizes and thread counts which is "good", e.g. I just did this for 8 threads and got best performance for 4,8 , for 4 threads it's 8,8 and for 2 threads it's 8,16

Good, then for now we just change the benchmarks JSON config to have those options to the pass, so that we "document" that these are the optimal configs for each number of threads.

> Perhaps we want to change the parameter to #workers in each dimensions instead of tiles/worker in each dimension... However, this we can also fix in a future PR.

Indeed, for a future PR. But having the options in the benchmarks JSON files will help whoever tackles this to get there faster, as well as show our best performance for each benchmark.

alheinecke:
I did a fully clean build... and following the instructions from our ReadME (have them in a script, where I just set the branch name and origin). It seems something went sideways that pass was in the LinAlgXTransform library?!

rengolin:
> I did a fully clean build... and following the instructions from our ReadME (have them in a script, where I just set the branch name and origin). It seems something went sideways that pass was in the LinAlgXTransform library?!

It was, wrong place. Not sure how the CI build it either. I get this error with the PR:
```
lib/TPP/Dialect/Transform/SCFParallelLoopTiling.cpp:266:9: error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]
        tileParallelLoop(ploop, tileSizes, noMinMaxBounds);
        ^~~~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
But that's unrelated to your error / fix, I think.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Reverted

lib/TPP/DefaultPipeline.cpp
KavithaTipturMadhu: Fixed

test/Passes/pass-parallel-tile.mlir
KavithaTipturMadhu: I have added gemm and mlp test cases with different grid sizes generated from mlir-gen to ensure that the splits are correct.

rengolin (approved):
The code itself is very confusing, with manual copying the basic block, `#if 0` and dead variables, but as discussed, it should be fine until we gather more examples and understand all the behaviour we want form it, including the dynamic case.

See if you can add some comment to the confusing parts trying to explain what you were trying to do or didn't do because of another problem (for example, the basic block copy), so that we can more easily read it later and fix it.

KavithaTipturMadhu:
> The code itself is very confusing, with manual copying the basic block, `#if 0` and dead variables, but as discussed, it should be fine until we gather more examples and understand all the behaviour we want form it, including the dynamic case.
> 
> See if you can add some comment to the confusing parts trying to explain what you were trying to do or didn't do because of another problem (for example, the basic block copy), so that we can more easily read it later and fix it.

I have added a couple of comments for now and gotten rid of #if 0 block and the dead code.

PR #887, opened 2024-02-21
Benchmark script

rengolin:
Trying to create a script that can help external users to build and benchmark our project. Preparing for the AWS runs before we book the instances and start paying for it, so testing locally for now.

Some fixes for OpenMP arguments that were out of sync between `tpp-run` and `xsmm_dnn_mlp` which made a difference on a newer Ubuntu.

@alheinecke

### DIFF
--- a/scripts/benchmarks/build_and_run.sh
+++ b/scripts/benchmarks/build_and_run.sh
@@ -59,7 +59,7 @@ echo_run ./driver.py -vv \
 echo " ========= PyTorch Benchmarks ==========="
 echo_run ./driver.py -vv \
          -n ${NUM_ITER} \
-         -c "${CONFIG_DIR}pytorch/torch_dynamo.json" \
+         -c "${CONFIG_DIR}/pytorch/torch_dynamo.json" \
          --build "${BUILD_DIR}"
 
 echo " ========= OpenMP Benchmarks ==========="
--- a/scripts/benchmarks/build_and_run.sh
+++ b/scripts/benchmarks/build_and_run.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+#
+# This script is meant to be used in a new machine, to build LLVM, TPP-MLIR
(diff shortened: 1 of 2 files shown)


rengolin:
The benchmark script looks ok, works on my Arch box, but not on my Ubuntu server box, which I'm trying to get in sync with AWS. the OpenMP problem is what's creating all this. I'll see if I can get an older Ubuntu if the problem persists.

rengolin:
> The benchmark script looks ok, works on my Arch box, but not on my Ubuntu server box, which I'm trying to get in sync with AWS. the OpenMP problem is what's creating all this. I'll see if I can get an older Ubuntu if the problem persists.

Ubuntu Server 23.10 is completely broken, I got 22.04 LTS and it seems better. Still some issues with symbols for some benchmarks, not all, but almost there.

rengolin:
> Ubuntu Server 23.10 is completely broken, I got 22.04 LTS and it seems better. Still some issues with symbols for some benchmarks, not all, but almost there.

all good, got a full run on my ubuntu server similar to aws

scripts/benchmarks/build_and_run.sh
```diff
@@ -0,0 +1,73 @@
...
+# and run all benchmarks. This should work in the same way as our local tests
+# and reproduce our numbers on local/cloud machines.
+
+# Include common utils
+SCRIPT_DIR=$(realpath $(dirname $0)/..)
+source ${SCRIPT_DIR}/ci/common.sh
+
+# Install packages needed
+if [ "$(is_linux_distro Ubuntu)" == "YES" ]; then
+  sudo apt update && \
```
alheinecke, line 13: I suggest using conda, this is much more portable.

rengolin: Conda faffles with `.bashrc` and needs reloading stuff. I found the `dnf` install line for "Amazon Linux" that just worked on different instances on AWS.

PR #888, opened 2024-02-22
GPU launch constants inlining

adam-smnk:
GPU launch constants inlining

Adds a pass that inlines constants into gpu.launch body.

This reduces the number of GPU kernel parameters after kernel outlining and allows for further constant propagation within the kernel.


lib/TPP/GPU/GpuInlineConstants.cpp
```diff
@@ -0,0 +1,93 @@
...
+namespace tpp {
+#define GEN_PASS_DEF_GPUINLINECONSTANTS
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+namespace {
+
+// Returns true if the operation represents a constant value.
+static bool isConstantOp(Operation *op) {
```
rengolin, line 35: is this really needed?

lib/TPP/GPU/GpuInlineConstants.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // Clone the constants into the gpu.launch body.
+    OpBuilder::InsertionGuard guard(rewriter);
+    rewriter.setInsertionPointToStart(&launchOpBody.front());
+
+    for (auto *op : constantOps) {
+      auto *clonedOp = rewriter.clone(*op);
+
+      // Replace uses withing the body with the inlines values.
```
rengolin, line 67:
```suggestion
      // Replace uses within the body with the inlined values.
```

rengolin (approved):
Small nits, but LGTM, thanks!

lib/TPP/GPU/GpuInlineConstants.cpp
adam-smnk: In the end, nothing else is needed so, removed the helper.

PR #889, opened 2024-02-25
Benchmark script

rengolin:
Working on multiple AWS nodes and my Ubuntu box. We can add conda support later, for now, I just want to get it running on AWS to avoid manual steps taking extra time.


alheinecke (approved):
LGTM, should support now Ubuntu + clones, AWS + RH based. SuSE and ArchLinux, can be later.

PR #890, opened 2024-02-26
Tile configuration addition pass

KavithaTipturMadhu:
Test cases to be added, draft PR in progress. Runtime and code changes are done, licm hoisting is incomplete, working on it.

### DIFF
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -263,6 +263,15 @@ def Xsmm_GemmDispatchOp : Xsmm_GemmLikeOp<"gemm.dispatch"> {
 def Xsmm_BrgemmDispatchOp : Xsmm_GemmLikeOp<"brgemm.dispatch"> {
   let summary = "dispatch for brgemm operation.";
   let hasVerifier = 1;
+  
+}
+
+//===----------------------------------------------------------------------===//
+// TileConfigDispatchOp
+//===----------------------------------------------------------------------===//
+
+def Xsmm_TileConfigDispatchOp : Xsmm_GemmLikeOp<"tileConfig.dispatch"> {
+  let summary = "dispatch for tileConfig operation.";
 }
 
 //===----------------------------------------------------------------------===//
@@ -296,6 +305,16 @@ def Xsmm_FusedBrgemmDispatchOp : Xsmm_Op<"fused_brgemm.dispatch", [Pure]> {
   let hasCustomAssemblyFormat = 1;
 
   let hasVerifier = 1;
+}
+
+
+//===----------------------------------------------------------------------===//
+// TileConfigOp
+//===----------------------------------------------------------------------===//
+
+def Xsmm_TileConfigOp : Xsmm_Op<"tileConfig",  [MemoryEffects<[MemWrite, MemRead]>]> {
+  let summary = "invoke for tileConfig operation.";
+  let arguments = (ins I64:$dispatch, Variadic<AnyMemRef>:$inputs);
 }
 
 #endif // TPP_XSMM_OPS
--- a/include/TPP/Dialect/Xsmm/XsmmOps.td
+++ b/include/TPP/Dialect/Xsmm/XsmmOps.td
@@ -267,11 +267,11 @@ def Xsmm_BrgemmDispatchOp : Xsmm_GemmLikeOp<"brgemm.dispatch"> {
 }
 
 //===----------------------------------------------------------------------===//
-// TileConfigDispatchOp
+// IntelAMXTileConfigDispatchOp
 //===----------------------------------------------------------------------===//
 
-def Xsmm_TileConfigDispatchOp : Xsmm_GemmLikeOp<"tileConfig.dispatch"> {
-  let summary = "dispatch for tileConfig operation.";
+def Xsmm_IntelAMXTileConfigDispatchOp : Xsmm_GemmLikeOp<"IntelAMXtileConfig.dispatch"> {
+  let summary = "dispatch for Intel amx tileConfig operation.";
 }
 
 //===----------------------------------------------------------------------===//
@@ -309,11 +309,11 @@ def Xsmm_FusedBrgemmDispatchOp : Xsmm_Op<"fused_brgemm.dispatch", [Pure]> {
 
 
 //===----------------------------------------------------------------------===//
-// TileConfigOp
+// IntelAMXTileConfigOp
 //===----------------------------------------------------------------------===//
 
-def Xsmm_TileConfigOp : Xsmm_Op<"tileConfig",  [MemoryEffects<[MemWrite, MemRead]>]> {
-  let summary = "invoke for tileConfig operation.";
+def Xsmm_IntelAMXTileConfigOp : Xsmm_Op<"IntelAMXtileConfig",  [MemoryEffects<[MemWrite, MemRead]>]> {
+  let summary = "invoke for Intel AMX tileConfig operation.";
   let arguments = (ins I64:$dispatch, Variadic<AnyMemRef>:$inputs);
 }
 
--- a/lib/TPP/Transforms/IntelAMXTileConfigHoisting.cpp
+++ b/lib/TPP/Transforms/IntelAMXTileConfigHoisting.cpp
@@ -1,4 +1,4 @@
-//===- TileConfigHoisting.cpp ---------------------------------------------===//
+//===- IntelAMXTileConfigHoisting.cpp -------------------------------------===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
@@ -10,16 +10,14 @@
 //
 //===----------------------------------------------------------------------===//
 #include "TPP/Dialect/Xsmm/XsmmOps.h"
-#include "TPP/Dialect/Xsmm/XsmmUtils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
-#include "mlir/Dialect/SCF/Transforms/Transforms.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 namespace mlir {
 namespace tpp {
-#define GEN_PASS_DEF_TILECONFIGHOISTINGPASS
+#define GEN_PASS_DEF_INTELAMXTILECONFIGHOISTINGPASS
 #include "TPP/Passes.h.inc"
 } // namespace tpp
 } // namespace mlir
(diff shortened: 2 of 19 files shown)


lib/TPP/Transforms/TileConfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+namespace mlir {
+namespace tpp {
+
+template <typename InvokeOpTy, typename DispatchOpTy>
+static void appendBrgemmFlags(SmallVector<Attribute> &attributes,
+                              PatternRewriter &rewriter, InvokeOpTy opTy) {
+  auto flags =
+      dyn_cast<DispatchOpTy>(opTy.getOperand(0).getDefiningOp()).getFlags();
+  for (auto flagItr : flags) {
```
rengolin, line 39: We have this same pattern twice in the code already (Combine XSMM pass and Linalg to XSMM). 

They're not identical but easily convertible. We can do this in a later PR, but please add a TODO here so that we don't forget, and please do this on a following PR ASAP.

lib/TPP/Transforms/TileConfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+    auto tileConfigSetup = rewriter.create<xsmm::TileConfigDispatchOp>(
+        op.getLoc(), rewriter.getI64Type(),
+        DenseI64ArrayAttr::get(
+            rewriter.getContext(),
+            dyn_cast<DispatchOpTy>(op.getOperand(0).getDefiningOp())
+                .getInputs()),
+        rewriter.getArrayAttr(attributesSetup),
+        xsmm::utils::getDataType(rewriter, op.getOperand(1).getType()));
+
+    SmallVector<Attribute> attributesReset;
```
rengolin, line 87: Why are you adding both here, all to the same `op`?

lib/TPP/Transforms/TileConfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+                .getInputs()),
+        rewriter.getArrayAttr(attributesReset),
+        xsmm::utils::getDataType(rewriter, op.getOperand(1).getType()));
+
+    SmallVector<Attribute> attributesBrgemm;
+    attributesBrgemm.push_back(xsmm::GemmFlagsAttr::get(
+        rewriter.getContext(), xsmm::GemmFlags::NO_RESET_TILECONFIG));
+    attributesBrgemm.push_back(xsmm::GemmFlagsAttr::get(
+        rewriter.getContext(), xsmm::GemmFlags::NO_SETUP_TILECONFIG));
+    appendBrgemmFlags<InvokeOpTy, DispatchOpTy>(attributesBrgemm, rewriter, op);
```
rengolin, line 105: I don't understand, you've just added (independently) the two flags above, why add the pair here?

lib/TPP/Transforms/TileConfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+    
+    ValueRange tileConfigInputs{alloca};
+    rewriter.create<mlir::xsmm::TileConfigOp>(
+        op.getLoc(), tileConfigSetup, tileConfigInputs);
+    
+    SmallVector<Value> invokeOperands;
+    invokeOperands.push_back(dispatch);
+    auto opItr = op->getOperands().begin();
+    std::advance(opItr, 1);
+    invokeOperands.append(opItr, op->getOperands().end());
```
rengolin, line 122: Do you need to re-create the original GEMM/BRGEMM op? They're independent of the tile config, you can just use the rewriter's ability to change the insertion point (before/after) and add each tile config invoke.

rengolin (commented):
You really need tests.

include/TPP/Dialect/Xsmm/XsmmOps.td
```diff
@@ -263,6 +263,15 @@ def Xsmm_GemmDispatchOp : Xsmm_GemmLikeOp<"gemm.dispatch"> {
...
   let summary = "dispatch for brgemm operation.";
   let hasVerifier = 1;
+  
+}
+
+//===----------------------------------------------------------------------===//
+// TileConfigDispatchOp
+//===----------------------------------------------------------------------===//
+
+def Xsmm_TileConfigDispatchOp : Xsmm_GemmLikeOp<"tileConfig.dispatch"> {
```
adam-smnk, line 273: It'd be best to have a separate base class or generalize the `Xsmm_GemmLikeOp` if it fits here.

lib/TPP/Transforms/TileConfig.cpp
```diff
@@ -0,0 +1,154 @@
+//
```
adam-smnk, line 1: Missing header with file name.

lib/TPP/Transforms/TileConfig.cpp
```diff
@@ -0,0 +1,154 @@
...
+      return failure();
+    auto flags =
+        dyn_cast<DispatchOpTy>(op.getOperand(0).getDefiningOp()).getFlags();
+    for (auto flagItr : flags) {
+      if (flagItr == xsmm::GemmFlagsAttr::get(
+                         rewriter.getContext(),
+                         mlir::xsmm::GemmFlags::NO_RESET_TILECONFIG) ||
+          flagItr == xsmm::GemmFlagsAttr::get(
+                         rewriter.getContext(),
+                         mlir::xsmm::GemmFlags::NO_SETUP_TILECONFIG)) {
```
adam-smnk, line 69: Do these flag have some meaning for xsmm calls or is the only use for them just pattern convergence?
If it's the latter, then op annotation would be simpler. Although, in this case perhaps `PatternRewriter` is not the best tool overall but the pass could just once walk over all xsmm ops in a function.

lib/TPP/Transforms/TileConfig.cpp
KavithaTipturMadhu: The NO_RESET_TILECONFIG flag is added, followed by the rest of the flags of the `op` appended via the `appendBrgemmFlags` call.

lib/TPP/Transforms/TileConfig.cpp
rengolin: It's a bit convoluted but they do. 

Basically, for correctness issues, *every* GEMM-like op carries the flags `SETUP` and `RESET`. Otherwise, you run the risk of running an uninitialized AMX instruction. 

These ones are saying "don't setup" (ie. `RESET`) or "don't reset" (ie. `SETUP`), and they're executed before/after the outer loops (OMP thread). The inner loop's GEMM now has *both* `NO SETUP` and `NO RESET`, to make sure it does not incurr in the penalty of setting up AMX unnecessarily. This is the heart of the optimization.

The current code in the PR is hoping that LICM will hoist the setup/reset invokes into the outer parallel loop, but it doesn't, because it's not side-effect free (there's an alloca in there). So we agreed to move the ops in the right place (between `scf.parallel` and the first `scf.for`) by hand for now.

As why they're flags, because libxsmm was modified to work that way. May not be the best API but it is what it is now. We can change that later (and I'd prefer a simpler API too).

lib/TPP/Transforms/TileConfig.cpp
KavithaTipturMadhu: `appendBrgemmFlags` attributes are used in dispatch operation, whereas the other two attribute sets are used in tileconfigdispatch operations.

lib/TPP/Transforms/TileConfig.cpp
KavithaTipturMadhu: I could just modify the existing op's dispatch, yes. I'll update the code.

lib/TPP/Transforms/TileConfig.cpp
alheinecke: @rengolin : this is just as we break API for performance, hence we need in total 3 ops and to not set it all time, it's negated.

lib/TPP/Transforms/TileConfig.cpp
alheinecke: > They're independent of the tile config, you can just use the rewriter's ability to change the insertion point (before/after) and add each tile config invoke.

No! The tileconfig/tilerreset need to be around the BRGEMM. Hence, I think fiddling with inserting points is brittle, just let's rewrite and replace.

lib/TPP/Transforms/TileConfig.cpp
alheinecke: This API is breaking the ABI. Hence it's negated. 

I still think it's a bug / too conservative that LICM is not moving the stuff. the alloca has nothing to do with the loops nest. We can proof that the memory we are allocating/accessing has nothing to do with induction variables and that the call we are making doesn't consider the induction variables.

include/TPP/Dialect/Xsmm/XsmmOps.td
adam-smnk: Without any libxsmm API knowledge, tile config sounds generic. So, does it configure/affect other operations too?
If not, then maybe GEMM (tile) config would be more accurate and at least that would fit the current inheritance scheme.

I just don't wanna further the spaghetti like the fused brgemm which already is not a GemmLikeOp for whatever reason.
Overall, I'd need better op documentation or at least to see some IR example to have any concrete suggestions.

rengolin: It is a 1-to-1 copy of the libxsmm API. Once we improve the API, we can change it here

lib/TPP/Transforms/TileConfigHoisting.cpp
```diff
@@ -0,0 +1,101 @@
...
+                                  rewriter.getContext(),
+                                  mlir::xsmm::GemmFlags::NO_SETUP_TILECONFIG)) {
+          secondTileConfig = dyn_cast<xsmm::TileConfigOp>(user);
+        }
+      }
+    }
+
+    scf::ParallelOp parallelOpParent = NULL;
+    auto op = alloca.getOperation();
+    while (true) {
```
KavithaTipturMadhu, line 64: @alheinecke @rengolin suggested that I flatten this into a two level check instead of a loop, however, as we discussed, the scf fors could be nested deeper than 1 level. should we keep this loop?

alheinecke: We don't have any guarantee that we have only 2 scf.for. We have to push up and push down until we reach the scf.parallel. Worst case I can see up to 7 scf.for (for full cache blocking), so there will be any number of 0 to 7 scf.for.

alheinecke:
@KavithaTipturMadhu On a second thought on the sames and what might have created also confusion on @adam-smnk side: l suggest renaming TileConfig to IntelAMXTileConfig in both pass names, etc. Then it;s clear this has nothing to do with tiling in general but with Intel AMX in particular.

rengolin (approved):

lib/TPP/Transforms/IntelAMXTileConfigHoisting.cpp
```diff
@@ -0,0 +1,101 @@
...
+} // namespace tpp
+} // namespace mlir
+
+using namespace mlir;
+using namespace mlir::xsmm;
+
+namespace mlir {
+namespace tpp {
+
+struct IntelAMXTileConfigHoisting : OpRewritePattern<memref::AllocaOp> {
```
adam-smnk, line 31: nit: seems a bit brittle to anchor it on alloca, I think it'd be safer to target `IntelAMXTileConfigOp`

alheinecke: I agree with both of you. This is not critical and not high prio. I have opened #896. Let's discuss a good solution over there and we track the open there.

PR #901, opened 2024-03-11
[mlir-gen] Add Float16 type support

adam-smnk:
Allows IR generation with f16 data type.

The new type is mostly relevant for GPU testing as f16 is a common target for hardware accelerated operations.

Required changes are propagated around as extra differentiation is now needed to select between f16 and bf16.

### DIFF
--- a/tools/mlir-gen/MLIRGen.cpp
+++ b/tools/mlir-gen/MLIRGen.cpp
@@ -19,6 +19,8 @@
 #include "MLIRGen.h"
 #include "llvm/Support/ErrorHandling.h"
 
+#include <optional>
+
 using namespace mlir;
 
 namespace {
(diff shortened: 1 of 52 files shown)


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -340,7 +338,7 @@ Value MLIRGenerator::lowerRelu(Value input, Value output) {
   if (!enableRelu)
     return input;
 
-  auto zero = getConstFloat(builder, 0.0, dataType.getIntOrFloatBitWidth());
+  auto zero = getConstFloat(builder, 0.0, dataType.cast<FloatType>());
```
rengolin, line 341: Isn't this always going to create the constant at `f32`? Won't that cause casts to appear?

tools/mlir-gen/generate.sh
```diff
@@ -126,7 +126,7 @@ debug "Random seed: $SEED"
 # Defaults
 
 # Command line to extract and run
-MLIR_GEN_ARGS="--float-width=$FLOAT_SIZE --seed=$SEED --batch=$MB --layers=$LAYER_SIZES --tiles=$TILE_SIZES"
+MLIR_GEN_ARGS="--float-type=$FLOAT_SIZE --seed=$SEED --batch=$MB --layers=$LAYER_SIZES --tiles=$TILE_SIZES"
```
rengolin, line 129: nit: Also change `$FLOAT_SIZE`

adam-smnk: Thanks, I completely missed that one.

tools/mlir-gen/MLIRGen.cpp
adam-smnk: No, `FloatType` carries the info about the specific float type f8, f16, f32, bf16 etc.
That is correctly propagated and the created constant is of the specified `dataType`

rengolin (approved):

PR #903, opened 2024-03-28
Allow to build TPP out-of-tree

Devjiu:
This commit allows to build tpp out-of-tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,21 +15,33 @@ set(CMAKE_CXX_STANDARD 17 CACHE STRING "C++ standard to conform to")
 set(TPP_GPU "" CACHE STRING "Enables GPU runtime (default: '')")
 set_property(CACHE TPP_GPU PROPERTY STRINGS "" "cuda" "vulkan")
 
-find_package(MLIR REQUIRED CONFIG)
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR OR TORCH_MLIR_OUT_OF_TREE_BUILD)
+  message(STATUS "TPP-MLIR out-of-tree build.")
 
-set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/bin)
-set(LLVM_LIBRARY_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/lib)
+  find_package(MLIR REQUIRED CONFIG)
+  message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
+  message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+
+  set(LLVM_RUNTIME_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/bin)
+  set(LLVM_LIBRARY_OUTPUT_INTDIR ${CMAKE_BINARY_DIR}/lib)
+
+  list(APPEND CMAKE_MODULE_PATH "${MLIR_CMAKE_DIR}")
+  list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
+  include(TableGen)
+  include(AddLLVM)
+  include(AddMLIR)
+  include(HandleLLVMOptions)
+else()
+  message(STATUS "TPP-MLIR in-tree build.")
+
+  set(MLIR_MAIN_SRC_DIR ${LLVM_MAIN_SRC_DIR}/../mlir)
+  set(MLIR_INCLUDE_DIR ${LLVM_MAIN_SRC_DIR}/../mlir/include)
+  set(MLIR_GENERATED_INCLUDE_DIR ${LLVM_BINARY_DIR}/tools/mlir/include)
+  set(MLIR_INCLUDE_DIRS "${MLIR_INCLUDE_DIR};${MLIR_GENERATED_INCLUDE_DIR}")
+endif()
+
 set(MLIR_BINARY_DIR ${CMAKE_BINARY_DIR})
-message(STATUS "Using MLIRConfig.cmake in: ${MLIR_DIR}")
-message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
-list(APPEND CMAKE_MODULE_PATH "${MLIR_CMAKE_DIR}")
-list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
-
-include(TableGen)
-include(AddLLVM)
-include(AddMLIR)
-include(HandleLLVMOptions)
 
 set(TPP_MAIN_INCLUDE_DIR "${PROJECT_SOURCE_DIR}/include")
 set(TPP_GEN_INCLUDE_DIR "${PROJECT_BINARY_DIR}/include")
@@ -140,4 +152,4 @@ add_custom_target(check-clang  ${CI_DIR}/lint-clang.sh)
 add_custom_target(check-format DEPENDS check-clang check-python)
 
 # Check-all is just check, for now
-add_custom_target(check-all DEPENDS check)
+add_custom_target(tpp-check-all DEPENDS tpp-check)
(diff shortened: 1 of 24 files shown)


Devjiu:
> Sorry for the delay. Long weeks behind. Some errors on CI indicate build issues.
> 
> Did you try to build locally? Any change to the project needs to allow it to be built both as a sub-module and stand alone, like CI. Can you make sure it builds both ways?

NP, I checked standalone build, build with submodule also should be fine, but currently it's a bit difficult to check as currently used different llvm versions for torch-mlir and tpp.

CMakeLists.txt
```diff
@@ -140,4 +152,4 @@ add_custom_target(check-clang  ${CI_DIR}/lint-clang.sh)
 add_custom_target(check-format DEPENDS check-clang check-python)
 
 # Check-all is just check, for now
-add_custom_target(check-all DEPENDS check)
+add_custom_target(tpp-check-all DEPENDS tpp-check)
```
rengolin, line 155: As observed below, better to just remove this target and make the test one `check-tpp`.

include/TPP/Passes.td
```diff
@@ -388,7 +388,7 @@ def DefaultPipeline : Pass<"default-pipeline", "ModuleOp"> {
   ];
 }
 
-def SetSPIRVCapabilities : Pass<"set-spirv-capabilities", "ModuleOp"> {
+def SetSPIRVCapabilities : Pass<"tpp-set-spirv-capabilities", "ModuleOp"> {
```
rengolin, line 391: @adam-smnk is there a better name for this?

cmake/modules/xsmm.cmake
```diff
@@ -32,12 +32,12 @@ endif()
...
-target_include_directories(xsmm PUBLIC ${XSMM_INCLUDE_DIRS})
-target_compile_definitions(xsmm PRIVATE
-  __BLAS=0
+add_mlir_library(xsmm STATIC ${XSMM_SRCS})
+target_include_directories(xsmm PUBLIC
+  $<BUILD_INTERFACE:${XSMM_INCLUDE_DIRS}>
+  $<INSTALL_INTERFACE:include/xsmm>
 )
-add_definitions(-DLIBXSMM_DEFAULT_CONFIG -U_DEBUG)
+add_definitions(-DLIBXSMM_DEFAULT_CONFIG -U_DEBUG -D__BLAS=0)
```
rengolin, line 40: @alheinecke Please check this change.

lib/TPP/CMakeLists.txt
```diff
@@ -19,6 +19,21 @@ add_mlir_library(TPPPipeline
 
   LINK_LIBS PUBLIC
     MLIRIR
+    MLIRMeshDialect
```
rengolin, line 22: That's a whole lotta libraries we don't use. Why add them here?

lib/TPP/Transforms/CMakeLists.txt
```diff
@@ -34,8 +34,23 @@ add_mlir_library(TPPTransforms
...
+    MLIRSupport
     MLIRBufferizationToMemRef
     MLIRBufferizationPipelines
+    # TPPTppDialect
     TPPTransformsUtils
     TPPIR
+    TPPXsmmDialect
+    TPPCheckToLoops
+    TPPLinalgToFunc
+    # TPPLinalgToTpp
```
rengolin, line 47: Don't leave commented-out lines.

test/CMakeLists.txt
```diff
@@ -14,10 +14,10 @@ set(TPP_OPT_TEST_DEPENDS
         tpp-run
         )
 
-add_lit_testsuite(check "Running the regression tests"
+add_lit_testsuite(tpp-check "Running the regression tests"
```
rengolin, line 17: We used to have formatting checks into `check-all` but now we don't anymore, so `check` can become `check-tpp`. We don't need a check-all any more.

Please, just remove the last line in the main CMake file:
```
# Check-all is just check, for now
add_custom_target(check-all DEPENDS check)
```

and rename this target to `check-tpp` instead. Then change the build scripts to just call `check-tpp`.

rengolin (commented):
Thanks @Devjiu! This is looking really nice!

I have a few comments on targets, but nothing major. I've asked a few checks from Alex and Adam and it should be good to go.

include/TPP/Passes.td
adam-smnk: I think it's fine as it's a common name for this kind of pass

@Devjiu could you also update the corresponding tests?

rengolin: Oh, these tests weren't executed because our CI only runs on CPU, so no failures.

lib/TPP/CMakeLists.txt
Devjiu: it's from registerAllDialects in DefaultPipeline.cpp
Without this many linking errors generated for FAILED: bin/torch-mlir-opt
Like this: 
```
GpuPipeline.cpp:(.text._ZN4mlir15DialectRegistry6insertINS_4math11MathDialectENS_6memref13MemRefDialectEJNS_4mesh11MeshDialectENS_10ml_program16MLProgramDialectENS_3mpi10MPIDialectENS_5nvgpu12NVGPUDialectENS_4NVVM11NVVMDialectENS_3omp13OpenMPDialectENS_3pdl10PDLDialectENS_10pdl_interp16PDLInterpDialectENS_5quant19QuantizationDialectENS_5ROCDL12ROCDLDialectENS_3scf10SCFDialectENS_5shape12ShapeDialectENS_13sparse_tensor19SparseTensorDialectENS_5spirv12SPIRVDialectENS_6tensor13TensorDialectENS_4tosa11TosaDialectENS_9transform16TransformDialectENS_2ub9UBDialectENS_6vector13VectorDialectENS_9x86vector16X86VectorDialectENS_5xegpu12XeGPUDialectEEEEvv[_ZN4mlir15DialectRegistry6insertINS_4math11MathDialectENS_6memref13MemRefDialectEJNS_4mesh11MeshDialectENS_10ml_program16MLProgramDialectENS_3mpi10MPIDialectENS_5nvgpu12NVGPUDialectENS_4NVVM11NVVMDialectENS_3omp13OpenMPDialectENS_3pdl10PDLDialectENS_10pdl_interp16PDLInterpDialectENS_5quant19QuantizationDialectENS_5ROCDL12ROCDLDialectENS_3scf10SCFDialectENS_5shape12ShapeDialectENS_13sparse_tensor19SparseTensorDialectENS_5spirv12SPIRVDialectENS_6tensor13TensorDialectENS_4tosa11TosaDialectENS_9transform16TransformDialectENS_2ub9UBDialectENS_6vector13VectorDialectENS_9x86vector16X86VectorDialectENS_5xegpu12XeGPUDialectEEEEvv]+0x1b9): undefined reference to `mlir::detail::TypeIDResolver<mlir::mpi::MPIDialect, void>::id'
```
or in shorten:
```
GpuPipeline.cpp: undefined reference to `mlir::detail::TypeIDResolver<mlir::mpi::MPIDialect, void>::id'
```

So this target uses these libraries, I don't know why in standalone mode this problem doesn't occur.
I am not sure how to fix them it in different way.

lib/TPP/Transforms/CMakeLists.txt
Devjiu: yes, sorry

lib/TPP/CMakeLists.txt
Devjiu: I found a slightly better solution and used ${mlir_dialect_libs}. I don't know why I only noticed this variable now.

include/TPP/Passes.td
Devjiu: I found it in tpp-mlir/test/GPU/Vulkan/set-spirv-capabilities.mlir. Updated but not verified. Do I have to run the lit tests manually to run this test?

test/CMakeLists.txt
Devjiu: done

include/TPP/Passes.td
adam-smnk: GPU tests are disabled by default as that requires extra dependencies. And we don't have CI setup for Vulkan atm.
But it's fine, I'll rerun it locally.

rengolin: Thanks! @adam-smnk please resolve this conversation once the GPU tests pass.

Devjiu: Thanks)

cmake/modules/xsmm.cmake
rengolin: My reading is that this disables the BLAS interface in libxsmm. We don't want use this internally, so I guess it's safe to keep it like that for now. In the future, if we have conflicts with other users, we'll need to add an options to enable|disable it from the CMake side.

For now, let's keep it like that. Closing the comment. Thanks!

include/TPP/Passes.td
adam-smnk: Works fine, thanks 👍

adam-smnk (approved):
Looks pretty good, works both in CI and locally
Thanks +1

PR #905, opened 2024-04-09
MLIR Bench pass

adam-smnk:
Refactor MLIR Bench infrastructure into a pass.

The core logic is already covered by tpp-run tests.


adam-smnk:
For TPP+IMEX testing, I settled on tpp-run but running tpp-opt with desired part of the pipeline is overall cleaner.
Also, creating runner wrapper is pretty useful and best done before any other transformations. So, this refactor exposes kernel wrapper creation to tpp-opt too.

adam-smnk:
> Do you aim to upstream this?

No, I don't think it's generic enough for that. It stays internal for now.

> If the latter, why change?

Mostly for future integration with external tools. Change is motivated by the current use case  of TPP -> IMEX pipeline:
`tpp-run INPUT FLAGS | sed '/Error:.*/d' | IMEX`.
The error is produced because we try to run an incomplete lowering but I want to use `tpp-run` for simplicity.

This separate runner wrapper pass allows to retain our handy wrappers and avoid silly errors when terminating lowering pipeline early. With this PR, the following is possible:
`tpp-opt -mlir-bench -default-pipeline | EXTERNAL CONSUMER`

It might or might not be useful in the future. But the worst case, this change simplifies our runner a bit by moving the wrapper into a separate logic (pass).

Perhaps, we can give it a nicer name.

rengolin (approved):
I see. It's fine as is. If we want to upstream later, we can clean up.

adam-smnk:
NFC rebase on main + renamed to `tpp-runner-wrapper`

PR #906, opened 2024-04-24
Retire method style casts

adam-smnk:
Aligns with MLIR style guideline on casting style.
The method style casts are now deprecated.


adam-smnk:
A small prep as I'll need to bump LLVM soon.

rengolin (approved):

PR #908, opened 2024-04-25
Remove LLVM build lock on failure

adam-smnk:
Removes install dir, that acts as a synchronization point to prevent parallel builds of the same LLVM, on failure.

This should allow LLVM builds to be rerun after early exits in case of errors.


adam-smnk:
That should make next LLVM bump smoother. Time will tell ;)

rengolin (approved):

PR #909, opened 2024-05-07
Remove ageing transform dialect extensions

rengolin:
Our own transforms are now being used as passes and the transform ops are unmaintained for a while and not helpful any more. We plan to restart working on transforms later on, but we'll certainly approach it form an upstream-first mindset, so likely not to use the custom ops we created.

However, they have served their purpose. They allowed us to experiment with IR at a time when we didn't have such a mature dialect / pass environment.

For those wanting to get these back, they can just revert this commit on a branch and work with the code locally, but I strongly suggest to only use these as a baseline, not as forward looking code.


adam-smnk (approved):
Agreed on the strategy - transforms need clean reevaluation in the future

PR #910, opened 2024-05-13
Bump LLVM

adam-smnk:
Updates LLVM version, fixes more deprecated casts, and updates IR syntax.


rengolin (approved):
@KavithaTipturMadhu this brings changes to `expand_shape` and reassociation maps, you'll need to rebase for your current work.

PR #911, opened 2024-05-14
Use upstream matmul pack

adam-smnk:
Retires downstream matmul packing logic and uses upstream block pack matmul instead.

### DIFF
--- a/test/Passes/DefaultPipeline/default-tpp-passes.mlir
+++ b/test/Passes/DefaultPipeline/default-tpp-passes.mlir
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -default-tpp-passes -split-input-file | FileCheck %s
+// RUN: tpp-opt %s -default-tpp-passes -split-input-file 2>/dev/null | FileCheck %s
 
 // CHECK: func.func @matmul(
 // CHECK-SAME:  %[[ARG0:.+]]: memref<4x8xf32>,
--- a/include/TPP/Transforms/Transforms.h
+++ b/include/TPP/Transforms/Transforms.h
@@ -51,14 +51,6 @@ FailureOr<linalg::GenericOp>
 packConv2DNhwcHwcfOp(RewriterBase &rewriter, linalg::Conv2DNhwcHwcfOp linalgOp,
                      ArrayRef<OpFoldResult> tiles);
 
-// Attempt to block a MatmulOp or a BatchMatmulOp.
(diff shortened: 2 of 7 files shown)


adam-smnk:
Identity permutation `outer_dims_perm = [0, 1]` in `tensor.pack` introduced by the upstream matmul packing prevents folding pairs of `tensor.unpack` and `tensor.pack`. Which then prevents complete layout propagation throughout IR graph.

Investigation in progress.

adam-smnk:
Fixed upstream. This PR should be able to continue after the next LLVM bump.

adam-smnk:
All works as before now. No regression in performance.

test/Passes/DefaultPipeline/default-tpp-passes.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tpp-opt %s -default-tpp-passes -split-input-file | FileCheck %s
+// RUN: tpp-opt %s -default-tpp-passes -split-input-file 2>/dev/null | FileCheck %s
```
rengolin, line 1: Why is this necessary?

rengolin (approved):

test/Passes/DefaultPipeline/default-tpp-passes.mlir
adam-smnk: The `linalg::detail::verifyContractionInterface` used in upstream generic matcher generates tons of diagnostics straight into stderr and makes output unusable.
I'll have to take a look at it upstream.

rengolin: This sounds like a bug. We don't want debug output unless explicitly requested. Diagnostics usually are enabled via some debug flag, maybe the macro is broken upstream?

adam-smnk: It just straight up dumps info through `op->emitError`. It's a questionable design I think but the call in question could be replaced with `isContractionInterfaceImpl` to avoid the issue.

rengolin: That's not right, we can fix it upstream.

adam-smnk: Opened PR upstream with a fix.

adam-smnk: I'll land this for now and try to remember to clean it up after the next LLVM bump.

PR #915, opened 2024-05-23
Linalg to XeGPU lowering

adam-smnk:
Adds direct lowering from Linalg to XeGPU and extends GPU runner support with 'intel' target.
The Intel GPU pipeline is designed with IMEX and IGC compatibility in mind.

The lowering targets tiled operations and assumes that the input shapes are nicely divisible by hardware supported sizes e.g., tiles 32x32, 16x16 etc.
This is the first step toward bridging XeGPU with higher abstraction dialect. Common patterns used in this conversion can be later split into more progressive lowering through other dialects like vector and memref.

Supported conversion:
- targets Vector Compute mode of XeGPU (subgroup-level kernel)
- eltwise operations of any type split into SIMD sized computations
- DPAS implementation for F16 matmul with output precision conversion

### DIFF
--- a/lib/TPP/GPU/LinalgToXeGPU.cpp
+++ b/lib/TPP/GPU/LinalgToXeGPU.cpp
@@ -860,8 +860,8 @@ static LogicalResult createDPASKernel(linalg::LinalgOp linalgOp,
       VectorType::get(dpasTypeC.getShape(), FloatType::getF32(ctx));
 
   // Extend the accumulation values if needed.
-  auto isOutF16 = typeC.getElementType().isF16();
-  if (isOutF16) {
+  auto convOutPrecision = !typeC.getElementType().isF32();
+  if (convOutPrecision) {
     for (size_t i = 0; i < loadVecC.size(); i++) {
       auto extOp =
           rewriter.create<arith::ExtFOp>(loc, dpasResType, loadVecC[i]);
@@ -1013,8 +1013,13 @@ static LogicalResult createDPASKernel(linalg::LinalgOp linalgOp,
       },
       /*elseBuilder=*/nullptr);
 
-  // TODO: Make the VNNI factor a flexible parameter.
-  const int vnniFactor = 2;
+  // TODO: Add more possible types.
+  int vnniFactor = TypeSwitch<Type, int>(typeA.getElementType())
+                       .Case([](Float16Type type) { return 2; })
+                       .Default([](Type type) { return -1; });
+  if (vnniFactor == -1)
+    return failure();
+
   VnniConfig vnniConfA{.vnniFactor = vnniFactor, .vnniAxis = 1};
   VnniConfig vnniConfB{.vnniFactor = vnniFactor, .vnniAxis = 0};
 
@@ -1114,9 +1119,9 @@ static LogicalResult createDPASKernel(linalg::LinalgOp linalgOp,
   }
 
   // Truncate the result values if needed.
-  if (isOutF16) {
+  if (convOutPrecision) {
     auto truncType =
-        VectorType::get(dpasTypeC.getShape(), FloatType::getF16(ctx));
+        VectorType::get(dpasTypeC.getShape(), typeC.getElementType());
     for (size_t i = 0; i < results.size(); i++) {
       auto truncOp =
           rewriter.create<arith::TruncFOp>(loc, truncType, results[i]);
--- a/lib/TPP/GPU/LinalgToXeGPU.cpp
+++ b/lib/TPP/GPU/LinalgToXeGPU.cpp
@@ -1,0 +1,1567 @@
+//===- LinalgToXeGPU.cpp -----------------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Passes.h"
+
+#include "TPP/IR/MatcherUtils.h"
+#include "TPP/IR/StructuredOpMatcher.h"
+#include "TPP/Transforms/Utils/ValueUtils.h"
+
+#include "mlir/Conversion/Passes.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/GPU/TransformOps/Utils.h"
+#include "mlir/Dialect/GPU/Transforms/Passes.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Passes.h"
+#include "mlir/Dialect/Math/IR/Math.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/Dialect/XeGPU/IR/XeGPU.h"
+#include "mlir/IR/Dialect.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/Passes.h"
+
+#include "llvm/ADT/TypeSwitch.h"
+
+#include <numeric>
+#include <optional>
+
+using namespace mlir;
+using namespace mlir::tpp;
+using namespace mlir::xegpu;
+
+namespace mlir {
+namespace tpp {
+#define GEN_PASS_DEF_LINALGTOXEGPU
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+namespace {
+
+struct VnniConfig {
+  int vnniFactor;
+  int vnniAxis;
+};
+
+struct TilesArray {
+  TilesArray() = delete;
+  TilesArray(int numRows, int numCols) {
+    for (int i = 0; i < numRows; i++) {
+      tileMatrix.push_back(SmallVector<Value>{});
+      for (int j = 0; j < numCols; j++)
+        tileMatrix[i].push_back(Value{});
+    }
+  }
+  ~TilesArray() = default;
+
+  Value getTile(int row, int col) { return tileMatrix[row][col]; }
+
+  void setTile(int row, int col, Value val) { tileMatrix[row][col] = val; }
+
+  SmallVector<Value> toFlatVector() {
+    SmallVector<Value> flatVector;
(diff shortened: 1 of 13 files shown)


adam-smnk:
Cleaned up lowering to XeGPU from `linalg-to-xegpu` branch.
The logic has been validated and benchmarked using IMEX runner. Currently, there is no upstream consumer of XeGPU dialect so the Intel GPU pipeline terminates at the XeGPU level.

The plan is to start upstreaming with a bit more simplified version of the current matmul lowering.
Long term plan would be to create more progressive lowering through vector dialect using OneToN conversion and/or identify missing infrastructure pieces required to do so.

rengolin:
@Jianhui-Li
@kurapov-peter

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1394 @@
...
+    return std::nullopt;
+
+  auto mDim = cType.getShape()[0];
+  auto nDim = cType.getShape()[1];
+  auto kDim = aType.getShape().back();
+
+  // DPAS hardware sizes in MxNxK format.
+  // TODO: In case more hardware configurations are available,
+  //       add some automatic selection for optimal sizes.
+  SmallVector<int64_t> dpasTile{8, 16, 16};
```
Jianhui-Li, line 118: this is the DPAS register block size for PVC. For ATS, it should be {8, 16, 8}.

adam-smnk: I'll expose it as a pass param for now. Later this could be pulled from target descriptor.

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1394 @@
...
+  int dpasTileK = dpasTile[2];
+  if ((mDim % dpasTileM != 0) || (nDim % dpasTileN != 0) ||
+      (kDim % dpasTileK != 0) || (kTile % dpasTileK != 0)) {
+    return std::nullopt;
+  }
+
+  return dpasTile;
+}
+
+// Verify if linalg operands fulfill XeGPU constraints.
```
Jianhui-Li, line 135: This is the constraints of the lowering, not XeGPU's constraints.

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1394 @@
...
+// tile.
+//
+// The loaded tiles must be perfectly divisible by the specified subtiles.
+// Assumes row-major ordering for both the loaded tiles and the original tile.
+//
+// If the loaded tiles use VNNI layout, corresponding VNNI configuration must be
+// provided.
+static TilesArray
+extractVecSubTiles(PatternRewriter &rewriter, Location loc,
+                   ValueRange loadVecTiles, ArrayRef<int64_t> totalTileShape,
```
Jianhui-Li, line 754: can we use sgTotalTile, loadTile, dpasTile here?  Just to make it easy to understand

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1394 @@
...
+  void setTile(int row, int col, Value val) { tileMatrix[row][col] = val; }
+
+  SmallVector<Value> toFlatVector() {
+    SmallVector<Value> flatVector;
+    for (auto row : tileMatrix)
+      flatVector.append(row);
+    return flatVector;
+  }
+
+  SmallVector<SmallVector<Value, 8>, 8> tileMatrix;
```
Jianhui-Li, line 80: So this mean that the largest size of subgroup level Tile (sgTotalTile) can only be 8x8 of dpas level Tile?

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1394 @@
...
+              /*sizes=*/ArrayRef<int64_t>{subTileShapeSize},
+              /*strides=*/ArrayRef<int64_t>{1});
+          auto castTile =
+              rewriter.create<vector::ShapeCastOp>(loc, vecSubTileType, slice);
+
+          // Insert the sub-tiles in their position relative to the whole
+          // thread/workitem tile.
+          int rowIdx = m * subTilesPerLoadRow + i;
+          int colIdx = k * subTilePerLoadCol + j;
+          subTiles.setTile(rowIdx, colIdx, castTile);
```
Jianhui-Li, line 819: what happens if rowIdx and colIdx are larger than 8 here? The TilesArray will automatically expand?

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: No, it's just a small allocation optimization. It indicates that enough space will be stack allocated for 8x8 values which a guess whether should be +/- enough for typical use case. But if more is needed, it will automatically expand.

To be fair, the size is a bit random guess derived for my testing so maybe it could be omitted.

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: At this point `subTiles` is just a 2D array, its constructor ensures the storage is as large as requested which is known as all relevant sizes are statically known. But it won't resize any further.

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: Removed the vector sizes to avoid confusion.

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: Updated naming and comments.

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: Added pass param.

adam-smnk:
#### Code updates:
Improved readability a bit and made DPAS sizes configurable.
I tried adding support for more data types but that touches more code than I expected. So, it is better to leave that to future iterations if needed.

#### Future steps update:
The general consensus is that it is fine to keep this PoC within TPP-MLIR.
The current lowering is hard to maintain to benefit wider upstream community or internal users. In the current form, it is difficult to propagate different decision due to large internal coupling (reduction dim tiling, HW features etc.). This is better address with separate (upstream) transformations.

With increasing interest in upstream GPU codegen, it will be better to shift the focus to the next step: progressive lowering through Vector dialect. The knowledge gained from this exercise and XeTile design should allow us to iterate on GPU-oriented vectorization scheme. This will allow to encapsulate different tuning parameters (prefetching scheme, optimal load/compute sizes, SIMD vs SIMT), reuse available upstream tools, and defer lowering to XeGPU.

adam-smnk:
@Jianhui-Li Please let me know if there is anything else unclear in the lowering logic. Of course, the main issue - the overall codegen complexity - can't be reduced here. It'll need different more progressive strategy.

If it is fine, I'd propose we wrap up with PoC and move on to Vector lowering work.

Jianhui-Li:
> @Jianhui-Li Please let me know if there is anything else unclear in the lowering logic. Of course, the main issue - the overall codegen complexity - can't be reduced here. It'll need different more progressive strategy.
> 
> If it is fine, I'd propose we wrap up with PoC and move on to Vector lowering work.

The PR looks to me, and I agree to move on to Vector lowering.

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1402 @@
...
+        .create<arith::AddFOp>(loc, resType, operands[0], operands[1])
+        .getResult();
+  }
+
+  return std::nullopt;
+}
+
+// Lower an elementwise operation to an XeGPU compatible op.
+// Returns the result of the lowered op or nullopt, otherwise.
+static std::optional<Value> lowerEltwiseOp(linalg::LinalgOp linalgOp,
```
rengolin, line 219: Can you not just generalize (upstream) and then only use the above method?

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1402 @@
...
+      auto ubVal = getConstVal(ub);
+      auto stepVal = getConstVal(step);
+      if (!lbVal || !ubVal || !stepVal)
+        return failure();
+
+      int64_t blockSize = (*ubVal - *lbVal) / *stepVal;
+
+      // Assume that at least one subgroup is created for the given
+      // dimension. Otherwise, outlining will fail anyway.
+      blockSizes.push_back(blockSize < 0 ? 1 : blockSize);
```
rengolin, line 427: When would this ever be `< 0`?

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -0,0 +1,1402 @@
...
+    }
+
+    // Too many dimensions, something went wrong. Bail out.
+    if (blockSizes.size() > 3)
```
rengolin, line 431: This is checked elsewhere, right? Could this perhaps be an assert, to crash the compiler and give use some more info?

rengolin (approved):
Some nits, but otherwise, LGTM. Thanks!

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: In case of somewhat ill defined IR like `lb > ub` (overall valid but wrong in this context) or one of them negative.
I think returning failure might be a bit better here.

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: Overall, this shouldn't be necessary to extract info out of loops. But I'll have to improve GPU pipeline and how the passes are ordered i.e., first outline kernels only then lower ops within `gpu.launch` than has this info.

If it is really `> 3` then kernel outlining will fail which should be easier to understand than hard crashing here.

lib/TPP/GPU/LinalgToXeGPU.cpp
adam-smnk: I think something comparable (Linalg -> arith) can be achieved using Linalg vectorizer and vector utilities. So, such explicit conversion like here shouldn't be needed. If I notice something missing there, then it can be fixed directly upstream.

lib/TPP/GPU/LinalgToXeGPU.cpp
rengolin: or even an assert. This sounds like broken IR to me.

adam-smnk: Tweaked the check.

adam-smnk: Went with failure like in the `if (blockSizes.size() > 3)` case. Further GPU passes will fail anyway.

rengolin (commented):
LGTM, thanks!

PR #916, opened 2024-05-23
[NFC] Separate local dialect lowering from low-level parallelization

rengolin:
The 2D parallelization and AMX tile config stuff were bundled into the local dialect lowering passes because that's "an XSMM thing". But that's changing, as we start working on multi-level loop parallelization, this will raise to tensor eventually and we need to separate the transform from the dialect.

This PR however, does none of that. It just separates the "bundles" to de-couple dialect lowering from the loop passes. Should not change a iota on the output.

Also removing a duplicated call to lower the `perf` dialect in the main pipeline.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -280,6 +280,22 @@ def Cleanup : Pass<"cleanup", "func::FuncOp"> {
   let summary = "General IR cleanup e.g., canonicalization, CSE etc.";
 }
 
+def LowLevelParallelization : Pass<"low-level-paralle", "ModuleOp"> {
+  let summary = "Low level parallelization (multi-threading, AMX config).";
+  let dependentDialects = ["affine::AffineDialect",
+                           "arith::ArithDialect",
+                           "func::FuncDialect",
+                           "memref::MemRefDialect",
+                           "scf::SCFDialect",
+                           "xsmm::XsmmDialect",
+                           "LLVM::LLVMDialect"];
+  let options = [
+    ListOption<"parallelTaskGrid", "parallel-task-grid",
+           "unsigned", "Grid-sizes for parallel tasks.">
+
+  ];
+}
+
 def LocalDialectsLowering : Pass<"lower-local-dialects", "ModuleOp"> {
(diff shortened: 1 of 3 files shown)


rengolin:
Benchmarks show no regressions.

include/TPP/Passes.td
```diff
@@ -280,24 +280,34 @@ def Cleanup : Pass<"cleanup", "func::FuncOp"> {
   let summary = "General IR cleanup e.g., canonicalization, CSE etc.";
 }
 
-def LocalDialectsLowering : Pass<"lower-local-dialects", "ModuleOp"> {
-  let summary = "Lower all local dialects (XSMM, check etc.).";
+def LowLevelParallelization : Pass<"low-level-paralle", "ModuleOp"> {
```
adam-smnk, line 283:
```suggestion
def LowLevelParallelization : Pass<"low-level-parallel", "ModuleOp"> {
```

adam-smnk (commented):
Could you add a small test for the new pass?

include/TPP/Passes.td
```diff
@@ -280,24 +280,34 @@ def Cleanup : Pass<"cleanup", "func::FuncOp"> {
...
                            "memref::MemRefDialect",
-                           "check::CheckDialect",
-                           "perf::PerfDialect",
                            "scf::SCFDialect",
-                           "tensor::TensorDialect",
                            "xsmm::XsmmDialect",
                            "LLVM::LLVMDialect"];
   let options = [
     ListOption<"parallelTaskGrid", "parallel-task-grid",
            "unsigned", "Grid-sizes for parallel tasks.">
```
adam-smnk, line 294: nit: some default value like `1`

adam-smnk:
> Could you add a small test for the new pass?

Actually, never mind. It's just a pure bundle. Options is unused but even if it is just passed through to a pass that consumes it.

adam-smnk (approved):

include/TPP/Passes.td
rengolin: I'm not sure the list option has a default value, none of the others do. Since this is just an NFC, I'll leave it for later.

PR #918, opened 2024-06-04
Pass bundles

adam-smnk:
Separates existing passes into two types:
  - passes - standard passes that transform IR
  - bundles - passes that have no custom logic and only run pass pipelines

Additionally, the existing bundles are moved to individual source files to make repo navigation easier. Plus some minor cleanup.

The goal is to improve pipeline modularization and make it easier to navigate available tools within TPP-MLIR.
`DefaultPipeline` and `DefaultTppPasses` stay at the top directory to promote their visibility. `GpuPipeline` stays together with all GPU sources.


tools/tpp-opt/tpp-opt.cpp
```diff
@@ -33,6 +34,7 @@
 int main(int argc, char **argv) {
   mlir::registerAllPasses();
   mlir::tpp::registerTppCompilerPasses();
+  mlir::tpp::registerTppBundlePasses();
```
rengolin, line 37: Shouldn't the bundle pass now replace the tpp-compiler registration?

adam-smnk: No, these calls only register what's declared in their respective tablegens.
Bundles only contain the bundle passes.

adam-smnk:
No performance regressions in benchmarks.

rengolin (approved):

PR #921, opened 2024-06-05
Retire downstream constant packing folder

adam-smnk:
Replaces our downstream implementation of constant packing folder with upstream patterns. The existing pass selects pack ops to fold and otherwise acts as a small wrapper for the upstream logic.

Also, there was a bug in the element index calculation. After the replacement, folded constants match the outputs produced by runtime packing.


adam-smnk:
Rewrite problems fixed. In the initial version, chains of packs e.g., `pack_vnni(pack(constant))` would not get folded after a single `ConstantFoldPack` pass execution. This is because now the rewrite is more iterative: select constant packs -> lower to linalg -> let other patterns constant fold and cleanup the ops.

The lowering to linalg is now a pattern such that greedy rewriter can iterate over the graph and step by step fold all the packs.
The core of the folder logic now relies fully on upstream passes. The only extra bit is operation selection (which packs to fold). However, that is opinionated and very user-specific so, it can probably remain downstream. 

Added extra test case to cover this case. No performance regression in benchmarks.

rengolin (approved):

PR #922, opened 2024-06-07
Loop tiling, shuffle and expansion passes

KavithaTipturMadhu:


alheinecke:
I haven't done a detailed review, but it seems the old 2d parallelization is disabled/removed from tpp-run. I suggest for now we keep the old one functional in the default pass and make the new one available through a different knob. Then we can compare both and retire the old one, when the new one completely covers the old usecases.

KavithaTipturMadhu:
I am working on unit tests for the passes as well as MLP integration tests, in the meantime we can start the review.

lib/TPP/DefaultPipeline.cpp
```diff
@@ -124,9 +142,16 @@ struct DefaultPipeline : public tpp::impl::DefaultPipelineBase<DefaultPipeline>,
...
-                                                parallelTaskGrid};
-      pm.addPass(createDefaultTppPasses(tppDefaultOptions));
+      if (!tileShapeM.empty() && !tileShapeN.empty()) {
+        DefaultTppPassesOptions tppDefaultOptions{
+            linalgToLoops, parallelTaskGrid, tileShapeM,
+            tileShapeN,    shuffleOrder,     outerParallelLoops};
+        pm.addPass(createDefaultTppPasses(tppDefaultOptions));
+      } else {
+        DefaultTppPassesOptions tppDefaultOptions{linalgToLoops,
+                                                  parallelTaskGrid};
```
adam-smnk, line 152: This triggers `-Wmissing-field-initializers`.
I'd suggest creating `DefaultTppPassesOptions` with defaults first, then modifying different options, and adding `createDefaultTppPasses` only once (no branching paths).
This should make it simpler and improve readibility.

include/TPP/Passes.td
```diff
@@ -429,6 +429,45 @@ def ConvertAddInplacePass: Pass<"linalg-convert-add-in-place",
...
+}
+
+def LoopShufflePass: Pass<"loop-shuffle-pass">{
+  let summary = "Shuffle brgemm parallel op";
+  let description = [{
+    Shuffle nd brgemm parallel loop.
+  }];
+  let options = [
+    ListOption<"shuffleOrder", "shuffle-order", "unsigned",
+               "Order to shuffle the parallel loop by">
```
adam-smnk, line 465: Could add some description how the order is interpreted? Is this a simple permutation list?

include/TPP/Passes.td
```diff
@@ -429,6 +429,45 @@ def ConvertAddInplacePass: Pass<"linalg-convert-add-in-place",
...
 
+def LoopInsertionPass: Pass<"loop-insertion-pass">{
+  let summary = "Insert loop around brgemm parallel op";
+  let description = [{
+    Insert nd parallel loop around brgemm parallel loop.
+  }];
+  let dependentDialects = ["scf::SCFDialect" , "xsmm::XsmmDialect"];
+  let options = [
+    ListOption<"tileShapeM", "M-tile-shape", "unsigned",
+               "Shape to reshape the M tensor into">,
```
adam-smnk, line 440: nit: just `tensor` is a bit overloaded (with the dialect and tensor type coming to mind first), I'd be a bit more specific here e.g., XY dim or n D input etc.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,276 @@
...
+  LoopInsertionPass(ArrayRef<unsigned> tileShapeM,
+                    ArrayRef<unsigned> tileShapeN) {
+    this->tileShapeM = tileShapeM;
+    this->tileShapeN = tileShapeN;
+  };
+
+  LoopInsertionPass(const tpp::LoopInsertionPassOptions &options) {
+    tileShapeM = options.tileShapeM;
+    tileShapeN = options.tileShapeN;
+  };
```
adam-smnk, line 262: These are autogenerated from TD.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,276 @@
...
+  xsmm::BrgemmOp brgemmOp = NULL;
+  OpBuilder b(op);
+  for (auto oper = op.getBody()->getOperations().begin();
+       oper != op.getBody()->getOperations().end(); oper++)
+    if (dyn_cast<xsmm::BrgemmOp>(oper)) {
+      brgemmOp = dyn_cast<xsmm::BrgemmOp>(oper);
+      break;
+    }
+  if (brgemmOp == NULL)
+    return failure();
```
adam-smnk, line 114: Is it safe to use when loop body is more complex?
Just wondering if there is anything about loop usage to be validated before we start modifying it.

lib/TPP/Transforms/LoopExpansion.cpp
```diff
@@ -0,0 +1,143 @@
...
+} // namespace mlir
+
+using namespace mlir;
+using namespace mlir::scf;
+using namespace std;
+
+namespace mlir {
+namespace tpp {
+
+struct LoopExpansion : OpRewritePattern<scf::ForallOp> {
```
adam-smnk, line 35: Just curious, why `scf.forall` and not `scf.parallel`?

No preference either way for me.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,276 @@
...
+          brgemmOp.getOperand(1).getDefiningOp()->getOperand(0).getType())
+          .getShape();
+
+  // Validate the input tile sizes against the operand shapes
+  long multipleM = 1;
+  for (size_t i = 0; i < tileShapeM.size(); i++)
+    multipleM = multipleM * tileShapeM[i];
+
+  if (mShape[0] != multipleM)
+    return failure();
```
adam-smnk, line 128: nit: add some `notifyMatchFailure` message to make it easier to test and debug

I tried the pass on simple example and I have no clue why nothing happens.

adam-smnk (commented):
Initial general comments.

lib/TPP/Transforms/LoopExpansion.cpp
KavithaTipturMadhu: Forall is updated in the loopinsertion pass, hence we expand it in this pass. There's also a transform pass that expands forall to for that we eventually want to use instead of this pass. That's the reason we're using forall instead of parallel.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: We check the shapes for tile shape mismatches as the next step, the op we look for is the innermost scf forall. We can probably add a few more checks like subview inputs for pytorch like models instead of mlir-gen like simple models, which adds to the complexity of our verifiers.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: I have removed the arrayref initializers but the options initializer needs to stay because the compiler complains otherwise.

lib/TPP/DefaultPipeline.cpp
KavithaTipturMadhu: Fixed with empty Arrayref initial values because I have empty checks in the DefaultTppPasses file. Does that work?

adam-smnk: I'd collapse the if else:
```mlir
DefaultTppPassesOptions opts;
opts.loops = loops;
...
if(!tileM.empty())
  opts.tileM = tileM;
...
pm.addPass(createDefaultTppPasses(opts));
```

lib/TPP/Transforms/LoopInsertion.cpp
adam-smnk: You can still remove these but you need to add `using LoopInsertionPassBase::LoopInsertionPassBase;` (or whatever base class name is for others).

KavithaTipturMadhu:
Performance numbers are comparable.

adam-smnk:
I'm playing with the passes to get a better idea of the transforms but I'm running into an error I don't really get why.
Could you help me out with this case?

Input kernel:
```mlir
#map = affine_map<(d0) -> (d0 * 32)>
func.func @entry(%arg0: memref<32x32x32x32xbf16>, %arg1: memref<32x32x16x32x2xbf16>, %arg2: memref<1024x1024xbf16>) {
  %c32_i64 = arith.constant 32 : i64
  scf.forall (%arg3, %arg4) in (32, 32) {
    %0 = affine.apply #map(%arg3)
    %1 = affine.apply #map(%arg4)
    %subview = memref.subview %arg0[%arg3, 0, 0, 0] [1, 32, 32, 32] [1, 1, 1, 1] : memref<32x32x32x32xbf16> to memref<32x32x32xbf16, strided<[1024, 32, 1], offset: ?>>
    %subview_0 = memref.subview %arg1[%arg4, 0, 0, 0, 0] [1, 32, 16, 32, 2] [1, 1, 1, 1, 1] : memref<32x32x16x32x2xbf16> to memref<32x16x32x2xbf16, strided<[1024, 64, 2, 1], offset: ?>>
    %subview_1 = memref.subview %arg2[%0, %1] [32, 32] [1, 1] : memref<1024x1024xbf16> to memref<32x32xbf16, strided<[1024, 1], offset: ?>>
    %2 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 1024, 1024, 1024] flags = (vnni_b) data_type = bf16
    xsmm.brgemm(data_type = bf16, %2, %subview, %subview_0, %subview_1, %c32_i64) : (i64, memref<32x32x32xbf16, strided<[1024, 32, 1], offset: ?>>, memref<32x16x32x2xbf16, strided<[1024, 64, 2, 1], offset: ?>>, memref<32x32xbf16, strided<[1024, 1], offset: ?>>, i64) -> ()
  }
  return
}
```

Pipeline:
`tpp-opt ../test.mlir -loop-insertion-pass="M-tile-shape=2,4 N-tile-shape=4,8"`

I end up with an error: `require m tile shape to match tensor shape`

adam-smnk:
> The tile shapes are expected to match the leading dimensions of the tensors. In this case, arg0's leading dimension is 32, therefore the M-tile-shape must be factors of 32. The same is true for N-tile-shape i.e., the tile factors are expected to be factors of 32 which is the leading dimension of arg1.

Thanks, that helped me go a bit further.

There is some issue with the pipeline currently. When I try to apply `loop-insertion-pass` on its own with tpp-opt, then it fails to tile. However, `tpp-run` works with any random `M-tile-shape` and `N-tile-shape` arguments.
Since the pass does not signal any failures, I think pass manager simply continues. I can see some changes in IR at the end with `print-mlir=mid` so, other passes still kick in. But I think `loop-insertion` currently does nothing.

KavithaTipturMadhu:
> ```mlir
> #map = affine_map<(d0) -> (d0 * 32)>
> func.func @entry(%arg0: memref<32x32x32x32xbf16>, %arg1: memref<32x32x16x32x2xbf16>, %arg2: memref<1024x1024xbf16>) {
>   %c32_i64 = arith.constant 32 : i64
>   scf.forall (%arg3, %arg4) in (32, 32) {
>     %0 = affine.apply #map(%arg3)
>     %1 = affine.apply #map(%arg4)
>     %subview = memref.subview %arg0[%arg3, 0, 0, 0] [1, 32, 32, 32] [1, 1, 1, 1] : memref<32x32x32x32xbf16> to memref<32x32x32xbf16, strided<[1024, 32, 1], offset: ?>>
>     %subview_0 = memref.subview %arg1[%arg4, 0, 0, 0, 0] [1, 32, 16, 32, 2] [1, 1, 1, 1, 1] : memref<32x32x16x32x2xbf16> to memref<32x16x32x2xbf16, strided<[1024, 64, 2, 1], offset: ?>>
>     %subview_1 = memref.subview %arg2[%0, %1] [32, 32] [1, 1] : memref<1024x1024xbf16> to memref<32x32xbf16, strided<[1024, 1], offset: ?>>
>     %2 = xsmm.brgemm.dispatch [32, 32, 32, 32, 32, 1024, 1024, 1024] flags = (vnni_b) data_type = bf16
>     xsmm.brgemm(data_type = bf16, %2, %subview, %subview_0, %subview_1, %c32_i64) : (i64, memref<32x32x32xbf16, strided<[1024, 32, 1], offset: ?>>, memref<32x16x32x2xbf16, strided<[1024, 64, 2, 1], offset: ?>>, memref<32x32xbf16, strided<[1024, 1], offset: ?>>, i64) -> ()
>   }
>   return
> }
> ```

You might be seeing the shuffle and expansion passes doing something in tpp-run despite tiling failing to do anything. The current implementation is such that the other passes should continue to work despite loop insertion not working, which happens in case of tpp-opt as well, with multiple passes. Should this behavior change? @rengolin @alheinecke

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: Checks have been added for const buffers as well, as in the case of benchmarks.

adam-smnk:
> The current implementation is such that the other passes should continue to work despite loop insertion not working, which happens in case of tpp-opt as well, with multiple passes.

That's perfectly fine for the pass to be opportunistic.

Looking at it a bit closer, my point it that more transformations kick in for `mlir-gen --kernel=args --batch=256 --layers=1024,1024 --tiles=32,32,32` compared to `mlir-gen --kernel=args --batch=256 --layers=1024,1024`.
I think there's an issue(?) handling cases when not all BRGEMM arguments are prepacked. I would expect it to work for either but maybe it's intentional?

I think I'll wait for lit tests with further review as I'm not sure what I should expect currently.

KavithaTipturMadhu:
> > The current implementation is such that the other passes should continue to work despite loop insertion not working, which happens in case of tpp-opt as well, with multiple passes.
> 
> That's perfectly fine for the pass to be opportunistic.
> 
> Looking at it a bit closer, my point it that more transformations kick in for `mlir-gen --kernel=args --batch=256 --layers=1024,1024 --tiles=32,32,32` compared to `mlir-gen --kernel=args --batch=256 --layers=1024,1024`. I think there's an issue(?) handling cases when not all BRGEMM arguments are prepacked. I would expect it to work for either but maybe it's intentional?
> 
> I think I'll wait for lit tests with further review as I'm not sure what I should expect currently.

BRGEMM arguments are expected to be tiled, yes. I have added a check to ensure that arg2 has the same rank as arg0, I had missed that, thanks for reminding me @adam-smnk.

KavithaTipturMadhu:
Unit tests have been updated. I have propogated the failures in loopexpansion and loopshuffle passes as warnings so that the pipeline does not stall, as there are loops that do not have same dimensions as the brgemm loops in mlp. Please review this @rengolin @adam-smnk @alheinecke.

lib/TPP/DefaultPipeline.cpp
```diff
@@ -124,8 +145,16 @@ struct DefaultPipeline : public tpp::impl::DefaultPipelineBase<DefaultPipeline>,
       pm.addPass(createGpuPipeline(GpuPipelineOptions{gpuBackend}));
     } else {
       // Apply the default preprocessing pass
-      DefaultTppPassesOptions tppDefaultOptions{linalgToLoops,
-                                                parallelTaskGrid};
+      DefaultTppPassesOptions tppDefaultOptions;
+      if (!tileShapeM.empty())
```
rengolin, line 149: why the check?

benchmarks/config/omp/mlir-bf16.json
```diff
@@ -5,28 +5,28 @@
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
       "environment": { "OMP_NUM_THREADS": "2", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
-      "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=8,16'" ],
+      "flags": [ "-n", "100", "-run-args='--def-parallel -M-tile-shape=8,1 -N-tile-shape=2,16 --loop-shuffle-order=2,0,1,3 --num-outer-parallel=1'" ],
```
rengolin, line 8: @alheinecke please double check these values

lib/TPP/PassBundles/LowLevelParallelization.cpp
```diff
@@ -62,14 +62,27 @@ struct LowLevelParallelization
...
+    mlir::tpp::LoopExpansionPassOptions loopExpansionPassOptions;
+    loopExpansionPassOptions.numOuterParallel = outerParallelLoops;
+    pm.addPass(createLoopExpansionPass(loopExpansionPassOptions));
 
     pm.addNestedPass<func::FuncOp>(createIntelAMXTileConfigInsertionPass());
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createIntelAMXTileConfigHoistingPass());
+    pm.addPass(createCombineXsmmOpPass());
```
rengolin, line 83: These last passes need to go to a different "bundle" as they're not related to low level parallelization. But this can be done in a follow-up PR.

lib/TPP/Transforms/CombineXsmmPass.cpp
```diff
@@ -126,19 +126,31 @@ struct CombineXsmmOp : public OpRewritePattern<xsmm::BrgemmOp> {
 
     // Replace and delete the old invokes and their dispatches
     rewriter.create<xsmm::FusedBrgemmOp>(loc, dtype, invokeOperands);
+    assert(brgemmOp.use_empty());
     rewriter.eraseOp(brgemmOp);
-    rewriter.eraseOp(brgemmOp.getOperand(0).getDefiningOp());
+    if (brgemmOp.getOperand(0).getDefiningOp()->use_empty()) {
```
rengolin, line 131: Please cache `brgemmOp.getOperand(0)` and the following repeated pattern below for clarity.

test/Integration/gemm-loop-tiling-e2e-bf16.mlir
```diff
@@ -0,0 +1,132 @@
+// RUN: mlir-gen --kernel=args --batch=32 --layers=16,16 --tiles=4,4,4 -float-type=bf16 | tpp-run --M-tile-shape=2,4 --N-tile-shape=2,2 --loop-shuffle-order=0,2,1,3 --num-outer-parallel=2  -e=entry -entry-point-result=void -seed 123 -print | FileCheck %s
+// RUN: mlir-gen --kernel=args --batch=32 --layers=16,16 --tiles=4,4,4 -float-type=bf16 | tpp-run --def-parallel --parallel-task-grid=2,2  -e=entry -entry-point-result=void -seed 123 -print | FileCheck %s
+// CHECK:	( 0.0922852, 0.0820312, 0.114258, 0.149414 )
```
rengolin, line 3: This will probably not pass on AMD/Arm, need the same kind of regex as the other file below.

lib/TPP/DefaultPipeline.cpp
adam-smnk: `DefaultTppPassesOptions` could have some defaults.
It doesn't right now and probably shouldn't but it's not a bad practice.

KavithaTipturMadhu: We can't provide defaults for shuffle and tiling as they are dependent on the tile shapes. I have added a default value for num-outer-parallel. The check is added because we take the older path in case parallel-task-grid option is provided and not tile shapes.

test/Integration/gemm-loop-tiling-e2e-bf16.mlir
KavithaTipturMadhu: The regex alone is not sufficient because in some cases, the differences could be something like 0.09 and 0.10, which I had to manually change in some cases.

rengolin:
> Benchmark regression for OMP 16 threads (~5%). May be nothing, as the DNN benchmarks also change for 16 threads (~6%). I'll trigger the run again to make sure this is a fluke.

New run is 7% slower on MLP 16 x bf16 only, there may be something to it. Could be bf16 fusion. Did you compare the IR emitted between `main` and your PR?

adam-smnk:
> New run is 7% slower on MLP 16 x bf16 only

OMP 16 usually varies the most but this is larger than usual. I'd suggest double checking by manually running benchmarks on main branch and this PR directly on cluster.

KavithaTipturMadhu:
> > Benchmark regression for OMP 16 threads (~5%). May be nothing, as the DNN benchmarks also change for 16 threads (~6%). I'll trigger the run again to make sure this is a fluke.
> 
> New run is 7% slower on MLP 16 x bf16 only, there may be something to it. Could be bf16 fusion. Did you compare the IR emitted between `main` and your PR?

Fusion seems to work just fine. I measured the time taken by my PR and parallel task grid configuration and the times are comparable on spr machine. In fact, parallel-task-grid configuration seems to be taking more time. Parallel task grid time is 0.000100212 whereas tiling time is 9.32289e-05.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,293 @@
...
+      {tileSizes}, dyn_cast<MemRefType>(subviewType).getElementType(),
+      StridedLayoutAttr::get(b.getContext(), ShapedType::kDynamic,
+                             originalStride));
+  auto subview = b.create<memref::SubViewOp>(
+      op.getLoc(), dyn_cast<MemRefType>(subviewType), expandShape.getResult(),
+      offsets, sizes, strides);
+  brgemmOp.getOperand(operandNumber).replaceAllUsesWith(subview);
+}
+
+LogicalResult insertParallelLoop(ForallOp op, unsigned mTileShape,
```
adam-smnk, line 101: I think the current approach disregards other IV users within the loop and can change their iteration space.
Consider this example:
```mlir
func.func @entry(%arg0: memref<8x4x4x4xbf16>, %arg1: memref<4x4x2x4x2xbf16>, %arg2: memref<8x4x4x4xbf16>,
    %arg5: memref<4x4xbf16>) -> memref<4x4xbf16> {
  %c4_i64 = arith.constant 4 : i64
  scf.forall (%arg3, %arg4) in (8, 4) {
    %subview = memref.subview %arg0[%arg3, 0, 0, 0] [1, 4, 4, 4] [1, 1, 1, 1] : memref<8x4x4x4xbf16> to memref<4x4x4xbf16, strided<[16, 4, 1], offset: ?>>
    %subview_0 = memref.subview %arg1[%arg4, 0, 0, 0, 0] [1, 4, 2, 4, 2] [1, 1, 1, 1, 1] : memref<4x4x2x4x2xbf16> to memref<4x2x4x2xbf16, strided<[16, 8, 2, 1], offset: ?>>
    %subview_1 = memref.subview %arg2[%arg3, %arg4, 0, 0] [1, 1, 4, 4] [1, 1, 1, 1] : memref<8x4x4x4xbf16> to memref<4x4xbf16, strided<[4, 1], offset: ?>>
    %0 = xsmm.brgemm.dispatch [4, 4, 4, 4, 4, 4, 16, 16] flags = (vnni_b) data_type = bf16
    xsmm.brgemm(data_type = bf16, %0, %subview, %subview_0, %subview_1, %c4_i64) : (i64, memref<4x4x4xbf16, strided<[16, 4, 1], offset: ?>>, memref<4x2x4x2xbf16, strided<[16, 8, 2, 1], offset: ?>>, memref<4x4xbf16, strided<[4, 1], offset: ?>>, i64) -> ()
    %subview_2 = memref.subview %arg0[%arg3, %arg4, 0, 0] [1, 1, 4, 4] [1, 1, 1, 1] : memref<8x4x4x4xbf16> to memref<4x4xbf16, strided<[4, 1], offset: ?>>
    linalg.add ins(%subview_2, %arg5 
      : memref<4x4xbf16, strided<[4, 1], offset: ?>>, memref<4x4xbf16>)
      outs(%arg5 : memref<4x4xbf16>)
  }
  return %arg5 : memref<4x4xbf16>
}
```

lib/TPP/Transforms/LoopShuffle.cpp
```diff
@@ -0,0 +1,111 @@
...
+        tempIndexMap[i] = j;
+        break;
+      }
+    }
+  }
+  for (size_t i = 0; i < op.getInductionVars().size(); i++) {
+    replaceAllUsesInRegionWith(
+        tempValueMap[i], op.getInductionVar(tempIndexMap[i]), op.getRegion());
+    rewriter.eraseOp(tempValueMap[i].getDefiningOp());
+  }
```
adam-smnk, line 88: I have a feeling there should be easier way to do it. Nothing off top of my head though.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -100,14 +101,32 @@ struct DefaultTppPasses
...
+
+      // Convert forAll to parallel loops should run after bufferization
+      // as scf.parallel does not handle tensor.
+      pm.addPass(createConvertForAllToParallelOp());
+    } else {
+      // FIXME remove as soon as the above code is fixed
+      pm.addPass(createConvertForAllToParallelOp());
+      mlir::tpp::SCFParallelLoopTilingOptions tilingOptions;
+      tilingOptions.tileSizes = parallelTaskGrid;
+      pm.addPass(createSCFParallelLoopTiling(tilingOptions));
```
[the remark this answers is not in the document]
KavithaTipturMadhu: Retired 2d tiling

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: Fixed this, working on improving this to use to not use mul-add chains and indexing instead.

lib/TPP/Transforms/LoopShuffle.cpp
KavithaTipturMadhu: I can't think of a simpler way to do this either.

test/Passes/pass-convert-gemm-to-parallel-tile.mlir
adam-smnk, line 1: I think we should keep these tests and verify that the new lowering produces the same IR.
Unless these cases are already covered by some other tests.

adam-smnk:
Some benchmark errors and regressions.
Reminder to revisit these later.

test/Passes/pass-convert-gemm-to-parallel-tile.mlir
KavithaTipturMadhu: I have already added tests to cover these so no point keeping them.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -100,14 +101,25 @@ struct DefaultTppPasses
...
+      pm.addPass(createLowLevelParallelization(LowLevelParallelization));
+    }
     // Convert forAll to parallel loops should run after bufferization
     // as scf.parallel does not handle tensor.
     pm.addPass(createConvertForAllToParallelOp());
-
-    // Low leve parallelization passes.
-    LowLevelParallelizationOptions LowLevelParallelization{parallelTaskGrid};
-    pm.addPass(createLowLevelParallelization(LowLevelParallelization));
+    pm.addPass(createCombineXsmmOpPass());
```
rengolin, line 114: In a follow-up PR, this needs to be in a new bundle, both Xsmm lowering and AMX passes.
Please add a TODO.

lib/TPP/Transforms/LoopExpansion.cpp
```diff
@@ -0,0 +1,113 @@
...
+  size_t i = 0;
+  for (auto lb = op.getStaticLowerBound().begin();
+       lb != op.getStaticLowerBound().end() &&
+       ub != op.getStaticUpperBound().end() && step != op.getStaticStep().end();
+       lb++, ub++, step++) {
+    auto lowerBound = rewriter.create<arith::ConstantIndexOp>(op.getLoc(), *lb);
+    auto upperBound = rewriter.create<arith::ConstantIndexOp>(op.getLoc(), *ub);
+    auto stepVal = rewriter.create<arith::ConstantIndexOp>(op.getLoc(), *step);
+
+    if (i++ < numOuterParallel) {
```
rengolin, line 57: This looks weird. Best if `i++` goes together with the other increments and here just check for `i < numOuter`.

lib/TPP/Transforms/LoopExpansion.cpp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+struct LoopExpansionPass
+    : public impl::LoopExpansionPassBase<LoopExpansionPass> {
+
+  using LoopExpansionPassBase::LoopExpansionPassBase;
+
+  void runOnOperation() override {
+    getOperation()->walk([&](scf::ForallOp forallOp) {
+      if (failed(loopExpand(forallOp, numOuterParallel)))
```
rengolin, line 106: Not sure why this got outlined into a static function. This is a small file, could have been here.

lib/TPP/Transforms/LoopExpansion.cpp
```diff
@@ -0,0 +1,113 @@
...
+                                               upperBound, stepVal);
+      rewriter.setInsertionPoint(&forOp.getBody()->front());
+      forOpList.push_back(forOp);
+    }
+  }
+
+  IRMapping mapping;
+  for (auto oper = op.getBody()->getOperations().begin();
+       oper != op.getBody()->getOperations().end(); oper++) {
+    if (!isa<scf::InParallelOp>(oper)) {
```
rengolin, line 74: What happens to the ones `in_ parallel`?

lib/TPP/Transforms/LoopExpansion.cpp
```diff
@@ -0,0 +1,113 @@
...
+      parallelOpList.push_back(parallelOp);
+    } else {
+      auto forOp = rewriter.create<scf::ForOp>(op.getLoc(), lowerBound,
+                                               upperBound, stepVal);
+      rewriter.setInsertionPoint(&forOp.getBody()->front());
+      forOpList.push_back(forOp);
+    }
+  }
+
+  IRMapping mapping;
```
rengolin, line 71: It took me a while to realise you change the insertion point above, so the cloning below adds it to the right place (the inner-most loop). A comment here would help future us to remember what happens.

rengolin (commented):
Finish review on loop expansion pass. All about the code, not the logic.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+struct LoopInsertionPass
+    : public tpp::impl::LoopInsertionPassBase<LoopInsertionPass> {
+
+  using LoopInsertionPassBase::LoopInsertionPassBase;
+
+  void runOnOperation() override {
+    auto *parentOp = getOperation();
+    SmallVector<ForallOp> innermostForAllloops;
+    getInnermostForLoops(parentOp, innermostForAllloops);
+    for (ForallOp loop : innermostForAllloops)
```
rengolin, line 301: Why isn't this one a walk like the others?

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  return success();
+}
+
+bool getInnermostForLoops(Operation *rootOp,
+                          SmallVectorImpl<scf::ForallOp> &result) {
+  assert(rootOp != nullptr && "Root operation must not be a nullptr.");
+  bool rootEnclosesForAllloops = false;
+  for (Region &region : rootOp->getRegions()) {
```
rengolin, line 274: Isn't there a better way to walk the ops of an enclosing op?

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+    for (Block &block : region.getBlocks()) {
+      for (Operation &op : block) {
+        bool enclosesPloops = getInnermostForLoops(&op, result);
+        rootEnclosesForAllloops |= enclosesPloops;
+        if (auto ploop = dyn_cast<scf::ForallOp>(op)) {
+          rootEnclosesForAllloops = true;
+
+          // Collect forall loop if it is an innermost one.
+          if (!enclosesPloops)
```
rengolin, line 283: This will return true for more than one inner loop at different nest levels, I'm not sure that's what you want.

I'd focus on the parallel loops at the function level and not recurse into further next levels for now.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+        inductionVarBase = 0;
+      insertSubview(shape, operandType, resultType, reassociationIndex, operand,
+                    op, rewriter, definingOp, inductionVarBase);
+    }
+  }
+
+  return success();
+}
+
+bool getInnermostForLoops(Operation *rootOp,
```
rengolin, line 270:
```suggestion
static bool getInnermostForLoops(Operation *rootOp,
```

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+  for (auto oper = op.getBody()->getOperations().begin();
+       oper != op.getBody()->getOperations().end(); oper++)
+    if (dyn_cast<xsmm::BrgemmOp>(oper)) {
+      brgemmOp = dyn_cast<xsmm::BrgemmOp>(oper);
+      break;
+    }
+  if (brgemmOp == NULL)
+    return rewriter.notifyMatchFailure(op, "require brgemm op");
+
+  if (mTileShape == 0 || nTileShape == 0)
```
rengolin, line 119: This is not enough. We need to check more stuff like if the tile sizes evenly divide the current lb/ub/step and the tile sizes of the inner loop ops.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+  xsmm::BrgemmOp brgemmOp = NULL;
+  OpBuilder b(op);
+  IRRewriter rewriter(b.getContext());
+  for (auto oper = op.getBody()->getOperations().begin();
+       oper != op.getBody()->getOperations().end(); oper++)
+    if (dyn_cast<xsmm::BrgemmOp>(oper)) {
+      brgemmOp = dyn_cast<xsmm::BrgemmOp>(oper);
+      break;
+    }
+  if (brgemmOp == NULL)
```
rengolin, line 116: As discussed, this needs more variations (matmul, batched_matmul, batched_reduction_matmul).

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+    return rewriter.notifyMatchFailure(op, "require brgemm op");
+
+  if (mTileShape == 0 || nTileShape == 0)
+    return rewriter.notifyMatchFailure(op, "require tile shape to not be zero");
+
+  ArrayRef<int64_t> mShape;
+  auto mDefiningOp = brgemmOp.getOperand(1).getDefiningOp();
+  if (isa<memref::SubViewOp>(mDefiningOp))
```
rengolin, line 124: This repetition calls for automation. Either a static function or a lambda would be fine.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+    kShape =
+        dyn_cast<ShapedType>(kDefiningOp->getResult(0).getType()).getShape();
+
+  if (mShape.size() != kShape.size())
+    return rewriter.notifyMatchFailure(
+        op, "require m tensor shape and k tensor shape to match");
+
+  SmallVector<unsigned> tileShapeM;
+  tileShapeM.push_back(mTileShape);
+  tileShapeM.push_back(mShape[0] / mTileShape);
```
rengolin, line 155: Verify compatibility of divisions.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+  long multipleN = 1;
+  for (size_t i = 0; i < tileShapeN.size(); i++)
+    multipleN = multipleN * tileShapeN[i];
+
+  if (nShape[0] != multipleN)
+    return rewriter.notifyMatchFailure(
+        op, "require n tile shape to match tensor shape");
+
+  if ((multipleM * multipleN) != (kShape[0] * kShape[1]))
```
rengolin, line 178: This is assuming the `kShape` is 2D. No more, no less. If that's the requirement, then we need to verify that this is the case for m, n and k.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+  SmallVector<unsigned> tileShapeN;
+  tileShapeN.push_back(nTileShape);
+  tileShapeN.push_back(nShape[0] / nTileShape);
+
+  // Validate the input tile sizes against the operand shapes
+  long multipleM = 1;
+  for (size_t i = 0; i < tileShapeM.size(); i++)
+    multipleM = multipleM * tileShapeM[i];
+
+  if (mShape[0] != multipleM)
```
rengolin, line 166: Easier to check with modulo before pushing back into the vector above.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+  SmallVector<ArrayRef<int64_t>> originalShapes{mShape, nShape, kShape};
+  SmallVector<SmallVector<SmallVector<unsigned>>> tilingVectors{
+      {tileShapeM}, {tileShapeN}, {tileShapeM, tileShapeN}};
+
+  // Replace old args with newly computed args
+  for (auto oper = op.getBody()->getOperations().begin();
+       oper != op.getBody()->getOperations().end(); oper++) {
+    int operandIndex = 0;
+    for (auto arg : oper->getOperands()) {
```
rengolin, line 213: This logic is very confusing. What are you trying to do here?

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+        Value add = op.getBody()->getArgument(tileOffsets[oldArgIndex]);
+        Value mul;
+        for (int j = tileOffsets[oldArgIndex] + 1;
+             j < tileOffsets[oldArgIndex + 1]; j++) {
+          unsigned cumulativeUpperBound = 1;
+          for (int k = j; k < tileOffsets[oldArgIndex + 1]; k++)
+            cumulativeUpperBound *= op.getStaticUpperBound()[j];
+          Value upperBound = rewriter.create<arith::ConstantIndexOp>(
+              op.getLoc(), cumulativeUpperBound);
+          mul = rewriter.create<arith::MulIOp>(op.getLoc(), b.getIndexType(),
```
rengolin, line 231: Let's not introduce indexing arithmetic where we can just use subviews, expand shape and tiling.

lib/TPP/Transforms/LoopInsertion.cpp
```diff
@@ -0,0 +1,307 @@
...
+          add = rewriter.create<arith::AddIOp>(
+              op.getLoc(), b.getIndexType(), mul, op.getBody()->getArgument(j));
+        }
+        oper->setOperand(operandIndex, add);
+      }
+      operandIndex++;
+    }
+  }
+
+  for (int i = 1; i <= 3; i++) {
```
rengolin, line 242: `3`? Shouldn't this be `gemmOp.getNumOperands()` or something?

rengolin (commented):
Review on loop insertion. This is more like parallel tiling, why is is called "insertion"?

lib/TPP/Transforms/LoopShuffle.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+static LogicalResult loopShuffle(scf::ForallOp op,
+                                 ArrayRef<unsigned> shuffleOrder) {
+  OpBuilder b(op);
+  IRRewriter rewriter(b.getContext());
+  if (shuffleOrder.size() != op.getInductionVars().size())
+    return rewriter.notifyMatchFailure(op, "Number of indices incorrect");
+  for (size_t i = 0; i < op.getInductionVars().size(); i++) {
+    bool match = false;
+    for (size_t j = 0; j < shuffleOrder.size(); j++)
```
rengolin, line 45: Easier to sort the list and check if `shuffleOrder[i] = i`.

lib/TPP/Transforms/LoopShuffle.cpp
```diff
@@ -0,0 +1,109 @@
...
+        break;
+      }
+    if (!match) {
+      return rewriter.notifyMatchFailure(op, "Indices missed");
+    }
+  }
+
+  SmallVector<int64_t> lbs, ubs, steps;
+
+  for (size_t i = 0; i < op.getStaticLowerBound().size(); i++) {
```
rengolin, line 57: If lb/ub/step needs to have the same size as `shuffleOrder`, you can collapse all this in a single loop. 

If you use iterators, you can even use `llvm::zip_equals`.

lib/TPP/Transforms/LoopShuffle.cpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+  for (size_t i = 0; i < op.getStaticStep().size(); i++) {
+    steps.push_back(op.getStaticStep()[shuffleOrder[i]]);
+  }
+
+  op.setStaticLowerBound(lbs);
+  op.setStaticUpperBound(ubs);
+  op.setStaticStep(steps);
+
+  SmallVector<Value> tempValueMap(op.getInductionVars().size());
```
rengolin, line 71: Cache `op.getInductionVars().size()` for readability.

rengolin (commented):
On the loop shuffle pass, IIUC, this just reorders the induction variables in the `scf.forall` list and replaces all users with. 

Technically, this makes no difference, since the order in which the tiles are executed do not matter (no order guarantee). But because after this you run the loop expansion, and you do that in IR order, it "works".

This is fragile, but it should be fine for now. For example, a canonicalization could reorder again for some unknown reason. When we move this to transform, the canonicalization reason would be less pertinent, because the schedule should make it explicit, and not implicit like in a pass pipeline.

But we need to be clear that this is the case, perhaps with a comment describing this succinctly.

KavithaTipturMadhu:
> Review on loop insertion. This is more like parallel tiling, why is is called "insertion"?

Earlier, the plan was the "insert" an scf.parallel loop around the brgemm op, that was why I started with the name Loop insertion instead of Loop parallelization.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: We need to parallelize the innermost forall loop, which is hard to detect with a walk. The upstream parallelization pass also does something similar, look for the innermost loop and invoke tiling on it.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: We can use walk but that will also translate to a recursive walk eventually, because we look for the innermost scf forall loop.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: This will return true for the innermost level of a nested forall loop, for each loop nest.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: It's generally not a good idea to make recursive functions static, but I guess we're stateless in this case, so we can make an exception.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: I have added a comment to explain this.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: I have added expand shape wherever possible, however there are affine maps which use induction variables, which cannot be updated to use expand shapes in case of torch mlp model for example. We need mul-add chains in such cases as fallback. I have simplified this code as much as possible so that it stays bug free. It now uses getStridesAndOffset call to compute strides and computes and mul-add chain of induction variables. Please take a look.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: Updated.

lib/TPP/Transforms/LoopShuffle.cpp
KavithaTipturMadhu: We'd have to make a copy of shuffle order and then sort it.

lib/TPP/Transforms/LoopInsertion.cpp
KavithaTipturMadhu: updated a test

lib/TPP/Transforms/LoopExpansion.cpp
KavithaTipturMadhu: I forgot to clone the body of in_parallelOps, will update the patch to clone.

rengolin: Ok, this is my last comment. Once you update this, please add a test to make sure you're catching it.

KavithaTipturMadhu:
Closing this PR since we're moving this pass to vector dialect.

PR #927, opened 2024-06-13
Retire FoldExpandShapeInParallelInsertOp pass

adam-smnk:
Replaces downstream pass with upstreamed tensor pattern.

### DIFF
--- a/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
@@ -704,27 +704,6 @@ struct PackOfReshape : public OpRewritePattern<tensor::PackOp> {
   }
 };
 
-struct FoldExpandShapeInParallelInsertOp
-    : public OpRewritePattern<tensor::ParallelInsertSliceOp> {
-  using OpRewritePattern<tensor::ParallelInsertSliceOp>::OpRewritePattern;
-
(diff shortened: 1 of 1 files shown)


lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
```diff
@@ -964,8 +943,7 @@ void mlir::tpp::populateSimplifyPacking(RewritePatternSet &patterns) {
   tensor::ParallelInsertSliceOp::getCanonicalizationPatterns(patterns, ctx);
   ctx->getLoadedDialect<tensor::TensorDialect>()->getCanonicalizationPatterns(
       patterns);
-  patterns.add<PackOfReshape, FoldExpandShapeInParallelInsertOp,
-               FoldCollapseShapeInExtractSliceOp, FoldUnPackIntoInsertSlice,
-               ForAllIterArgsFolder>(ctx);
+  patterns.add<PackOfReshape, FoldCollapseShapeInExtractSliceOp,
```
rengolin, line 946: You're not adding an upstream pass here. Does the upstream pattern runs automatically or was already added?

adam-smnk: The replacement pattern is part of `tensor::populateReassociativeReshapeFoldingPatterns` which is already present here.

adam-smnk:
No benchmark regressions.

rengolin (approved):

PR #928, opened 2024-06-13
Retire FoldCollapseShapeInExtractSliceOp pass

adam-smnk:
Replaces downstream pass with upstreamed tensor reshape folding pattern.


adam-smnk:
Some small fluctuations on highly parallel OMP benchmarks but seems within expected bounds.
No regressions otherwise.

adam-smnk:
> Some small fluctuations on highly parallel OMP benchmarks

They look to be in line with results a few commits back.

PR #930, opened 2024-06-14
Retire downstream scf.forall to scf.parallel conversion

adam-smnk:
Removes custom conversion logic and reuses the existing pass as a simple wrapper around upstream transform API.


adam-smnk:
No regressions.

rengolin (approved):

PR #938, opened 2024-06-24
New README.md for llvm_version.txt

### LINKED ISSUE #936: build_tools/llvm_version.txt vs `llvm-project` as a submodule?
I was a bit surprised by the instructions in the [current README](https://github.com/plaidml/tpp-mlir/blob/f7ea8fd05acbafbbba3ff840cd4fa16cc8c5124f/README.md) regarding checking out the designated `llvm-project` version.

The instructions are to:
```
# Clone
git clone https://github.com/llvm/llvm-project.git

# checking out a tpp-mlir compatible version of llvm-project
wget https://raw.githubusercontent.com/plaidml/tpp-mlir/main/build_tools/llvm_version.txt
pushd llvm-project
git checkout `cat ../llvm_version.txt`
popd
rm llvm_version.txt
```

To me this seems to reproduce what `git submodule`s are good for though does so manually. As I currently don't see advantages to this approach, my suggestions are as follows:

1. Switch to just using a submodule to keep track of the designated commit.
2. Add a short document `build_tools/llvm_version.readme.txt` documenting why the current approach is preferable over using submodules.

Either solution would have stopped me from making this issue. Other takes on this are of course appreciated as well.

rengolin:
Fixes #936

### DIFF
--- a/build_tools/README.md
+++ b/build_tools/README.md
@@ -1,0 +1,34 @@
+## How to build LLVM for tpp-mlir
+
+The [instructions](https://github.com/plaidml/tpp-mlir?tab=readme-ov-file#how-to-build-llvm) to build LLVM are meant for cases where you only need to build once. 
+In those cases, a sub module would be more helpful, but those cases are rare, with so many other cases where sub modules would be detrimental.
+
+The main issues with using sub modules are:
+* Cloning the LLVM repo takes a long time, and it's redundant if you already have it elsewhere (very likely).
+* Building LLVM is an order of magnitude longer than tpp-mlir, and you probably already have the build elsewhere (personal or CI).
+* CI would have one LLVM copy (source + build) per builder (different directories).
+* Incompatible with [worktree](https://github.com/plaidml/tpp-mlir/wiki/Multiple-LLVM-Workspaces), which is the most sane way of developing in LLVM.
+
+The main way we use `llvm_version.txt` are listed below.
+
+### CI loop
+
+In CI we have a separate LLVM build clones the PR's version and installs in a shared disk, where any tpp-mlir build (CI or not) can use.
+This build has a _mutex_ so it only happens once per new hash. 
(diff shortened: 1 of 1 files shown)


build_tools/README.md
```diff
@@ -0,0 +1,34 @@
...
+### Local builds
+
+If you are hacking on tpp-mlir, chances are you are also hacking on LLVM directly, and already have an LLVM tree (worktree even) elsewhere, and a build laying around.
+You can create a branch for tpp-mlir and install in a local directory (with the commit hash as name) and set-up your tpp-mlir build to pick up on that.
+This also helps when you go back and forth in tpp-mlir crossing LLVM version boundaries, there's no need to rebuild anything, as the previous hash install will still be there.
+
+### Benchmark builds
+
+If you're trying to reproduce our numbers from the paper, then you'll have a bit more work than recurring users to get everything built.
+Luckily, we have written a script for that task, which will do most of the manual steps for you, and get the benchmarks running.
```
adam-smnk, line 34: nit: link to the script or some example - otherwise, it's a bit off topic

adam-smnk (approved):
Looks good, thanks

rolfmorel:
Thanks for the explanation and documenting it in-repo, @rengolin!

It will take some time to become acquainted with the development experience laid out in this document (e.g. I am new to `git worktree`). All the more, thanks for spelling it out.

(When working this way, I'll try to see if I can address the question for myself if saving the commit id in a `.txt` is really that different from just having it in a `.gitmodules` file. That is, as far as I understand, a `git submodule` does not commit you to certain workflows. You could just use it to document a current designated commit for some other project somewhere else (though it has a default location according to `.gitmodules`) . Maybe there's too much baggage associated with `git submodule`s to go this route anyway.)

rengolin:
FYI, we're not the only ones using this method. [StableHLO](https://github.com/openxla/stablehlo/blob/7bd4e772b07867de074c3306e1b23cdbc95bb859/build_tools/llvm_version.txt) does a similar thing.

Issue #933, opened 2024-06-21
Replicate MLP examples in Linalg Named Ops


rengolin:
Replicate tests and benchmarks written in Linalg generic to use named ops, for example, by changing `mlir-gen` to output named ops instead of generics.

We should compare the output between `mlir-gen -output=generic` with `mlir-gen -output=named | mlir-opt -linalg-generalize` and see that they're the same.

The output should pass through `tpp-opt`, tile and fuse in the same way. This may not work out of the box, so may need to be implemented. This may involve changing the upstream packing, tiling, fusing passes to cope with named ops.

The end result is a complete replacement of all benchmarks to named ops, with the same performance as current generic ones, and additional tests matching IR generation between the natively generic forms and the generalized via upstream pass.

Most of the logic should go upstream, other than the "usage wrappers" like our current framework. This work will be a mix of upstream and downstream PRs, LLVM bumps and fix-ups, until converging to a fixed point.

PR #943, opened 2024-07-12
Extend mlir-gen tool for linalg named ops #933

shahidact:
Mlir-Gen tool currently generates linalg generic ops for basic MLP operators. We want to extend it to also generate
linalg named ops whenever required based on command line option.

This is the 1st PR to a larger goal to ensure we have enough test coverage which passes through TPP-OPT pipeline.


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -264,6 +309,54 @@ std::string MLIRGenerator::createMetadata() {
...
+    namedMatmul = builder
+                      .create<linalg::MatmulOp>(
+                          loc, TypeRange{output.getType()},
+                          ValueRange{input, weight}, ValueRange{output})
+                      .getResult(0);
+  } else if (inputShape.getRank() == 4) {
+    SmallVector<OpFoldResult, 4> dims =
+        tensor::getMixedSizes(builder, loc, weight);
+    Value emptyTensor = builder.create<tensor::EmptyOp>(
+        loc, dims, weightShape.getElementType());
```
adam-smnk, line 342: Builder might not complain but I believe you have to swap reorder dimensions to match for the transpose output.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -264,6 +309,54 @@ std::string MLIRGenerator::createMetadata() {
...
+  auto weightShape = cast<ShapedType>(weight.getType());
+  auto outputShape = cast<ShapedType>(output.getType());
+  Value namedMatmul;
+  if (inputShape.getRank() == 2) {
+    namedMatmul = builder
+                      .create<linalg::MatmulOp>(
+                          loc, TypeRange{output.getType()},
+                          ValueRange{input, weight}, ValueRange{output})
+                      .getResult(0);
+  } else if (inputShape.getRank() == 4) {
```
adam-smnk, line 338: I can see long term value in going DAG route but I don't think we have necessary tools for that atm.
I'm not sure we should bother with named ops in these cases right now when we don't have proper ops to generate.

Also, this doesn't cover the VNNI case (mlir-gen opt `--vnni=2|4`) when rank is 5.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -264,6 +309,54 @@ std::string MLIRGenerator::createMetadata() {
   return data;
 }
 
+void MLIRGenerator::computeFlops(ShapedType inputShape,
```
adam-smnk, line 312: +1 to the helper.
Could you also insert it in other lowerings?

test/Integration/tiling-add-named-op.mlir
adam-smnk, line 1: What do this and `tiling-relu` tests validate?
Seems unrelated to this PR.

adam-smnk (commented):
Could you also extend `mlir-gen-(matmul|fc).mlir` tests?
One test case with named ops for each should be enough.

test/Passes/tile-and-fuse_named-op.mlir
```diff
@@ -0,0 +1,131 @@
...
+// CHECK: func.func @matmul_eletwise_matmul_and_relu
+// CHECK-DAG: %[[C32:.+]] = arith.constant 32 : index
+// CHECK-DAG: %[[C0:.+]] = arith.constant 0 : index
+// CHECK-DAG: %[[C2:.+]] = arith.constant 2 : index
+// CHECK: %[[LOOP:.+]] = scf.for %{{.+}} = %[[C0]] to %[[C32]] step %[[C2]]
+// CHECK-NEXT: %[[LOOP1:.+]] = scf.for %{{.+}} = %[[C0]] to %[[C32]] step %[[C2]]
+// CHECK: linalg.matmul
+// CHECK-NEXT: tensor.empty()
+// CHECK-NEXT: linalg.fill
+// CHECK: linalg.generic
```
rengolin, line 45: Why is the matmul tiled as a named op and the max isn't?

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -139,15 +169,30 @@ Value MLIRGenerator::createLayer(LayerArgs &args) {
   OpBuilder::InsertionGuard guard(builder);
 
   Value chain;
-  chain = lowerMatmul(args.input.value, args.weight.value, args.output.value);
+  if (outputOpKind == OutputOpKind::Generic) {
```
rengolin, line 172: Since `outputKind` is a field of the class, the `lowerMatmul` function can just use it directly, so the if can be inside the `lower*` functions, not here.

tools/mlir-gen/MLIRGen.cpp
rengolin: One aspect of this work is to understand what we want and how we get `tpp-opt` to work with that. Doing both at the same time would be too much. So, if there's an intermediate state we can reach, good. If not, we may have to lower to a named ops format that we can't cope with it (if we know we can change the passes to do).

shahidact: > I can see long term value in going DAG route but I don't think we have necessary tools for that atm. I'm not sure we should bother with named ops in these cases right now when we don't have proper ops to generate.
> 
Another way we could lower this is by tiling + matmul
scf.for
    scf.for
       linalg.matmul ..
however, the subsequent passes need to handle it accordingly.

shahidact: > One aspect of this work is to understand what we want and how we get `tpp-opt` to work with that. Doing both at the same time would be too much. So, if there's an intermediate state we can reach, good. If not, we may have to lower to a named ops format that we can't cope with it (if we know we can change the passes to do).

Even we could have multiple ways of lowering controlled with option which will give more flexibility to play around.

test/Passes/tile-and-fuse_named-op.mlir
shahidact: While doing the fusion in tile-and-fuse pass, at one point we are generalizing the NamedOps as "populateFoldUnitExtentDimsViaSlicesPatterns" works only with linalg.generic.
Hence linalg.max is getting generalized.

This is what I had pointed in our last discussion IIRC.

tools/mlir-gen/MLIRGen.cpp
rengolin: > Also, this doesn't cover the VNNI case (mlir-gen opt `--vnni=2|4`) when rank is 5.

Let's just make sure we cover the VNNI case.

test/Passes/tile-and-fuse_named-op.mlir
rengolin: Ok, this is good information that we'll need to fix upstream eventually. Can you add this as a TODO/FIXME somewhere in the code, so we don't forget? A Github issue would also work.

tools/mlir-gen/MLIRGen.cpp
adam-smnk: Okay maybe to capture my point a bit differently.

We want to switch to named ops to make things easier like matching, having clearer semantics, simplifying transforms and/or codegen. I don't see added value in anchoring around `mmt4d` which: is an arbitrary choice, needs more effort in matching compared to an equivalent generic, doesn't support VNNI cases anyway.

To me, this case shows we have some missing tooling around that space which might need to be addressed at Linalg level. But I'm not sure what it adds to `mlir-gen` as a tool.

shahidact:
> Could you also extend `mlir-gen-(matmul|fc).mlir` tests? One test case with named ops for each should be enough.

Earlier I had tried to cover this case too but landed in cases as shown below which had irregular tile sizes as "tiles=64,48,64"
and it seems it is not a priority for now. Moreover, it requires to do as DAG matching using "reduce_batch_matmul ->  64x48xbf16 ->  broadcast -> 2x16x64x48xbf16".

func.func @entry(%arg0: tensor<2x36x64x64xbf16>, %arg1: tensor<16x36x64x48xbf16>, %arg2: tensor<2x16x64x48xbf16>) -> tensor<2x16x64x48xbf16> {
    %0 = linalg.generic {indexing_maps = [#map, #map1, #map2], iterator_types = ["parallel", "parallel", "reduction", "parallel", "parallel", "reduction"]} ins(%arg0, %arg1 : tensor<2x36x64x64xbf16>, tensor<16x36x64x48xbf16>) outs(%arg2 : tensor<2x16x64x48xbf16>) {
    ^bb0(%in: bf16, %in_0: bf16, %out: bf16):
      %1 = arith.mulf %in, %in_0 : bf16
      %2 = arith.addf %out, %1 : bf16
      linalg.yield %2 : bf16
    } -> tensor<2x16x64x48xbf16>
    return %0 : tensor<2x16x64x48xbf16>
  }

adam-smnk:
> Earlier I had tried to cover this case too but landed in cases as shown below which had irregular tile sizes as "tiles=64,48,64"
and it seems it is not a priority for now. Moreover, it requires to do as DAG matching using "reduce_batch_matmul -> 64x48xbf16 -> broadcast -> 2x16x64x48xbf16".

These two tests `mlir-gen-(matmul|fc).mlir` are not integration in the sense that we run them through `tpp-run`. They only `CHECK` generated IR to ensure it is as we expect it to be i.e., ensure that the IR is valid, to catch possible `mlir-gen` regressions etc.

test/Passes/tile-and-fuse_named-op.mlir
shahidact: Actually, this is our downstream pass "tile-consumer-and-fuse-producers".

test/Integration/tiling-add-named-op.mlir
shahidact: It is just trying to validate that these named ops are going through the opt pipeline smoothly.

adam-smnk: That's way too general for a test. Pipeline is flexible enough to push through and do "something" with these ops but not necessarily what you'd expect.

Tiling itself won't kick in for 2 reasons: tiling anchors only on matmul, the default tile sizes are 32x32 so larger than this input.
Also, as you mentioned in another comment, there is also linalg ops generalization pattern somewhere within the pipeline so we're probably back to generics.

I think it's a good idea to validate how our pipeline behaves with named ops but I would suggest small unit tests with individual ops piped through `-default-pipeline` or some parts of it (depending what we care about). Full integration test with `tpp-run` doesn't tell us much here.

adam-smnk: Such testing is probably most suitable for a separate PR.

shahidact: > That's way too general for a test. Pipeline is flexible enough to push through and do "something" with these ops but not necessarily what you'd expect.
> 
> Tiling itself won't kick in for 2 reasons: tiling anchors only on matmul, the default tile sizes are 32x32 so larger than this input. Also, as you mentioned in another comment, there is also linalg ops generalization pattern somewhere within the pipeline so we're probably back to generics.
> 
> I think it's a good idea to validate how our pipeline behaves with named ops but I would suggest small unit tests with individual ops piped through `-default-pipeline` or some parts of it (depending what we care about). Full integration test with `tpp-run` doesn't tell us much here.

Good points!! I will see if I can add specific case here, otherwise I will remove this test from this PR.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -355,13 +508,24 @@ Value MLIRGenerator::lowerRelu(Value input, Value output) {
...
-  int64_t reluFlops = 1;
+  computeBiasOrReluFlops(outTy);
+  return relu;
+}
+
+Value MLIRGenerator::lowerNamedSoftmax(Value input, Value output) {
+  if (!enableSoftmax)
+    return input;
+
+  // TODO: Add lowering of softmax to sequence of named Ops
```
adam-smnk, line 519: nit: assert on unsupported path

test/Passes/tile-and-fuse_named-op.mlir
```diff
@@ -0,0 +1,131 @@
...
+// CHECK-SAME:  iterator_types = ["parallel", "parallel"]
+// CHECK-SAME:  outs({{.+}} : tensor<2x2xf32>)
+// CHECK-NEXT: ^bb0(
+// CHECK-NEXT: arith.maximumf
+// CHECK: scf.yield %{{.+}} : tensor<32x32xf32>
+// CHECK-NEXT: }
+// CHECK: scf.yield %{{.+}} : tensor<32x32xf32>
+// CHECK-NEXT: }
+
+// -----
```
adam-smnk, line 131: nit: missing new lines

test/Passes/tile-and-fuse_named-op.mlir
adam-smnk, line 1: nit: replace `_` with `-` in the file name for consistency

adam-smnk (approved):

test/Passes/tile-and-fuse_named-op.mlir
adam-smnk, line 1: Duplicated file ;)

shahidact: Oh, its due to rebase, will update.

shahidact: Hold on.. I don't have it locally though.

adam-smnk: My guess is that you deleted it locally but didn't commit that change.

shahidact: > My guess is that you deleted it locally but didn't commit that change.

Just found, thanks

PR #944, opened 2024-07-17
Transposed matmul block packing

adam-smnk:
Enables pack-matmul driver to handle transpose matmul variants.

PyTorch nn.Linear layer by default transposes matrix B (weights). This change allows accelerating also these workload by using upstream block packing capability to handle multiple matmul variants. After packing, matmuls are represented by the same generic format accepted by the current XSMM lowering.

A small performance hit compared to standard linalg.matmul is due to a linalg.transpose of a matmul_tranpose_a|b input matrix not being lowered to XSMM.

### DIFF
--- a/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
@@ -454,7 +454,9 @@ getDefaultBlockingFactors(linalg::LinalgOp linalgOp) {
     return {32, 32};
   }
   assert(isa<linalg::MatmulOp>(linalgOp) ||
-         isa<linalg::BatchMatmulOp>(linalgOp));
+         isa<linalg::BatchMatmulOp>(linalgOp) ||
+         isa<linalg::MatmulTransposeAOp>(linalgOp) ||
+         isa<linalg::MatmulTransposeBOp>(linalgOp));
   return {32, 32, 32};
 }
 
@@ -482,6 +484,8 @@ struct PackMatmul : public tpp::impl::PackMatmulBase<PackMatmul> {
 
       // Pack only these two named matmul variants.
       if (!(isa<linalg::MatmulOp>(linalgOp) ||
+            isa<linalg::MatmulTransposeAOp>(linalgOp) ||
+            isa<linalg::MatmulTransposeBOp>(linalgOp) ||
             isa<linalg::BatchMatmulOp>(linalgOp))) {
         return std::nullopt;
       }
--- a/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
+++ b/lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
@@ -482,7 +482,7 @@ struct PackMatmul : public tpp::impl::PackMatmulBase<PackMatmul> {
         -> std::optional<linalg::BlockPackMatmulOptions> {
       linalg::BlockPackMatmulOptions options;
 
-      // Pack only these two named matmul variants.
+      // Pack only these named matmul variants.
       if (!(isa<linalg::MatmulOp>(linalgOp) ||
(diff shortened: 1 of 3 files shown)


adam-smnk:
In general, our infrastructure can't handle matmul_transpose. However, by prepacking matmuls, which we do anyway, the transpose variants can be converted to a form compatible with the rest of our pipeline.
While it might not be optimal, it does speed up the transposed matmul variants and should allow for easier testing with OV.

lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
```diff
@@ -482,6 +484,8 @@ struct PackMatmul : public tpp::impl::PackMatmulBase<PackMatmul> {
 
       // Pack only these two named matmul variants.
```
rengolin, line 485: comment

test/Passes/DefaultPipeline/linalg-matmul-variants.mlir
```diff
@@ -0,0 +1,95 @@
...
+// CHECK-LABEL: @matmul(
+// CHECK-SAME: %[[ARG0:[a-zA-Z0-9]+]]: memref<2048x2048xbf16>,
+// CHECK-SAME: %[[ARG1:[a-zA-Z0-9]+]]: memref<2048x2048xbf16>,
+// CHECK-SAME: %[[ARG2:[a-zA-Z0-9]+]]: memref<2048x2048xbf16>
+// CHECK: memref.subview %[[ARG0]]
+// CHECK: call @xsmm_unary_invoke
+// CHECK: memref.subview %[[ARG1]]
+// CHECK: call @xsmm_unary_invoke
+// CHECK: memref.subview %[[ARG2]]
+// CHECK: call @xsmm_intel_amx_tile_config_invoke
```
rengolin, line 20: This should only work in SPR machines

test/Passes/DefaultPipeline/linalg-matmul-variants.mlir
```diff
@@ -0,0 +1,95 @@
...
+    -> tensor<2048x2048xbf16>
+  return %0 : tensor<2048x2048xbf16>
+}
+
+// CHECK-LABEL: @matmul_transpose_a(
+// CHECK-SAME: %[[ARG0:[a-zA-Z0-9]+]]: memref<2048x2048xbf16>,
+// CHECK-SAME: %[[ARG1:[a-zA-Z0-9]+]]: memref<2048x2048xbf16>,
+// CHECK-SAME: %[[ARG2:[a-zA-Z0-9]+]]: memref<2048x2048xbf16>
+// CHECK: memref.subview %[[ARG0]]
+// CHECK: linalg.transpose
```
rengolin, line 39: Shouldn't we just change the leading dimensions in the brgemm dispatch? @alheinecke

test/Passes/DefaultPipeline/linalg-matmul-variants.mlir
adam-smnk: Do we have any logic that controls when these are generated?
Or is it just noop on other platforms?

test/Passes/DefaultPipeline/linalg-matmul-variants.mlir
adam-smnk: Currently the block packed layout is the same for all variants, only packing changes. I think no changes should be needed for dispatch.

test/Passes/DefaultPipeline/linalg-matmul-variants.mlir
rengolin: We do have a check for `x86` vs `arm` using libxsmm's CPUID, but that assumes host arch == target arch, so we didn't abuse of it. It should be fine for now, but we need to fix this once this logic goes to graph compilers.

rengolin (approved):

PR #945, opened 2024-07-22
Improve tile-and-fuse for named ops (#933)

shahidact:
Linalg named ops supports destination style argument passing which makes hasOneUse()
API imprecise, which prevents tiling and fusion. This patch extends this logic to handle
destination style argument passing.

*Adds an option to keep-named-op which prevent generalization of named ops.

### DIFF
--- a/lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
@@ -566,6 +566,18 @@ getTileForEltWiseConsumer(Operation *consumer, Operation *producer,
   return eltWiseTiles;
 }
 
+static bool hasOneUser(Value op) {
+  if (op.hasOneUse())
+    return true;
+  auto users = op.getUsers();
+  Operation *firstUser = *users.begin();
+  if (firstUser)
+    return llvm::all_of(users,
+                        [&](Operation *user) { return user == firstUser; });
+
+  return false;
+}
+
 static Operation *getLastFusableEltWiseConsumer(
     linalg::LinalgOp linalgOp,
     llvm::SmallDenseSet<Operation *> &visitedConsumers,
@@ -574,7 +586,7 @@ static Operation *getLastFusableEltWiseConsumer(
   Value linalgOpRes = linalgOp->getResult(0);
   // If we allow use, we may end up doing recomputation. Unclear if it is
   // profitablem thus disallow for now.
-  if (!linalgOpRes.hasOneUse())
+  if (!hasOneUser(linalgOpRes))
     return linalgOp;
 
   // Start checking consumers.
@@ -606,8 +618,7 @@ static Operation *getLastFusableEltWiseConsumer(
         getTileForEltWiseConsumer(currentConsumer, linalgOp, tiles[linalgOp]);
     visitedConsumers.insert(currentConsumer);
     // Require each eltwise to have a single user.
-    if (std::distance(resNextConsumer.getUsers().begin(),
-                      resNextConsumer.getUsers().end()) != 1) {
+    if (!hasOneUser(resNextConsumer)) {
       break;
     }
     nextConsumer = *(resNextConsumer.getUsers().begin());
--- a/test/Passes/tile-and-fuse-fill-named-op.mlir
+++ b/test/Passes/tile-and-fuse-fill-named-op.mlir
@@ -1,5 +1,7 @@
 // RUN: tpp-opt %s -tile-consumer-and-fuse-producers | FileCheck %s
 
+
+// This test checks for the fusion of bias and relu with matmul.
 func.func @fuse_fill(%arg0: tensor<8x32x32x32xf32>, %arg1: tensor<32x32x32x32xf32>, %arg4: tensor<32x32x32x32xf32>) -> tensor<8x32x32x32xf32> {
     %cst_d = arith.constant dense<1.000000e+00> : tensor<32x32x32x32xf32>
     %cst = arith.constant 0.000000e+00 : f32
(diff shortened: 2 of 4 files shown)


lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -574,7 +587,7 @@ static Operation *getLastFusableEltWiseConsumer(
   Value linalgOpRes = linalgOp->getResult(0);
   // If we allow use, we may end up doing recomputation. Unclear if it is
   // profitablem thus disallow for now.
-  if (!linalgOpRes.hasOneUse())
+  if (!hasActualOneUse(linalgOpRes))
```
adam-smnk, line 590: Tile-and-fuse is performed at the tensor level where tensor are immutable and there's no concept of overwriting buffer values. If the linalg op had memref semantics, it would have no results.

I'm not sure how `hasOneUse` could be insufficient. Do you have any specific example?

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -724,7 +736,8 @@ struct TileConsumerAndFuseProducers
 
         // TODO: Remove the generalization of named ops after resolving the
         // above dependency with "populateFoldUnitExtentDimsViaSlicesPatterns".
-        linalg::populateLinalgNamedOpsGeneralizationPatterns(patterns);
+        if (!this->keepNamedOp)
+          linalg::populateLinalgNamedOpsGeneralizationPatterns(patterns);
```
adam-smnk, line 740: I'm not against a flag for control per se but I fail to see how this addresses or impacts anything.
Is it a setup for some other long-term changes? Could this be added to some other PR that utilizes this change?

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
shahidact: You are right to raise this question and it basically relates to why with such a tensor semantic we have this DPS representation.
However, given that one can have "linalg.add ins(%A, %B) outs(%A)", hasOneUse() API, which is based on SSA
representation, will count it as >1 use while in fact the user is only one.

I stumbled upon this test case "matmul_sequence_fusion_with_eltwise" in tile-and-fuse.mlir. This was not getting fused for the named op equivalent test case.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
shahidact: Yes, I was thinking to add more test case where we could see the impact of tile-and-fuse in named ops. 

I can take it up in a separate PR too. In fact, I think this generalization is redundant now as no test was failing but not sure if this is the case also with CI and benchmarks. Let me know if you have any info.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
adam-smnk: Thanks for the explanation! Now I see clearly what you're trying to solve.
You're right, this is a case we should be able to handle.

test/Passes/tile-and-fuse-fill-named-op.mlir
```diff
@@ -1,4 +1,18 @@
...
+    // %{{.+}} = linalg.transpose ins(%{{.+}} : tensor<32x32x32x32xf32>) outs(%{{.+}} : tensor<32x32x32x32xf32>) permutation = [0, 1, 3, 2]
+    // %{{.+}} = scf.forall (%{{.+}}, %{{.+}}) in (8, 32) shared_outs(%{{.+}} = %{{.+}}) -> (tensor<8x32x32x32xf32>) {
+    //     %{{.+}} = linalg.fill ins(%{{.+}} : f32) outs(%{{.+}} : tensor<32x32xf32>) -> tensor<32x32xf32>
+    //     %{{.+}} = linalg.generic
+    //        %{{.+}} = linalg.mulf
+    //        %{{.+}} = linalg.addf
+    //     %{{.+}} = linalg.generic
+    //        %{{.+}} = linalg.addf
+    //     %{{.+}} = linalg.generic 
+    //        %{{.+}} = arith.maximumf
```
adam-smnk, line 15: I'd remove the IR example, IMO it makes things more confusing.
Let the CHECK cases do the talking but the top level comment can still be present next to relevant test case.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -566,6 +566,19 @@ getTileForEltWiseConsumer(Operation *consumer, Operation *producer,
   return eltWiseTiles;
 }
 
+static bool hasActualOneUse(Value op) {
```
adam-smnk, line 569: More accurately, I'd call it `hasOneUser`

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -566,6 +566,19 @@ getTileForEltWiseConsumer(Operation *consumer, Operation *producer,
...
+  if (op.hasOneUse())
+    return true;
+  llvm::SmallSet<Operation *, 2> users;
+  for (auto &use : op.getUses()) {
+    users.insert(use.getOwner());
+  }
```
adam-smnk, line 575: It should be more efficient to just iterate over users and compare them directly. Sth like:
```
auto users = op.getUsers();
Operation *firstUser = *users.begin();
llvm::all_of(users, [&](Operation *user) { return user == firstUser; });
```

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -566,6 +566,19 @@ getTileForEltWiseConsumer(Operation *consumer, Operation *producer,
...
+
+  if (users.size() == 1)
+    return true;
+  return false;
```
adam-smnk, line 579:
```suggestion
  return users.size() == 1;
```

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
adam-smnk: I don't think generalization destroys the test case you're looking for i.e. `linalg.add ins(%A, %B) outs(%A)` generalizes to `linalg.generic ins(%A, %B) outs(%A)`. So, let's remove this change to the tile-and-fuse pass from this PR.

If you want to, let's try removing generalization in a separate PR. Then we can focus on impact on benchmarks etc.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -574,7 +587,7 @@ static Operation *getLastFusableEltWiseConsumer(
   Value linalgOpRes = linalgOp->getResult(0);
   // If we allow use, we may end up doing recomputation. Unclear if it is
   // profitablem thus disallow for now.
-  if (!linalgOpRes.hasOneUse())
```
adam-smnk, line 577: Please add simple test cases that validate this change to `tile-and-fuse.mlir` and `tile-and-fuse-named-op.mlir`.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -566,6 +566,13 @@ getTileForEltWiseConsumer(Operation *consumer, Operation *producer,
   return eltWiseTiles;
 }
 
+static bool hasOneUser(Value op) {
+  auto users = op.getUsers();
+  Operation *firstUser = *users.begin();
```
adam-smnk, line 571: It might crash on zero users.

include/TPP/Passes.td
```diff
@@ -155,7 +155,8 @@ def TileConsumerAndFuseProducers : Pass<"tile-consumer-and-fuse-producers",
            "Run fusion for the given number of iterations">,
     Option<"useForAll", "use-for-all", "bool", "true", "Use parallel forAll">,
     Option<"minTileFactor", "min-tile-factor", "int64_t", "2",
-           "Minimum factor between dimension size and a tile size">
+           "Minimum factor between dimension size and a tile size">,
+    Option<"keepNamedOp", "keep-named-op", "bool", "false", "Don't generalize named ops">
```
adam-smnk, line 159: Looks like the option is no longer really needed for anything. Could you remove it from this PR?
I would recommend opening a separate PR that removes `populateLinalgNamedOpsGeneralizationPatterns` all together from tile and fuse.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
adam-smnk: nit: there could be also early exit on `hasOneUse`

test/Passes/tile-and-fuse-named-op.mlir
```diff
@@ -129,3 +129,34 @@ func.func @matmul_sequence_fusion_with_relu(%arg0: tensor<32x64xf32>, %arg1: ten
 // CHECK-NEXT: }
 
 // -----
+
+func.func @matmul_chain_multi_use_into_relu(%arg0: tensor<32x64xf32>, %arg1: tensor<64x32xf32>,
```
adam-smnk, line 133: Is there any negative case with two users that would prevent fusion?
If not, it'd be nice to add to validate that path too.

shahidact: added a negative test.

adam-smnk (approved):

PR #947, opened 2024-07-23
Generalize before packing propagation

adam-smnk:
Shifts tpp-mapping named ops generalization step to allow packing propagation when input IR contains named Linalg ops. Currently, upstream propagation passes only work on linalg.generics.


adam-smnk:
Regressions in PyTorch benchmarks.

adam-smnk:
Requires more investigation, closing for now.

PR #950, opened 2024-08-06
[DLTI] Enable obtaining tile size from module-associated attr

rolfmorel:
By way of demonstration, this change only enables obtaining the tile size in the -tile-consumer-and-fuse-producers pass.

### DIFF
--- a/lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
+++ b/lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
@@ -449,7 +449,32 @@ static FailureOr<scf::SCFTileAndFuseResult> fuseWithEltwise(
 // Trivial tile selection. If the dimension is statically known, it perfectly
 // divides the tile, and we have enough iterations return a default of 32.
 static int64_t getTileForDim(linalg::LinalgOp linalgOp, unsigned dim) {
-  const int64_t tile = 32;
+  int64_t tile = 32;
+
+  // Check if a tile size hint is associated to the IR via DLTI.
+  auto deriveFromDLTI = [&](ModuleOp moduleOp) {
+    if (!moduleOp)
+      return;
+    TargetSystemSpecInterface sysSpec = moduleOp.getTargetSystemSpec();
+    if (!sysSpec)
+      return;
+    auto deviceId = StringAttr::get(linalgOp->getContext(), "CPU");
+    auto deviceSpec = sysSpec.getDeviceSpecForDeviceID(deviceId);
+    if (!deviceSpec)
+      return;
+    auto tileSizeId = StringAttr::get(linalgOp->getContext(), "tile_size");
+    DataLayoutEntryInterface entry =
+        (*deviceSpec).getSpecForIdentifier(tileSizeId);
+    if (!entry)
+      return;
+    Attribute value = entry.getValue();
+    if (auto intAttr = llvm::dyn_cast<IntegerAttr>(value))
+      tile = intAttr.getInt();
+    // TODO: might want to print a warning if tile_size exists as a key but the
+    //       associated attribute has an unexpected type.
+  };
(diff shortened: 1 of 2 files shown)


lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -449,7 +449,32 @@ static FailureOr<scf::SCFTileAndFuseResult> fuseWithEltwise(
 // Trivial tile selection. If the dimension is statically known, it perfectly
 // divides the tile, and we have enough iterations return a default of 32.
 static int64_t getTileForDim(linalg::LinalgOp linalgOp, unsigned dim) {
-  const int64_t tile = 32;
+  int64_t tile = 32;
+
+  // Check if a tile size hint is associated to the IR via DLTI.
+  auto deriveFromDLTI = [&](ModuleOp moduleOp) {
```
rengolin, line 455: I'd move most of that to a helper where you pass device id, spec id and it returns an attribute. This should eventually be upstream in some query helper.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -449,7 +449,32 @@ static FailureOr<scf::SCFTileAndFuseResult> fuseWithEltwise(
...
+      return;
+    auto tileSizeId = StringAttr::get(linalgOp->getContext(), "tile_size");
+    DataLayoutEntryInterface entry =
+        (*deviceSpec).getSpecForIdentifier(tileSizeId);
+    if (!entry)
+      return;
+    Attribute value = entry.getValue();
+    if (auto intAttr = llvm::dyn_cast<IntegerAttr>(value))
+      tile = intAttr.getInt();
+    // TODO: might want to print a warning if tile_size exists as a key but the
```
rengolin, line 473: I'd add a simple checker for `tile % 2 == 0` and `2 < tile < 128` or something.

test/Passes/tile-and-fuse-default.mlir
```diff
@@ -729,3 +729,58 @@ func.func @contraction(%arg0: tensor<16x1xf32>, %arg1: tensor<1x32xf32>) -> tens
 
 // CHECK-LABEL: contraction
 // CHECK-NOT: scf.for
+
+// -----
+
+// CHECK-LABEL: dlti_tile_size_32
+module attributes { dlti.target_system_spec = #dlti.target_system_spec<"CPU" : #dlti.target_device_spec<#dlti.dl_entry<"tile_size", 32 : i32>>> } {
```
rengolin, line 736: We usually want to pass more than one tile dim. How about `tile_size` gets replicated across all dims, and `tile_0`, `tle_1`, etc are specific? @adam-smnk

adam-smnk: Ideally, I agree we should have more fine-grained control. But it depends how far we want to push this integration right now.
I think it's good enough for now. For basic benchmarking with nice shapes it should be fine to set all tiles to 64, for example.

To do it really "properly" we'd have to define what convention we want to follow, what these tiles even mean etc.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
rolfmorel: Yes, that's what I would like as well as an API.The issue I was having is with communicating back why a lookup failed. A utility function like that has many failure modes. Should we have two util functions, or a flag, whereby one behaviour is to just return a `Optional<Attribute>`/`FailureOr<Attribute>` and the other one does that and emits messages for each distinct failure case.

rengolin: Fair enough, we can start this way for now and evolve into that.

rengolin (approved):

PR #951, opened 2024-08-06
Generalize before propagation

adam-smnk:


adam-smnk:
Most important generalization is now covered by the added folding patterns.

Technically, this further named ops generalization opens up more fusion opportunities when named ops are present.
However, current performance numbers don't reflect much change and there's some regression in the pytorch benchmarks.

For now keeping it as a draft for further investigation.

adam-smnk:
Not really needed, closing for now.

PR #952, opened 2024-08-07
mlir-gen - use empty as outs for named eltwise operations

adam-smnk:
Adjusts mlir-gen named ops generation to use tensor.empty as output destination.

For linalg named eltwise operations, outs value is write only. Using empty as a destination placeholder instead of zero initialized value improves bufferization and overall IR quality.


adam-smnk:
This also aligns our named benchmarks with more realistic IR like one generated by OV.

rengolin:
> This also aligns our named benchmarks with more realistic IR like one generated by OV.

Any change in benchmark numbers?

adam-smnk:
> > This also aligns our named benchmarks with more realistic IR like one generated by OV.
> 
> Any change in benchmark numbers?

Not as of right now that is: no changes in the older benchmarks and no impact on the named case.
This is due to other limitations in lowering named MLPs.

PR #953, opened 2024-08-07
Fold into eltwise pass

adam-smnk:
Adds pass that folds operations into elementwise Linalg ops. Combined operations are rewritten into a single linalg.generic.

Initially, a pattern for folding broadcast producer into an elementwise consumer is added.
This allows to improve layout propagation, allows for fusion after tiling, and reduces the number of temporary buffers by eliminating explicit broadcast.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -504,4 +504,15 @@ def LinalgToXeGPU : Pass<"linalg-to-xegpu", "func::FuncOp"> {
   ];
 }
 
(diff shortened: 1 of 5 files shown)


rengolin (approved):
Nice. I guess this could be an upstream pass. But it would be easier to get it merged once we have the broadcast semantics added to element-wise named ops.

adam-smnk:
> this could be an upstream pass.

Yes, it should be pretty generic but

> But it would be easier to get it merged once we have the broadcast semantics added to element-wise named ops.

Agreed here. With these new semantics it would be much cleaner and we can avoid going down to generic.

PR #954, opened 2024-08-08
Fold fill into max

adam-smnk:
Adds pattern that folds linalg.fill into linalg.max and outputs combined linalg.generic.

A constant filled buffer is replaced by a single constant used directly in max operation on the elements of the other operand.
This allows to eliminate potential temporary buffer allocation and value initialization.

### DIFF
--- a/lib/TPP/Transforms/FoldIntoEltwise.cpp
+++ b/lib/TPP/Transforms/FoldIntoEltwise.cpp
@@ -112,11 +112,75 @@ struct BroadcastIntoEltwise
 // into:
 //   linalg.generic
 //     arith.max %in, %cst
+//
+// NOTE: This could be achieved by controlled generalization + linalg eltwise
+//       operation fusion using the upstream linalg fusion pass and the upstream
+//       `linalg-inline-scalar-operands` pass extended to support scalar
+//       non-shaped inputs.
+// TODO: Linalg eltwise fusion pass could be generalized to operate on LinalgOp
+//       instead of GenericOp.
+// TODO: Extend linalg scalar inlining pass.
 struct FillIntoMax : public OpRewritePattern<linalg::MaxOp> {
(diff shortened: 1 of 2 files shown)


rengolin (commented):
Why is this restricted to max and not just any element-wise op?

adam-smnk:
> Why is this restricted to max and not just any element-wise op?

I think this could be easily relax to all named ops. Processing generics is also doable but hassle.
All in all, I just didn't want to bother for now.

adam-smnk:
Fill folding together with broadcast folding finally improves named ops benchmarks.
The remaining slowdown is caused by one more temporary buffer allocation. It most likely caused by the folded max generic which is not in in-place format (outs to tensor.empty). The next step is to rewrite the max generic same as we do with `linalg-convert-add-in-place` for generic adds.

lib/TPP/Transforms/FoldIntoEltwise.cpp
```diff
@@ -105,10 +105,93 @@ struct BroadcastIntoEltwise
...
+        inputs.push_back(input);
+        indexingMaps.push_back(
+            maxOp.getMatchingIndexingMap(maxOp.getDpsInputOperand(idx)));
+        continue;
+      }
+
+      // TODO: Check if there is any benefit in this over passing scalar
+      //       as an input argument.
+      // Store the constant separately to be later inserted directly into
+      // generic's body.
```
adam-smnk, line 155: For now I wanted to have the constant inlined but actually I'm not sure if it makes any real difference.
If it doesn't, this pass could be simplified and easy generalized.

rengolin (approved):

adam-smnk:
Looks like the two folders could be merged, it just requires more testing to see if I missed any edge cases.
I'll follow up on that a bit later.

PR #955, opened 2024-08-08
Convert unary generic to inplace form

adam-smnk:
Adds pattern to rewrite unary linalg.generic op to operate inplace that is to replace the output with its input when possible.
Additionally, `linalg-convert-add-in-place` is refactored and generalized into `convert-linalg-to-inplace` to enable the pass to hold various patterns.

The new pattern allows to favor inplace bufferization when possible.


adam-smnk:
@rengolin some troubles with CI atm.
But with this rewrite it should be enough input IR preprocessing to reach the same lowering for both named and generic MLPs. Of course, for the baseline model without transposition.

rengolin (commented):
We also need to make sure that the `empty` does not have other users, otherwise the other user will lose its DPS buffer.

adam-smnk:
> We also need to make sure that the `empty` does not have other users, otherwise the other user will lose its DPS buffer.

That shouldn't matter at the tensor-level. `outs` only represent the initial values which can be ignored if they're not read by the op. So, I'm free to change it as data flow follows results.
Plus the fact that CSE can collapse all empties into one also supports that IMO.

rengolin:
> That shouldn't matter at the tensor-level. `outs` only represent the initial values which can be ignored if they're not read by the op. So, I'm free to change it as data flow follows results. Plus the fact that CSE can collapse all empties into one also supports that IMO.

Right, you're not deleting the `empty`, that should be fine, yes.

rengolin (approved):

adam-smnk:
Rebased on top of #956 to validate benchmarks.
This PR fixes PyTorch performance regression of #956 and improves performance of the named ops benchmarks (`mlp_named_ops`). The performance now matches their generic counterparts (`mlp_models`).

adam-smnk:
One more rebase back on top of current main and validation.

PR #956, opened 2024-08-09
Support inplace generic in compare-select to max rewrite

adam-smnk:
Simplifies conversion to only change generic's body operation without affecting operation operands. This prevents crashes when generic has inplace format.

Originally the pass would also implicitly rewrite generic operation to be in place by replacing output with the input. This is removed to improve robustness and better separate concerns. Inplace rewrite will be handled by a separate pass later on.


adam-smnk:
PyTorch benchmarks will take a temporary hit in performance. This will be fixed by #955.

rengolin (approved):
Good, the less we change in generics the better. Cleanups are better as a separate pass, like #955

PR #957, opened 2024-08-19
Bump LLVM

adam-smnk:
Fixes for upstream API changes and improves tests' checks.

Improves CI build scripts and recovery from LLVM build failures.
Moves build directory to prevent running out of space on local disk during concurrent builds.


adam-smnk:
Finally works. No benchmark regressions.

rengolin (approved):

PR #958, opened 2024-08-29
Linalg->Vector dialect pass and pipeline changes

KavithaTipturMadhu:
Working on test cases, will update shortly, only small tensors lower right now as vectorization lowers them to extract-insert operations, need to investigate what happens at lowering. Also, options need to be added for vector length and other command line arguments supported upstream. Working on those, will update the PR asap.


KavithaTipturMadhu:
One set of tests added, pr is good for review, will add e2e tests for correctness and cli options next.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -97,22 +98,38 @@ struct DefaultTppPasses
...
+      // Low leve parallelization passes.
+      pm.addPass(createLowLevelParallelization(LowLevelParallelization));
+      pm.addNestedPass<func::FuncOp>(createIntelAMXTileConfigInsertionPass());
+      pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+      pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
+      pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+      pm.addNestedPass<func::FuncOp>(createIntelAMXTileConfigHoistingPass());
+
+      // Covert all local TPP-related dialects.
+      pm.addPass(createLocalDialectsLowering());
```
adam-smnk, line 131: Even when going to vectors, we still need that for perf and check.

include/TPP/Passes.td
```diff
@@ -43,6 +43,28 @@ def ConvertLinalgToFunc : Pass<"convert-linalg-to-func", "ModuleOp"> {
...
+  let summary = "convert linalg to memref/vector";
+  let description = [{
+    Vectorization pass to convert linalg to memref/vector instead of Xsmm.
+  }];
+
+  let dependentDialects = [ "memref::MemRefDialect", "linalg::LinalgDialect", "vector::VectorDialect" ];
+}
+
+
+def VectorContractPass : Pass<"vector-contract-pass",
```
adam-smnk, line 57: Do we ever want to not run these patterns?
I'd merge it with `VectorizationPass`.

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+
+struct VectorizationPass
+    : public impl::VectorizationPassBase<VectorizationPass> {
+
+  void populateCombinePatterns(RewritePatternSet &patterns) {
+    patterns.add<LinalgToVector<linalg::BatchReduceMatmulOp>,
+                 LinalgToVector<linalg::FillOp>>(patterns.getContext());
+    patterns.add<LinalgGenericToVector>(patterns.getContext());
```
adam-smnk, line 99: These patterns need more preconditions e.g., check for buffer semantics, currently there are hardcoded assumptions on the affine maps.

include/TPP/Passes.td
```diff
@@ -43,6 +43,28 @@ def ConvertLinalgToFunc : Pass<"convert-linalg-to-func", "ModuleOp"> {
...
+                                     "func::FuncOp"> {
+  let summary = "convert vector rductions to contractions";
```
rolfmorel, line 59: Typo: "rductions" -> "reductions"

test/Passes/pass-vectorization.mlir
```diff
@@ -0,0 +1,64 @@
+// RUN: tpp-opt %s --tile-consumer-and-fuse-producers -bufferize --vectorization-pass --split-input-file | FileCheck %s
```
adam-smnk, line 1: Tests should pull only what's necessary i.e., `-vectorization-pass` and canonicalization/cse if desired.
Could you update the test cases that they are in such shape the vectorization pass can consume directly?
Also a few negative test cases could be useful.

test/Passes/pass-vector-contract.mlir
adam-smnk, line 1: Same comment as the other test, it should call only `-vector-contract-pass`.
Overall, it is only a wrapper for upstream passes so, testing can be very symbolic.

adam-smnk: Or to be removed all together if this pass is not needed separately.

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace mlir {
+namespace tpp {
+
+struct LinalgGenericToVector : OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (xsmm::utils::getDataType(rewriter, linalgOp.getOperand(0).getType()) ==
```
adam-smnk, line 38: I take this is about VNNI layout?
It should be a separate pass as it is unrelated to vectorization itself.

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return linalg::vectorize(rewriter, linalgOp);
+  }
+};
+
+struct VectorizationPass
+    : public impl::VectorizationPassBase<VectorizationPass> {
+
+  void populateCombinePatterns(RewritePatternSet &patterns) {
+    patterns.add<LinalgToVector<linalg::BatchReduceMatmulOp>,
+                 LinalgToVector<linalg::FillOp>>(patterns.getContext());
```
adam-smnk, line 98: Do we want to be selective with vectorization or should be vectorize all linalg ops?
If the latter, you could have a more generic pattern using `OpInterfaceRewritePattern<linalg::LinalgOp>`

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,109 @@
+//===- Vectorization.cpp -----------------------------------------*- C++-*-===//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements tile configuration hoisting on parallel loops.
```
adam-smnk, line 8: nit: outdated?

lib/TPP/Transforms/VectorContract.cpp
```diff
@@ -0,0 +1,43 @@
+//===- VectorContractPass.cpp -----------------------------------*- C++-*-===//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements tile configuration hoisting on parallel loops.
```
adam-smnk, line 8: nit: outdated?

test/Passes/pass-vectorization.mlir
```diff
@@ -0,0 +1,64 @@
...
+
+#map = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
+module {
+  func.func @entry(%arg0: tensor<4x4x4x4xf32>, %arg1: tensor<4x4x4x4xf32>, %arg2: tensor<4x4x4x4xf32>) -> tensor<4x4x4x4xf32> {
```
adam-smnk, line 7: I'd suggest using different sizes for each dimension for readability and it's easier to spot if everything works as intended.

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,109 @@
...
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (xsmm::utils::getDataType(rewriter, linalgOp.getOperand(0).getType()) ==
+            xsmm::DataTypeAttr::get(rewriter.getContext(),
+                                    xsmm::DataType::BF16) &&
+        linalgOp.getIteratorTypes().size() >= 5) {
+      SmallVector<int64_t> shape;
+      SmallVector<ReassociationIndices> indices;
+      int index = 0;
+      for (size_t i = 0; i < dyn_cast<ShapedType>(linalgOp.getOperand(0).getType())
```
rolfmorel, line 45: Style-guide [nit](https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop): 
`for (size_t i = 0, end = dyn_cast<ShapedType>(linalgOp.getOperand(0).getType()).getShape().size(); i < end; i++)`

lib/TPP/Transforms/Vectorization.cpp
KavithaTipturMadhu: As discussed yesterday, this is specifcally for VNNI for vectorization, hence I am not moving this out.

lib/TPP/Transforms/Vectorization.cpp
KavithaTipturMadhu: Adding this in a future PR for other ops that are vectorizable.

lib/TPP/Transforms/Vectorization.cpp
KavithaTipturMadhu: Added bufferizability check, currently works with and without tiling pass.

test/Passes/pass-vectorization.mlir
KavithaTipturMadhu: Removed other passes, will add negative tests.

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,118 @@
...
+namespace tpp {
+
+struct LinalgGenericToVector : OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (!linalgOp.hasPureBufferSemantics())
+      return failure();
+    if (xsmm::utils::getDataType(rewriter, linalgOp.getOperand(0).getType()) ==
```
adam-smnk, line 42: This should further restrict what is matched. I suspect it expects only contraction.
Otherwise, the pass crashes on eltwise generic with 5 dims.

lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace mlir {
+namespace tpp {
+
+struct LinalgGenericToVector : OpRewritePattern<linalg::GenericOp> {
+  using OpRewritePattern<linalg::GenericOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(linalg::GenericOp linalgOp,
+                                PatternRewriter &rewriter) const override {
+    if (!linalgOp.hasPureBufferSemantics())
```
adam-smnk, line 40: Please add a negative test that covers this case.

test/Passes/pass-vector-contract.mlir
adam-smnk, line 1: Update file name.

lib/TPP/Transforms/Vectorization.cpp
KavithaTipturMadhu: Its not just contractions, but vnni transformation expects at least 2 operands with the second operand in VNNI layout. I have added a check with your test case added into the lit test.

KavithaTipturMadhu:
Added bf16 end to end test case. The pr is good to go @adam-smnk @rolfmorel. Thanks @rengolin for the pointers on fpcmp.

tools/fpcmp/fpcmp.c
```diff
@@ -0,0 +1,446 @@
...
+  double V1 = 0.0, V2 = 0.0;
+
+  // Fast character-by-character comparison of the numbers.
+  if (F1Len == F2Len && memcmp(F1P, F2P, F1Len) == 0)
+    return false;
+
+  // Note that some ugliness is built into this to permit support for numbers
+  // that use "D" or "d" as their exponential marker, e.g. "1.234D45".  This
+  // occurs in 200.sixtrack in spec2k.
+  V1 = strtod(F1P, (char **)(&F1NumEnd));
```
adam-smnk, line 134: These casts produce warnings `drops const qualifier`

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -97,19 +98,35 @@ struct DefaultTppPasses
...
 
     // Convert forAll to parallel loops should run after bufferization
     // as scf.parallel does not handle tensor.
     pm.addPass(createConvertForAllToParallelOp());
-
-    // Low leve parallelization passes.
     LowLevelParallelizationOptions LowLevelParallelization{parallelTaskGrid};
-    pm.addPass(createLowLevelParallelization(LowLevelParallelization));
 
+    if (linalgToVector) {
```
rengolin, line 116: Please add a TODO here to disentangle this mess. We need to move away from heavy branching in the default pipeline or it will be hard to compose with other pipelines/transform schedules.

lib/TPP/PassBundles/LocalDialectsLowering.cpp
```diff
@@ -58,7 +58,6 @@ struct LocalDialectsLowering
   void constructPipeline() override {
     pm.addNestedPass<func::FuncOp>(createConvertCheckToLoops());
     pm.addNestedPass<func::FuncOp>(createConvertPerfToLoops());
-    pm.addPass(createConvertXsmmToFunc());
     pm.addPass(createConvertPerfToFunc());
```
rengolin, line 61: Why do you remove this here but add it in the test?

lib/TPP/PassBundles/LowLevelParallelization.cpp
```diff
@@ -65,11 +65,5 @@ struct LowLevelParallelization
     mlir::tpp::SCFParallelLoopTilingOptions tilingOptions;
     tilingOptions.tileSizes = parallelTaskGrid;
     pm.addPass(createSCFParallelLoopTiling(tilingOptions));
-
-    pm.addNestedPass<func::FuncOp>(createIntelAMXTileConfigInsertionPass());
-    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-    pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
```
rengolin, line 71: We need to avoid moving passes out of bundles into the pipeline itself. Perhaps this is just an artifact of the duality xsmm/vectorization but it needs understanding if this is the best way through this.

test/Integration/vectorization-mlp.mlir
```diff
@@ -0,0 +1,5 @@
+// RUN: export vectorFile=$(mktemp)
+// RUN: mlir-gen   --kernel=const --bias --relu --layers=16,16 --tiles=4,4,4 -float-type=bf16 --vnni=2  | tpp-run -linalg-to-vector -e entry --entry-point-result=void -print -n 1 --seed=123 2>&1 > $vectorFile
```
adam-smnk, line 2: nit: maybe this shouldn't use `--vnni` flag to also check correct layout changes

adam-smnk (approved):
Looks fine 👍
We can do more cleanup as part of and after the next few vectorization PRs.

lib/TPP/PassBundles/LocalDialectsLowering.cpp
KavithaTipturMadhu: Added a comment in the file.

rengolin (approved):

PR #961, opened 2024-09-04
Allow packing matmul dims equal to block size

adam-smnk:
Relaxes matmul packing blocked dimensions validation to allow packing when `matmul dims >= block factor`.

PR #962, opened 2024-09-04
Removed unstable e2e test

KavithaTipturMadhu:
https://github.com/plaidml/tpp-mlir/actions/runs/10701176077/job/29666570826


KavithaTipturMadhu:
We need to discuss how to fix this, till then deleting this test from tpp-mlir.

adam-smnk:
The test initially succeeded on #961 and then failed after no changes to overall pipeline logic (push of lit test case). So, we suspect instability to be somewhere around this test itself. It might be caused by the test's temp file creation.

rengolin:
We had this problem before, we need some better way to do this, because that's a very important type of test. Perhaps a small Python script that does that internally in a transactional and safe way.

rolfmorel:
@KavithaTipturMadhu, @adam-smnk, @rengolin 

How about using `llvm-lit`'s temp file facility for this? I think the following might work:
`mlir-gen ... | tpp-run > %t` and the other does `mlir-gen ... | tpp-run ... | fp-comp - %t`

Disclaimer: I have not used this before. The documentation on the feature is minimal, e.g. whether it does proper clean-up: https://llvm.org/docs/CommandGuide/lit.html#substitutions

PR #965, opened 2024-09-06
Deprecate Vulkan support

adam-smnk:
Removes Vulkan support and related passes.
Lowers maintenance cost and allows future GPU pipeline cleanup.

The Vulkan GPU execution path has shown limitation and has been largely unused for a long time now. For Nvidia, the native CUDA backend is more complete and performant. For Intel, various alternative backends are now available and will become the primary way to execute on Intel GPUs.


adam-smnk:
Bump

PR #966, opened 2024-09-07
Add a fold-add-into-dest pattern and pass

rolfmorel:
Achieves the sum of linalg.add by relying on the accumulation semantics of contraction ops w.r.t. their out/dest argument. Main value is likely to come from that a buffer allocation can be elided, in case the contraction op had its own buffer.

### DIFF
--- a/lib/TPP/Transforms/FoldAddIntoDest.cpp
+++ b/lib/TPP/Transforms/FoldAddIntoDest.cpp
@@ -9,8 +9,10 @@
 #include "TPP/Passes.h"
 #include "TPP/Transforms/Utils/ValueUtils.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/IR/Dominance.h"
+#include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 using namespace mlir;
 
@@ -62,14 +64,19 @@ struct FoldAddIntoDestRewrite : public OpRewritePattern<linalg::AddOp> {
     // When the dominated op has a single-result and is a contraction, the op
     // accumulates on the out argument, starting from the supplied out argument.
     // E.g., AddOp is not a contraction and hence ignores its out arg's value.
+    auto dominatedDestOp =
+        dyn_cast<DestinationStyleOpInterface>((Operation *)dominatedOp);
     if (dominatedOp->getNumResults() != 1 ||
-        !linalg::isaContractionOpInterface(dominatedOp))
+        !linalg::isaContractionOpInterface(dominatedOp) ||
+        (!dominatedDestOp || dominatedDestOp.getNumDpsInits() != 1))
       return rewriter.notifyMatchFailure(
-          dominatedOp, "expected dominated op to be single-result contraction");
+          dominatedOp, "expected dominated op to be single-result "
+                       "destination-passing contraction");
 
     // As the dominated op was already accumulating on its out argument, it is
     // only safe to discard its current out arg when it is the additive zero.
-    auto *dominatedDest = dominatedOp->getOperand(2).getDefiningOp();
+    auto *dominatedDest =
+        dominatedDestOp.getDpsInitOperand(0)->get().getDefiningOp();
     if (!mlir::utils::isZeroOp(dominatedDest))
       return rewriter.notifyMatchFailure(
           dominatedOp, "expected dominated op's dest to be additive zero");
--- a/lib/TPP/Transforms/FoldAddIntoDest.cpp
+++ b/lib/TPP/Transforms/FoldAddIntoDest.cpp
@@ -73,16 +73,14 @@ struct FoldAddIntoDestRewrite : public OpRewritePattern<linalg::AddOp> {
                        "destination-passing contraction");
 
     // To change the contraction's result, `addOp` must be its only user.
-    auto contractionUsers = dominatedOp->getResult(0).getUsers();
-    if (std::distance(contractionUsers.begin(), contractionUsers.end()) != 1)
+    if (!dominatedOp->getResult(0).hasOneUse())
       return rewriter.notifyMatchFailure(
           dominatedOp,
-          "expected linalg.add op to be single user of contraction's result");
+          "expected linalg.add to be single user of contraction's result");
 
     // As `dominatedOp` was already accumulating on its out argument, it is only
     // safe to no longer use its current out arg when it is the additive zero.
     auto *destOperand = dominatedDestOp.getDpsInitOperand(0);
-    ;
     if (!mlir::utils::isZeroOp(destOperand->get().getDefiningOp()))
       return rewriter.notifyMatchFailure(
           dominatedOp, "expected dominated op's dest to be additive zero");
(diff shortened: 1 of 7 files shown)


rolfmorel:
This PR is now ready for review.

Do let me know if adding it to the default pipeline is appropriate and whether I added it in the right place.

adam-smnk:
Could you add some lit tests?

rolfmorel:
> Could you add some lit tests?

Oops - forget to push. Pushed them now.

lib/TPP/Transforms/FoldAddIntoDest.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+namespace {
+
+/// Replace a linalg.add where its linalg-contraction operand - with a
+/// zero-filled destination - is dominated by the `other` linalg operand,
+/// by passing `other` as the destination of the contraction.
+struct FoldAddIntoDestRewrite : public OpRewritePattern<linalg::AddOp> {
```
adam-smnk, line 31: I think the add can't be folded when the contraction has other users. Consider this example:
```
%0 = linalg.sub %A, %B
%1 = linalg.matmul %C, %D, %zero
%2 = linalg.add %0, %1
%3 = linalg.mul %1, %2
```

rolfmorel: Good catch. This is fixed now, with a test case.

lib/TPP/Transforms/FoldAddIntoDest.cpp
```diff
@@ -0,0 +1,110 @@
...
+      //     region there is no need to perform checks on addOp's out argument.
+    }
+
+    // When the dominated op has a single-result and is a contraction, the op
+    // accumulates on the out argument, starting from the supplied out argument.
+    // E.g., AddOp is not a contraction and hence ignores its out arg's value.
+    auto dominatedDestOp =
+        dyn_cast<DestinationStyleOpInterface>((Operation *)dominatedOp);
+    if (dominatedOp->getNumResults() != 1 ||
+        !linalg::isaContractionOpInterface(dominatedOp) ||
```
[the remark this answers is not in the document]
rolfmorel: Another good catch.

Of course `linalg.add` indeed uses an identity mapping for its operands. 

As far as I understand, for the dest/out of the contraction, this corresponds to an ordered projection (as some of the affine maps' inputs will be contracted out). Hence I have now implemented a check for whether the dest's affine map is an ordered projection.

If you could check if this indeed captures what you had in mind/the semantics for affine_maps on contractions - that would be great!

test/Passes/fold-add-into-dest.mlir
```diff
@@ -0,0 +1,135 @@
...
+!type = tensor<2048x2048xf32>
+func.func @expect_no_fold_as_operands_do_not_dominate_each_other(%arg0: !type, %arg1: !type) -> !type {
+  %0 = arith.constant dense<1.111111e+00> : !type
+  %cst = arith.constant 0.000000e+00 : f32
+  %1 = tensor.empty() : !type
+  %2 = linalg.fill ins(%cst : f32) outs(%1 : !type) -> !type
+  %3 = linalg.matmul_transpose_b ins(%arg0, %0 : !type, !type) outs(%2 : !type) -> !type
+  %4 = linalg.add ins(%3, %3 : !type, !type) outs(%1 : !type) -> !type
+  return %4 : !type
+}
```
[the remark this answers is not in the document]
rolfmorel: Fixed.

lib/TPP/Transforms/FoldAddIntoDest.cpp
```diff
@@ -0,0 +1,131 @@
...
+        (!dominatedDestOp || dominatedDestOp.getNumDpsInits() != 1))
+      return rewriter.notifyMatchFailure(
+          dominatedOp, "expected dominated op to be single-result "
+                       "destination-passing contraction");
+
+    // To change the contraction's result, `addOp` must be its only user.
+    auto contractionUsers = dominatedOp->getResult(0).getUsers();
+    if (std::distance(contractionUsers.begin(), contractionUsers.end()) != 1)
```
adam-smnk, line 77: nit: use `hasOneUse` instead

lib/TPP/Transforms/FoldAddIntoDest.cpp
```diff
@@ -0,0 +1,131 @@
...
+      return rewriter.notifyMatchFailure(
+          dominatedOp,
+          "expected linalg.add op to be single user of contraction's result");
+
+    // As `dominatedOp` was already accumulating on its out argument, it is only
+    // safe to no longer use its current out arg when it is the additive zero.
+    auto *destOperand = dominatedDestOp.getDpsInitOperand(0);
+    ;
```
adam-smnk, line 85: nit: extra char

lib/TPP/Transforms/FoldAddIntoDest.cpp
```diff
@@ -0,0 +1,131 @@
...
+    SmallVector<AffineMap> indexMaps = dominatedOp.getIndexingMapsArray();
+    int prevDimPos = -1;
+    for (auto expr : indexMaps[destOperand->getOperandNumber()].getResults()) {
+      auto dim = dyn_cast<AffineDimExpr>(expr);
+      if (!dim || prevDimPos >= (int)dim.getPosition())
+        return rewriter.notifyMatchFailure(
+            dominatedOp, "expected index_map for contraction's dest to be an "
+                         "ordered projection");
+      prevDimPos = dim.getPosition();
+    }
```
adam-smnk, line 105: nit: you can use `AffineMap::isIdentity()`

lib/TPP/Transforms/FoldAddIntoDest.cpp
```diff
@@ -0,0 +1,131 @@
...
+    // safe to no longer use its current out arg when it is the additive zero.
+    auto *destOperand = dominatedDestOp.getDpsInitOperand(0);
+    ;
+    if (!mlir::utils::isZeroOp(destOperand->get().getDefiningOp()))
+      return rewriter.notifyMatchFailure(
+          dominatedOp, "expected dominated op's dest to be additive zero");
+    // TODO: If the other op is a contraction and has additive zero as dest, we
+    // can swap the dests and achieve the proper sum, given suitable dominance.
+
+    // As an operand to `addOp`, `dominatingOperand` has an identity affine_map.
```
adam-smnk, line 92: Could you add a test case for that?

lib/TPP/Transforms/FoldAddIntoDest.cpp
adam-smnk: Ah never mind, I thought there's a variant that could do it but there's only `minorIdentity`.

rolfmorel: Tried this yesterday already. This is not the case as for a contraction the affine_map would never be the identity map for the whole iteration space.

See my comment from yesterday up above.

lib/TPP/Transforms/FoldAddIntoDest.cpp
rolfmorel: Done.

lib/TPP/Transforms/FoldAddIntoDest.cpp
rolfmorel: I was looking for an named op which had this property but found none (
All the `ContractionOpInterface`-implementing linalg ops have that their dest/out argument has an ordered projection as indexing map (check for `LinalgContractionOpInterface` in `mlir/include/mlir/Dialect/Linalg/IR/LinalgNamedStructuredOps.yaml`).)

Have now implemented a matmul-like test case with a linalg.generic that does a transpose on C. Added a matmul-as-linalg.generic test case as well to show linalg.generics also work, even though they do not implement the ContractionOpInterface.

adam-smnk (approved):

PR #970, opened 2024-09-13
GPU vectorize pass

adam-smnk:
Adds GPU kernel vectorization pass.
Extends CUDA lowering pass to support vector operations.

GPU-specific vectorization pass that guides upstream Linalg vectorizer to process operations within a GPU kernel or prepared for outlining. CUDA-specific pass gets extended to allow lowering of vector ops within GPU kernel.

The vectorization is for now disabled within the GPU pipeline due to lack of vector operation unrolling. When vector sizes exceed hardware supported lengths, pipeline gets stuck on GPU binary compilation step. This will be addressed by a separate transformation pass in the future.

### DIFF
--- a/lib/TPP/GPU/GpuToCuda.cpp
+++ b/lib/TPP/GPU/GpuToCuda.cpp
@@ -73,13 +73,8 @@ private:
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertNVGPUToNVVMPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertGpuOpsToNVVMOps());
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertVectorToLLVMPass());
-    pm.addNestedPass<gpu::GPUModuleOp>(createConvertSCFToCFPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertNVVMToLLVMPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertFuncToLLVMPass());
-    pm.addNestedPass<gpu::GPUModuleOp>(
-        memref::createExpandStridedMetadataPass());
-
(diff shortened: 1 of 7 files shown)


lib/TPP/GPU/GpuToCuda.cpp
```diff
@@ -67,18 +67,34 @@ struct GpuToCuda : public tpp::impl::GpuToCudaBase<GpuToCuda>,
...
+    pm.addNestedPass<gpu::GPUModuleOp>(createConvertVectorToSCFPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertSCFToCFPass());
 
-    // Create CUDA kernels.
-    pm.addNestedPass<gpu::GPUModuleOp>(createStripDebugInfoPass());
+    pm.addNestedPass<gpu::GPUModuleOp>(createConvertNVGPUToNVVMPass());
     pm.addNestedPass<gpu::GPUModuleOp>(createConvertGpuOpsToNVVMOps());
-    pm.addNestedPass<gpu::GPUModuleOp>(createReconcileUnrealizedCastsPass());
+    pm.addNestedPass<gpu::GPUModuleOp>(createConvertVectorToLLVMPass());
+    pm.addNestedPass<gpu::GPUModuleOp>(createConvertSCFToCFPass());
```
rengolin, line 76: Duplicate pass?

rengolin (approved):
Looks good with comment. Nice that the GPU pipeline is getting closer to the CPU one.

lib/TPP/GPU/GpuToCuda.cpp
adam-smnk: The first one is necessary to "finalize" vector lowering.
This one doesn't seem needed as of right now but I left it for completeness if any other passes generate scf.

rengolin: I'd remove it if not needed, so we can detect if we ever need and why in the future.

PR #972, opened 2024-09-16
Transform to lower packs/unpacks without transpose, except when the packing occurs on a constant

rolfmorel:
A transform that enables selectively "reverting" packing on `linalg.generic` operands and unpacking on the generic's results by lowering `tensor.pack` to `tensor.expand_shape` with identity permutation and `tensor.unpack` to `tensor.collapse_shape` with identity permutation.

The new pass `-lower-packs-unpacks-without-transpose` does this reverting on all (compatible) packed/unpacked `linalg.generic` operands/results, except if they are constants. Similarly, the CLI flag `--lower-pack-unpack-without-transpose` enables this pass in the default pipeline.


lib/TPP/Transforms/ConvertPackToExpandShape.cpp
```diff
@@ -0,0 +1,123 @@
...
+    associationIndices.emplace_back(SmallVector<int64_t>());
+    associationIndices.back().push_back(curDimIdx++);
+    if (llvm::is_contained(innerDimPos, idx))
+      associationIndices.back().push_back(curDimIdx++);
+  }
+
+  auto expandShape = rewriter.create<tensor::ExpandShapeOp>(
+      genericOp.getLoc(), normalizedType, packOp.getOperand(0),
+      ArrayRef(associationIndices));
+  rewriter.replaceAllOpUsesWith(packOp, expandShape);
```
adam-smnk: I don't think we can update other users. It might be best to only update specific generic's operand.
Or maybe it doesn't make sense to rewrite pack at all if there are multiple users. But then if these are multiple generic contractions then maybe it's still fine.

For us, restriction to only one user might be enough, in general I'd leave that for a user to decide.

lib/TPP/Transforms/ConvertPackToExpandShape.cpp
```diff
@@ -0,0 +1,123 @@
...
+  auto packInverseMap = AffineMap::getPermutationMap(
+      mlir::tensor::getPackInverseDestPerm(packOp), rewriter.getContext());
+  auto normalizedShape = applyPermutationMap(packInverseMap, packedShape);
+
+  auto normalizedType = packedType.clone(normalizedShape);
+  auto normalizedIndexingMap = packInverseMap.compose(affineMap);
+
+  auto innerDimPos = SmallVector<unsigned int>(packOp.getInnerDimsPos());
+
+  SmallVector<SmallVector<int64_t, 2>> associationIndices;
```
adam-smnk, line 55: nit: maybe use `ReassociationIndices` alias

lib/TPP/Transforms/ConvertPackToExpandShape.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+      SmallVector<Attribute> indexingMaps =
+          llvm::to_vector(genericOp.getIndexingMaps());
+      indexingMaps[idx] = AffineMapAttr::get(res->second);
+
+      genericOp.setIndexingMapsAttr(
+          ArrayAttr::get(rewriter.getContext(), indexingMaps));
+    }
+
+    if (auto unpackOp = llvm::dyn_cast<tensor::UnPackOp>(
```
adam-smnk, line 135: I'd move it to a separate pattern

rolfmorel: On the one hand, yes, it might be nice to have it available as a separate pattern. On the other hand, when we modify the out parameter of the linalg, the result type changes/needs to be changed as well. This in turn means the unpack afterwards needs to have _something_ done to it. In the context of a single pattern it is clear what this _something_ is: rewrite it to a collapse_shape. 

In case of separate patterns, do you have an idea what to do with the result of the linalg when we change the shape of the out parameter? If we don't do anything the IR is most likely left in an illegal state (i.e. types will not match up).

rolfmorel: When it is not the case that there is a single unpack user of the linalg, we probably shouldn't do the rewrite of the out parameter any way (as we don't know how the users are impacted by the shape change of the type). The only sensible thing seems to be to insert some op in between the linalg's result and all the users to invert the shape change.

lib/TPP/Transforms/ConvertPackToExpandShape.cpp
rolfmorel: Have now added requirement for single user for there to be a match.

We can choose to relax later.

lib/TPP/Transforms/ConvertPackToExpandShape.cpp
adam-smnk: For outs, we definitely need the whole chain pack->generic->unpack. If there's any user other than unpack then it probably doesn't make sense to change outs shape.
Also this case might require extra checks on the pack-unpack pair to see if they are compatible.

So my thinking is mainly that it'd be cleaner to move this extra logic to a separate pattern that focuses on outs, and this one focuses on ins. But maybe it doesn't make much difference.

rolfmorel: Done.

rolfmorel: If you could give an example of a compatibility issue for pack-unpack pairs, that would much appreciated. I had a think and came to the conclusion that as the generic's result type is shared by the pair, this is enough to force compatability.

adam-smnk: Not sure if you can fold a pair in which pack has different perms than unpack `outer_dims_perm = [1, 0] inner_dims_pos = [0, 1]`.

adam-smnk: There are also probably cases when you can fold pack but need to keep unpack. For example, maybe pack can have permutation which then is performed by the generic but unpack has to stay as you need extra data movement anyway.

But let's ignore these for now.

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
```diff
@@ -0,0 +1,125 @@
...
+  // CHECK: %[[COL:.+]] = tensor.collapse_shape %[[RES]] {{\[}}[0, 1], [2, 3]{{\]}} : tensor<4x32x8x32xf32> into tensor<128x256xf32>
+  // CHECK: return %[[COL]]
+
+// -----
+
+#map = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+#map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
+#map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
+func.func @only_keep_constant_packed(%arg0: tensor<128x512xf32>, %arg1: tensor<128x256xf32>) -> tensor<128x256xf32> {
+  %cst = arith.constant dense<1.000000e-03> : tensor<8x16x32x32xf32>
```
adam-smnk, line 42: I think this constant should have a pack op. Prepacked one doesn't test anything.

rolfmorel:
@adam-smnk , @rengolin 

This is now ready for final review.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+  LowerPackOnInputsWithoutTranspose(MLIRContext *context,
+                                    ControlFn controlFn = nullptr,
+                                    PatternBenefit benefit = 1)
+      : OpRewritePattern(context, benefit), controlFn(std::move(controlFn)) {}
+
+  LogicalResult matchAndRewrite(linalg::GenericOp genericOp,
+                                PatternRewriter &rewriter) const override {
+    bool modifiedAnOperand = false;
+    for (auto operandIdx : {0, 1}) {
```
adam-smnk, line 80: That's an oddly specific pattern. I think the pattern should be either more precise or more generic.

rolfmorel: This is based on your suggestion to have separate patterns.

adam-smnk: I mean the explicit operands 0 and 1

rolfmorel: I see - the current thing is incorrect in case of non-binary input generics. Before I had the guard that this only applied to ContractionInterface ops but I removed it as I thought it was not necessary.

Will fix this so that it iterates over all inputs of the generic (and correspondingly fix the output/unpack pattern).

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
rolfmorel: The pre-transform IR for this test is just `-pack-matmul` applied to a matmul where one of the arguments is a constant. The difference between this test and the test after it is exactly the difference between the pack having already been applied or not.

If this test is not useful (as has more to do with the integration between `-pack-matmul` and the new pass, let me know.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
```diff
@@ -0,0 +1,173 @@
...
+    bool destHasStaticShape = unpackDest.getType().hasStaticShape();
+
+    lowerPackWithoutTranspose(packOp, genericOp, /*operandIdx=*/2, rewriter);
+    auto res = linalg::lowerUnPack(rewriter, unpackOp);
+
+    // Set genericOp's result type to the adjusted type of the out parameter.
+    genericOp->getOpResult(0).setType(genericOp.getOperand(2).getType());
+
+    if (auto transposeOp = res->transposeOp) {
+      // Forget about the transpose introduced by lowerUnPack.
```
adam-smnk, line 132: This might not be safe to do if unpack has different permutations than pack.

adam-smnk: Or to put it more simply if you have sth like:
```
%packed = pack(<128x64>)
generic outs(%packed)
unpack into <64x128>
```

include/TPP/PassBundles.td
```diff
@@ -37,7 +37,10 @@ def DefaultTppPasses : Pass<"default-tpp-passes", "ModuleOp"> {
            "unsigned", "Grid-sizes for parallel tasks.">,
     Option<"linalgToVector", "linalg-to-vector",
            "bool", /*default=*/"false",
-           "Lower linalg directly to vector.">
+           "Lower linalg directly to vector.">,
+    Option<"lowerPackUnpackWithoutTranspose", "lower-pack-unpack-without-transpose-to-vector",
```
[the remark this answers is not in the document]
rolfmorel: Done.

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
rolfmorel: Converted it into actually running `-pack-matmul` on a matmul operating on a constant . This has a bit of integration test flavour, but not on the same scale as most tests in `test/Integration`.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
rolfmorel: Made both patterns agnostic to the arity. If you could check if this looks right, that would be appreciated.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
rolfmorel: Have made it such that the pack on the out param and the unpack on the corresponding result must specify the same permutation. Else the pattern does not apply.

lib/TPP/Transforms/ConvertPackToExpandShape.cpp
rolfmorel: Moved to two separate patterns. There's discussion below as to the safe permutations on unpack.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
```diff
@@ -0,0 +1,185 @@
...
+        continue;
+
+      auto packOp =
+          dyn_cast_if_present<tensor::PackOp>(outOperand.getDefiningOp());
+      auto unpackOp =
+          llvm::dyn_cast<tensor::UnPackOp>(*(result.getUsers().begin()));
+
+      if (!packOp || !packOp->hasOneUse() || !unpackOp ||
+          unpackOp.getInnerDimsPos() != packOp.getInnerDimsPos() ||
+          unpackOp.getOuterDimsPerm() != packOp.getOuterDimsPerm() ||
```
adam-smnk, line 128: nit: in case of identity `getOuterDimsPerm` can be empty so it's not completely reliable

adam-smnk (approved):
Logic looks solid, I'd still cleanup the tests a bit.

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
adam-smnk: That makes tests more unpredictable as unrelated tests start breaking when one of the passes changes due to coupling.

IMO, it's nice to have one (perhaps cleaned up) matmul test case as that's what we care about, otherwise you can verify edge cases on much simpler examples.

rolfmorel:
Performance on the lora-fragment benchmark in use for Openvino:

```
$ python -m xonsh lora-runner.xsh 
CONFIGS [[8], [16], [32], [64], [128], [256], [512], [1024], [2048], [4096], [8192]]
OV NO-MLIR [0.84, 1.38, 2.48, 4.7, 9.62, 19.14, 37.54, 74.51, 147.15, 311.04, 628.24]
OV MLIR [3.82, 4.43, 5.35, 8.06, 13.45, 23.87, 44.33, 85.15, 167.73, 340.55, 690.18]
NO-OV MLIR [4.05, 4.54, 5.57, 8.2, 13.47, 23.31, 44.07, 83.64, 166.25, 333.87, 662.45]
MANUAL MLIR [0.54, 1.1, 2.2, 4.43, 8.82, 17.39, 35.26, 70.28, 138.55, 279.32, 561.73]
```

- OV NO-MLIR vs. MANUAL MLIR compares plain OV vs TPP-MLIR with this new rewrite.
- NO-OV MLIR vs. MANUAL MLIR compares TPP-MLIR vs TPP-MLIR with this new rewrite.
- OV MLIR vs. NO-OV MLIR compares TPP-MLIR vs TPP-MLIR (both without the new rewrite) to infer overhead of calling MLIR-generated code from OV.

adam-smnk:
> NO-OV MLIR vs. MANUAL MLIR compares TPP-MLIR vs TPP-MLIR with this new rewrite.

Kind of, NO-OV MLIR still has runtime packing due to weights passed as arguments.

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
rolfmorel: Only the first test case uses `-pack-matmul` when run - the other tests use IR obtained by manually running `-pack-matmul`. It is also as simple as we can go to test a matmul with a constant and covers the situation that we care about. Hence I believe the tests already are as you suggest.

I think in this case it is also not a bad idea to have such a "coupled" test - though possibly we need to move it to  `test/Integration` - as we would like to know when changes to either pass results in this part of the pipeline no longer doing what we expect.

rolfmorel:
> `> NO-OV MLIR vs. MANUAL MLIR compares TPP-MLIR vs TPP-MLIR with this new rewrite.
> 
> Kind of, NO-OV MLIR still has runtime packing due to weights passed as arguments.

Disambiguation issue: NO-OV MLIR vs. MANUAL MLIR compares (TPP-MLIR) vs (TPP-MLIR with this new rewrite).

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
adam-smnk: > Only the first test case uses -pack-matmul when run - the other tests use IR obtained by manually running -pack-matmul

Then the one obtained from a manual run should be enough. Otherwise, the pass can be tested with more readable 1 input 1 output cases + maybe sth for multiple results if you wanna stress that.

> I think in this case it is also not a bad idea to have such a "coupled" test - though possibly we need to move it to test/Integration

It's one of those things that are good in theory but super annoying in practice. And that's what we've started with as seen by the current collection of integration and bundle tests. In reality, they end up being brittle and whenever sth breaks I try to at least simplify or tweak them. Although, most could be removed with no loss.

I'd be all for integration tests that validate full execution by examining actual computed non-trivial results to double check correctness. I think it's sth that we lack currently.

> as we would like to know when changes to either pass results in this part of the pipeline no longer doing what we expect.

I think it'd be more helpful to verify behavior of the whole pipeline on more known kernels.
Overall, that is good goal but I don't think manually made "test pipelines" truly achieve it.

adam-smnk: But all in all, it's subjective. I'm not gonna block on that - if you prefer to have such multi pass example, I'm fine with it in integration tests.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
rolfmorel: Fixed.

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
rolfmorel: Made an integration test to test functional correctness in case of a matmul with a constant as argument. Replaced the `-pack-matmul` test case with a single argument linalg.generic test.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
rolfmorel: A sanity check on this would be appreciated - thanks!

rolfmorel:
@adam-smnk, @rengolin, in case of no further comments I will merge this tomorrow, Sept 25th.

test/Integration/lower-pack-unpack-folding-transpose.mlir
```diff
@@ -0,0 +1,44 @@
+// RUN: tpp-run %s -e entry --entry-point-result=void -print -n 1 --seed=123 2>&1 > %S/matmul_48x64x96-default-tpp-passes.out
```
adam-smnk, line 1: nit: a bit outdated output name ;)
Maybe sth more specific to the test name?

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
```diff
@@ -0,0 +1,151 @@
+// : tpp-opt %s -pack-matmul -lower-packs-unpacks-without-transpose -canonicalize -split-input-file | FileCheck %s --check-prefix PACK-CHECK
```
adam-smnk, line 1: Should this line be removed?

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+      if (!packOp || !packOp->hasOneUse() || !unpackOp)
+        continue;
+
+      // Normalize empty outer_dims_perm to its corresponding identity map.
+      auto packOuterDimsPerm = SmallVector<long>(packOp.getOuterDimsPerm());
+      if (packOuterDimsPerm.empty()) {
+        packOuterDimsPerm =
+            SmallVector<long>(packOp.getSource().getType().getRank());
+        std::iota(packOuterDimsPerm.begin(), packOuterDimsPerm.begin(), 0);
```
adam-smnk, line 135:
```suggestion
        std::iota(packOuterDimsPerm.begin(), packOuterDimsPerm.end(), 0);
```

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
```diff
@@ -0,0 +1,202 @@
...
+      }
+      auto unpackOuterDimsPerm = SmallVector<long>(unpackOp.getOuterDimsPerm());
+      if (unpackOuterDimsPerm.empty()) {
+        unpackOuterDimsPerm =
+            SmallVector<long>(unpackOp.getResult().getType().getRank());
+        std::iota(unpackOuterDimsPerm.begin(), unpackOuterDimsPerm.begin(), 0);
```
adam-smnk, line 141:
```suggestion
        std::iota(unpackOuterDimsPerm.begin(), unpackOuterDimsPerm.end(), 0);
```

adam-smnk: Could you add a test case for this?

test/Passes/lower-packs-and-unpacks-without-transpose.mlir
adam-smnk, line 1: You should verify affine maps in at least one case to check that transpose is folded correctly.

lib/TPP/Transforms/LowerPacksAndUnpacksWithoutTranspose.cpp
adam-smnk: Approach is fine, just a small fix in the logic is needed.

rolfmorel:
Thank you for the thorough review, @adam-smnk !

I think I have addressed everything, though will wait a couple more hours before I merge.

test/Integration/lower-pack-unpack-without-transpose.mlir
```diff
@@ -0,0 +1,55 @@
+// RUN: tpp-run %s -e big_matmul --entry-point-result=void -print -n 1 --seed=123 2>&1 > %S/lower-pack-unpack-without-transpose-big_matmul-only-default-passes.out
+// RUN: tpp-run --lower-pack-unpack-without-transpose %s -e big_matmul  --entry-point-result=void -print -n 1 --seed=123 2>&1 > %S/lower-pack-unpack-without-transpose-big_matmul-default-passes-lower-pack-unpack-without-transpose.out
+// RUN: fpcmp -r 0.001 %S/lower-pack-unpack-without-transpose-big_matmul-only-default-passes.out %S/lower-pack-unpack-without-transpose-big_matmul-default-passes-lower-pack-unpack-without-transpose.out
+// RUN: rm %S/lower-pack-unpack-without-transpose-big_matmul-only-default-passes.out %S/lower-pack-unpack-without-transpose-big_matmul-default-passes-lower-pack-unpack-without-transpose.out
+
+func.func @big_matmul(%A: tensor<1024x2048xf32>, %B: tensor<2048x4096xf32>, %C: tensor<1024x4096xf32>) -> tensor<1024x4096xf32> {
```
adam-smnk, line 6: That's way too large for CI, single run alone on my local machine takes 25% time of whole test suite.

adam-smnk (commented):
Thanks for the tweaks 👍
Looks good

PR #973, opened 2024-09-17
Lower vector.contract to vector.outerproduct.

shahidact:
Implements the lowering of vector contraction op to vector outerproduct wrapped inside an scf.forloop. with iterargs to accumulate the result of each outerproduct corresponding to the K dimension size. The idea is to exploit the AVX feature to generate optimal vector code. Outerproduct gets lowered to FMAs X86 assembly through llvm intrinsic "llvm.intr.fmuladd".

### DIFF
--- a/lib/TPP/Transforms/VectorContractToOuterproduct.cpp
+++ b/lib/TPP/Transforms/VectorContractToOuterproduct.cpp
@@ -1,0 +1,283 @@
+//===--------------- VectorContractToOuterproduct.cpp ------------*- C++-*-===//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements lowering of vector contraction to vector outerproduct.
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/Transforms/Transforms.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Vector/IR/VectorOps.h"
+#include "mlir/Dialect/Vector/Transforms/LoweringPatterns.h"
+#include "mlir/Dialect/Vector/Transforms/VectorRewritePatterns.h"
+#include "mlir/Dialect/Vector/Transforms/VectorTransforms.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypeInterfaces.h"
+#include "mlir/IR/TypeUtilities.h"
+#include "mlir/Interfaces/VectorInterfaces.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/LogicalResult.h"
+#include <cmath>
+#include <cstdint>
+
+#define DEBUG_TYPE "vector-contract-to-outerproduct"
+
+namespace mlir {
+namespace tpp {
+#define GEN_PASS_DEF_VECTORCONTRACTTOOUTERPRODUCT
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+using namespace mlir;
+using namespace mlir::tpp;
+
+namespace {
+/// Returns true if the \p map is transposed.
+static bool isTransposed(AffineMap map) {
+  auto results = map.getResults();
+  // Assert if the map does not have 3 inputs (m, n, k).
+  assert(map.getNumInputs() > 2 && "Al least 3 input dim expected");
+  // Assert if the result is not 2D.
+  assert(map.getNumResults() == 2 && "Only 2 output dim expected");
+
+  // Check the last two dimensions for transposition.
+  auto dimExpr0 = dyn_cast<AffineDimExpr>(results[0]);
+  auto dimExpr1 = dyn_cast<AffineDimExpr>(results[1]);
+  assert((dimExpr0 && dimExpr1) && "Unexpected dim expression");
+
+  // Exclude output map result.
+  bool isOutputResultMap =
+      dimExpr0 == mlir::getAffineDimExpr(0, map.getContext()) &&
+      dimExpr1 == mlir::getAffineDimExpr(1, map.getContext());
+  assert(!isOutputResultMap && "Output result map not expected");
+
+  // It's transposed if result found as (k, m) or (n, k), else not transposed.
+  if ((dimExpr0 == mlir::getAffineDimExpr(2, map.getContext()) &&
+       dimExpr1 == mlir::getAffineDimExpr(0, map.getContext())) ||
+      (dimExpr0 == mlir::getAffineDimExpr(1, map.getContext()) &&
+       dimExpr1 == mlir::getAffineDimExpr(2, map.getContext())))
+    return true;
+  return false;
+}
+} // namespace
+
+namespace mlir {
+namespace tpp {
+// Enum to represent the type of matmul operation
+enum class MatMulType { Standard, Batch, BatchReduce };
+
+struct VectorContractToOuterproductPattern
+    : public OpRewritePattern<vector::ContractionOp> {
+  using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
+                                PatternRewriter &rewriter) const override {
+
+    if (contractOp.getKind() != vector::CombiningKind::ADD)
+      return rewriter.notifyMatchFailure(
+          contractOp,
+          "Unsupported combining kind, only supports ADD at the moment)");
+
+    SmallVector<AffineMap, 3> maps = contractOp.getIndexingMapsArray();
+    if (llvm::any_of(
+            maps, [](AffineMap map) { return !map.isProjectedPermutation(); }))
+      return rewriter.notifyMatchFailure(contractOp, "Unexpected map");
+
+    // Check for the variant of matrix multiply.
+    auto iteratorTypes = contractOp.getIteratorTypesArray();
+    MatMulType matmulType;
+    unsigned outerDimIndex = 0;
+    if (iteratorTypes.size() > 3) {
+      outerDimIndex = iteratorTypes.size() - 4;
+      matmulType =
+          iteratorTypes[outerDimIndex] == vector::IteratorType::parallel
+              ? MatMulType::Batch
+              : MatMulType::BatchReduce;
+      outerDimIndex++;
+    } else if (iteratorTypes.size() == 3) {
+      matmulType = MatMulType::Standard;
+    } else {
+      return rewriter.notifyMatchFailure(contractOp, "Not a gemm");
+    }
+
+    if (matmulType == MatMulType::Batch)
+      return rewriter.notifyMatchFailure(contractOp,
+                                         "Batch matmul not supported");
+    if (iteratorTypes[outerDimIndex] != vector::IteratorType::parallel ||
+        iteratorTypes[outerDimIndex + 1] != vector::IteratorType::parallel ||
+        iteratorTypes[outerDimIndex + 2] != vector::IteratorType::reduction)
+      return rewriter.notifyMatchFailure(contractOp, "Not a gemm");
+
+    Value acc = contractOp.getAcc();
+    // Find the original tensor operands
+    auto lhsDefiningOp =
+        contractOp.getLhs().getDefiningOp<vector::TransferReadOp>();
+    auto rhsDefiningOp =
+        contractOp.getRhs().getDefiningOp<vector::TransferReadOp>();
+    auto accDefiningOp = acc.getDefiningOp<vector::TransferReadOp>();
+    if (!lhsDefiningOp || !rhsDefiningOp || !accDefiningOp)
+      return failure();
+
+    // Make sure the inputs being read are whole tensor or subview.
+    if (!llvm::all_of(lhsDefiningOp.getIndices(), isZeroIndex) ||
+        !llvm::all_of(rhsDefiningOp.getIndices(), isZeroIndex)) {
+      return failure();
+    }
+
+    auto lhsType = cast<ShapedType>(lhsDefiningOp.getType());
+    auto rhsType = cast<ShapedType>(rhsDefiningOp.getType());
+    auto accType = cast<ShapedType>(accDefiningOp.getType());
+
+    if (matmulType == MatMulType::BatchReduce &&
+        (lhsType.getRank() != 3 || rhsType.getRank() != 3))
+      return failure();
+
+    if (matmulType == MatMulType::Standard &&
+        (lhsType.getRank() != 2 || rhsType.getRank() != 2))
+      return failure();
+
+    // Only 2-D output expected.
+    if (accType.getRank() != 2)
+      return failure();
+
+    // Handle 3D subviews
+    auto mapLHS = maps[0];
+    auto mapRHS = maps[1];
+    if (matmulType == MatMulType::BatchReduce) {
+      mapLHS = mapLHS.dropResult(outerDimIndex);
+      mapRHS = mapRHS.dropResult(outerDimIndex);
+    }
+
+    int64_t M = accType.getDimSize(0);
+    int64_t N = accType.getDimSize(1);
+    int64_t K = !isTransposed(mapLHS)
+                    ? lhsType.getDimSize(lhsType.getRank() - 1)
+                    : lhsType.getDimSize(lhsType.getRank() - 2);
+
+    // Create constants
+    Location loc = contractOp.getLoc();
+    Value c0 = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+    Value c1 = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+    Value cK = rewriter.create<arith::ConstantIndexOp>(loc, K);
+
+    auto elementType = lhsType.getElementType();
+    FloatType floatType = cast<FloatType>(elementType);
+    Value f0 = rewriter.create<arith::ConstantFloatOp>(
+        loc, APFloat::getZero(floatType.getFloatSemantics()), floatType);
+
+    // Create the outer scf.for loop
+    auto forOp = rewriter.create<scf::ForOp>(
+        loc, c0, cK, c1, ValueRange{acc},
+        [&](OpBuilder &nestedBuilder, Location nestedLoc, Value iv,
+            ValueRange iterArgs) {
+          // Prepare indices and map to iterate over rows/colums and read
+          // slices of lhs/rhs input operands.
+          SmallVector<Value, 3> lhsIndices, rhsIndices;
+          AffineMap lhsMap, rhsMap;
+          for (int i = 0; i < lhsType.getRank() - 2; ++i)
+            lhsIndices.push_back(c0);
+          // LHS operand
+          if (!isTransposed(mapLHS)) {
+            // If not transposed, iterate over colums and read each column
+            // using map.
+            lhsIndices.push_back(c0);
+            lhsIndices.push_back(iv);
+            lhsMap = AffineMap::get(lhsType.getRank(), 0,
+                                    {nestedBuilder.getAffineDimExpr(0)},
+                                    nestedBuilder.getContext());
+          } else {
+            // If transposed, iterate over rows and read each row with default
+            // map.
+            lhsIndices.push_back(iv);
+            lhsIndices.push_back(c0);
+            lhsMap = AffineMap::get(lhsType.getRank(), 0,
+                                    {nestedBuilder.getAffineDimExpr(1)},
+                                    nestedBuilder.getContext());
+          }
+
+          for (int i = 0; i < rhsType.getRank() - 2; ++i)
+            rhsIndices.push_back(c0);
+          // RHS operand
+          if (!isTransposed(mapRHS)) {
+            // If not transposed, iterate over rows and read each row using
+            // default map.
+            rhsIndices.push_back(iv);
+            rhsIndices.push_back(c0);
+            rhsMap = AffineMap::get(rhsType.getRank(), 0,
+                                    {nestedBuilder.getAffineDimExpr(1)},
+                                    nestedBuilder.getContext());
+          } else {
+            // If transposed, iterate over columns and read each column with
+            // default map.
+            rhsIndices.push_back(c0);
+            rhsIndices.push_back(iv);
+            rhsMap = AffineMap::get(rhsType.getRank(), 0,
+                                    {nestedBuilder.getAffineDimExpr(0)},
+                                    nestedBuilder.getContext());
+          }
+
+          Value lhsTensor = lhsDefiningOp.getSource();
+          Value rhsTensor = rhsDefiningOp.getSource();
+          // Read vector slices using TransferReadOp
+          auto lhsSlice = nestedBuilder.create<vector::TransferReadOp>(
+              nestedLoc, VectorType::get({M}, lhsType.getElementType()),
+              lhsTensor, lhsIndices, AffineMapAttr::get(lhsMap), f0, Value(),
+              rewriter.getBoolArrayAttr({true}));
+
+          auto rhsSlice = nestedBuilder.create<vector::TransferReadOp>(
+              nestedLoc, VectorType::get({N}, rhsType.getElementType()),
+              rhsTensor, rhsIndices, rhsMap, f0, Value(),
+              rewriter.getBoolArrayAttr({true}));
+
+          // Perform outer product
+          auto outerProduct = nestedBuilder.create<vector::OuterProductOp>(
+              nestedLoc, accType, lhsSlice, rhsSlice, iterArgs[0],
+              vector::CombiningKind::ADD);
+
+          // Yield the result
+          nestedBuilder.create<scf::YieldOp>(nestedLoc,
+                                             ValueRange{outerProduct});
+        });
+
+    // Replace the original contraction with the result of the loop
+    rewriter.replaceOp(contractOp, forOp.getResults());
+
+    return success();
+  }
+};
+
+struct VectorContractToOuterproduct
+    : public tpp::impl::VectorContractToOuterproductBase<
+          VectorContractToOuterproduct> {
+
+  using VectorContractToOuterproductBase::VectorContractToOuterproductBase;
+
+  void runOnOperation() override {
+    auto funcOp = getOperation();
+    MLIRContext *context = &getContext();
+
+    RewritePatternSet patterns(context);
+    patterns.add<VectorContractToOuterproductPattern>(context);
+
+    if (failed(applyPatternsAndFoldGreedily(funcOp, std::move(patterns)))) {
+      signalPassFailure();
+    }
+  }
+};
+
+} // namespace tpp
+} // namespace mlir
--- a/test/Passes/pass-vector-contract-to-outerproduct.mlir
+++ b/test/Passes/pass-vector-contract-to-outerproduct.mlir
@@ -1,0 +1,319 @@
+// RUN: tpp-opt %s  --vector-contract-to-outerproduct --split-input-file | FileCheck %s
+
+
+#map = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+
+module {
+  func.func @matmul_k_one(%arg0: tensor<4x1xf32>, %arg1: tensor<1x64xf32>, %arg2: tensor<4x64xf32>) -> tensor<4x64xf32> {
+    %c0 = arith.constant 0 : index
+    %cst = arith.constant 0.000000e+00 : f32
+    %0 = vector.transfer_read %arg0[%c0, %c0], %cst {in_bounds = [true, true]} : tensor<4x1xf32>, vector<4x1xf32>
+    %1 = vector.transfer_read %arg1[%c0, %c0], %cst {in_bounds = [true, true]} : tensor<1x64xf32>, vector<1x64xf32>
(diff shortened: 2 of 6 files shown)


lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  void runOnOperation() override {
+    auto funcOp = getOperation();
+    MLIRContext *context = &getContext();
+
+    RewritePatternSet patterns(context);
+    patterns.add<VectorContractToOuterproductPattern>(context);
+
+    if (failed(applyPatternsAndFoldGreedily(funcOp, std::move(patterns)))) {
+      failure();
```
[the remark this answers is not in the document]
shahidact: Ah right.

test/Passes/pass-vector-contract-to-outerproduct.mlir
```diff
@@ -0,0 +1,238 @@
...
+
+// CHECK-LABEL:   func.func @matmul_k_one(
+// CHECK-SAME:                      %[[VAL_0:.*]]: tensor<4x1xf32>,
+// CHECK-SAME:                      %[[VAL_1:.*]]: tensor<1x64xf32>,
+// CHECK-SAME:                      %[[VAL_2:.*]]: tensor<4x64xf32>) -> tensor<4x64xf32> {
+// CHECK:           %[[VAL_3:.*]] = arith.constant 1 : index
+// CHECK:           %[[VAL_4:.*]] = arith.constant 0 : index
+// CHECK:           %[[VAL_5:.*]] = arith.constant 0.000000e+00 : f32
+// CHECK:           %[[VAL_6:.*]] = vector.transfer_read %[[VAL_2]]{{\[}}%[[VAL_4]], %[[VAL_4]]], %[[VAL_5]] {in_bounds = [true, true]} : tensor<4x64xf32>, vector<4x64xf32>
+// CHECK:           %[[VAL_7:.*]] = scf.for %[[VAL_8:.*]] = %[[VAL_4]] to %[[VAL_3]] step %[[VAL_3]] iter_args(%[[VAL_9:.*]] = %[[VAL_6]]) -> (vector<4x64xf32>) {
```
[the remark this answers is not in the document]
shahidact: I would prefer to have it as it is being introduced by this pass.

test/Passes/pass-vector-contract-to-outerproduct.mlir
```diff
@@ -0,0 +1,238 @@
...
+// CHECK:             %[[VAL_10:.*]] = vector.transfer_read %[[VAL_0]]{{\[}}%[[VAL_4]], %[[VAL_8]]], %[[VAL_5]] {in_bounds = [true, false]} : tensor<4x1xf32>, vector<4x1xf32>
+// CHECK:             %[[VAL_11:.*]] = vector.transfer_read %[[VAL_1]]{{\[}}%[[VAL_8]], %[[VAL_4]]], %[[VAL_5]] {in_bounds = [false, true]} : tensor<1x64xf32>, vector<1x64xf32>
```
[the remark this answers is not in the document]
shahidact: I am not sure about dynamic size case but I think it is better to not make it out of bound.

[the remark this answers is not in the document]
shahidact: I will update.

test/Passes/pass-vector-contract-to-outerproduct.mlir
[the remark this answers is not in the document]
adam-smnk: Fair enough

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,158 @@
...
+              lhsTensor, lhsIndices, f0);
+
+          auto rhsSlice = nestedBuilder.create<vector::TransferReadOp>(
+              nestedLoc, VectorType::get({N}, rhsType.getElementType()),
+              rhsTensor, rhsIndices, f0);
+
+          // Perform outer product
+          auto outerProduct = nestedBuilder.create<vector::OuterProductOp>(
+              nestedLoc, accType, lhsSlice, rhsSlice, iterArgs[0],
+              vector::CombiningKind::ADD);
```
adam-smnk, line 124: Combining kind should be propagated from the original `vector.contract` or lowering restricted to only match with `add`.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,158 @@
...
+    Value rhsTensor = rhsDefiningOp.getSource();
+
+    auto lhsType = cast<ShapedType>(lhsTensor.getType());
+    auto rhsType = cast<ShapedType>(rhsTensor.getType());
+    auto accType = cast<VectorType>(acc.getType());
+
+    // Handle 3D subviews
+    int64_t M = accType.getDimSize(0);
+    int64_t N = accType.getDimSize(1);
+    int64_t K = lhsType.getDimSize(lhsType.getRank() - 1);
```
adam-smnk, line 83: Indexing maps need checking if they are as expected.
For example, one of the operands could be transposed.

adam-smnk:
I see large diff in results after applying this lowering.

Test example:
```
#map = affine_map<(d0, d1, d2) -> (d0, d2)>
#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
module {
  func.func @entry(%arg0: tensor<16x16xf32>, %arg1: tensor<16x16xf32>, %arg2: tensor<16x16xf32>) -> tensor<16x16xf32> {
    %0 = vector.transfer_read %arg0[%c0, %c0], %cst{in_bounds = [true, true]} : tensor<16x16xf32>, vector<16x16xf32>
    %1 = vector.transfer_read %arg1[%c0, %c0], %cst {in_bounds = [true, true]} : tensor<16x16xf32>, vector<16x16xf32>
    %2 = vector.transfer_read %arg2[%c0, %c0], %cst {in_bounds = [true, true]} : tensor<16x16xf32>, vector<16x16xf32>
    %3 = vector.contract {indexing_maps = [#map, #map1, #map2],
      iterator_types = ["parallel", "parallel", "reduction"],
      kind = #vector.kind<add>} %0, %1, %2
      : vector<16x16xf32>, vector<16x16xf32> into vector<16x16xf32>
    %4 = vector.transfer_write %3, %arg2[%c0, %c0] {in_bounds = [true, true]} : vector<16x16xf32>, tensor<16x16xf32>
    return %4 : tensor<16x16xf32>
  }
}
```
Baseline:
`tpp-opt ../test.mlir  | tpp-run -e entry --entry-point-result=void -seed 123 -print`
After rewrite:
`tpp-opt ../test.mlir --vector-contract-to-outerproduct  | tpp-run -e entry --entry-point-result=void -seed 123 -print`

adam-smnk:
> I see large diff in results after applying this lowering.

Ah okay spotted the issue.
After `vector.contract` lowering:
```
%1 = scf.for %arg3 = %c0 to %c2 step %c1 iter_args(%arg4 = %0) -> (vector<4x64xf32>) {
  %3 = vector.transfer_read %arg0[%c0, %arg3], %cst : tensor<4x2xf32>, vector<4xf32>
  %4 = vector.transfer_read %arg1[%arg3, %c0], %cst {in_bounds = [true]} : tensor<2x64xf32>, vector<64xf32>
  %5 = vector.outerproduct %3, %4, %arg4 {kind = #vector.kind<add>} : vector<4xf32>, vector<64xf32>
  scf.yield %5 : vector<4x64xf32>
}
```
to make reading from A
`%3 = vector.transfer_read %arg0[%c0, %arg3], %cst : tensor<4x2xf32>, vector<4xf32>`
a column load, a permutation maps needs to be added like:
`%3 = vector.transfer_read %arg0[%c0, %arg3], %cst {permutation_map = affine_map<(d0, d1) -> (d0)>} : tensor<4x2xf32>, vector<4xf32>`

Otherwise, it just reads contiguous elements.

shahidact:
> > I see large diff in results after applying this lowering.
> 
> Ah okay spotted the issue. After `vector.contract` lowering:
> 
> ```
> %1 = scf.for %arg3 = %c0 to %c2 step %c1 iter_args(%arg4 = %0) -> (vector<4x64xf32>) {
>   %3 = vector.transfer_read %arg0[%c0, %arg3], %cst : tensor<4x2xf32>, vector<4xf32>
>   %4 = vector.transfer_read %arg1[%arg3, %c0], %cst {in_bounds = [true]} : tensor<2x64xf32>, vector<64xf32>
>   %5 = vector.outerproduct %3, %4, %arg4 {kind = #vector.kind<add>} : vector<4xf32>, vector<64xf32>
>   scf.yield %5 : vector<4x64xf32>
> }
> ```
> 
> to make reading from A `%3 = vector.transfer_read %arg0[%c0, %arg3], %cst : tensor<4x2xf32>, vector<4xf32>` a column load, a permutation map needs to be added like: `%3 = vector.transfer_read %arg0[%c0, %arg3], %cst {permutation_map = affine_map<(d0, d1) -> (d0)>} : tensor<4x2xf32>, vector<4xf32>`
> 
> Otherwise, it just reads contiguous elements.

Thanks, I assumed it could infer the map using indices.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: For the moment only need ADD, will add an assert for it.

adam-smnk: Alright, just return failure on contract without `kind<add>`

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,199 @@
...
+          SmallVector<Value, 3> lhsIndices, rhsIndices;
+          for (int i = 0; i < lhsType.getRank() - 2; ++i)
+            lhsIndices.push_back(c0);
+          lhsIndices.push_back(c0);
+          lhsIndices.push_back(iv);
+
+          for (int i = 0; i < rhsType.getRank() - 2; ++i)
+            rhsIndices.push_back(c0);
+          rhsIndices.push_back(iv);
+          rhsIndices.push_back(c0);
```
adam-smnk, line 143: Does it account for shifting K-dim position in case the operands are transposed?

shahidact: No, but it seems I need to reorder the indices accordingly. Thanks.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,234 @@
...
+    auto rhsDefiningOp = rhs.getDefiningOp<vector::TransferReadOp>();
+    if (!lhsDefiningOp || !rhsDefiningOp)
+      return failure();
+
+    Value lhsTensor = lhsDefiningOp.getSource();
+    Value rhsTensor = rhsDefiningOp.getSource();
+
+    auto lhsType = cast<ShapedType>(lhsTensor.getType());
+    auto rhsType = cast<ShapedType>(rhsTensor.getType());
+    auto accType = cast<VectorType>(acc.getType());
```
adam-smnk, line 111: Current lowering can't handle non-2D output e.g., `batch_matmul`.
I'd just add a check for that and fail otherwise.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,234 @@
...
+      return rewriter.notifyMatchFailure(contractOp, "Not a gemm");
+
+    Location loc = contractOp.getLoc();
+    Value lhs = contractOp.getLhs();
+    Value rhs = contractOp.getRhs();
+    Value acc = contractOp.getAcc();
+
+    // Find the original tensor operands
+    auto lhsDefiningOp = lhs.getDefiningOp<vector::TransferReadOp>();
+    auto rhsDefiningOp = rhs.getDefiningOp<vector::TransferReadOp>();
```
adam-smnk, line 102: Current lowering seems to assume that producers have zero offsets and load whole tensors.
When that's not true, it results in incorrect new loads and loop range.

Please add further checks for any assumptions made.

adam-smnk: Could you add a test case for this too?

And also try this example:
```
#map = affine_map<(d0, d1, d2) -> (d0, d2)>
#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
module {
  func.func @entry(%arg0: tensor<4x4xf32>, %arg1: tensor<8x64xf32>,
      %arg2: tensor<8x64xf32>) -> tensor<8x64xf32> {
    %c1 = arith.constant 1 : index
    %0 = vector.transfer_read %arg0[%c0, %c0],
      %cst {in_bounds = [true, true]} : tensor<4x4xf32>, vector<4x2xf32>
    %1 = vector.transfer_read %arg1[%c0, %c0],
      %cst {in_bounds = [true, true]} : tensor<8x64xf32>, vector<2x64xf32>
    %2 = vector.transfer_read %arg2[%c0, %c0],
      %cst {in_bounds = [true, true]} : tensor<8x64xf32>, vector<4x64xf32>
    %3 = vector.contract {indexing_maps = [#map, #map1, #map2],
      iterator_types = ["parallel", "parallel", "reduction"],
      kind = #vector.kind<add>} %0, %1, %2 : vector<4x2xf32>, vector<2x64xf32>
      into vector<4x64xf32>
    %4 = vector.transfer_write %3, %arg2[%c0, %c0] {in_bounds = [true, true]}
      : vector<4x64xf32>, tensor<8x64xf32>
    return %4 : tensor<8x64xf32>
  }
}
```

shahidact: The above example is not a gemm, though such case is not expected i will add a check for that.

adam-smnk: It is a gemm just on slices of the source buffers.

adam-smnk: It's fine if such version is not supported but we can't produce invalid IR.

adam-smnk: Just to add to it, some vector operations have pretty high level of abstraction which might not be desirable or what we aim to consume for our rewrites. However, these variants and semantics have to be accounted for even if only to reject such (unexpected) patterns at the matching step.

shahidact: Agree

test/Passes/pass-vector-contract-to-outerproduct.mlir
```diff
@@ -0,0 +1,319 @@
...
+// CHECK:           return
+// CHECK:         }
+
+// -----
+
+#map = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+module {
+  func.func @entry(%arg0: tensor<4x4xf32>, %arg1: tensor<8x64xf32>,
```
adam-smnk, line 240:
```suggestion
  func.func @gemm_source_slices(%arg0: tensor<4x4xf32>, %arg1: tensor<8x64xf32>,
```

adam-smnk (approved):
I think it's good enough for now.
There are still unhandled `vector.transfer_read` semantics but that's probably unnecessary when we decide to generalize this lowering to handle presence of GEMM prologue etc.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    // Handle 3D subviews
+    auto mapLHS = maps[0];
+    auto mapRHS = maps[1];
+    unsigned matrixDimsLHS = mapLHS.getNumResults();
+    unsigned matrixDimsRHS = mapRHS.getNumResults();
+
+    int64_t M = accType.getDimSize(0);
+    int64_t N = accType.getDimSize(1);
+    int64_t K = !isTransposed(mapLHS, matrixDimsLHS)
```
rengolin, line 134: `matrixDimsLHS` come from `mapLHS`, you don't need to pass both arguments to the function.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+namespace mlir {
+namespace tpp {
+
+/// Returns true if the \p map is transposed for the given operand in context of
+/// matrix multiply. \p 'isRHS = false' flag indicates thet the check is being
+/// performed on LHS operand, otherwise RHS operand.
+bool isTransposed(AffineMap map, unsigned matrixDims, bool isRHS = false) {
+  auto results = map.getResults();
+  if (results.size() != matrixDims)
```
rengolin, line 53: This check is tautological, given you just created `matrixDims` with `map.getResults()` on the caller.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+  // Check the last two dimensions for transposition
+  auto secondLast = dyn_cast<AffineDimExpr>(results[matrixDims - 2]);
+  auto last = dyn_cast<AffineDimExpr>(results[matrixDims - 1]);
+
+  if (!secondLast || !last)
+    return false;
+
+  // For LHS, If the last dimension comes before the second last, it's
+  // transposed. its opposite for RHS.
+  return isRHS ? last.getPosition() > secondLast.getPosition()
```
rengolin, line 65: You're assuming this is a 3D operation on 2D shapes of a matmul kind, but not checking any of it. Calling the function `isTransposed` doesn't help to identify what this is, and may trick future developers into using it without checking.

Here you want to guarantee the following properties:
* This function only handles matmul maps `(m, n, k)`. So, the map's input _must_ be 3D.
* It only handles 2D shapes, so the result _must_ be 2D.
* There's only 3 combinations (`m, n`, 'm, k', `k, n'). You can restrict just to inputs.

Clearly, with just inputs, there are only four options:
* `m, k` and `k, n` are non-transposed.
* `k, m` and `n, k` are transposed.

Everything else is broken. IF you just check for those, and call is `isMatmulInputTransposed`, the function will be simple to read, won't be misused and we can extend in the future if we need anything else.

Also, please make it `static`.

rengolin (commented):
ongoing review, have a few comments for now...

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
rengolin: One more thing. Returning `false` here has the meaning "it's not transposed", but also "there's been a problem". We need to separate them into transposed, non-transposed, failure.

For the sake of brevity, and because this will be a local static function, you can use asserts for the errors and return `true` for transposed and `false` for non-transposed. But you _have_ to make sure there will be no other returns.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    auto maps = contractOp.getIndexingMapsArray();
+    if (llvm::any_of(
+            maps, [](AffineMap map) { return !map.isProjectedPermutation(); }))
+      return rewriter.notifyMatchFailure(contractOp, "Unexpected map");
+
+    // Check if this is a matrix multiply
+    auto iteratorTypes = contractOp.getIteratorTypesArray();
+    if (iteratorTypes.size() < 3 ||
+        iteratorTypes[iteratorTypes.size() - 3] !=
```
rengolin, line 89: Not clear why you decided to go backwards here. Can you compare the `iteratorTypes` with a static initializer like `{ parallel, parallel, reduction }`?

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+            vector::IteratorType::parallel ||
+        iteratorTypes[iteratorTypes.size() - 2] !=
+            vector::IteratorType::parallel ||
+        iteratorTypes[iteratorTypes.size() - 1] !=
+            vector::IteratorType::reduction)
+      return rewriter.notifyMatchFailure(contractOp, "Not a gemm");
+
+    Location loc = contractOp.getLoc();
```
rengolin, line 97: Move the variable before the first use, or it can be left over on future refactoring.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+    Value lhs = contractOp.getLhs();
+    Value rhs = contractOp.getRhs();
+    Value acc = contractOp.getAcc();
```
rengolin, line 100: These three variables are not used later, you can combine them with the three lines below and just keep the `DefiningOp` variants

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    // Find the original tensor operands
+    auto lhsDefiningOp = lhs.getDefiningOp<vector::TransferReadOp>();
+    auto rhsDefiningOp = rhs.getDefiningOp<vector::TransferReadOp>();
+    auto accDefiningOp = acc.getDefiningOp<vector::TransferReadOp>();
+    if (!lhsDefiningOp || !rhsDefiningOp || !accDefiningOp)
+      return failure();
+
+    // Make sure the inputs being read are whole tensor or subview.
+    if (!llvm::all_of(lhsDefiningOp.getIndices(), isZeroIndex) ||
```
rengolin, line 110: Where is `isZeroIndex` coming from?

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+        !llvm::all_of(rhsDefiningOp.getIndices(), isZeroIndex)) {
+      return failure();
+    }
+
+    Value lhsTensor = lhsDefiningOp.getSource();
```
rengolin, line 115: Move the variable before the first use, or it can be left over on future refactoring.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
```diff
@@ -0,0 +1,247 @@
...
+    Value rhsTensor = rhsDefiningOp.getSource();
+
+    auto lhsType = cast<ShapedType>(lhsDefiningOp.getType());
+    auto rhsType = cast<ShapedType>(rhsDefiningOp.getType());
+    auto accType = cast<ShapedType>(accDefiningOp.getType());
+
+    // Only 2-D output expected.
+    if (accType.getRank() != 2)
```
rengolin, line 123: Don't we expect all types to be 2D?

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: This pass is expected to work on a tiled batch_reduce_matmul too. For example:
```
#map = affine_map<(d0, d1, d2, d3) -> (d0, d1, d3)>
#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d3, d2)>
#map2 = affine_map<(d0, d1, d2, d3) -> (d1, d2)>
module {
  func.func @batch_reduce_matmul_tiled(%arg0: memref<16x32x16x32xf32>, %arg1: memref<32x32x32x32xf32>, %arg2: memref<16x32x16x32xf32>) {
    %c8 = arith.constant 8 : index
    %c1 = arith.constant 1 : index
    %c32 = arith.constant 32 : index
    %c4 = arith.constant 4 : index
    %c16 = arith.constant 16 : index
    scf.forall (%arg3, %arg4) in (16, 32) {
      %subview = memref.subview %arg0[%arg3, 0, 0, 0] [1, 32, 16, 32] [1, 1, 1, 1] : memref<16x32x16x32xf32> to memref<32x16x32xf32, strided<[512, 32, 1], offset: ?>>
      %subview_0 = memref.subview %arg1[%arg4, 0, 0, 0] [1, 32, 32, 32] [1, 1, 1, 1] : memref<32x32x32x32xf32> to memref<32x32x32xf32, strided<[1024, 32, 1], offset: ?>>
      %subview_1 = memref.subview %arg2[%arg3, %arg4, 0, 0] [1, 1, 16, 32] [1, 1, 1, 1] : memref<16x32x16x32xf32> to memref<16x32xf32, strided<[32, 1], offset: ?>>
      scf.for %arg5 = %c0 to %c16 step %c4 {  // m-loop
        scf.for %arg6 = %c0 to %c32 step %c16 { //n-loop
          scf.for %arg7 = %c0 to %c32 step %c1 { //BRGEMM-loop
            scf.for %arg8 = %c0 to %c32 step %c8 { //k-loop
              %subview_2 = memref.subview %subview[%arg7, %arg5, %arg8] [1, 4, 4] [1, 1, 1] : memref<32x16x32xf32, strided<[512, 32, 1], offset: ?>> to memref<1x4x4xf32, strided<[512, 32, 1], offset: ?>>
              %subview_3 = memref.subview %subview_0[%arg7, %arg8, %arg6] [1, 4, 2] [1, 1, 1] : memref<32x32x32xf32, strided<[1024, 32, 1], offset: ?>> to memref<1x4x2xf32, strided<[1024, 32, 1], offset: ?>>
              %subview_4 = memref.subview %subview_1[%arg5, %arg6] [4, 2] [1, 1] : memref<16x32xf32, strided<[32, 1], offset: ?>> to memref<4x2xf32, strided<[32, 1], offset: ?>>
              %0 = vector.transfer_read %subview_2[%c0, %c0, %c0], %cst {in_bounds = [true, true, true]} : memref<1x4x4xf32, strided<[512, 32, 1], offset: ?>>, vector<1x4x4xf32>
              %1 = vector.transfer_read %subview_3[%c0, %c0, %c0], %cst {in_bounds = [true, true, true]} : memref<1x4x2xf32, strided<[1024, 32, 1], offset: ?>>, vector<1x4x2xf32>
              %2 = vector.transfer_read %subview_4[%c0, %c0], %cst {in_bounds = [true, true]} : memref<4x2xf32, strided<[32, 1], offset: ?>>, vector<4x2xf32>
              %3 = vector.contract {indexing_maps = [#map, #map1, #map2], iterator_types = ["reduction", "parallel", "parallel", "reduction"], kind = #vector.kind<add>} %0, %1, %2 : vector<1x4x4xf32>, vector<1x4x2xf32> into vector<4x2xf32>
              vector.transfer_write %3, %subview_4[%c0, %c0] {in_bounds = [true, true]} : vector<4x2xf32>, memref<4x2xf32, strided<[32, 1], offset: ?>>
            }
          }
        }
      }
    }
    return
  }
}
```
Here the map `affine_map<(d0, d1, d2, d3) -> (d0, d1, d3)>` results are 3, `d0` is the batch dimension. The last two dimensions, d1 and d3, are the one related to matmul which I to wanted to check for transposition.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: I could also get `iterator_types = ["reduction", "parallel", "parallel", "reduction"]` in the contraction ops, So I wanted to only check last 3 iterator types from this  `iterator_types = ["reduction", "parallel", "parallel", "reduction"]`.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: So silly, thanks

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: So this `iteratorTypes[iteratorTypes.size() - 3]` is essentially iteratorTypes[1].

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: It is referred from StaticValueUtils.h

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: No, as we may be working on higher dimension subviews as shown in the example above.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
rengolin: I see, so I propose we decompose the problem. 

First we detect the `batch` part, which shows 4 dimensions instead of 3. Then we detect the `reduce` part, which shows that the output has one less dimension than the inputs. Then we match against a `matmul` on the three inner-most dims in the way I expose above.

For the `matmul` part, we only need to pass the three inner-most dims, so in the `batch` case we remove the outer-dim and continue. This way the checks can be reused and we identify the shapes and dims, then apply the right outer product (or match the right XSMM uKernel).

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
rengolin: Same comment as above: let's decompose the problem. First we match the outer-dim iterator type, select between `batch` and `batch_reduce` (`reduction` or `parallel` outer) or just `matmul` (no outer), then remove the first dim and compare the three inner dims with a fixed pattern.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
rengolin: But in the case of the `batch_matmul`, the output is 3D too, right? So this check (and the checks on the inputs) need to be conditional on what flavour of GEMM we're doing.

shahidact: Yes, updated.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: For now, batch matmul is not supported. Updated it as exposed.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: Batch matmul is not supported for now, not sure if it is required. Updated accordingly.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: Done.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: done.

lib/TPP/Transforms/VectorContractToOuterproduct.cpp
shahidact: done.

rengolin (approved):

PR #975, opened 2024-09-20
Split reduction dim pass

adam-smnk:
Adds a pass to allow tiling contraction's innermost reduction dimension using serial loop with in-place accumulation.

Compared to other available transformations, this rewrite computes reduction sequentially with in-place accumulation which avoids temporary allocation and separate reduction operation. This tiling strategy is more friendly in terms of register and memory pressure more suitable for low-level GPU kernel generation. Similarly, restriction to the innermost dimension is there to simplify both usage and pass logic as the rewrite is geared toward progressive GEMM lowering.

Additionally, a GPU vectorization control flag is added to allow grouping of passes which target lowering through vector operations and might not be compatible with other existing lowering strategies.
Effectively, this pass perform separate K-dim split which is currently baked into linalg-to-xegpu lowering, and the two are not fully compatible.

### DIFF
--- a/lib/TPP/Transforms/SplitReductionDim.cpp
+++ b/lib/TPP/Transforms/SplitReductionDim.cpp
@@ -22,6 +22,7 @@
 #include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/SCF/Transforms/TileUsingInterface.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/IR/Dialect.h"
 #include "mlir/Pass/Pass.h"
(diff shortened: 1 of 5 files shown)


test/Passes/split-reduction-dim.mlir
```diff
@@ -0,0 +1,332 @@
+// RUN: tpp-opt %s -split-reduction-dim=tile=8 -canonicalize -split-input-file | FileCheck %s
```
shahidact, line 1: Pls add a test without canonicalize.

adam-smnk: I don't think there's anything interesting without it.
Canonicalization just makes IR succinct and the pass has to work anyway with post-cleanup present.

shahidact: In that case, its ok.

lib/TPP/Transforms/SplitReductionDim.cpp
```diff
@@ -0,0 +1,248 @@
+//===- SplitReductionDim.cpp -------------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
```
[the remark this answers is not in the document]
adam-smnk: Not a big fan of those as I think it just duplicates the same info as other comments and descriptions at best.
But added.

lib/TPP/Transforms/SplitReductionDim.cpp
```diff
@@ -0,0 +1,248 @@
...
+        [&](OpBuilder &builder, Location loc, Value iv, ValueRange args) {
+          MLIRContext *ctx = builder.getContext();
+
+          OpFoldResult kDimBound;
+          std::optional<int64_t> dimK =
+              linalgx::utils::getConstantRange(reductionRange);
+          if (dimK && (*dimK % options.tileSize == 0)) {
+            kDimBound = builder.getIndexAttr(options.tileSize);
+          } else {
+            // The tile size to use (to avoid out of bounds access) is  minimum
```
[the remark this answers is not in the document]
adam-smnk: The size of the split dimension is kept dynamic. In practice, there are N-1 iterations using the specified tile size and the last iteration using a smaller tile with whatever leftover data.
This is implemented through a simple min operation over an affine map that computes tile size on each iteration.

adam-smnk:
> Now that I think more about it, it should be possible to largely simplify this logic using `scf::tileUsingSCF`. The concept of the pass can probably stay the same but the whole loop creation could be delegated to TilingInterface.

Yup, that was really straight forward. Overall the pass behavior remains the same but now named ops are preserved after tiling.

shahidact (approved):
Thanks, looks good to me.

PR #977, opened 2024-11-11
Hoist vector transfers outside reduction and k loop

arun-thmn:
Hoists the vector transfer read and write operations of resultant matrix outside the reduction and k loop for a brgemm operation. This pass should be applied after the BrgemmLinalgTiling pass.


test/Integration/hoist-vector-transfer-brgemm.mlir
```diff
@@ -3,38 +3,30 @@
 // RUN: diff %t.1 %t.2 | FileCheck %s --check-prefix=DIFF --allow-empty
 
 // DIFF-NOT: {{.}}
```
rengolin, line 5: You don't need this. You can just run `diff` and if there's anything it returns non-zero and fails the test.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  HoistVectorTransferOp(MLIRContext *ctx)
+      : OpRewritePattern(ctx) {}
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
+                                PatternRewriter &rewriter) const override {
+
+        // Code to hoist vector transfer read before the reduction and k loop
+	if (auto vectorReadOp = contractOp.getOperand(contractOp.getNumOperands()-1).getDefiningOp()) {
+          auto subviewOp = vectorReadOp->getOperand(0).getDefiningOp();
```
adam-smnk, line 48: It needs some invariance check.

include/TPP/Passes.td
```diff
@@ -53,6 +53,18 @@ def VectorizationPass : Pass<"vectorization-pass",
...
 }
 
+
+
+def HoistVectorTransfers : Pass<"hoist-vector-transfer"> {
+  let summary = "Hoist vector transfer operation outside of reduction and k loop";
+  let description = [{
+    Hoists the vector transfer read and write operations of the resultant  matrix outside the reduction and k loop for a brgemm operation. This pass should be applied after the BrgemmLinalgTiling Pass.
+  }];
+  let dependentDialects = [ "vector::VectorDialect" ];
```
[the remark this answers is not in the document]
arun-thmn: Added scf dialect to the list.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,155 @@
...
+using namespace vector;
+
+namespace mlir {
+namespace tpp {
+
+struct HoistVectorTransferOp : OpRewritePattern<vector::ContractionOp> {
+  using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
+
+  HoistVectorTransferOp(MLIRContext *ctx)
+      : OpRewritePattern(ctx) {}
```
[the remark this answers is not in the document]
arun-thmn: Yes. Removed those lines.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
+                                PatternRewriter &rewriter) const override {
+
+        // Code to hoist vector transfer read before the reduction and k loop
+	if (auto vectorReadOp = contractOp.getOperand(contractOp.getNumOperands()-1).getDefiningOp()) {
```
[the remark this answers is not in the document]
arun-thmn: Done.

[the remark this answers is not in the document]
arun-thmn: Done. Updated the defining p[ with vector transfer read and subview op.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+struct HoistVectorTransferOp : OpRewritePattern<vector::ContractionOp> {
+  using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
+                                PatternRewriter &rewriter) const override {
+
+        // Code to hoist vector transfer read before the reduction and k loop
+	if (auto retriveVectorReadOp = contractOp.getAcc().getDefiningOp<mlir::vector::TransferReadOp>()) {
+          auto subviewOp = retriveVectorReadOp.getOperand(0).getDefiningOp<memref::SubViewOp>();
```
adam-smnk, line 45: It should be checked too, it might not be a subview as expected

adam-smnk (commented):
Could you add a few negative test cases?
Pattern can be very specific but we shouldn't crash compiler on unexpected inputs.

test/Integration/hoist-vector-transfer-brgemm.mlir
arun-thmn: Okay. Removed the FileCheck.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Okay. Updated the code to check subview or args.

arun-thmn:
Add three negative test-cases. Two in unit tests and one in integration test.

lib/TPP/Transforms/HoistVectorTransfers.cpp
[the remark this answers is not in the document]
arun-thmn: Done, with the `notifyMatchFailure` method.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,164 @@
...
+        
+        auto subviewOp = retriveVectorReadOp.getOperand(0).getDefiningOp<memref::SubViewOp>();
+        if (subviewOp == NULL)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+        
+        auto ReductionForOp = llvm::dyn_cast<mlir::scf::ForOp>(subviewOp->getNextNode());
+        if (ReductionForOp == NULL)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+
+        auto KForOp = llvm::dyn_cast<mlir::scf::ForOp>(ReductionForOp.getBody()->front());
```
rengolin, line 56: This needs to verify that the loop is actually `K` by looking at the indices in the vector contract

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,164 @@
...
+        if (KForOp == NULL)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+
+	// Move the vector transfer read before the reduction and k loop
+        rewriter.setInsertionPointAfter(subviewOp);
+        auto *cloneVectorReadOp = rewriter.clone(*retriveVectorReadOp);
+        retriveVectorReadOp.replaceAllUsesWith(cloneVectorReadOp);
```
rengolin, line 63: It's usually better to `replaceAllUsesWith` after everything has been checked and pass.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        // Code to re-create the reduction and k loop with iter args
+        auto *nextOp = (*cloneVectorReadOp).getNextNode();
```
rengolin, line 66: This is dangerous. There's no guarantee that the next op in the IR is what you expect. If not, the cast below will yield a null pointer and the compiler will crash.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,164 @@
...
+                oldKForOp.getStep(), iterArgsNewReductionForOp,
+                [&](OpBuilder &rewriterNewKForOp, Location locNewKForOp, Value ivNewKForOp,
+                ValueRange iterArgsNewKForOp) {
+
+                        mlir::IRMapping mapper;
+                        mapper.map(oldReductionForOp.getInductionVar(), ivNewReductionForOp);
+                        mapper.map(oldKForOp.getInductionVar(), ivNewKForOp);
+
+                        for (auto [origArgReduction, newArgReduction] :
+                        llvm::zip(oldReductionForOp.getRegionIterArgs(), iterArgsNewReductionForOp)) {
```
rengolin, line 90: prefer `zip_equals` to make sure the two lists have the same number of elements.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,164 @@
...
+                vectorWriteOperation->moveBefore(oldReductionForOp);
+        }
+
+        // Erase the old vector contract operation
+        for (auto result : contractOp->getResults()) {
+                for (auto *userOp : result.getUsers()) {
+                        userOp->erase();
+                }
+        }
+        contractOp.erase();
```
rengolin, line 139: Technically, the other ops will be cleaned up with DCE, so removing only the contract should be fine. But if you know you have `replacedAllUsesWith` for other ops, you can also remove them here and avoid DCE being smart.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Sure. Updated the code accordingly.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: This lines of code is not required, as the old loops don't have any iter args. So, `llvm::zip` and its related code is removed.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Agreed. The above code first removes the use of `contract` op followed by `contact` itself.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: We do check the `nextOp` in the beginning of the pass. So, it won't be a `null` pointer.
code:
 `auto ReductionForOp = llvm::dyn_cast<mlir::scf::ForOp>(subviewOp->getNextNode());
        if (ReductionForOp == NULL)
                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");`

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,164 @@
...
+	// Check whether the linalg tiling + vector contract pattern matches 
+        auto retriveVectorReadOp = contractOp.getAcc().getDefiningOp<mlir::vector::TransferReadOp>();
+        if (retriveVectorReadOp == NULL)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+        
+        auto subviewOp = retriveVectorReadOp.getOperand(0).getDefiningOp<memref::SubViewOp>();
+        if (subviewOp == NULL)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+        
+        auto ReductionForOp = llvm::dyn_cast<mlir::scf::ForOp>(subviewOp->getNextNode());
```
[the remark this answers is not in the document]
arun-thmn: Ohh okay, sorry I'm not aware of it. 
Do you have reference to a `pass` which does it.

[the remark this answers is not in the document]
arun-thmn: I updated the pass to follow parent/child relationships

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Added an additional check `k == 1`

test/Integration/hoist-vector-transfer-brgemm.mlir
```diff
@@ -0,0 +1,59 @@
+// RUN: tpp-opt %s  | tpp-run -e entry --entry-point-result=void  -print > %t.1
+// RUN: tpp-opt %s --loop-invariant-code-motion  --vectorization-pass --loop-invariant-code-motion --hoist-vector-transfer | tpp-run -e entry --entry-point-result=void  -print > %t.2
+// RUN: diff %t.1 %t.2
```
rengolin, line 3: Make sure you remove `%t.1` and `%t.2`

test/Integration/hoist-vector-transfer-brgemm.mlir
```diff
@@ -0,0 +1,59 @@
...
+
+// RUN: tpp-opt %s  | tpp-run -e nomatch --entry-point-result=void -seed 123 -print > %t.1
+// RUN: tpp-opt %s  --hoist-vector-transfer  | tpp-run -e nomatch --entry-point-result=void -seed 123 -print > %t.2
+// RUN: diff %t.1 %t.2 
+
+#permA0 = affine_map<(d0, d1, d2) -> (d2, d0)>
+#permA1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#permA2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+
+func.func @nomatch(%arg0: tensor<4x4xf32>, %arg1: tensor<4x4xf32>, %arg2: tensor<4x4xf32>) -> tensor<4x4xf32> {
```
rengolin, line 46: If this is a no match, don't run the IR, just create a pass test with CHECK lines that show unchanged IR.

test/Integration/hoist-vector-transfer-brgemm.mlir
```diff
@@ -0,0 +1,59 @@
+// RUN: tpp-opt %s  | tpp-run -e entry --entry-point-result=void  -print > %t.1
```
rengolin, line 1: `tpp-opt` here is unnecessary, as `tpp-run` already runs the default pipeline

test/Integration/hoist-vector-transfer-brgemm.mlir
```diff
@@ -0,0 +1,59 @@
...
+// RUN: tpp-opt %s --loop-invariant-code-motion  --vectorization-pass --loop-invariant-code-motion --hoist-vector-transfer | tpp-run -e entry --entry-point-result=void  -print > %t.2
```
rengolin, line 2: Add a new RUN line that prints the IR and make sure that the contract read/write gets hoisted. Otherwise we won't know if your pass actually did anything, and the result will be the same.

test/Passes/pass-hoist-vector-transfer-operation-brgemm.mlir
```diff
@@ -0,0 +1,382 @@
...
+// CHECK:           }
+// CHECK:           return %[[VAL_10]] : memref<8x24x32x64xf32>
+// CHECK:         }
+
+
+
+
+// -----
+
+// RUN: tpp-opt %s  --hoist-vector-transfer --split-input-file  | FileCheck %s
```
rengolin, line 97: You don't need multiple RUN lines. You already use `split-input-file`.

test/Passes/pass-hoist-vector-transfer-operation-brgemm.mlir
```diff
@@ -0,0 +1,382 @@
...
+
+
+#map = affine_map<(d0, d1, d2, d3) -> (d0, d1, d3)>
+#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d3, d2)>
+#map2 = affine_map<(d0, d1, d2, d3) -> (d1, d2)>
+module {
+  memref.global "private" constant @__constant_24x64x64xf32 : memref<24x64x64xf32> = dense<1.000000e+00> {alignment = 64 : i64}
+  func.func @entry(%arg0: memref<8x24x32x64xf32>) -> memref<8x24x32x64xf32> {
```
rengolin, line 9: please name the functions as to what they're doing. It's hard to see the difference between the multiple entries in this test

test/Passes/pass-hoist-vector-transfer-operation-brgemm.mlir
```diff
@@ -0,0 +1,382 @@
...
+// CHECK:             }
+// CHECK:           }
+// CHECK:           return %[[VAL_9]] : memref<8x48x32x32xf32>
+// CHECK:         }
+
+
+
+// -----
+
+// RUN: tpp-opt %s  --hoist-vector-transfer --split-input-file  | FileCheck %s
```
rengolin, line 288: No new RUN line

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+enum class MatMulType { Standard, Batch, BatchReduce };
+
+struct HoistVectorTransferOp : OpRewritePattern<vector::ContractionOp> {
+  using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
+                                PatternRewriter &rewriter) const override {
+
+        // Check whether the linalg tiling + vector contract pattern matches for the 4-nested loop structure
+        auto oldKForOp =  contractOp->getParentOfType<scf::ForOp>();
```
rengolin, line 45: This is better. However, you're not really sure these are the actual M, N, K loops. The order may be swapped (ex. parlooper).

Try to extract this repetition to a function that recursively get the parent loop and saves the induction variable, like:
```
  SmallVector<tuple<ForOp, Value>> getNestedoop(contractOp, 4);
```

Then you can _find_ which is the M, N, K loop by iterating through the vector and checking which args are used in which `subview`, mapping them to each dim.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+        auto oldReductionForOp = oldKForOp->getParentOfType<scf::ForOp>();
+        if (!oldReductionForOp)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+
+        auto oldNForOp = oldReductionForOp->getParentOfType<scf::ForOp>();
+        if (!oldNForOp)
+                return rewriter.notifyMatchFailure(contractOp, "Not a linalg tile + vector contract operation");
+
+        auto oldMForOp = oldNForOp->getParentOfType<scf::ForOp>();
```
rengolin, line 57: These two variables are unused, just here for validation (which is good), but reinforces my point above that we can collapse all that into an external function and only use the variables we need from the list.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+                return failure();
+
+
+
+        // Check the operation type MatMul, B-MatMul, or BR-MatMul
+        auto contractIteratorTypes = contractOp.getIteratorTypesArray();
+        MatMulType matmulType;
+
+        if (contractIteratorTypes.size() > 3) {
+                matmulType = contractIteratorTypes[contractIteratorTypes.size() - 4] == vector::IteratorType::parallel ? MatMulType::Batch : MatMulType::BatchReduce;
```
rengolin, line 87: A simpler approach here would be to count the number of `reduction` iterators.

1 reduction = batch, 2 = batch reduce, 3+ = error.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+                if (matmulType == MatMulType::Batch)
+                        return rewriter.notifyMatchFailure(contractOp, "Batch matmul not supported");
```
rengolin, line 89:
```suggestion
                        return rewriter.notifyMatchFailure(contractOp, "Batch matmul not supported yet");
```

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+} // namespace tpp
+} // namespace mlir
+
+using namespace mlir;
+using namespace vector;
+
+namespace mlir {
+namespace tpp {
+
+enum class MatMulType { Standard, Batch, BatchReduce };
```
rengolin, line 36: We may have this already in some helper

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+                matmulType = contractIteratorTypes[contractIteratorTypes.size() - 4] == vector::IteratorType::parallel ? MatMulType::Batch : MatMulType::BatchReduce;
+                if (matmulType == MatMulType::Batch)
+                        return rewriter.notifyMatchFailure(contractOp, "Batch matmul not supported");
+        } else if (contractIteratorTypes.size() == 3) {
+                matmulType = MatMulType::Standard;
+        } else {
+                return rewriter.notifyMatchFailure(contractOp, "The vector contract operation is not a gemm");
+        }
+
+        auto vectorReadOpLHSType = cast<ShapedType>(vectorReadOpLHS.getType());
```
rengolin, line 96: this is not saving a lot of typing. Since you're mostly using `getRank`, you can cache that instead.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,198 @@
...
+                return failure();
+
+        //Check the K-dim to be 1
+        int64_t K = vectorReadOpLHSType.getDimSize(vectorReadOpLHSType.getRank() - 1);
+        if (K != 1)
+                return rewriter.notifyMatchFailure(contractOp, "K dim is not 1");
+
+
+        // Move the vector transfer read before the reduction and k loop
+        rewriter.setInsertionPointAfter(subviewOp);
```
rengolin, line 112: That's dangerous. You want to move it relative to the loop region, not some arbitrary operation in there.

rengolin (commented):
Please run clang-format on your code.

test/Integration/hoist-vector-transfer-brgemm.mlir
arun-thmn: Thanks. Added command to remove `%t.1` and `%t.2`.

test/Integration/hoist-vector-transfer-brgemm.mlir
arun-thmn: Okay. The check is already being done in the `unit` test-case. There is no point this test-case being in Integration test.
So, removing it.

test/Integration/hoist-vector-transfer-brgemm.mlir
arun-thmn: Correct. Removed `tpp-opt`.

test/Integration/hoist-vector-transfer-brgemm.mlir
arun-thmn: Same. The check is already done in `unit` tests.

test/Passes/pass-hoist-vector-transfer-operation-brgemm.mlir
arun-thmn: Okay. Good to learn. Removed multiple RUN lines.

test/Passes/pass-hoist-vector-transfer-operation-brgemm.mlir
arun-thmn: Yes, Named the functions.

test/Passes/pass-hoist-vector-transfer-operation-brgemm.mlir
arun-thmn: Removed.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Committed the changes

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: The `subviewOp` is the operand of `vectorTransferRead`. So, the `setInsertionPointAfter` is set on `subviewOp` to move the  `vectorTransferReadOp` of `C`. But, it's okay now the `setInsertionPoint` is set to the `Reduction` loop.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: I think, we don't have such an `enum`.

lib/TPP/Transforms/HoistVectorTransfers.cpp
rengolin: Oh, so if the subview is still a producer of the transfer read's argument, then your original line was not wrong, but incomplete. We need to make sure that you add it to the right level AND after all its producers.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Sure. Have replaced the code with your logic.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: For `vectorReadOpLHSType`  we are later calling `getDimSize()` to do a check on the `k-dim`.
For `vectorReadOpRHSType` we do only `getRank()`, So I cached it.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+    // 4-nested loop structure
+    auto oldKForOp = contractOp->getParentOfType<scf::ForOp>();
+    if (!oldKForOp)
+      return rewriter.notifyMatchFailure(
+          contractOp, "Not a linalg tile + vector contract operation");
+
+    auto oldReductionForOp = oldKForOp->getParentOfType<scf::ForOp>();
+    if (!oldReductionForOp)
+      return rewriter.notifyMatchFailure(
+          contractOp, "Not a linalg tile + vector contract operation");
```
rengolin, line 105: repetitive notification message doesn't help identify what wen't wrong, when it does. Follow my main comment strategy on error messages.

rengolin (commented):
There's still a lot of repetition in the code.

Both operands and types from the `contractOp` and the nested loop structure can be extracted into separate helper (static) functions that return a list of operations.

```
// Extracts all three operands from the contract op as transfer_read
static
FailureOr<SmallVector<TransferReadOp>> getContractOperands (Op* contractOp) {
  SmallVector<TransferReadOp> list;
  for (int i=0; i<3; i++) {
    auto op = contractOp.getODSOperands(i).getDefiningOp<vector::TransferReadOp>();
    if (!op)
      return failure();
    list.push_back(op);
  }
  return list;
}

// Extracts all four nested loops as ForOp
static
FailureOr<SmallVector<ForOp>> getNestedLoop (Op* contractOp) {
  SmallVector<ForOp> list;
  Op* current = contractOp;
  for (int i=0; i<4; i++) {
    Op* parent = op->getParentOfType<scf::ForOp>();
    if (!parent)
      return failure();
    list.push_back(parent);
    op = parent;
  }
  return list;
}
```

Then you can use them in your function like:
```
  auto operands = getContractOperands(contractOp);
  if (failed(operands))
    return rewriter.notifyMatchFailure(
          contractOp, "Invalid operands for contract op");
  vectorReadOpLhs = operands[0];
  vectorReadOpRhs = operands[1];
  vectorReadOpAcc = operands[2];
  
  ...
  
  auto loops = getNestedLoop(contractOp);
  if (failed(loops))
    return rewriter.notifyMatchFailure(
          contractOp, "Invalid loop nest in contract pattern");
  oldKForOp = loops[0];
  oldReductionForOp = loops[1];
  oldNForOp = loops[2];
  oldMForOp = loops[3];
```

_Please note, the code above was written on a text editor, not tested, not compiled and is here for illustration only. Types and API calls may vary._

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+                                PatternRewriter &rewriter) const override {
+
+    // Check the vector contract operation satisfies the required pattern.
+    // Check the ACC, LHS, and RHS of contract operation
+    auto vectorReadOpAcc =
+        contractOp.getAcc().getDefiningOp<mlir::vector::TransferReadOp>();
+    if (!vectorReadOpAcc)
+      return failure();
+
+    auto subviewOp =
```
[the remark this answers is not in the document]
arun-thmn: Re-structured with new code. Now, the subview is used for checking `induction variable` of nested loops.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+        vectorReadOpAcc.getOperand(0).getDefiningOp<memref::SubViewOp>();
+    if (!subviewOp)
+      return failure();
+
+    auto vectorReadOpLHS =
+        contractOp.getLhs().getDefiningOp<mlir::vector::TransferReadOp>();
+    if (!vectorReadOpLHS)
+      return failure();
+
+    auto vectorReadOpRHS =
```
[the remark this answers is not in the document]
arun-thmn: Yes. The pass is updated for vector read op and loop nests with a function call.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+struct HoistVectorTransferOp : OpRewritePattern<vector::ContractionOp> {
+  using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
+                                PatternRewriter &rewriter) const override {
+
+    // Check the vector contract operation satisfies the required pattern.
+    // Check the ACC, LHS, and RHS of contract operation
+    auto vectorReadOpAcc =
+        contractOp.getAcc().getDefiningOp<mlir::vector::TransferReadOp>();
```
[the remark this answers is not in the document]
arun-thmn: removed.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+namespace mlir {
+namespace tpp {
+#define GEN_PASS_DEF_HOISTVECTORTRANSFERS
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+using namespace mlir;
+using namespace vector;
```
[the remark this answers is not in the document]
arun-thmn: Removed.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (!oldMForOp)
+      return rewriter.notifyMatchFailure(
+          contractOp, "Not a linalg tile + vector contract operation");
+
+    // Move the vector transfer read before the reduction and k loop
+    rewriter.setInsertionPoint(oldReductionForOp);
+    auto *cloneVectorReadOp = rewriter.clone(*vectorReadOpAcc);
+
+    // Code to re-create the reduction and k loop with iter args
+    auto vectorReadOpValue = (*cloneVectorReadOp).getResult(0);
```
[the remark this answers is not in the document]
arun-thmn: Agreed. Updated the code accordingly.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Yes. Updated the pass with functions as said in the main comment.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,209 @@
...
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+namespace mlir {
+namespace tpp {
+#define GEN_PASS_DEF_HOISTVECTORTRANSFERS
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+using namespace mlir;
```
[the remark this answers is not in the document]
arun-thmn: Removed

arun-thmn:
@rengolin Re-structured the pass with functions to do the pattern matching checks for `vector read op`, `subviews`, and `loops`.
Also, addressed the other comments. Please have a look.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+        reductionCount++;
+    }
+
+    auto vectorReadOpLhsType = cast<ShapedType>(vectorReadOpLhs.getType());
+    auto vectorReadOpRhsRank =
+        (cast<ShapedType>(vectorReadOpRhs.getType())).getRank();
+
+    if (reductionCount == 2 &&
+        (vectorReadOpLhsType.getRank() != 3 || vectorReadOpRhsRank != 3))
+      return failure();
```
rengolin, line 143: Couldn't this be a `notifyMatchFailure` to?

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+    vector::TransferWriteOp vectorWriteOperation;
+    Block *bodyBlock = newKForOp.getBody();
+    for (auto &op : bodyBlock->getOperations()) {
+      if (auto vectorContractOp = llvm::dyn_cast<vector::ContractionOp>(op)) {
+        vectorContractOp.setOperand(vectorContractOp.getNumOperands() - 1,
+                                    newKForOp.getRegionIterArgs()[0]);
+        newcontractOpValue = vectorContractOp.getResult();
+      }
+      if (auto yieldOp = llvm::dyn_cast<scf::YieldOp>(op)) {
+        if (newcontractOpValue != NULL)
```
rengolin, line 216: Is there a case where the new K loop does not have a contract op?

If not, then this should be an assert.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+          yieldOp.setOperand(0, newcontractOpValue);
+      }
+      if (auto vectorWriteOp = llvm::dyn_cast<vector::TransferWriteOp>(op)) {
+        vectorWriteOperation = vectorWriteOp;
+      }
+    }
+
+    if (vectorWriteOperation != NULL) {
```
rengolin, line 224: Same here, it the loop has no contract, then it shouldn't have gotten this far. Assert.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+      vectorWriteOperation.setOperand(0, newReductionForOp.getResult(0));
+      vectorWriteOperation->moveBefore(reductionForOp);
+    }
+
+    // Erase the old vector contract operation
```
rengolin, line 229: Don't you need to erase the old K loop, too?

rengolin (commented):
Thanks! Looking a lot better! A few minor comments this round, which should be the last.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+      return failure();
+    list.push_back(vectorReadOp);
+  }
+  return list;
+}
+
+static FailureOr<SmallVector<memref::SubViewOp>>
+getReadOperands(SmallVector<vector::TransferReadOp> readOp) {
+  SmallVector<memref::SubViewOp> list;
+  for (int i = 0; i < 3; i++) {
```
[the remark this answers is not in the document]
arun-thmn: Learning. Updated the code with `range` loops.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+        contractOp.getOperand(i).getDefiningOp<vector::TransferReadOp>();
+    if (!vectorReadOp)
+      return failure();
+    list.push_back(vectorReadOp);
+  }
+  return list;
+}
+
+static FailureOr<SmallVector<memref::SubViewOp>>
+getReadOperands(SmallVector<vector::TransferReadOp> readOp) {
```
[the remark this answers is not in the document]
arun-thmn: Corrected

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    // Check the vector contract operation satisfies the required pattern.
+    // Check the Acc, Lhs, and Rhs of contract operation
+
+    auto operands = getContractOperands(contractOp);
+    if (failed(operands))
+      return rewriter.notifyMatchFailure(contractOp,
+                                         "Invalid operands for contract op");
+
+    auto readOp = *operands;
```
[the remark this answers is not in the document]
arun-thmn: Corrected. Thanks.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+    auto subViewOp = readOp[i].getOperand(0).getDefiningOp<memref::SubViewOp>();
+    if (!subViewOp)
+      return failure();
+    list.push_back(subViewOp);
+  }
+  return list;
+}
+
+static FailureOr<SmallVector<scf::ForOp>>
+getNestedLoop(vector::ContractionOp contractOp,
```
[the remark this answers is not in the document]
arun-thmn: Separated into two function.
`getNestedLoop` - gets all the loops,
`checkNestedLoop` - to validate the loops.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+    auto readOp = *operands;
+    auto vectorReadOpAcc = readOp[2];
+    auto vectorReadOpLhs = readOp[0];
+    auto vectorReadOpRhs = readOp[1];
+
+    // Check whether the operand of vector transfer read is a subview
+    auto subviews = getReadOperands(readOp);
+    if (failed(subviews))
+      return rewriter.notifyMatchFailure(
+          contractOp, "Vector read op operand(s) is/are not a subview");
```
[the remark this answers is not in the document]
arun-thmn: corrected.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    // Check the operation type MatMul, B-MatMul, or BR-MatMul
+    auto contractIteratorTypes = contractOp.getIteratorTypesArray();
+    size_t reductionCount = 0;
+    for (size_t i = 0; i < contractIteratorTypes.size(); i++) {
```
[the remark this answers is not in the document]
arun-thmn: Learning. Updated the pass with `std::count`

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Okay. Replaced with `notifyMatchFailure `

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: We check these in the beginning of the pass after re-structuring. So, this check is not required, removing it.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Same, the check are done already. Removing this condition as well.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: No, I let the DCE to eliminate it in a smart way.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: This change is done via the pass re-structuring.

lib/TPP/Transforms/HoistVectorTransfers.cpp
```diff
@@ -0,0 +1,261 @@
...
+      return failure();
+    list.push_back(subViewOp);
+  }
+  return list;
+}
+
+static FailureOr<SmallVector<scf::ForOp>>
+getNestedLoop(vector::ContractionOp contractOp) {
+  SmallVector<scf::ForOp> list;
+  scf::ForOp current = contractOp->getParentOfType<scf::ForOp>();
```
rengolin, line 62: This is the same check as inside the loop. You can just update the the trip count to 4 and remove this extra logic

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: This change is done via the pass re-structuring.

lib/TPP/Transforms/HoistVectorTransfers.cpp
rengolin: Have you had problems deleting it? I recommend you delete if possible, to make it easier to chain transforms without needing a full-scan DCE in between them.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Ahh, I missed it. The count is set to `4` and the check is moved inside.

lib/TPP/Transforms/HoistVectorTransfers.cpp
arun-thmn: Yes, While I tried erasing the K loops it resulted in `seg faults` with no error msg. So, I took the help of DCE to remove those loops.

arun-thmn:
@rengolin Please find the comments below:
1) Merged the extra same login inside the loop.
2) Yes, While I tried erasing the K loop it resulted in `seg faults` with no error msg. So, I took the help of DCE to remove those loops.

rengolin (approved):
LGTM, thanks!

PR #978, opened 2024-11-11
Lower vector.contract to chain of vector.fma for matrix multiply.

shahidact:
Implements the lowering of vector contraction op for GEMM of size MxN to sequence of vector FMAs wrapped inside scf.for loop with iterargs to accumulate the result of FMAs.The K dimension size is expected to be 1. The idea is to exploit the AVX-512 feature to generate optimal vector code.

### DIFF
--- a/test/Integration/vector-contract-to-fma.mlir
+++ b/test/Integration/vector-contract-to-fma.mlir
@@ -1,6 +1,7 @@
 // RUN: tpp-opt %s  | tpp-run -e entry --entry-point-result=void -seed 123 -print > %t.1
 // RUN: tpp-opt %s  --vector-contract-to-fma  | tpp-run -e entry --entry-point-result=void -seed 123 -print > %t.2
-// RUN: diff %t.1 %t.2 | FileCheck %s --check-prefix=DIFF --allow-empty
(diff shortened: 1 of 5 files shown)


test/Integration/vector-contract-to-fma.mlir
```diff
@@ -0,0 +1,49 @@
+// RUN: tpp-opt %s  | tpp-run -e entry --entry-point-result=void -seed 123 -print > %t.1
+// RUN: tpp-opt %s  --vector-contract-to-fma  | tpp-run -e entry --entry-point-result=void -seed 123 -print > %t.2
+// RUN: diff %t.1 %t.2 | FileCheck %s --check-prefix=DIFF --allow-empty
```
rengolin, line 3: Don't need to call `FileCheck`, if the diff is not empty it will return non-zero and the test will fail.

shahidact: Done as expected.

rengolin (approved):
This is an isolated pass, let's move it in-tree and evolve from there.

I'll create the new bundle after Kavitha's pass is merged and add these to your bundle.

PR #979, opened 2024-11-15
fix: converting cmd output to float can generate a ValueError

AndradeKaio:


adam-smnk:
@AndradeKaio Hey, sorry the PR slipped under my radar.
Could you add description and some reproducer for the issue? Or at least some context for any errors you encountered?

Overall, this test check is pretty naive - perhaps there's a better way to handle it altogether.

PR #980, opened 2024-11-18
Transpose pdll

KavithaTipturMadhu:
tests pending:pr1
Edit: I have updated the PR with unit tests and rebased it on top of the new pipeline, its good for review.
Edit2: Removed the transpose and gemm code, need to rebase on top of the bundle pr, will convert back to non-draft mode once that is done and tests are added.
Edit3: moved to new pipeline,  unit tests and integration tests updated.


rengolin (commented):
Some initial comments

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -115,11 +115,13 @@ struct DefaultTppPasses
         // Vectorizes the remaining Linalg operations
         pm.addNestedPass<func::FuncOp>(createVectorizationPass());
         pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-
-        // TODO: Add a flag for this pass so it doesn't conflate with other options.
-        pm.addNestedPass<func::FuncOp>(createVectorContractToOuterproduct());
+        if (outerProduct) {
```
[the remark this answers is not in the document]
rengolin: After #983 you can just add your stuff to the `VectorToXSMM` bundle.

include/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.h
```diff
@@ -0,0 +1,9 @@
+#include "mlir/Pass/Pass.h"
+using namespace std;
+using namespace mlir;
+
+namespace mlir {
+namespace tpp {
+std::unique_ptr<mlir::Pass> createInsertTranspose();
```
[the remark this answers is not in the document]
rengolin: Actually, scratch that. This should reuse the bundle introduced in #983.

include/TPP/Transforms/Utils/VNNIUtils.h
```diff
@@ -40,12 +45,19 @@ std::optional<int64_t> getVnniBlockingFactor(Type type);
 // Return true if the memref is in VNNI layout with rank `expectedRank`.
 bool isInVnniLayout(VnniOperandRank expectedRank, MemRefType memref);
 
+// Return true if the memref is in VNNI layout with rank `expectedRank`.
+bool isInVnniLayout(VnniOperandRank expectedRank, VectorType vector);
```
alheinecke, line 49: The we need VNNI Utils for transpose. I would suggest adding Norm->VNNI function call lowering in a different PR to keep PRs small.

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -21,6 +26,22 @@ class ArrayAttr;
 class Operation;
 
 namespace xsmm {
+
+struct BrgemmInfo {
```
alheinecke, line 30: Why do we need the BRGEMM Info struct when lowering Transpose.

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -77,6 +104,9 @@ FailureOr<UnaryFlags> getUnaryFlags(Type inputType, Type outputType);
 
 // Compute the broadcasting flags for 'operandType' based on 'outputType'.
 enum class OperandPos { LHS = 0, RHS = 1 };
+FailureOr<BinaryFlags> getBinFlags(ArrayRef<int64_t> shapeOutput,
```
alheinecke, line 107: Transpose is unary, no addition on Binary is needed.

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -92,6 +122,40 @@ template <typename DispatchOpTy>
...
+
+FailureOr<vector::ContractionOp>
+makeMinorDimensionsInnerMost(RewriterBase &rewriter,
+                             vector::ContractionOp contractOp, unsigned m,
+                             unsigned n, unsigned k, xsmm::DataTypeAttr type);
+
+std::optional<unsigned>
+getPosInCodomain(unsigned dim, vector::ContractionOp contractOp, AffineMap map);
+
+LogicalResult checkAccess(PatternRewriter &rewriter,
```
alheinecke, line 140: This seems to be GEMM related as well, no need in my eyes.

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -92,6 +122,40 @@ template <typename DispatchOpTy>
...
+SmallVector<Value> getOperands(OpBuilder &builder, Location loc,
+                               ValueRange operands, IntegerAttr dataTypeAttr);
+
+FailureOr<vector::ContractionOp>
+makeMinorDimensionsInnerMost(RewriterBase &rewriter,
+                             vector::ContractionOp contractOp, unsigned m,
+                             unsigned n, unsigned k, xsmm::DataTypeAttr type);
+
+std::optional<unsigned>
+getPosInCodomain(unsigned dim, vector::ContractionOp contractOp, AffineMap map);
```
alheinecke, line 138: This seems to be GEMM related as well, no need in my eyes.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,405 @@
...
+    populateGeneratedPDLLPatterns(patternList);
+    patterns = std::move(patternList);
+    return success();
+  }
+
+  void runOnOperation() final {
+    PatternRewriter rewriter(&getContext());
+    // Enable conversion for linalg.generic to XSMM Brgemm if possible.
+    auto res =
+        getOperation()->walk([&](mlir::vector::ContractionOp contractOp) {
```
alheinecke, line 286: We want to rewrite transpose, why are we walking over contraction ops?

alheinecke (commented):
It seems there is also some BRGEMM matching in this PR which I don't understand. I thought the idea is to rewrite transpose.

include/TPP/Transforms/Utils/VNNIUtils.h
KavithaTipturMadhu: We need this for when contraction inputs are not compatible to be lowered into brgemm. Should we seperate that out into a different patch? If yes, I'll move this out and keep just the transpose lowering in this patch for now. That makes the pr quite clean and small.

include/TPP/Dialect/Xsmm/XsmmUtils.h
KavithaTipturMadhu: Agreed, will remove this. It got left out while moving the code.

include/TPP/Dialect/Xsmm/XsmmUtils.h
KavithaTipturMadhu: Same comment as above, will remove.

include/TPP/Dialect/Xsmm/XsmmUtils.h
KavithaTipturMadhu: Will remove, same comment as above.

include/TPP/Dialect/Xsmm/XsmmUtils.h
KavithaTipturMadhu: same comment as above.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
KavithaTipturMadhu: Like i said earlier, I thought I'd have the code to insert transposes when gemms cannot be performed in this patch itself. If we want to keep this in the brgemm/gemm lowering patch, I'll move this to the next patch.

include/TPP/Transforms/Utils/VNNIUtils.h
rengolin: > If yes, I'll move this out and keep just the transpose lowering in this patch for now.

Yes, please. Keep here _only_ transpose logic.

lib/TPP/GPU/GpuVectorize.cpp
```diff
@@ -29,6 +29,7 @@
 #include "mlir/Transforms/Passes.h"
 
 using namespace mlir;
+using namespace mlir::vector;
```
rengolin, line 32: Why is this here? Please keep the unrelated changes for further PRs. This carries the `mlir::vector` changes everywhere that are not necessary.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,299 @@
...
+    operandRange.push_back(dispatched.getResult(0));
+    operandRange.push_back(input->getOperand(0));
+    operandRange.push_back(output->getOperand(1));
+    SmallVector<Value> inputs;
+    SmallVector<Value> preceedingOperands;
+    auto invokeCall = xsmm::utils::buildInvokeCall(
+        rewriter, output, module, inputs, preceedingOperands, -1, operandRange,
+        invokeName, dtype, true);
+
+    return std::make_pair(&*dispatched, &*invokeCall);
```
rengolin, line 129: This is a repeat of the code below. Please condense the two transposes into one function with different arguments for VNNI and non-VNNI.

lib/TPP/GPU/GpuVectorize.cpp
KavithaTipturMadhu: Its not an unrelated change, it is a build fix, vector is not disambiguated otherwise. std::vector and mlir:vector are both visible to the .cpp file. We need to explicitly make mlir::vector visible for build to go through.

rengolin: This should only happen if both `std` and `mlir` namespaces are in scope. You should not have `using namespace std` in LLVM source files.

include/TPP/Passes.h
```diff
@@ -98,6 +98,8 @@ class XeGPUDialect;
 
 } // namespace mlir
 
+#include "TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.h"
```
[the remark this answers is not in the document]
KavithaTipturMadhu: We need to make the headers visible to tpp-opt somehow in order to make the passes available in standalone mode. We can include them directly in both tpp-opt and in VectorToXsmm.cpp, or we could include them in Passes.h. I thought it made sense to include them in Passes.h because  they are passes after all, and they ought to be visible wherever the rest of the passes are, but if you think that needn't be, we can do what I said prreviously i.e., include the headers in two places.

include/TPP/Passes.td
```diff
@@ -54,7 +54,18 @@ def VectorizationPass : Pass<"vectorization-pass",
     Vectorization pass to convert linalg to memref/vector instead of Xsmm.
   }];
 
-  let dependentDialects = [ "memref::MemRefDialect", "linalg::LinalgDialect", "vector::VectorDialect" ];
+  let dependentDialects = [ "memref::MemRefDialect", "linalg::LinalgDialect", "mlir::vector::VectorDialect" ];
+}
+
+def InsertTranspose : Pass<"insert-transpose-pass", "func::FuncOp"> {
```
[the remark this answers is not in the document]
KavithaTipturMadhu: This pass just converts a transpose, the next pass converts the rest of the constructs to xsmm. Isn't it an overkill to match transpose pattern in the pass and then invoke the pass, when the pass itself does that for you? The pass is currently translating just transposes, nothing else, so it pattern matches only  transposes. It achieves the same effect as having to match transpose pattern and lowering only those.

lib/TPP/DefaultPipeline.cpp
```diff
@@ -64,17 +64,19 @@ llvm::cl::opt<bool> lowerPackUnpackWithoutTranspose(
...
 llvm::cl::list<unsigned>
-    rhsTile("rhsTile",
-                     llvm::cl::desc("Rhs tile size for brgemm operation"),
-                     llvm::cl::list_init<unsigned>(SmallVector<unsigned>{8, 16}),
-                     llvm::cl::CommaSeparated);
+    rhsTile("rhsTile", llvm::cl::desc("Rhs tile size for brgemm operation"),
+            llvm::cl::list_init<unsigned>(SmallVector<unsigned>{8, 16}),
+            llvm::cl::CommaSeparated);
+
+llvm::cl::opt<bool> vectorToXSMM("vector-To-XSMM",
```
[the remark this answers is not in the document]
KavithaTipturMadhu: I tried to use the same flag as the one in passbundles, I get an error saying there's no such option in tpp-run. However, if I try to add another command with the same name, it complains that the command already exists. I didn't know to resolve it immediately and added a command with a different name for now.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,299 @@
...
+        dyn_cast<MemRefType>(output->getOperand(1).getType()),
+        dyn_cast<VectorType>(input->getResult(0).getType()),
+        dyn_cast<VectorType>(output->getOperand(0).getType()),
+        xsmm::UnaryFlags::NONE);
+
+    if (failed(unaryInfo)) {
+      return failure(transposeOp);
+    }
+  }
+  return success(transposeOp);
```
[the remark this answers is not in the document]
KavithaTipturMadhu: I cant return just transposeOp because that invokes the private constructor of LogicalResult and results in a compiler error.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,299 @@
...
+        xsmm::UnaryFlags::NONE);
+    if (failed(unaryInfo)) {
+      return failure(transposeOp);
+    }
+  } else {
+    if (!xsmm::utils::isTwoDTransposeOp(
+            dyn_cast<mlir::vector::TransposeOp>(transposeOp))) {
+      return failure(transposeOp);
+    }
+    auto unaryInfo = xsmm::utils::getVectorUnaryInfo(
```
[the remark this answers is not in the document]
KavithaTipturMadhu: This was actually a mistake on my part, we dont need to check unary info call in case of vnni transpose, I have fixed the validation call.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,299 @@
...
+  patterns.getPDLPatterns().registerRewriteFunction("BuildTranspose",
+                                                    buildTransposeOp);
+  patterns.getPDLPatterns().registerConstraintFunction("ValidateTranspose",
+                                                       validateTransposeOpImpl);
+}
+
+namespace mlir {
+namespace tpp {
+
+struct ConvertTranspose
```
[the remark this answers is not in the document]
KavithaTipturMadhu: Actually, this one is just the transpose conversion, because we want to prioritize transpose conversions over other ops, and we don't know how else to force conversion of transposes first before converting other ops. That's the reason transposes are in a separate pdll and the rest of the ops are in a subsequent pdll.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTransposePDL.pdll
```diff
@@ -0,0 +1,22 @@
+#include "mlir/Dialect/Func/IR/FuncOps.td"
+#include "mlir/Dialect/Vector/IR/VectorOps.td"
+#include "mlir/Dialect/Arith/IR/ArithOps.td"
+#include "mlir/IR/OpBase.td"
+#include "mlir/IR/BuiltinTypes.td"
+
+
+Rewrite BuildTranspose(op:Op<vector.transpose>, input0:Op<vector.transfer_read>, output: Op<vector.transfer_write>, outputType:TypeRange)->(dispatch:Op<func.callOp>, invoke:Op<func.callOp>);
```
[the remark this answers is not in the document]
KavithaTipturMadhu: As I mentioned, this one is just transpose converter and not general xsmm cnverter, the brgemm patch has the general xsmm converter :  https://github.com/plaidml/tpp-mlir/pull/985

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,299 @@
...
+      SymbolRefAttr::get(transposeOp->getContext(), dispatchName));
+
+  SmallVector<Value> operandRange;
+  operandRange.push_back(rewriter.create<arith::ConstantOp>(
+      loc, integer64, cast<TypedAttr>(dtype)));
+  operandRange.push_back(dispatched.getResult(0));
+  operandRange.push_back(input->getOperand(0));
+  operandRange.push_back(output->getOperand(1));
+  SmallVector<Value> inputs;
+  SmallVector<Value> preceedingOperands;
```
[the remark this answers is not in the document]
KavithaTipturMadhu: The function takes smallvector arguments, however, I can initialize directly without push backs, I have changed that.

include/TPP/Passes.td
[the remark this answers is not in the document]
rengolin: We're talking about public names. The final objective of the pass is to convert the whole set of operations, even if it just does transposes for now. We should not change the name of the pass, classes and functions as we add more functionality.

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -88,6 +98,28 @@ ArrayAttr getUnaryDispatchFlags(UnaryOp op);
 
 ArrayAttr getBinaryDispatchFlags(BinaryOp op);
 
+int64_t getOredFlags(ArrayAttr flags);
```
rengolin, line 101: This already exists in `ConvertXsmmToFunc`, please reuse there.

lib/TPP/Conversion/ConvertVectorToXsmm/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+add_mlir_pdll_library (
```
rengolin, line 1: Do we really need one PDLL library per pattern? This sounds like a huge scalability problem.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
rengolin: I can't see the changes here, did you push them to the branch?

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
rengolin: Code is still duplicated, at least on PR, please don't mark as done until we can all see the changes.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
rengolin: > we want to prioritize transpose conversions over other ops

I'm not sure about this. Can you expand?

> That's the reason transposes are in a separate pdll

I understood we were just adding transpose first because it's easier and allows us to review the infrastructure before the complex ops. I don't see any special reason why the transpose should be before any other.

But even if so, there has to be an order to PDL patterns that is not individually running them. For instance, we want to look at fused GEMMs before plain GEMMs, softmax before a sequence of element-wise and reduction, etc.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
rengolin: I'm not sure what you expect passing the op will do. I expected you wanted the actual operations, but if you really just want the `LogicalResult`, then adding arguments will convert them to a `bool` and change the meaning of `success` / `failure` if the argument converts to `true` or `false`.

lib/TPP/DefaultPipeline.cpp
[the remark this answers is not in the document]
rengolin: Please don't just create a new option if you can't use an existing one.

For `tpp-opt`, because you have to pass `-default-tpp-passes`, you need to add the flag to that one, for example `-default-tpp-passes="vector-to-xsmm"`. In `tpp-run`, because the default TPP passes will run by default, you can just call `-vector-to-xsmm`.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -8,18 +8,67 @@
...
 #include "mlir/IR/TypeUtilities.h"
 
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SetOperations.h"
 #include "llvm/Support/Compiler.h"
+#include "llvm/Support/Debug.h"
+#include <map>
+#define DEBUG_TYPE "xsmm-utils"
+
+using namespace mlir;
```
rengolin, line 30: this is unnecessary, as below you're already in the `mlir` namespace.

Please don't use `using namespace` in these files, as it complicates namespace resolution. Use the existing pattern of entering and leaving namespaces as needed.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +358,175 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+    return failure();
+  // fail if the strides are non-constant
+  if (llvm::any_of(strides, [](int64_t stride) {
+        return stride == ShapedType::kDynamic;
+      }))
+    return failure();
+  return strides[pos];
+}
+
+bool isTwoDTransposeOp(vector::TransposeOp transposeOp) {
```
rengolin, line 383: Sounds like a simpler approach would be to create a function to get the rank of a vector (in view of VNNI layout, which just needs the operand).

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +358,175 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+  auto operandType = dyn_cast<VectorType>(transposeOp.getOperand().getType());
+  bool isVnni = vnni::utils::isInVnniLayout(operandType.getRank(), operandType);
+  if (isVnni || operandType.getRank() != 2 ||
+      dyn_cast<VectorType>(transposeOp.getResult().getType()).getRank() != 2 ||
+      (isa<scf::ForallOp>(transposeOp->getParentOp()) &&
```
rengolin, line 388: This is unstable, as the parent op could be `scf::parallel` or `scfp:for` or `func::func` and you'd have to multiply the checks here.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +358,175 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+      results.push_back(basePtrType);
+      results.push_back(builder.getIndexType()); // offset
+    } else {
+      results.push_back(operand.getType());
+    }
+  }
+  return results;
+}
+
+int64_t getOredFlags(ArrayAttr flags) {
```
rengolin, line 432: This sounds like a GEMM function, but you use for the unary check below. I'm not sure what's going on.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +358,175 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+                               ArrayRef<Type> dispatchOperandTypes,
+                               ModuleOp module, FlatSymbolRefAttr fnName) {
+  auto libFnType = rewriter.getFunctionType(
+      dispatchOperandTypes, IntegerType::get(rewriter.getContext(), 64));
+
+  if (!module.lookupSymbol(fnName.getAttr())) {
+    OpBuilder::InsertionGuard guard(rewriter);
+    // Insert before module terminator.
+    rewriter.setInsertionPoint(module.getBody(),
+                               std::prev(module.getBody()->end()));
```
rengolin, line 467: Code repetition, please adapt `BuilderUtils::createFunction`.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +358,175 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+    func::FuncOp funcOp =
+        rewriter.create<func::FuncOp>(loc, fnName.getValue(), libFnType);
+    funcOp.setPrivate();
+  }
+
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPointAfter(dispatchOperands.back().getDefiningOp());
```
rengolin, line 474: This doesn't seem necessary. If you're building this function now, then all the producers have already been lowered into IR and the insertion point is at the right place?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +358,175 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+  }
+  finalOperands.append(extraOperands.begin(), extraOperands.end());
+  SmallVector<Type> invokeTypes =
+      xsmm::utils::extractInvokeOperandTypes(rewriter, finalOperands);
+  auto loc = parentOp->getLoc();
+  auto libFnType = rewriter.getFunctionType(invokeTypes, {});
+  FlatSymbolRefAttr fnName =
+      SymbolRefAttr::get(rewriter.getContext(), invokeName);
+
+  if (!module.lookupSymbol(fnName)) {
```
rengolin, line 510: Code repetition, please adapt `BuilderUtils::createFunction`.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,253 @@
...
+  SmallVector<Value, 10> dispatchOperands;
+  SmallVector<Type, 10> dispatchOperandTypes;
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(&*functionOp.getBody().op_begin());
+  xsmm::UnaryKindAttr kind =
+      xsmm::UnaryKindAttr::get(rewriter.getContext(), unaryKind);
+  IntegerType integer64 = IntegerType::get(rewriter.getContext(), 64);
+  dispatchOperands.push_back(rewriter.create<arith::ConstantOp>(
+      loc, integer64, cast<TypedAttr>(kind)));
+  dispatchOperandTypes.push_back(integer64);
```
rengolin, line 65: Can't you really get the type from the value list?

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+  auto functionOp = transposeOp->getParentOfType<func::FuncOp>();
+  SmallVector<Value, 10> dispatchOperands;
+  SmallVector<Type, 10> dispatchOperandTypes;
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(&*functionOp.getBody().op_begin());
+  xsmm::UnaryKindAttr kind =
+      xsmm::UnaryKindAttr::get(rewriter.getContext(), unaryKind);
+  IntegerType integer64 = IntegerType::get(rewriter.getContext(), 64);
+  dispatchOperands.push_back(rewriter.create<arith::ConstantOp>(
```
rengolin, line 63: `BuilderUtils::getConstInt`

lib/TPP/GPU/LinalgToXeGPU.cpp
```diff
@@ -196,7 +196,7 @@ static std::optional<Value> lowerGenericOp(linalg::GenericOp genericOp,
     }
 
     auto zeroVec =
-        rewriter.create<vector::BroadcastOp>(loc, resType, zeroConst);
+        rewriter.create<mlir::vector::BroadcastOp>(loc, resType, zeroConst);
```
rengolin, line 199: If this change is necessary now (it wasn't before), then you changed something in the namespace resolution. Make sure you remove the change and these lines can continue as they were.

Always revisit your assumptions when a file that should remain untouched needs to be touched. Often you're doing something unnecessary elsewhere.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -69,6 +69,17 @@ FailureOr<AffineDimExpr> isInVnniLayout(linalg::GenericOp linalgOp,
   return failure();
 }
 
+bool isInVnniLayout(VnniOperandRank expectedRank, VectorType vector) {
+  return isInVnniLayout((int64_t)expectedRank, vector);
```
rengolin, line 73: this wrapper seems unnecessary, this C-style cast is not appropriate.

lib/TPP/Conversion/ConvertVectorToXsmm/CMakeLists.txt
KavithaTipturMadhu: I didn't say we need one per pattern, we just need one for transpose and then one for the rest of them to make sure transpose runs before the others do. This is so that we can insert and translate transposes before gemms/brgemms.

test/Conversion/VectorToXsmm/vector-to-transpose.mlir
```diff
@@ -0,0 +1,129 @@
...
+// CHECK-NEXT: %[[indexCast1:.*]] = arith.index_cast %[[intptr]]
+// CHECK-NEXT: %[[inttoptr:.*]] = llvm.inttoptr %[[indexCast1]]
+// CHECK: %[[base_buffer_0:.*]], %[[offset_1:.*]], %[[sizes_2:.*]]:3, %[[strides_3:.*]]:3 = memref.extract_strided_metadata %[[arg1]]
+// CHECK-DAG:  %[[intptr0:.*]] = memref.extract_aligned_pointer_as_index %[[arg1]]
+// CHECK-NEXT: %[[indexCast2:.*]] = arith.index_cast %[[intptr0]]
+// CHECK-NEXT: %[[inttoptr2:.*]] = llvm.inttoptr %[[indexCast2]]
+// CHECK:   call @xsmm_unary_invoke(%[[c2_i64]], %[[dispatch]], %[[inttoptr]], %[[offset]], %[[inttoptr2]], %[[offset_1]])
+
+// -----
+func.func @not_vnni_packing_1(%arg0: memref<32x32xf32, strided<[512, 1], offset: ?>>, %arg1: memref<16x32x2xf32, strided<[64, 2, 1], offset: ?>>) {
```
rengolin, line 74: took me a while to realize this was `f32`. Please update the function name with `not_vnni_in_f32` or something.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: I thought you were referring to the operandRange SmallVector. I will update the other arguments.

test/Integration/tpp-transpose.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: tpp-run --vector-To-XSMM %s -e entry -entry-point-result=void -print --seed 123 2>&1 | FileCheck %s
+// RUN: tpp-run --linalg-to-loops %s -e entry -entry-point-result=void -print --seed 123 2>&1 | FileCheck %s
+
+module {
+  func.func @entry(%arg0: tensor<3x5xf32>, %arg1: tensor<5x3xf32>)->tensor<5x3xf32> {
+    %out = linalg.transpose ins(%arg0 : tensor<3x5xf32>) outs(%arg1 : tensor<5x3xf32>) permutation = [1, 0]
```
rengolin, line 6: From my interpretation, your pattern matching above needs vector read/write to match the pattern. Will this really be converted into XSMM calls?

Can you add another RUN like that prints the MLIR=late and CHECK for xsmm calls?

tools/tpp-opt/tpp-opt.cpp
```diff
@@ -30,6 +30,7 @@
 #include "TPP/Dialect/Xsmm/XsmmDialect.h"
 #include "TPP/PassBundles.h"
 #include "TPP/Passes.h"
+#include "TPP/Passes.h.inc"
```
rengolin, line 33: Nope. If you need this, something is really wrong elsewhere.

test/Integration/tpp-transpose-bf16.mlir
```diff
@@ -0,0 +1,525 @@
+// RUN: tpp-run --vector-To-XSMM %s -e entry -entry-point-result=void -print --seed 123 2>&1 | FileCheck %s
+// RUN: tpp-run --linalg-to-loops %s -e entry -entry-point-result=void -print --seed 123 2>&1 | FileCheck %s
```
rengolin, line 2: Same comment as the other integration test: please add a RUN line to make sure we continue to match vector to XSMM.

test/Integration/tpp-transpose-bf16.mlir
```diff
@@ -0,0 +1,525 @@
...
+// CHECK:	( 0, 0 )
+// CHECK:	( 0, 0.0256348 )
+// CHECK:	( 0.15625, 0 )
+// CHECK:	( 0, 0.104492 )
+// CHECK:	( 0, 0 )
+// CHECK:	( 0.131836, 0 )
+// CHECK:	( 0.207031, 0 )
+// CHECK:	( 0, 0.0991211 )
+// CHECK:	( 0.0649414, 0.242188 )
+// CHECK:	( 0.154297, 0.171875 )
```
rengolin, line 44: you don't need to check against ALL values, just the first dozen or so.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: Which code are you referring to? I thought you meant the unary info call.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: Change the meaning how? The diagnostic engine is expected to propagate the value of transposeop, not change the return value of success/failure, yes? Unless my understanding is completely wrong.

include/TPP/Dialect/Xsmm/XsmmUtils.h
KavithaTipturMadhu: I did not want to add dependences between the two conversion modules, and we were planning to retire ConvertXsmmToFunc after this. But if you suggest that I reuse this function, I will for now, We can move it to xsmmutils when we retire it, I suppose.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: I realize that, the flag is always unary::none so we don't need oredflags function. I'll get rid of it.

lib/TPP/GPU/LinalgToXeGPU.cpp
KavithaTipturMadhu: My bad, that was a stray change, I'll get rid of it.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
KavithaTipturMadhu: Agree, will get rid of it.

test/Integration/tpp-transpose.mlir
KavithaTipturMadhu: Will add print-mid/print-late

tools/tpp-opt/tpp-opt.cpp
KavithaTipturMadhu: This isn't necessary, I forgot to remove this.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
KavithaTipturMadhu: I'll change it to get the type from the value list.

lib/TPP/GPU/LinalgToXeGPU.cpp
KavithaTipturMadhu: I have reverted the changes to both LinalgToXeGPU and GPUVectorize, theyre no longer needed as I've changed the dialect dependence to mlir::vector::VectorDialect instead of vector::VectorDialect.

lib/TPP/DefaultPipeline.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: I tried invoking tpp-run with --vector-to-xsmm. I get an error "Unknown command line argument". Can you please let me know how to pass vector-to-xsmm flag to tpp-run? I don't see it in help either.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: We run scf:foralltoparallelloop conversion after this pass. As for the funcop, its taken care of in this check already.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Function creation is right before, so that moves the insertion point. Thats why I think I set the insertion point explicitly, nevertheless I'll verify if I can remove this line safely.

include/TPP/Passes.td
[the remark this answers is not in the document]
KavithaTipturMadhu: As discusssed, I'm not implementing transposes for brgemms for now. Hence, merging transposes and other vector ops in one and renaming these pdll and cpp files as ConvertVectorToXsmm.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: As discussed, we'll be moving transposes and other vector ops into a single pdll file for now.

include/TPP/Dialect/Xsmm/XsmmUtils.h
KavithaTipturMadhu: Not depending on ConvertXsmmToFunc now, because this function is no longer required.

lib/TPP/Conversion/ConvertVectorToXsmm/CMakeLists.txt
KavithaTipturMadhu: As discussed, we'll be moving transposes and other vector ops into a single pdll file for now.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: I didn't realize this immediately, but we miss out on licm for dispatch calls if we adapt BuilderUtils::createFunction in place of this function. What do you suggest we do here?

test/Integration/tpp-transpose-bf16.mlir
KavithaTipturMadhu: I didnt see this comment before, we might end up passing incorrect size arguments to xsmm call which might go unnoticed if we dont check ALL values. Am i missing something?

lib/TPP/DefaultPipeline.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: Can you please tell me how to pass the right argument? I cant run tests and I have no clue how to pass the argument, I tried everything, all variants of vector-to-xsmm, nothing seems to work.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: This funciton belongs in the brgemm patch. Removing this.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: @rengolin can you please help me understand what I am missing here?

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: I have updated the function to take just one SmallVector of arguments, I have kept it as SmallVector initialized statically and hence it gets allocated on stack.

lib/TPP/DefaultPipeline.cpp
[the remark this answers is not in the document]
KavithaTipturMadhu: I had to add the argument to tpp-run.cpp, just like gpuBackend commandline argument, and that is working.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Removed the namespace mlir in XsmmUtils.cpp. However, I haven't removed it in other files because that meant namespace resolution for a bunch of objects there.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Just as a side note, getOredFlags is used on unaryFlags in ConvertXsmmToFunc which is why I ended up using the same utility. We'll need to update that as well, till such a time that we retire that conversion.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Removed the insertion point, as this wasnt necessary as rightly pointed out by Renato.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Done.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertTranspose.cpp
KavithaTipturMadhu: Done.

test/Conversion/VectorToXsmm/vector-to-transpose.mlir
```diff
@@ -0,0 +1,129 @@
+// RUN: tpp-opt --vector-to-xsmm %s --split-input-file | FileCheck %s
+
+func.func @transpose_op_0(%arg0: memref<3x5xf32>, %arg1: memref<5x3xf32>) {
```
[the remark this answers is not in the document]
KavithaTipturMadhu: Done.

test/Conversion/VectorToXsmm/vector-to-transpose.mlir
KavithaTipturMadhu: Done.

test/Integration/tpp-transpose.mlir
KavithaTipturMadhu: Done.

tools/tpp-opt/tpp-opt.cpp
KavithaTipturMadhu: Done.

test/Integration/tpp-transpose-bf16.mlir
KavithaTipturMadhu: Done.

test/Integration/tpp-transpose-bf16.mlir
KavithaTipturMadhu: @rengolin what do you think? Am i missing something here?

KavithaTipturMadhu: as discussed, I have reduced the problem size so that the results are smaller and can be fully compared.

KavithaTipturMadhu:
I have mostly addressed all the review comments as per my understanding. I havent marked the conversations as closed yet since they may have opens in your opinion @rengolin Please review it wn you have the time. The CI is green.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -8,18 +8,63 @@
...
 #include "mlir/IR/TypeUtilities.h"
 
+#include "TPP/Transforms/Utils/BuilderUtils.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/Compiler.h"
+#include "llvm/Support/Debug.h"
+#define DEBUG_TYPE "xsmm-utils"
+
+using namespace mlir::linalg;
+using namespace mlir::structured_match;
```
rengolin, line 28: Do you really need this?

include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -11,7 +11,9 @@
 
 #include "TPP/Dialect/Xsmm/XsmmEnum.h"
 #include "TPP/Dialect/Xsmm/XsmmOps.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "TPP/IR/StructuredOpMatcher.h"
```
rengolin, line 14: Do not add more headers to headers if you're not using it in the actual header. Add it to the CPP file that uses it.

include/TPP/Passes.td
```diff
@@ -54,7 +54,7 @@ def VectorizationPass : Pass<"vectorization-pass",
     Vectorization pass to convert linalg to memref/vector instead of Xsmm.
   }];
 
-  let dependentDialects = [ "memref::MemRefDialect", "linalg::LinalgDialect", "vector::VectorDialect" ];
+  let dependentDialects = [ "memref::MemRefDialect", "linalg::LinalgDialect", "mlir::vector::VectorDialect" ];
```
rengolin, line 57: It's still unclear to me why you need this now since we didn't before. There is no clash with `std::vector`.

lib/TPP/Transforms/Utils/BuilderUtils.cpp
```diff
@@ -35,13 +35,15 @@ arith::ConstantOp getConstant(OpBuilder &builder, Type type, ValueT value) {
 } // anonymous namespace
 
 func::FuncOp createFunction(OpBuilder &builder, ModuleOp module, StringRef name,
-                            TypeRange args, TypeRange ret) {
+                            TypeRange args, TypeRange ret, bool createBody) {
```
rengolin, line 38: IIRC, your removed code had private visibility, while this function gives public visibility. It should not matter for our purposes, just pointing out because you manually set it. Making sure you're not missing anything.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmmPDL.pdll
```diff
@@ -0,0 +1,22 @@
...
+#include "mlir/Dialect/Arith/IR/ArithOps.td"
+#include "mlir/IR/OpBase.td"
+#include "mlir/IR/BuiltinTypes.td"
+
+
+Rewrite ConvertTranspose(op:Op<vector.transpose>, input0:Op<vector.transfer_read>, output: Op<vector.transfer_write>, outputType:TypeRange)->(dispatch:Op<func.callOp>, invoke:Op<func.callOp>);
+
+Constraint ValidateTranspose(op:Op<vector.transpose>, input0:Op<vector.transfer_read>, output:Op<vector.transfer_write>, outputType:TypeRange);
+
+Pattern ConvertTransposeiPattern{
```
rengolin, line 12: typo

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmmPDL.pdll
```diff
@@ -0,0 +1,22 @@
...
+     let input0 = op<vector.transfer_read>(alloc0:Value, indices0:ValueRange, const0:Value, constIndex:ValueRange)->(output:TypeRange);
+     let transpose = op<vector.transpose>(input0)->(transposeOutput0:Type);
+     let output0 = op<vector.transfer_write>(transpose, alloc1:Value, outindices:ValueRange, constIndex2:ValueRange)->(typeRange:TypeRange);
+     ValidateTranspose(transpose, input0, output0, transposeOutput0);
+     rewrite transpose with{
+        let replacement = ConvertTranspose(transpose, input0, output0, transposeOutput0);
+        replace transpose with (replacement.dispatch, replacement.invoke);
+        erase output0;
```
rengolin, line 20: shouldn't it erase `input0` too?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
rengolin: you still seem to be using in the code? Is this really just for the brgemm patch?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +354,76 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+                           FlatSymbolRefAttr fnName, Operation *insertBefore) {
+  auto operandTypes = xsmm::utils::extractOperandTypes(rewriter, operands);
+  func::FuncOp funcOp = createFunction(rewriter, module, fnName.getValue(),
+                                       operandTypes, results, false);
+  funcOp.setPrivate();
+
+  func::CallOp call;
+  //  if (insertAfter) {
+  //  rewriter.setInsertionPointAfter(insertAfter);
+  if (insertBefore) {
```
rengolin, line 414: Avoid the pattern of long if/else blocks. Use the if/else to build the argument (`operands`) then emit a single `call = ` line after the block.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +354,76 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+  }
+  return results;
+}
+
+static SmallVector<Value> getOperands(OpBuilder &builder, Location loc,
+                                      ValueRange operands,
+                                      IntegerAttr dataTypeAttr,
+                                      Operation *parentOp) {
+  SmallVector<Value> res;
+  builder.setInsertionPoint(parentOp);
```
rengolin, line 388: Please do _not_ use `setInsertionPoint` without an insertion guard.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -287,6 +354,76 @@ FailureOr<BinaryFlags> getBinaryFlags(Type operandType, Type outputType,
...
+  func::FuncOp funcOp = createFunction(rewriter, module, fnName.getValue(),
+                                       operandTypes, results, false);
+  funcOp.setPrivate();
+
+  func::CallOp call;
+  //  if (insertAfter) {
+  //  rewriter.setInsertionPointAfter(insertAfter);
+  if (insertBefore) {
+    rewriter.setInsertionPoint(insertBefore);
+    auto finalOperands = xsmm::utils::getOperands(
```
rengolin, line 416: This seems an arbitrary decision?

Why is it that, when `insertBefore` is non null, you extract the operands, but when it's null, you don't?

Is this just the difference between `dispatch` and `invoke`? If so, then this should be explicit in the code and not veiled behind the existence of an unrelated (to this code at least) argument.

tools/tpp-run/tpp-run.cpp
```diff
@@ -147,6 +147,10 @@ llvm::cl::opt<bool>
                llvm::cl::desc("Kernel buffers are allocated on GPU"),
                llvm::cl::init(true));
 
+llvm::cl::opt<bool> defVectorToXSMM("vector-to-xsmm",
```
rengolin, line 150: Please do not put this here. This is the wrong place. None of the other options are here because this is not a tool option, but a pipeline option. All of the other pipeline options are in the pipeline code. You have already mentioned `lhsTile` and `rhsTile`, and they are not here either.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+  Location loc = transposeOp->getLoc();
+
+  ModuleOp module = transposeOp->getParentOfType<ModuleOp>();
+  auto dtype =
+      xsmm::utils::getDataType(rewriter, transposeOp->getOperand(0).getType());
+
+  auto functionOp = transposeOp->getParentOfType<func::FuncOp>();
+  SmallVector<Value, 10> dispatchOperands;
+  OpBuilder::InsertionGuard guard(rewriter);
+  rewriter.setInsertionPoint(&*functionOp.getBody().op_begin());
```
rengolin, line 53: This change of insertion point doesn't need to be this early. Move it down with the use (`buildXsmmCall`) so that it's clear why you're changing it.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+      rewriter.getContext(), ArrayRef<int64_t>{unaryInfo.m, unaryInfo.n,
+                                               unaryInfo.ldi, unaryInfo.ldo});
+  for (auto idx = 0; idx < dims.size(); idx++) {
+    dispatchOperands.push_back(getConstInt(rewriter, dims[idx], 64));
+  }
+  IntegerAttr unaryFlags = dyn_cast<IntegerAttr>(
+      xsmm::UnaryFlagsAttr::get(rewriter.getContext(), xsmm::UnaryFlags::NONE));
+  dispatchOperands.push_back(getConstInt(rewriter, unaryFlags.getInt(), 64));
+
+  auto dispatched = xsmm::utils::buildXsmmCall(
```
rengolin, line 72: As discussed in the definition, `buildXsmmCall` is ambiguous. You need to make it clear which of them you're changing it (dispatch, invoke).

If you have two calls (or a type argument), then you don't even need to change the insertion point in this function. You can set it for _dispatch_ and the insertion guard will make it sure that when you build the `invoke`, it will be back in the right place.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+    auto stridesOnInput = mlir::utils::getStaticStrides(source);
+    unaryInfo.ldi = stridesOnInput->front();
+    auto stridesOnOutput = mlir::utils::getStaticStrides(output->getOperand(1));
+
+    // Adjust ldo based on the VNNI factor.
+    unaryInfo.ldo =
+        stridesOnOutput->front() /
+        *vnni::utils::getVnniBlockingFactor(expandShapeOp.getSrcType());
+    auto functionOp = transposeOp->getParentOfType<func::FuncOp>();
+    OpBuilder::InsertionGuard guard(rewriter);
```
rengolin, line 109: You're again setting the insertion point without reason and guards. Please do _not_ do that. This is very dangerous.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+                                 xsmm::UnaryKind::VNNI2, input, output,
+                                 transposeOp);
+  } else {
+    auto unaryInfo = xsmm::utils::getVectorUnaryInfo(
+        dyn_cast<MemRefType>(input->getOperand(0).getType()),
+        dyn_cast<MemRefType>(output->getOperand(1).getType()),
+        dyn_cast<VectorType>(input->getResult(0).getType()),
+        dyn_cast<VectorType>(output->getOperand(0).getType()),
+        xsmm::UnaryFlags::NONE);
+    auto temp = unaryInfo->n;
```
rengolin, line 124: `std::swap`?

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+    unaryInfo->n = unaryInfo->m;
+    unaryInfo->m = temp;
+
+    temp = unaryInfo->ldi;
+    unaryInfo->ldi = unaryInfo->ldo;
+    unaryInfo->ldo = temp;
+
+    return getTransposeOperation(rewriter, *unaryInfo, outputType,
```
rengolin, line 132: This is still duplicated code: `getTransposeOperation` is called with basically the same arguments on both sides of the branch.

As discussed before, use the if/else block to define the arguments then call the function only once.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+  LLVM_DEBUG(llvm::dbgs() << "BuildTransposeOp\n");
+  Value source = input->getOperand(0);
+  VectorType outType = cast<VectorType>(outputType);
+  if (vnni::utils::isInVnniLayout(vnni::utils::VnniOperandRank::TRANSPOSE,
+                                  outType)) {
+    memref::ExpandShapeOp expandShapeOp =
+        dyn_cast<memref::ExpandShapeOp>(source.getDefiningOp());
+    source = expandShapeOp.getSrc();
+    xsmm::UnaryInfo unaryInfo;
+    unaryInfo.m = expandShapeOp.getSrcType().getShape()[0];
```
rengolin, line 98: Why isn't this logic in `getVectorUnaryInfo`?

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,235 @@
...
+        xsmm::UnaryFlags::NONE);
+
+    if (failed(unaryInfo)) {
+      return failure();
+    }
+  }
+  return success();
+}
+
+void registerNativeRewrite(RewritePatternSet &patterns) {
```
rengolin, line 184:
```suggestion
static void registerNativeRewrite(RewritePatternSet &patterns) {
```

lib/TPP/Transforms/Utils/BuilderUtils.cpp
KavithaTipturMadhu: I didnt want to change the API.I can either add an argument, similarly as the createBody flag, or just set the function as private after creating the prototype. [Here's why](https://mlir.llvm.org/getting_started/Faq/#whats-with-op-symbol-declaration-cannot-have-public-visibility). What do you recommend?

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmmPDL.pdll
KavithaTipturMadhu: Will fix.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
KavithaTipturMadhu: I was trying to ensure that constants are created at the beginning of the function body. I didn't realize that the default insertion point would be the beginning of the function body. Should we not set the guard and insertion point explicitly to ensure the insertion point is set correctly or should we rely on default insertion points?

KavithaTipturMadhu: Will delete.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
KavithaTipturMadhu: I thought since the code was specific to vnni and wasnt gonna be used elsewhere, there was no need to move it to getVectorUnaryInfo. I'll move it there.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmmPDL.pdll
KavithaTipturMadhu: input0 will get dced.

tools/tpp-run/tpp-run.cpp
KavithaTipturMadhu: lhsTile and rhsTile are explicitly added as pipeline options and are not bundles, as you pointed out yourself. gpubackend is a bundle and is added this way, and the way to add bundles is this. The only other way to add options to DefaultPipeline is like lhsTile, which is with a different name, which does not match vector-to-xsmm.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
KavithaTipturMadhu: We'll then have to change the insertion point conditionally, as we're using buildXsmmCall for both invokes and dispatches, which I feel is confusing. What do you think?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: I'll add a flag for type, in order to distinguish between dispatch and invoke types.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Oh i thought you were referring to the line above it, you were referring to the twoDtransposeOp. I will update it to use ranks.

rengolin:
> I have mostly addressed all the review comments as per my understanding. I havent marked the conversations as closed yet since they may have opens in your opinion @rengolin Please review it wn you have the time. The CI is green.

Did you forget to push? I'm not seeing your changes.

KavithaTipturMadhu:
> > > > I have mostly addressed all the review comments as per my understanding. I havent marked the conversations as closed yet since they may have opens in your opinion @rengolin Please review it wn you have the time. The CI is green.
> > > 
> > > 
> > > Did you forget to push? I'm not seeing your changes.
> > 
> > 
> > This was my reply yesterday, I'm still working on a few more comments, gimme an hour or so, will push my changes.
> 
> I see. Then please avoid requesting reviews before pushing everything you have. I spent some time today working around Github thinking it was broken that it wasn't showing any changes. Try to time your reply _after_ you make the final push.

I commented this yesterday and thought you posted more comments afterwards, since your comments were on the latest changes. I was referring to the changes I was going to make based on those comments before I made the next push.

KavithaTipturMadhu:
@rengolin I think I have addressed all the comments barring the one which said I shouldn't add vector-to-xsmm flag to tpp-run.cpp. Please do take a look when you have the time. I apologize for the delay. Also, if I have missed something, my sincere apologies.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,220 @@
...
+    unaryInfo.ldi = unaryInfo.ldi / vnniFactor;
+    std::swap(unaryInfo.ldi, unaryInfo.ldo);
+  } else
+    std::swap(unaryInfo.m, unaryInfo.n);
+  return getTransposeOperation(rewriter, unaryInfo, outputType, opType, input,
+                               output, transposeOp);
+}
+
+static LogicalResult
+validateTransposeOpImpl(PatternRewriter &rewriter, Operation *transposeOp,
```
rengolin, line 119: Why `Impl` in the name? This is a local function that just validates something. `Impl` is a pattern in LLVM/MLIR (and general C++ projects) that imply it's the _hidden_ part of a public API that is not accessible to the users of that API.

Here, there's no API.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,220 @@
...
+      xsmm::UnaryFlags::NONE);
+  if (vnni::utils::isInVnniLayout(vnni::utils::VnniOperandRank::TRANSPOSE,
+                                  outType)) {
+    // Adjust ldi and ldo  based on the VNNI factor.
+    auto vnniFactor = *vnni::utils::getVnniBlockingFactor(outType);
+    unaryInfo.ldo = unaryInfo.ldo * vnniFactor;
+    unaryInfo.ldi = unaryInfo.ldi / vnniFactor;
+    std::swap(unaryInfo.ldi, unaryInfo.ldo);
+  } else
+    std::swap(unaryInfo.m, unaryInfo.n);
```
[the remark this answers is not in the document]
KavithaTipturMadhu: I thought the llvm coding standard said that we should not use a brace for single statement bodies, atleast thats what I remember Lorenzo telling me: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

lib/TPP/PassBundles/VectorToXSMM.cpp
```diff
@@ -6,10 +6,19 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
```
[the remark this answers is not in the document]
KavithaTipturMadhu: I have trimmed the list to a minimum.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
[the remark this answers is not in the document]
adam-smnk: That's true if all cases are simple one liners. When any requires braces (be it more code, comments etc.), all the other cases should have braces too to keep whole code block uniform.

include/TPP/PassBundles.td
```diff
@@ -10,6 +10,7 @@
 #define TPP_DIALECT_TPP_PASSBUNDLES
 
 include "mlir/Pass/PassBase.td"
+include "mlir/Dialect/PDL/IR/PDLDialect.td"
```
rolfmorel, line 13: Is this needed? I expect you don't need any symbols from this file in `PassBundles.td`.

KavithaTipturMadhu: You're right, removed.

KavithaTipturMadhu:
@rengolin I have addressed all comments, I haven't updated the integration tests the way we discussed but just kept the xsmm calls as check lines because we have a new flag along with linalg-to-vector flag which forces lowering to vector. I can add check lines to ensure that we lower to vector in the first place. What do you think?

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -90,8 +90,10 @@ struct DefaultTppPasses
     if (linalgToVector) {
       skipOperations.push_back("all");
     }
-    if (vectorToXSMM)
-      skipOperations.clear();
+    if (vectorToXSMM) {
```
rengolin, line 93: Careful. `linalg-to-vector` and `vector-to-xsmm` could both be set, and this would result in a vector `{ "all", "transpose", "vnni" }`.

lib/TPP/PassBundles/VectorToXSMM.cpp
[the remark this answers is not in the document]
rengolin: I'm still surprised with the number. @adam-smnk I just copy this file from the other examples. Now I noticed we include `"TPP/PassBundles.h.inc"`, which sounds wrong (why not `"TPP/PassBundles.cpp.inc"`?).

Not for this PR, but we need to have a harder look on these bundles again soon.

test/Integration/transpose-fp32.mlir
```diff
@@ -0,0 +1,21 @@
...
+
+module {
+  func.func @entry(%arg0: tensor<3x5xf32>, %arg1: tensor<5x3xf32>)->tensor<5x3xf32> {
+    %out = linalg.transpose ins(%arg0 : tensor<3x5xf32>) outs(%arg1 : tensor<5x3xf32>) permutation = [1, 0]
+    return %out: tensor<5x3xf32>
+  }
+}
+
+
+// XSMM: call @xsmm_unary_dispatch
```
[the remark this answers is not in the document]
rengolin: As discussed, let's leave this test as is, but create a new one that dumps the IR between passes and makes sure we get vector and xsmm in that order.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -8,18 +8,60 @@
 
 #include "TPP/Dialect/Xsmm/XsmmUtils.h"
 #include "TPP/Dialect/Xsmm/XsmmOps.h"
+#include "TPP/Transforms/Utils/VNNIUtils.h"
 #include "TPP/Transforms/Utils/ValueUtils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
```
rengolin, line 14: I'm surprised by the need of the LLVM dialect header here.

rengolin (commented):
Looking much better, thanks! Added a few more comments but nothing major. Just need to make sure we have the test that shows your pass is actually running and looks good.

lib/TPP/Conversion/ConvertVectorToXsmm/ConvertVectorToXsmm.cpp
```diff
@@ -0,0 +1,218 @@
...
+  }
+  return getTransposeXSMMCalls(rewriter, unaryInfo, outputType, opType, input,
+                               output, transposeOp);
+}
+
+static LogicalResult validateTransposeOp(PatternRewriter &rewriter,
+                                         Operation *transposeOp,
+                                         Operation *input, Operation *output,
+                                         Type outputType) {
+  LLVM_DEBUG(llvm::dbgs() << "validateTransposeOpImpl\n");
```
rengolin, line 119: Forgot one last `Impl` from the comment.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: It's for memref expansion into llvm ops.

lib/TPP/DefaultTppPasses.cpp
KavithaTipturMadhu: Fixed.

include/TPP/PassBundles.td
```diff
@@ -41,6 +41,9 @@ def DefaultTppPasses : Pass<"default-tpp-passes", "ModuleOp"> {
     Option<"vectorToXSMM", "vector-to-xsmm",
            "bool", /*default=*/"false",
            "Lower vector patterns to XSMM calls.">,
+    Option<"printVector", "print-vector-ir",
```
rengolin, line 44: This is not needed anymore

rengolin (commented):
Only one left-over and looks good, thanks!

rengolin (approved):
LGTM, thank you!

PR #983, opened 2024-11-20
Vector-to-{XSMM, Kernel} bundles

rengolin:
Adding two new bundles enabled by their own flags and turning on linalg vectorization on demand. Future PRs on each side should add their passes to the respective bundles and add benchmark updates to show their performance on CI.


rolfmorel (approved):
LGTM.

PR #984, opened 2024-11-21
[CMake] Add hack to find LLVM's OpenMP with GCC

### LINKED ISSUE #959: Build issues on Ubuntu 24.04 - incorrect openmp linkage in CMake

rengolin:
This should replace #960.

It worked on my Ubuntu for both Clang and GCC.

@dbabokin please check that this works for you, too.

Fixes #959.

### DIFF
--- a/cmake/modules/openmp.cmake
+++ b/cmake/modules/openmp.cmake
@@ -1,5 +1,19 @@
 option(USE_OpenMP "Use OpenMP" ON)
 
+# We don't want GOMP because its performance sinks for large core count, so we force libomp
+# This finds the library path from the system's clang for OpenMP
+#
+# On Fedora, it's at the same place as others, so we don't need to look elsewhere
+# On Ubuntu, it's in /usr/lib/llvm-${version}, so find_package finds GOMP for GCC instead.
+execute_process (
+    COMMAND bash -c "for lib in $(clang -lomp -### 2>&1); do echo $lib | grep -o \"\\/.*llvm.*\\w\"; done"
+    OUTPUT_VARIABLE LLVM_OMP_PATH
+)
+# Only if we found an "llvm" path that we need to add
+if (LLVM_OMP_PATH)
+  set(CMAKE_PREFIX_PATH ${LLVM_OMP_PATH})
+endif()
+
 if(USE_OpenMP)
   find_package(OpenMP)
   if(OPENMP_FOUND)


cmake/modules/openmp.cmake
```diff
@@ -1,5 +1,19 @@
 option(USE_OpenMP "Use OpenMP" ON)
 
+# We don't want GOMP because its performance sinks for large core count, so we force libomp
+# This finds the library path from the system's clang for OpenMP
+#
+# On Fedora, it's at the same place as others, so we don't need to look elsewhere
+# On Ubuntu, it's in /usr/lib/llvm-${version}, so find_package finds GOMP for GCC instead.
+execute_process (
+    COMMAND bash -c "for lib in $(clang -lomp -### 2>&1); do echo $lib | grep -o \"\\/.*llvm.*\\w\"; done"
```
dbabokin, line 9: It seem to not work as expected.

I tired running `for lib in $(clang -lomp -### 2>&1); do echo $lib | grep -o \"\\/.*llvm.*\\w\"; done` on both Ubuntu 22.04 and 24.04. I'm trying in docker: `docker run -it ubuntu:22.04 /bin/bash`. The only thing I do is `apt update; apt install clang libomp-dev`

On Ubuntu 22.04, I see that `clang -lomp -###` has `"-L/usr/lib/llvm-14/bin/../lib"` in it's output, but grep still doesn't catch it. On Ubuntu 24.04 there are no directories with `llvm` in it in clang's output.

rengolin: Thanks for checking!

My idea here is that, if `libomp` is in a "standard" directory (ex. `/lib64`), we don't need to add the extra library path. But if it's in an "LLVM specific" directory, we do. So, if `clang` adds any "LLVM specific" directory, we do, too.

So, even if the command line above doesn't find anything, it's still possible that we can continue to build with GCC as hoped.

rengolin: @dbabokin Just noticed, you quoted the grep when you ran (`\"` ), so your grep didn't _"catch"_ because it was broken. I just did this myself, too. :laughing: 

I'm running this on Ubuntu 22.04 and when I unquote the quote (as Bash does via CMake), it does catch the directory and the build works (where it previously didn't).

rengolin:
Ok, I'll merge this as is and let's see if you or more people have trouble with GCC + libomp.

PR #987, opened 2024-12-03
Vector-to-kernel pipeline bundle

arun-thmn:
The PR `Vector-to-kernel pipeline bundle` does the following:

1) Bundles the four passes: brgemm tiling (https://github.com/plaidml/tpp-mlir/pull/974), vectorization (https://github.com/plaidml/tpp-mlir/pull/958), hoisting vector transfers (https://github.com/plaidml/tpp-mlir/pull/977), and vector to fma (https://github.com/plaidml/tpp-mlir/pull/978) together for the `vector-to-kernel` pipeline.

2) New `json` files are added to benchmark folder for benchmarking the `vector-to-kernel` pipeline. 
            a) `/base/vector.json`: tests the base sequential version for mlp/gemm fp-32
            b) `/omp/mlp-fp32-vector.json`: tests the `omp` version for mlp/gemm-fp32
            c) `/omp/torch-dynamo-vector-to-kernel.json`: tests the `omp` vesion for torch mlp/gemm-fp32.
 
3) Updating the `brgemm` default tile size to `lhsTile:4,32` and `rhsTile:32,1`, as this size emits the better `gflops`.


benchmarks/config/base/vector-to-kernel.json
```diff
@@ -0,0 +1,53 @@
...
+    "mlp_fp32_mlir": {
+      "type": "IR-GEN",
+      "benchmark": [ "mlir-gen", "--kernel=const --bias --relu --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
+      "environment": {},
+      "flags": [ "-n", "100",  "-run-args='--def-parallel --vector-to-kernels --lhsTile=4,32 --rhsTile=32,1'" ],
+      "extensions": []
+    }
+  }},
+  {
+  "gemm_models_vector_kernel": {
```
rengolin, line 20: I'd just add these to `base.json`, so it's easy to compare by proximity. Call them `..._vector` (as opposed to `_mlir` or `_dnn_target`).

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -138,7 +140,6 @@ struct DefaultTppPasses
             BrgemmLinalgTilingOptions{lhsTile, rhsTile}));
         pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
         pm.addNestedPass<func::FuncOp>(createVectorizationPass());
-        pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
```
adam-smnk, line 141: Any reason to remove?

benchmarks/config/base/vector-to-kernel.json
```diff
@@ -0,0 +1,53 @@
...
+    "fp32_3x1024_const_mlir": {
+      "type": "IR-GEN",
+      "benchmark": [ "mlir-gen", "--kernel=const --float-type=f32 --batch=256 --layers=1024,1024,1024,1024" ],
+      "environment": {},
+      "flags": [ "-n", "100", "-run-args='--vector-to-kernels --lhsTile=4,32 --rhsTile=32,1'" ],
```
rengolin, line 25: The biggest problem with the register tile options is that, unlike the 2D parallelization ones that are related to the number of OMP threads, those are specific to the target extension availability.

They will be different to other architectures and we don't have a good way to select these options in the benchmark harness. We'd really benefit from target information here. @rolfmorel

adam-smnk (commented):
Overall looks good

lib/TPP/DefaultTppPasses.cpp
arun-thmn: We need `brgemm tiling -> licm -> vectorization -> hoisting -> canonicalizer -> vector to fma`  order for  the `vector-to-kernel` path.

And for Kavitha's `vector-to-xsmm` path the input IR is after `vectorization` pass. So, we moved the `hoisting -> canonicalizer -> vector to fma` into VectorToKernel.cpp file.

Please note, `canonicalizer` should be after `hoisting` pass because it fuses outer tiling loops and it results in `no pattern matching` for `hoisting` pass.

rengolin: I don't like that the three passes above are not a "bundle" per se, but the idea is to remove the `linalg-to-vector` flag and always vectorize whatever `linalg-to-xsmm` didn't catch. For a later PR, of course.

scripts/buildkite/tpp-mlir.yml
```diff
@@ -9,21 +9,21 @@ steps:
   - wait
 
   - label: "TPP-MLIR-gcc-rel"
-    command: "${SRUN} --partition=spr-all --time=0:30:00 -- \
+    command: "${SRUN} --partition=emr --time=0:30:00 -- \
```
rengolin, line 12: please rebase

arun-thmn: The main: https://github.com/plaidml/tpp-mlir/blob/main/scripts/buildkite/tpp-mlir.yml still points to `spr-all`, hence updated to `emr`.

rengolin: oh, we don't use buildkite any more, ignore those

arun-thmn: Rebase, remove this file.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -92,8 +92,10 @@ struct DefaultTppPasses
     }
     if (vectorToXSMM)
       skipOperations.clear();
-    if (vectorToKernel)
-      skipOperations.clear();
+    
+    if (vectorToKernel && !linalgToVector) {
+	 skipOperations.push_back("all");
```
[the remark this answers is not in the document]
arun-thmn: Okay. `vector-to-kernel` also needs `skipOperations.push_back("all");`.
So, for time being  added to the `linalg-to-vector` if-then.

lib/TPP/DefaultTppPasses.cpp
arun-thmn: @adam-smnk Sure. Added the note.

benchmarks/config/base/vector-to-kernel.json
arun-thmn: Okay. Moved mlp/gemm vector models to `base.json`

benchmarks/config/base/vector-to-kernel.json
arun-thmn: Agreed.

arun-thmn:
@rengolin Responded to your comments. Please have a look.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -141,8 +139,12 @@ struct DefaultTppPasses
             BrgemmLinalgTilingOptions{lhsTile, rhsTile}));
         pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
         pm.addNestedPass<func::FuncOp>(createVectorizationPass());
-        pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-        if (vectorToXSMM) {
+
+	//Please note, canonicalizer should be after hoisting pass because 
+	//it fuses outer tiling loops and it results in no pattern 
+	//matching for hoisting pass. Moved inside VectorToKernel Path.
```
rengolin, line 145: @KavithaTipturMadhu, does `vectorToXSMM` need the canonicalization pass? The behaviour here will be changed for that path. We may need to add it to your bundle.

benchmarks/config/base/base.json
```diff
@@ -133,5 +133,39 @@
       "flags": [ "-n", "100"],
       "extensions": [ "(avx2|asimd)" ]
     }
+  }},
+  {
+  "gemm_models_vector_kernel": {
```
rengolin, line 138: move these inside the block `gemm_models` above and call the runs:
* `fp32_3x1024_const_vector`
* `fp32_3x1024_args_vector`

Same below for MLP.

benchmarks/config/base/vector-to-kernel.json
```diff
@@ -0,0 +1,19 @@
+[
+  {
+  "prepacked_targets_vector_kernel": {
+    "gemm_fp32_mlir": {
```
rengolin, line 4: Move these into the block `prepacked_targets` in `base.json` and name them:
* `gemm_fp32_vector`
* `mlp_fp32_vector`

arun-thmn: Same as above. Moved into `prepacked_targets` and removed the `base\vector-to-kernel.json`

arun-thmn:
@rengolin Removed `base/vector-to-kernel.json` and moved  the examples into `base\base.json` as you mentioned.

lib/TPP/DefaultTppPasses.cpp
rengolin: Let's leave it at that, since the vector-to-xsmm work is not complete yet. Kavitha can fix this on her next PR, if needed.

PR #989, opened 2024-12-04
Segfault fix related to 'skipOperations' flag.

shahidact:
LinalgLowering pass now can accept 'skipOperations' which can be used to control Ops to be lowered to XSMM. Segfault were being observed when "linalg-to-vector" flag was used which tried to intialize 'skipOperations' ArrayRef object. This patch fixes this issue by changing this object to SmallVector.

Co-Authors: Kavitha, Arun, Shahid


rengolin (approved):
D'oh! My bad. Sorry.

PR #990, opened 2024-12-04
Common CLI flags registry

adam-smnk:
Moves CLI flags used to control pipelines behavior to a common registry.

This cleanup allows to remove the overhead of piping options through multiple layers of pass bundles. Now, the common CLI flags are globally accessible and can be pulled directly as a dependency by any pass bundle.

Note that individual passes should be self contained and define all necessary options.

### DIFF
--- a/include/TPP/Options/PipelineOptions.h
+++ b/include/TPP/Options/PipelineOptions.h
@@ -5,6 +5,10 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 //===----------------------------------------------------------------------===//
+//
+// Collection of common CLI flags controlling lowering pipeline.
+//
+//===----------------------------------------------------------------------===//
 
 #ifndef TPP_OPTIONS_PIPELINEOPTIONS_H
 #define TPP_OPTIONS_PIPELINEOPTIONS_H
--- a/include/TPP/Options/PipelineOptions.h
+++ b/include/TPP/Options/PipelineOptions.h
@@ -35,6 +35,10 @@ extern llvm::cl::list<unsigned> parallelTaskGrid;
 
 extern llvm::cl::opt<bool> linalgToVector;
 
+extern llvm::cl::opt<bool> vectorToXSMM;
+
+extern llvm::cl::opt<bool> vectorToKernel;
+
 extern llvm::cl::opt<bool> lowerPackUnpackWithoutTranspose;
 
 // Lhs tile sizes for linalg-to-vector.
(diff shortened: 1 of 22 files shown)


adam-smnk:
No benchmark regressions.

adam-smnk:
@KavithaTipturMadhu This should help with flags through pass bundles.
If still needed, we can also extend and add printing to different parts of the pipeline.

include/TPP/Options/PipelineOptions.h
```diff
@@ -0,0 +1,54 @@
...
+#include "llvm/Support/CommandLine.h"
+
+#include <string>
+
+namespace mlir {
+namespace tpp {
+namespace opt {
+
+// Print MLIR before lowering
+extern llvm::cl::opt<std::string> printMLIR;
```
rengolin, line 25: Wait, that's odd. `extern` is supposed to be used in a C file when another C file declares your variable, not in a header.

It's also bad form to initialize everything statically before `_main`. This is why we have things like pass options, etc.

I don't think this is a step forward.

adam-smnk: `extern` is normally rather redundant as that's header's job to provide declarations. However, explicit `extern` can be useful to make intention more obvious in certain cases. This seems to be a common pattern for CLI options across LLVM headers.

CLI flags are global object anyway, this doesn't change their initialization or lifetime - only helps with scope. The main idea is to have a central place that has all the flags gathered together and easily accessible by pass bundles. This eliminates the need to create multiple duplicate pass options only to pipe a flag to a single pass nested deep within 5 other bundles.

I agree it shouldn't be misused for controlling individual passes like convert X to Y.

adam-smnk: Perhaps to put this more simply, this header is a collection of available singletons.

rengolin: I'm worried we'll leak anyway because not there's no indication where this flag will have an effect.

adam-smnk: On the other hand, currently it takes jumping across multiple files to track down the origin of option's value.
Plus, CLI flags generally define desired values. Now if they have defaults, these values have to be propagated through all pass option defaults. Otherwise running `default-pipeline` vs other standalone pass can be much different.

I think between the two, this has potentially fewer points of failure and requires less maintenance but ultimately I'm fine with either.

adam-smnk: Well, it's just an experimental alternative - I'm not fully sure it's better.

If you don't feel convinced, then let's leave things as they are.

adam-smnk:
Not necessarily beneficial - closing.

adam-smnk:
Note for future reference: there is actually a nice utility `llvm::ManagedStatic` for managing globals.
The CLI flags could also be wrapped into structs to group them logically.

PR #994, opened 2024-12-05
VNNI gemm vectorization fix

KavithaTipturMadhu:
This PR fixes the VNNI contract lowering for gemms from linalg generics. There was a bug in the older code which did not consider the gemm iterators which would have size 4: 1 for vnni and 3 others corresponding to gemm. Also, the index setting of map was incorrect for VNNI, it was hardcoded to index 3, but should've been a function of the map's size, which I have changed now. I have left the strided VNNI test be, because it covers a positive VNNI example, although it would've worked with the older code as well. The test non_square_vnni_gemm covers both 4 iterator case, as well as setting the index of the map correctly.

### DIFF
--- a/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
+++ b/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
@@ -674,22 +674,13 @@ LogicalResult checkVNNIGemmStructure(PatternRewriter &rewriter,
   if (failed(contractionDims))
     return failure();
 
-  unsigned m = contractionDims->m.back();
-  unsigned n = contractionDims->n.back();
-
-  // m and n dimensions must be parallel dimensions
-  if (!linalg::isParallelIterator(iteratorTypes[m]) ||
-      !linalg::isParallelIterator(iteratorTypes[n])) {
-    return failure();
-  }
-
   // innermost dimension must be a reduction dimension for VNNI type operations
   if (!linalg::isReductionIterator(iteratorTypes[iteratorTypes.size() - 1])) {
     return failure();
   }
 
   // get the index of the iterator corresponding to the floordiv operation
-  auto k = contractionDims->k.size() > 0 ? contractionDims->k.back() : 0;
+  auto k = contractionDims->k.back();
   auto map1 = linalgOp.getIndexingMapsArray()[1];
   auto index = getAffineBinaryOpExprIndex(map1, k, linalgOp.getContext());
   if (!index)
@@ -698,12 +689,12 @@ LogicalResult checkVNNIGemmStructure(PatternRewriter &rewriter,
   // Ensure that the body of the generic operation is mul-add chain
   // clang-format off
   using namespace mlir::structured_match;
-  auto hasRightOpChain =
+  auto isMatmulChain =
     StructuredOpMatcher::make<linalg::GenericOp>()
       .region(MatchOne(0), WithOpChain<KindMul, KindAdd>(
                                      /*captures=*/nullptr));
   // clang-format on
-  if (!hasRightOpChain.match(linalgOp))
+  if (!isMatmulChain.match(linalgOp))
     return failure();
   return success();
 }
--- a/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
+++ b/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
@@ -677,21 +677,25 @@ LogicalResult checkVNNIGemmStructure(PatternRewriter &rewriter,
   unsigned m = contractionDims->m.back();
   unsigned n = contractionDims->n.back();
 
+  // m and n dimensions must be parallel dimensions
   if (!linalg::isParallelIterator(iteratorTypes[m]) ||
       !linalg::isParallelIterator(iteratorTypes[n])) {
     return failure();
   }
 
+  // innermost dimension must be a reduction dimension for VNNI type operations
   if (!linalg::isReductionIterator(iteratorTypes[iteratorTypes.size() - 1])) {
     return failure();
   }
 
+  // get the index of the iterator corresponding to the floordiv operation
   auto k = contractionDims->k.size() > 0 ? contractionDims->k.back() : 0;
   auto map1 = linalgOp.getIndexingMapsArray()[1];
   auto index = getAffineBinaryOpExprIndex(map1, k, linalgOp.getContext());
   if (!index)
     return failure();
 
+  // Ensure that the body of the generic operation is mul-add chain
   // clang-format off
   using namespace mlir::structured_match;
   auto hasRightOpChain =
--- a/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
+++ b/lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
@@ -8,20 +8,26 @@
 
 #include "TPP/Dialect/Xsmm/XsmmUtils.h"
 #include "TPP/Dialect/Xsmm/XsmmOps.h"
+#include "TPP/IR/StructuredOpMatcher.h"
+#include "TPP/Transforms/Utils/BuilderUtils.h"
 #include "TPP/Transforms/Utils/VNNIUtils.h"
 #include "TPP/Transforms/Utils/ValueUtils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/Linalg/Utils/Utils.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "mlir/IR/BuiltinTypeInterfaces.h"
 #include "mlir/IR/TypeUtilities.h"
-
-#include "TPP/Transforms/Utils/BuilderUtils.h"
 #include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SetOperations.h"
 #include "llvm/Support/Debug.h"
 #define DEBUG_TYPE "xsmm-utils"
+
+using namespace mlir;
+using namespace mlir::linalg;
 
 namespace mlir {
 namespace xsmm {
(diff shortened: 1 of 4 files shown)


lib/TPP/Transforms/Vectorization.cpp
```diff
@@ -42,7 +42,7 @@ struct LinalgGenericToVector : OpRewritePattern<linalg::GenericOp> {
     if (xsmm::utils::getDataType(rewriter, linalgOp.getOperand(0).getType()) ==
             xsmm::DataTypeAttr::get(rewriter.getContext(),
                                     xsmm::DataType::BF16) &&
-        linalgOp.getIteratorTypes().size() >= 5 &&
+        linalgOp.getIteratorTypes().size() >= 4 &&
```
adam-smnk, line 45: Which case this change covers? The test example has 5 iterators as I'd expect from VNNI BRGEMM

KavithaTipturMadhu: I'll update the test for a 4 iterators case.

KavithaTipturMadhu:
This PR fixes the VNNI contract lowering for gemms from linalg generics. There was a bug in the older code which did not consider the gemm iterators which would have size 4: 1 for vnni and 3 others corresponding to gemm. Also, the index setting of map was incorrect for VNNI, it was hardcoded to index 3, but should've been a function of the map's size, which I have changed now. I have left the strided VNNI test be, because it covers a positive VNNI example, although it would've worked with the older code as well. The test non_square_vnni_gemm covers both 4 iterator case, as well as setting the index of the map correctly.

lib/TPP/Transforms/Vectorization.cpp
rengolin: I don't think this is the right check. The VNNI shape is +1 on the non-VNNI shape and not just larger than some random constant.

I'm also not sure what the `numOperands` check is supposed to do, because all our operations have three operands.

KavithaTipturMadhu: Actually, it was wrong before, this is fixed now, because it is supposed to be 3 iterators for gemm: Parallel, Parallel, Reduction, and vnni introduces one extra reduction dimension which makes it 4. On the other hand, in case of fp32, you need at least 4 dimensions in order to introduce an expand shape. Therefore, this is not a random check.

adam-smnk: So this expands coverage to VNNI GEMM, alright. The logic is a bit iffy as we don't really validate that it is actually a VNNI format but I think in most mismatch cases the reassociations fails so nothing gets broken.

Overall, this wouldn't be needed as at all if we move away from this `floordiv 2` indexing map.

rengolin: I'm not arguing it doesn't work, I'm arguing it's not fixing the bug.

It's just nudging it to work on the extra case you have and will need more nudges for future cases, ad infinitum. We need to fix the actual bug by either adding a reasonable comparison (relative dimensions between inputs) or using a more appropriate affine map as Adam suggests.

> On the other hand, in case of fp32, you need at least 4 dimensions in order to introduce an expand shape.

Can you explain the reasoning here?

adam-smnk: Agreed - but a proper fix is much larger work. If a nudge is sufficient to unblock current vectorization PoC, I'd go with that.

KavithaTipturMadhu: > I'm not arguing it doesn't work, I'm arguing it's not fixing the bug.
> 
> It's just nudging it to work on the extra case you have and will need more nudges for future cases, ad infinitum. We need to fix the actual bug by either adding a reasonable comparison (relative dimensions between inputs) or using a more appropriate affine map as Adam suggests.
> 
> > On the other hand, in case of fp32, you need at least 4 dimensions in order to introduce an expand shape.
> 
> Can you explain the reasoning here?

>> Sorry I misspoke, there is no fp32 case, its only bf16 and it needs atleast 4 dimensions to be a gemm. However, as @rengolin pointed out, its not sufficient to check that the number of iteratortypes is 4, we need to check that the iterators match that of a gemm/brgemm and the operations correspond to a gemm/brgemm. I'll update the patch. Thanks!

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -564,6 +570,140 @@ FailureOr<SmallVector<Attribute>> getBrgemmFlags(PatternRewriter &rewriter,
...
+  if (failed(contractionDims))
+    return failure();
+
+  unsigned m = contractionDims->m.back();
+  unsigned n = contractionDims->n.back();
+
+  if (!linalg::isParallelIterator(iteratorTypes[m]) ||
+      !linalg::isParallelIterator(iteratorTypes[n])) {
+    return failure();
+  }
```
adam-smnk, line 683: nit: Isn't this guaranteed by inference logic? That is, I think M and N are always parallel

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -564,6 +570,140 @@ FailureOr<SmallVector<Attribute>> getBrgemmFlags(PatternRewriter &rewriter,
...
+
+  if (!linalg::isReductionIterator(iteratorTypes[iteratorTypes.size() - 1])) {
+    return failure();
+  }
+
+  auto k = contractionDims->k.size() > 0 ? contractionDims->k.back() : 0;
```
adam-smnk, line 689: Shouldn't it just fail on size 0?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -564,6 +570,140 @@ FailureOr<SmallVector<Attribute>> getBrgemmFlags(PatternRewriter &rewriter,
...
+    if (isa<AffineBinaryOpExpr>(result) &&
+        dyn_cast<AffineBinaryOpExpr>(result).getLHS() ==
+            getAffineDimExpr(index, context)) {
+      return i;
+    }
+  }
+  llvm_unreachable("invalid binary op index");
+}
+
+LogicalResult checkVNNIGemmStructure(PatternRewriter &rewriter,
```
adam-smnk, line 660: Should it also validate that the innermost dimension is a valid VNNI factor?

adam-smnk (commented):
Looks like a fine temporary workaround
Overall, we just need to fix our affine maps for VNNI layouts but I'd do this refactoring after current pdll PRs

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: On size 0, we still need to get vnni dimension, which we don't get currently as we don't have the reduction dimension present in both M and N maps. Since we know there's an inner reduction due to the previous check, we can safely say there's a 0th dimension corresponding to the k here.

adam-smnk: Oh right vnni dim won't show up at all.

adam-smnk: Wait I'm a bit lost now, wasn't this change suppose to catch the  floordiv dim?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -564,6 +570,144 @@ FailureOr<SmallVector<Attribute>> getBrgemmFlags(PatternRewriter &rewriter,
...
+  unsigned n = contractionDims->n.back();
+
+  // m and n dimensions must be parallel dimensions
+  if (!linalg::isParallelIterator(iteratorTypes[m]) ||
+      !linalg::isParallelIterator(iteratorTypes[n])) {
+    return failure();
+  }
+
+  // innermost dimension must be a reduction dimension for VNNI type operations
+  if (!linalg::isReductionIterator(iteratorTypes[iteratorTypes.size() - 1])) {
```
rengolin, line 687: VNNI must have two reductions, no? Why are you not checking in the same way as `m` and `n`?

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
rengolin: I don't think that's true. Just checking the last dim reduction isn't enough, as this could just be the `k` reduction and a non-VNNI bf16 matmul.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -564,6 +570,144 @@ FailureOr<SmallVector<Attribute>> getBrgemmFlags(PatternRewriter &rewriter,
...
+  auto k = contractionDims->k.size() > 0 ? contractionDims->k.back() : 0;
+  auto map1 = linalgOp.getIndexingMapsArray()[1];
+  auto index = getAffineBinaryOpExprIndex(map1, k, linalgOp.getContext());
+  if (!index)
+    return failure();
+
+  // Ensure that the body of the generic operation is mul-add chain
+  // clang-format off
+  using namespace mlir::structured_match;
+  auto hasRightOpChain =
```
rengolin, line 701: Nit: I'd call it `isMatmulChain`, because "right" is subjective and non descriptive.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: I am, the thing is contraction dims infer api does not return vnni dimension as one of the ks and thats why I need to check the innermost dimension as a reduction seperately, followed by the k as a floordiv, which is what I do here.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: Fair enough, will remove

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
KavithaTipturMadhu: > Wait I'm a bit lost now, wasn't this change suppose to catch the floordiv dim?

>>> Sorry I realize now the 0 assignment wasn't necessary. I have removed that.

lib/TPP/Dialect/Xsmm/XsmmUtils.cpp
```diff
@@ -564,6 +570,135 @@ FailureOr<SmallVector<Attribute>> getBrgemmFlags(PatternRewriter &rewriter,
...
+  }
+  return res;
+}
+namespace {
+auto par = mlir::utils::IteratorType::parallel;
+auto red = mlir::utils::IteratorType::reduction;
+} // namespace
+
+FailureOr<linalg::ContractionDimensions>
+inferContractionDims(linalg::GenericOp linalgOp) {
```
rengolin, line 604: Just realized, this is a copy of the upstream code. This really should not happen.

This fix need to be upstream. Let's stop work in this PR for now, please.

KavithaTipturMadhu: I had to copy the upstream code because the upstream code does not get the floordiv dimension if its an affinebinaryopdimexpr. I forgot to discuss this with you.

adam-smnk:
As discussed offline, we'll refactor our affine maps to avoid temporary fixes.

PR #995, opened 2024-12-11
Unary broadcast lowering

KavithaTipturMadhu:


include/TPP/Dialect/Xsmm/XsmmUtils.h
```diff
@@ -122,6 +122,7 @@ func::CallOp buildXsmmCall(RewriterBase &rewriter, XsmmCallType callType,
                            SmallVector<XsmmOperand> operands, TypeRange results,
                            FlatSymbolRefAttr fnName, Operation *parentOp,
                            Operation *insertBefore);
+
```
rengolin, line 125: avoid whitespace changes in files you're not touching otherwise.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -143,11 +144,11 @@ struct DefaultTppPasses
...
         pm.addNestedPass<func::FuncOp>(createVectorizationPass());
 
-	//Please note, canonicalizer should be after hoisting pass because 
-	//it fuses outer tiling loops and it results in no pattern 
-	//matching for hoisting pass. Moved inside VectorToKernel Path.
-        
-	if (vectorToXSMM) {
+        // Please note, canonicalizer should be after hoisting pass because
+        // it fuses outer tiling loops and it results in no pattern
+        // matching for hoisting pass. Moved inside VectorToKernel Path.
```
rengolin, line 149: You can remove this comment now. It was meant for you to make sure you didn't need it.

test/Integration/broadcast-2d.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: tpp-run --vector-to-XSMM %s -e columnBroadcast --entry-point-result=void -print --seed 123 2>&1 | FileCheck %s -check-prefix=COLUMNBROADCAST
+// RUN: tpp-run --linalg-to-loops %s -e columnBroadcast --entry-point-result=void -print -seed 123 2>&1 | FileCheck %s -check-prefix=COLUMNBROADCAST
+
+
+#map2 = affine_map<(d0, d1, d2) -> (d2)>
+#map3 = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
+
+func.func @columnBroadcast(%arg0: tensor<8xf32>, %arg1: tensor<2x4x8xf32>) -> tensor<2x4x8xf32> {
```
rengolin, line 8: Is this really a 2D column broadcast?

test/Integration/broadcast-2d.mlir
```diff
@@ -0,0 +1,27 @@
+// RUN: tpp-run --vector-to-XSMM %s -e columnBroadcast --entry-point-result=void -print --seed 123 2>&1 | FileCheck %s -check-prefix=COLUMNBROADCAST
+// RUN: tpp-run --linalg-to-loops %s -e columnBroadcast --entry-point-result=void -print -seed 123 2>&1 | FileCheck %s -check-prefix=COLUMNBROADCAST
```
rengolin, line 2: Same as transpose, you have no idea if your path is being selected for this pattern.

test/Integration/broadcast-row.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: tpp-run --vector-to-XSMM %s -e rowBroadcast --entry-point-result=void -print-mlir=mid 2>&1 | FileCheck %s -check-prefix=ROWBROADCAST
+// RUN: tpp-run %s -e rowBroadcast --entry-point-result=void -print-mlir=mid 2>&1 | FileCheck %s -check-prefix=ROWBROADCAST
+
+
+#map4 = affine_map<(d0, d1) -> (d1, 0)>
+#map5 = affine_map<(d0, d1) -> (d0, d1)>
+
+func.func @rowBroadcast(%arg0: tensor<128x1xf32>, %arg1: tensor<512x128xf32>) -> tensor<512x128xf32> {
```
rengolin, line 8: This looks odd. I was expecting `<128x1>` to broadcast to `<128x?>` and not transposed like that.

rengolin (commented):
You have huge gaps in your testing. You need to make sure the tests are correct (I'm not so sure about that), that they follow the path you expect (linalg-to-vector and vector-to-xsmm), and that the output is correct. 

You did some of those things for different integration tests. You need to do all of those things for all integration tests. You also need negative conversion tests (dims that don't match, reducing rank, etc), and a more comprehensive functionality testing (more patterns like 1D to 2D, 2D to 3D, 1D to 3D, etc).

It would also be interesting to have a more elaborate test, where you start adding all of the supported patterns. For now, one that has a transpose followed by a broadcast and another that is the other way round. You should match both patterns on both cases, and create two function calls on both.

test/Integration/broadcast-2d.mlir
KavithaTipturMadhu: I do actually, this was not supported earlier, it is now in the vector-to-xsmm lowering path.

test/Integration/broadcast-row.mlir
KavithaTipturMadhu: https://github.com/plaidml/tpp-mlir/blob/d7fcc28647aaf3ef853349296139414e856aac61/test/Conversion/LinalgToXsmm/linalg-to-unary.mlir#L280
This is the same behavior as earlier.

test/Integration/broadcast-2d.mlir
KavithaTipturMadhu: It is a column broadcast, which is what the function is called, the file should probably be renamed broadcast with 2D destination or something, can you please suggest an appropriate name for the test file?

test/Integration/broadcast-2d.mlir
KavithaTipturMadhu: In order to ensure correctness, I'll add tpp-opt -default-tpp-pipeline="--vector-to-xsmm" run line to print vector output and verify that vector output is indeed getting generated, does that sound reasonable? I could try printing from tpp-run -print-after but I don't know how to disable multithreading from commandline.

adam-smnk:
+1 on more test coverage or rather more specific matchers
Our tests right now surely don't cover all `vector.broadcast` semantics.

As an example:
```
func.func @entry(%arg0: memref<4x1xf32>, %arg1: memref<4x2xf32>) -> memref<4x2xf32> {
  %0 = vector.transfer_read %arg0[%c0, %c0], %cst {in_bounds = [true, true]} : memref<4x1xf32>, vector<4x1xf32>
  %1 = vector.broadcast %0 : vector<4x1xf32> to vector<4x2xf32>
  vector.transfer_write %1, %arg1[%c0, %c0] {in_bounds = [true, true]}
    : vector<4x2xf32>, memref<4x2xf32>
  return %arg1 : memref<4x2xf32>
}
```
This broadcast fails at runtime xsmm JIT step. I'm sure we don't need this variant currently but it'd be best not to open ourselves to new pitfalls.

adam-smnk:
As a side note, it could be that our current lowering from linalg is also incomplete. I don't want to feature/scope creep this PR.
Let's just agree on some sufficient state and possibly note down future necessary improvements.

test/Conversion/VectorToXsmm/vector-to-identity.mlir
```diff
@@ -0,0 +1,166 @@
...
+  %0 = vector.transfer_read %arg0[%c0], %cst {in_bounds = [true]} : memref<256xf32>, vector<256xf32>
+  %1 = vector.broadcast %0 : vector<256xf32> to vector<128x512xf32>
+  vector.transfer_write %1, %arg1[%c0, %c0] {in_bounds = [true, true]} : vector<128x512xf32>, memref<128x512xf32>
+  return
+}
+
+// ----
+
+// Vector dialect's Verifiers throw an error, can't expect-error this or handle this in our validation
+//XFAIL:*
```
rengolin, line 128: Move XFAIL tests over another file, or this test will always pass even if the rest of the tests get broken

rengolin (commented):
Small comment but looks good now. Just need to split the XFAIL tests out and we're good to merge, thanks!

test/Conversion/VectorToXsmm/vector-to-identity.mlir
```diff
@@ -0,0 +1,166 @@
...
+  %cst = arith.constant 0.000000e+00 : f32
+  %0 = vector.transfer_read %arg0[%c0], %cst {in_bounds = [true]} : memref<256xf32>, vector<256xf32>
+  %1 = vector.broadcast %0 : vector<256xf32> to vector<128x512xf32>
+  vector.transfer_write %1, %arg1[%c0, %c0] {in_bounds = [true, true]} : vector<128x512xf32>, memref<128x512xf32>
+  return
+}
+
+// ----
+
+// Vector dialect's Verifiers throw an error, can't expect-error this or handle this in our validation
```
adam-smnk, line 127: Why have a test for invalid IR at all?

KavithaTipturMadhu: I guess it makes sense to validate that the upstream vector verifiers behave consistently? I have moved the tests to a different file for now, we can get rid of them if we think they're redundant.

KavithaTipturMadhu:
@rengolin @adam-smnk I've updated the tests, please take a look.

test/Conversion/VectorToXsmm/vector-to-identity.mlir
rengolin: Invalid syntax should be a part of upstream validation and not downstream. If the compiler is already emitting an error, there's not much we can do here and don't need those tests.

But we still need negative tests, with valid syntax, that doesn't convert because of some other issue (wrong shape, extra ops in the middle, etc).

rengolin: I see that we already have the failed tests in the original test (`CHECK-NOT: xsmm`), so we don't need the syntax test file at all. Remove the test file and it's good to merge, thanks!

rengolin (approved):

PR #997, opened 2024-12-24
Python-generated schedules (just mirroring pipeline bundles)

rolfmorel:
Invoke either as `build/bin/tpp-sched` or `python -m mlir.tpp.sched` (which requires `PYTHONPATH` to point to `build/python_packages` - the `tools/tpp-sched/tpp-sched.py` script (installed as `tpp-sched`) takes care of that).

When `tpp-sched` is invoked it just dumps the default pipeline schedule to stdout. When a path is provided as a positional argument, i.e. `tpp-sched.py PAYLOAD_FILE`, the schedule gets generated and immediately applied to the payload and only the transformed payload gets dumped to stdout. Use `--bundles=BUNDLE1,...,BUNDLEN` to control which bundles make up the schedule. All the bundles' arguments available on `tpp-opt` are available on `tpp-sched`.

When invoked as `tpp-sched --payload PAYLOAD_FILE`, it will just print the payload and schedule together as one IR dump. This is meant to facilitate working directly with the IR before passing it on to, e.g., `tpp-opt` (to apply the schedule to the payload use `build/bin/tpp-sched --payload PAYLOAD_FILE | build/bin/tpp-opt --load-tpp-dialects --transform-interpreter` -- as why we need `--load-tpp-dialects`, see [here](https://discourse.llvm.org/t/dynamic-pass-scheduling-and-dialect-loading/85166/2)).

Introduces Python-bindings for our own dialects, for now mostly to be able to register them from Python before running passes that dependent on them.

Requires MLIR to be build with `-DMLIR_ENABLE_BINDINGS_PYTHON=1`.

Adds RUN lines which use `tpp-sched` to each test which used `tpp-opt` to apply a pipeline bundle.

For now we only have partial support for the GPU pipeline bundles.


tools/tpp-opt/tpp-opt.py
```diff
@@ -0,0 +1,408 @@
...
+    with InsertionPoint(named_sequence.body):
+      func = Match(named_sequence.bodyTarget, ops={"func.func"})
+      mod = transform.GetParentOp(
+        transform.AnyOpType.get(),
+        func,
+        op_name="builtin.module",
+        deduplicate=True,
+      )
+
+      try:
```
rolfmorel, line 356: @thiagotei, @fschlimb this (utterly standard) `try ... except ...` allows me to place a `raise PipelineInterrupt` line between any two transform ops above and it appropriately cuts off schedule generation right there. As we caught the exception, the `print(module)` line at the end of `main` still gets executed.

(In more complex setups, e.g. when generating named sequences this is unlikely to be as simple.)

python/mlir/tpp/sched/bundles.py
```diff
@@ -0,0 +1,335 @@
...
+        if linalg_to_vector or force_linalg_to_vector:
+            func = match(mod, ops={"func.func"})
+            options = "registerTileShape=" + ",".join(map(str, register_blocking))
+            func = apply_registered_pass(func, "brgemm-linalg-tiling", options=options)
+            func = apply_registered_pass(func, "loop-invariant-code-motion")
+            apply_registered_pass(func, "vectorization-pass")
+            # NB: canonicalizer should be after hoisting pass because
+            # it fuses outer tiling loops and it results in no pattern
+            # matching for hoisting pass. Moved inside VectorToKernel Path.
+            if vector_to_xsmm:
```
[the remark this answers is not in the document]
rolfmorel: Addressed.

python/mlir/tpp/sched/bundles.py
```diff
@@ -0,0 +1,335 @@
...
+        mod = cleanup(mod)
+        func = match(mod, ops={"func.func"})
+        # Decompose Aggregated operations. These ops currently do not bufferize.
+        # Once this is possible we can move this pass after bufferization.
+        apply_registered_pass(func, "decompose-aggregated-ops")
+        mod = apply_registered_pass(mod, "bufferize")
+        mod = linalg_lowering(mod, skip_operations=skip_ops, **config)
+        if linalg_to_vector or force_linalg_to_vector:
+            func = match(mod, ops={"func.func"})
+            options = "registerTileShape=" + ",".join(map(str, register_blocking))
```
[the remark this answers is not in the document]
rolfmorel: Addressed.

rolfmorel:
@rengolin, @adam-smnk, I would suggest starting with looking over `python/mlir/tpp/sched/main.py` and `python/mlir/tpp/sched/bundles.py`. And then playing a bit with `build/bin/tpp-sched.py`.

rengolin (approved):

rolfmorel:
Have now added RUN lines which use `tpp-sched` to each test which used `tpp-opt` to apply a pipeline bundle. And dealt with the final idiosyncrasies.

test/Integration/matmul_48x64x96.mlir
```diff
@@ -3,6 +3,7 @@
 // RUN: FileCheck %s
 
 // RUN: tpp-opt -default-tpp-passes %s | FileCheck %s -check-prefix=IR
+// RUN: tpp-sched --bundle=default-tpp-passes %s | tpp-opt -canonicalize | FileCheck %s -check-prefix=IR
```
adam-smnk, line 6: Why the extra round through tpp-opt with canonicalization is needed?

adam-smnk:
With `PYTHONPATH` set to `TPP/build/python_packages`, I still get error:
`ImportError: cannot import name 'ir' from 'mlir._mlir_libs._mlir'`

test/Integration/matmul_48x64x96.mlir
rolfmorel: The transform interpreter requires the transform IR module and the payload IR module to be in another module together. When we run canonicalize only on the payload module, it appears the canonicalizer recognizes there's another outer scope and hence will not move the affine_maps at all (so not to the payload module and also not to the "container" module containing it and the schedule module). As the test cases expect (i.e. CHECK) the affine_maps to be hoisted, without doing something more we get test failures.

So, upshot: canonicalize needs to run outside of the transform interpreter. Either we hack it into the Python script (e.g. add a `--canonicalize` flag that does a final invocation of the pass through the pass manager) or we run this "canonicalization that is only for tests" right before CHECK-ing. Pick your poison.

rolfmorel: Example of a file that requires the additional canonicalize: test/BF16/matmul-vnni.mlir

I am now removing the extraneous cases of the additional canonicalize, like the particular test this comment is on.

adam-smnk: I’d add it to the script directly to have the two equivalent by default.
Then there’s no need to run extra tpp-opt at all, right?

rolfmorel: Fixed now. Only `test/Conversion/VectorToXsmm/vector-to-transpose.mlir` and `test/BF16/matmul-vnni.mlir` require the affine_map canonicalization

rolfmorel: Turns out it is just a printing issue. That is, just running the IR through `tpp-opt` already pulls up the affine_maps. Latest commit makes it so that these two test cases just run the IR through tpp-opt.

At the moment I am not inclined to spend more time on these kinds of issues (as I know there are often subtleties that end up taking up disproportionate time).

adam-smnk (approved):
I still need to fix my python setup but logic looks fine, whole pipeline is accurate.
+1 to the concept

rolfmorel:
> With `PYTHONPATH` set to `TPP/build/python_packages`, I still get error: `ImportError: cannot import name 'ir' from 'mlir._mlir_libs._mlir'`

Is TPP the checkout of the repo? and `TPP/build` the build dir?

In general:
1. The llvm-project build needs to have the Python bindings enabled.
2. start from a clean build (`rm -rf *` in build dir) as changes to the Python stuff can be finicky.
3. `build/bin/tpp-sched` automagically finds the python modules. Maybe check if that works? (Also, the CI is happy - e.g. when running `tpp-check`.) 

Also, if there's another mlir python install present I am (attempting to) shadow(ing) that install. Basically by our python_packages dir being at the front of PYTHON_PATH. If this is not working as expected, I would like to know.

adam-smnk:
Problem solved - in my case it was a mix-up of python versions. MLIR bindings were built with python3.9 while I run 3.8 by default.  Because of that, my env couldn't pick up the modules.
After correcting version issues, all the modules get picked up automatically for me. No need to even setup `PYTHONPATH` env var (although that wouldn't hurt).

For future reference: check if `.cpython-XX` version under `~/LLVM/build/tools/mlir/python_packages/mlir_core/mlir/_mlir_libs` matches the used python's version.

PR #998, opened 2025-01-06
Fix VNNI affine maps

adam-smnk:
Retires incorrect VNNI affine map representation using `floordiv 2` and replaces it with an extra separate VNNI dimension in matrix operand A: `expand <M x K> into <M x K/VNNI x VNNI>`.

The updated VNNI representation is propagated through matchers, lowerings, and tests.
A new test is added to ensure correct results between Linalg to SCF loops and XSMM lowering.

### DIFF
--- a/lib/TPP/IR/MatcherUtils.cpp
+++ b/lib/TPP/IR/MatcherUtils.cpp
@@ -56,8 +56,8 @@ std::pair<bool, bool> isMatmulVnniOp(linalg::GenericOp linalgOp,
           .operation(NumOfLoops(_OR(EqualsTo(5), EqualsTo(4))))
           .input(MatchAll(), HasStaticShape())
           .output(MatchAll(), HasStaticShape())
-          .input(MatchOne(0), HasMap(Any(), &mapOperandA))
-          .input(MatchOne(1), HasMap(Any(), &mapOperandB))
+          .input(MatchOne(0), HasMap(ProjectedPermutation(), &mapOperandA))
+          .input(MatchOne(1), HasMap(ProjectedPermutation(), &mapOperandB))
           .output(MatchOne(0), HasMap(BroadcastableProjectedPermutation(), &mapOperandC))
           .region(MatchOne(0),
                   WithOpChain<arith::MulFOp, arith::AddFOp>(operands));


adam-smnk:
No benchmark regressions.

test/Conversion/LinalgToXsmm/linalg-to-brgemm.mlir
```diff
@@ -393,15 +401,17 @@ func.func @vnni_brgemm_require_transpose_on_C(%arg0: memref<16x32x32xbf16>, %arg
 
 // -----
 
-#map = affine_map<(d0, d1, d2, d3, d4) -> (d0, d2, d4)>
-#map1 = affine_map<(d0, d1, d2, d3, d4) -> (d0, d4 floordiv 5, d3, d1)>
+#map = affine_map<(d0, d1, d2, d3, d4) -> (d0, d2, d4, d1)>
+#map1 = affine_map<(d0, d1, d2, d3, d4) -> (d0, d4, d3, d1)>
 #map2 = affine_map<(d0, d1, d2, d3, d4) -> (d3, d2)>
```
adam-smnk, line 406: TODO: fix `map2` to not have transpose, change example VNNI factor to 5

adam-smnk: Fixed this and the equivalent test case in `linalg-to-gemm.mlir`.

After changes, this example would trigger an assertion in pass matcher. Thus, tweaked the pass logic to fail gracefully and simply leave operation unchanged.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -453,6 +453,20 @@ static void replaceOpWithGemmLikeOp(RewriterBase &rewriter,
   }
   auto flags = rewriter.getArrayAttr(gemmFlags);
   SmallVector<Value> invokeOperands;
+  SmallVector<Value> inputs = {linalgOp->getOperands()};
+
+  // Collapse VNNI factor dimension for matrix A.
```
rengolin, line 458: nit: example in the comment? Like `32x64x2` -> `64x64` or something.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -601,7 +615,8 @@ static FailureOr<BrgemmInfo> isMappableToBrgemm(linalg::LinalgOp linalgOp) {
   unsigned n = contractionDims->n[0];
   unsigned k = contractionDims->k.back();
   std::optional<unsigned> batch;
-  if (contractionDims->k.size() == 2)
+  unsigned extraVnniDim = vnni::utils::isInVnniLayout(linalgOp);
```
rengolin, line 618: This works for bf16 VNNI but will not work for bf8. Perhaps use `getVNNIBlockingFactor`?

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -1085,7 +1102,12 @@ struct ConvertGenericToVnniMatmulLikeOp
     int64_t kPos = 1;
     if (hasBatch)
       kPos++;
-    int64_t k = cast<ShapedType>(bufferA.getType()).getShape()[kPos];
+    // Take the whole reduction dim size. Account for the VNNI factor that
+    // splits the K dim in the shape.
+    std::optional<int64_t> vnniFactor =
+        vnni::utils::getVnniBlockingFactor(bufferB.getType());
+    int64_t k =
+        cast<ShapedType>(bufferA.getType()).getShape()[kPos] * *vnniFactor;
```
rengolin, line 1110: You haven't checked that the `optional` has a value. If it does not, this will crash.

lib/TPP/IR/MatcherUtils.cpp
```diff
@@ -82,17 +82,13 @@ std::pair<bool, bool> isBrgemmVnniOp(linalg::GenericOp linalgOp,
 
   llvm::SmallVector<int64_t> operandAPosIterRed = getIteratorPos(
       linalgOp, mapOperandA, mlir::utils::IteratorType::reduction);
-  if (operandAPosIterRed.size() != 2 && operandAPosIterRed.size() != 1)
+  if (operandAPosIterRed.size() != 3 && operandAPosIterRed.size() != 2)
```
rengolin, line 85: readability nit: maybe cache `size`?

lib/TPP/IR/MatcherUtils.cpp
```diff
@@ -112,10 +108,8 @@ std::pair<bool, bool> isBrgemmVnniOp(linalg::GenericOp linalgOp,
     return std::make_pair(false, hasBatch);
   }
 
-  auto vnniDim =
-      vnni::utils::isInVnniLayout(linalgOp, mapOperandB, *blockingFactor);
-  bool isBrgemmOp = succeeded(vnniDim) && vnniDim->getPosition() == kRedIter;
-  return std::make_pair(isBrgemmOp, hasBatch);
+  bool isBrgemmVnni = vnni::utils::isInVnniLayout(linalgOp, *blockingFactor);
```
rengolin, line 111: If this is guaranteed to be BRGEMM (previous checks), then at least a comment saying so would help

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -37,37 +38,73 @@ bool isInVnniLayout(VnniOperandRank expectedRank, MemRefType memref) {
...
-    return failure();
-  }
+  auto vnniDimA = dyn_cast<AffineDimExpr>(mapA.getResult(rankA - 1));
+  auto vnniDimB = dyn_cast<AffineDimExpr>(mapB.getResult(rankB - 1));
+  if (!vnniDimA || !vnniDimB || vnniDimA != vnniDimB ||
+      iteratorTypes[vnniDimA.getPosition()] !=
+          mlir::utils::IteratorType::reduction)
+    return false;
+  auto redDimA = dyn_cast<AffineDimExpr>(mapA.getResult(rankA - 2));
+  auto redDimB = dyn_cast<AffineDimExpr>(mapB.getResult(rankB - 3));
```
rengolin, line 82: need to make sure `rankA` and `rankB` are large enough to subtract `2` and `3`.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
adam-smnk: The VNNI layout is ensured by the earlier checks. Added comment plus assert just in case.

lib/TPP/IR/MatcherUtils.cpp
adam-smnk: This function got confusing as it also checks for GEMM. Refactored a bit to improve readability.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
adam-smnk: It doesn't check the dimension size but the number of reduction dimensions which is type independent.
Refactored to be more readable.

lib/TPP/Conversion/ConvertLinalgToXsmm/ConvertLinalgToXsmm.cpp
```diff
@@ -1085,7 +1109,13 @@ struct ConvertGenericToVnniMatmulLikeOp
     int64_t kPos = 1;
     if (hasBatch)
       kPos++;
-    int64_t k = cast<ShapedType>(bufferA.getType()).getShape()[kPos];
+    // Take the whole reduction dim size. Account for the VNNI factor (ensured
+    // by the earlier check) that splits the K dim in the shape.
+    std::optional<int64_t> vnniFactor =
+        vnni::utils::getVnniBlockingFactor(bufferB.getType());
+    assert(vnniFactor && "Must be in VNNI format");
```
rengolin, line 1116: assert only works in debug mode, this can still crash in release mode

adam-smnk: Fair, added proper check.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -375,16 +378,34 @@ Value MLIRGenerator::lowerNamedMatmul(Value input, Value weight, Value output) {
...
-  auto inputShape = cast<ShapedType>(input.getType());
-  auto outputShape = cast<ShapedType>(output.getType());
+  auto inputType = cast<ShapedType>(input.getType());
+  auto outputType = cast<ShapedType>(output.getType());
+
+  if (vnniPacked) {
+    SmallVector<int64_t> vnniShape{inputType.getShape()};
+    vnniShape.back() = vnniShape.back() / vnniFactor;
+    vnniShape.push_back(vnniFactor);
+    auto vnniType =
```
rengolin, line 388: shouldn't we also make sure that this is the same as the `weight` shape?

adam-smnk: Assert wouldn't hurt

rengolin (approved):

PR #1001, opened 2025-01-16
Make VNNI more robust

adam-smnk:
Improves VNNI validation and lit tests to be target independent w.r.t. VNNI blocking factor.

VNNI infrastructure is extended with DLTI support that allows to override target's default VNNI factor.
This allows to fix VNNI shape for lit testing to ensure that general pass logic works fine. This is primarily applied to tests that use prepacked VNNI shapes.

When possible, tests' checks are generalized to account for varying packing factor.
Test shapes are adjusted to prevent failure due to incompatible VNNI packing sizes compared to dimension size used for verification.

BF16 integration tests are now disabled on non-x86 systems to avoid runtime mismatch between prepacked shapes and microkernel requirements.


adam-smnk:
No performance regression.
Manually verified `check-tpp` passing on graviton.

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
```diff
@@ -455,10 +455,13 @@ LogicalResult GemmOp::verify() {
     auto memref = dyn_cast<MemRefType>(memrefOperands[idx].getType());
     assert(memref && (memref.getRank() == 2 || memref.getRank() == 3));
 
-    if (memref.getRank() == 3 &&
-        !vnni::utils::isInVnniLayout(vnni::utils::VnniOperandRank::GEMM,
-                                     memref)) {
-      return emitOpError() << "expect VNNI layout for operand: " << actualIdx;
+    if (memref.getRank() == 3) {
+      auto vnniFactor = vnni::utils::getVnniBlockingFactor(memref);
```
rengolin, line 459: here you can pass `self` as the operation pointer for DLTI, no?

include/TPP/Transforms/Utils/VNNIUtils.h
```diff
@@ -35,16 +36,25 @@ enum class VnniOperandRank {
   BRGEMM_OUTS = 3
 };
 
-// Return the VNNI blocking factor: 2 for BF16 and 4 for BF8.
-std::optional<int64_t> getVnniBlockingFactor(Type type);
+// Return the VNNI blocking factor.
+// Optionally, operation can be provided to give access to DLTI.
```
rengolin, line 40: @rolfmorel is this the expected API usage for DLTI?

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -20,22 +21,54 @@ namespace mlir {
 namespace vnni {
 namespace utils {
 
-std::optional<int64_t> getVnniBlockingFactor(Type type) {
+std::optional<int64_t> getVnniBlockingFactor(Type type, Operation *op) {
   auto elementType = getElementTypeOrSelf(type);
-  if (elementType.isBF16())
+  if (elementType.isBF16()) {
+    // Check if a VNNI factor hint is associated to the IR via DLTI.
+    auto deriveVnniFromDLTI = [&]() -> std::optional<int64_t> {
```
rengolin, line 28: I would extract this in to a proper method, for some future DLTI helper.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -20,22 +21,54 @@ namespace mlir {
...
+  }
   return std::nullopt;
 }
 
 // Until we have a better way to express the VNNI layout (see: #563), it is up
 // to the callee to specify the expected rank in the VNNI layout as the rank
 // depends on the operations we are dealing with.
-bool isInVnniLayout(VnniOperandRank expectedRank, MemRefType memref) {
+bool isInVnniLayout(VnniOperandRank expectedRank, MemRefType memref,
+                    std::optional<int64_t> blockingFactor) {
```
rengolin, line 63: `blockingFactor` can be `std::nullopt` meaning "there isn't one" instead of "I don't know", and here you will treat both equally.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -20,22 +21,54 @@ namespace mlir {
...
+      if (auto intAttr = llvm::dyn_cast<IntegerAttr>(value))
+        return intAttr.getInt();
+      return std::nullopt;
+    };
+    if (auto vnniFactor = deriveVnniFromDLTI())
+      return *vnniFactor;
+
     return libxsmm_cpuid_dot_pack_factor(LIBXSMM_DATATYPE_BF16);
+  }
   return std::nullopt;
```
rengolin, line 56: Shouldn't we return `1`  when there is no blocking factor? This would fix the issue below when we treat "no blocking" and "don't know" as the same answer.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -20,22 +21,54 @@ namespace mlir {
...
+      if (!moduleOp)
+        return std::nullopt;
+      TargetSystemSpecInterface sysSpec = moduleOp.getTargetSystemSpec();
+      if (!sysSpec)
+        return std::nullopt;
+      auto deviceId = StringAttr::get(moduleOp->getContext(), "CPU");
+      auto deviceSpec = sysSpec.getDeviceSpecForDeviceID(deviceId);
+      if (!deviceSpec)
+        return std::nullopt;
+      auto vnniId = StringAttr::get(moduleOp->getContext(), "vnni");
```
rengolin, line 41: "VNNI" is an Intel name. The Arm name is "BFDOT" or "BFMLA". This method is making it easier to use the Arm type packing, so we can't expect Arm will use "vnni" in its nomenclature.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -109,15 +142,21 @@ bool isInVnniLayout(linalg::LinalgOp linalgOp,
...
-  return isInVnniLayout(static_cast<int64_t>(expectedRank), vector);
+bool isInVnniLayout(VnniOperandRank expectedRank, VectorType vector,
+                    std::optional<int64_t> blockingFactor) {
+  return isInVnniLayout(static_cast<int64_t>(expectedRank), vector,
+                        blockingFactor);
 }
 
-bool isInVnniLayout(int64_t expectedRank, VectorType vector) {
-  if (vector.getRank() != expectedRank || !vector.getElementType().isBF16()) {
+bool isInVnniLayout(int64_t expectedRank, VectorType vector,
```
rengolin, line 151: This is near identical to the memref one. Can we not use `ShapedType`? Or a templated version?

test/BF16/Integration/lit.local.cfg
```diff
@@ -19,6 +19,29 @@ def has_support(feature):
...
+
+
 # AVX512 and SVE should have BF16 support
 if not has_support('avx512') and not has_support('avx2') and not has_support('sve'):
     config.unsupported = True
+
+# Enable only on x86
+# Other targets may use different VNNI blocking scheme that is not compatible with
+# prepacked shapes in some of the tests
+if not is_arch('x86'):
```
rengolin, line 46: Do we have a plan to introduce Arm tests in a follow up PR? Should be just a matter of fixing the blocking factors, no?

test/BF16/Integration/tpp-run-splat-shape.mlir
```diff
@@ -41,7 +41,7 @@ func.func @entry(%arg0: tensor<4x8x8x8xbf16>, %output: tensor<4x8x8x8xbf16>) ->
 // due to compile time packing.
 // CHECK-NOT: memref.global "private" constant @__constant_{{.*}}: memref<8x8xbf16>
 // CHECK-DAG: memref.global "private" constant @__constant_{{.*}}: memref<4x8x8x8xbf16>
-// CHECK-DAG: memref.global "private" constant @__constant_{{.*}}: memref<8x8x4x8x2xbf16>
+// CHECK-DAG: memref.global "private" constant @__constant_{{.*}}: memref<8x8x{{[4|2]}}x8x{{2|4}}xbf16>
```
rengolin, line 44: Lit won't run these tests on Arm, why do we need `4` here?

lib/TPP/Transforms/Utils/VNNIUtils.cpp
```diff
@@ -20,22 +21,54 @@ namespace mlir {
 namespace vnni {
 namespace utils {
 
-std::optional<int64_t> getVnniBlockingFactor(Type type) {
+std::optional<int64_t> getVnniBlockingFactor(Type type, Operation *op) {
```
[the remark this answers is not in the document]
rolfmorel: Reminder to self: introduce templated query method - that is: allow for `dlti::query<IntegerAttr>(op, keys)`. Would swallow up the common case of the dyn_cast directly after the query.

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
rolfmorel: Hmm - at the moment - without passing an Operation pointer - this does not compile, right?

include/TPP/Transforms/Utils/VNNIUtils.h
rolfmorel: Yes, makes sense to me. These kinds of wrappers give you a sensible default if the DLTI query they perform fails. These wrappers should take an `Operation*` to pass to `dlti::query` - if null, the query will fail/crash, so the wrapper should fallback on its default in this case as well.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: This optional argument means I want to check this is VNNI and with this specific factor.
Only reason to be this specific is to verify calls with kernel requirements like in `-verify-xsmm`.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: Factor of `1` would imply you still want 3D shape like `<8x16x1>`.
We could of course choose to do nothing if it's `1` but I see no reason to refactor everything right now.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: It is brute forcing our tests to behave predictably.
Our whole infrastructure is VNNI-centric already, no reason to be different here (at least for now).

test/BF16/Integration/lit.local.cfg
adam-smnk: I think we shouldn't have prepacked integration tests as clearly they're not scalable. That's fine for testing passes but not really runtime behavior.
I'd suggest cleaning up tests in general and reenabling them after.

So in some future modulo other priorities.

test/BF16/Integration/tpp-run-splat-shape.mlir
adam-smnk: This one should work but I ended up disabling whole dir.
We can reenable it later after some other cleanup.

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
adam-smnk: I think operation verifier should be highly localized. Going outside and parsing DLTI seems to violate this.
Our default value should be a good enough substitute and the subsequent check accepts range of VNNI factors (mod 2).

adam-smnk: Actually I shouldn't be pulling blocking factor at all here, just directly verify shape.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
rengolin: Could also be `0`. What I mean is that we're using `nullopt` to mean both states and that breaks assumptions on the user side.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
rengolin: There is: if we have tests for packing factor 4 (Arm), and we use DLTI, we'll have to call it "vnni".

Simply calling it "packing_factor" or "type_packing" would suffice.

test/BF16/Integration/lit.local.cfg
rengolin: The problem is that we want to start CI on Arm and this will skip all B16 integration tests on Arm. I recommend we either add some in this PR or in a quick follow up.

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
rengolin: agree

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: Added helper.
@rolfmorel I think such query on strings could also be upstreamed.

lib/TPP/Dialect/Xsmm/XsmmOps.cpp
adam-smnk: Removed factor, added check on memref directly.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: But then the test pass is still `-pack-vnni` 🙃 
(type_)packing is too generic. I don't mind renaming it per se but vnni is most consistent for now.

test/BF16/Integration/lit.local.cfg
adam-smnk: Let's start spinning the Arm CI with this PR. I'll follow up with test cleanup in the next few days.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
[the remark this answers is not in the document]
adam-smnk: Thanks for the snippet.
The newest DLTI API is really nice.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: Actually I could add some verification on the internal queries to make sure the value is reasonable.
But the way I see it, I ask for a blocking factor and if I get it, it means I'm good to continue.
If not, then it's up to me as a user to decide what I expect to do.

rengolin: Than don't use `std::optional`. Optional created a ternary state: { don't have it, have null, have value }. Here the usage conflates the first two, but the code that uses is is still required to make the difference.

So, either use `0` as the null set, or actually make use of `optional` and separate the three states.

test/BF16/Integration/lit.local.cfg
rengolin: We'll start Arm CI after we move to libxsmm, the runner is there

lib/TPP/Transforms/Utils/VNNIUtils.cpp
rengolin: ok, we can rename vnni to something else in a separate PR

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: After more offline discussions, I agree.
Refactored API to return `unsigned` and added more checks through the codebase.

lib/TPP/Transforms/Utils/VNNIUtils.cpp
adam-smnk: Refactored as well.
The checks now enforce that VNNI dimension is non-zero and divisible by 2.
Non-zero `blockingFactor` allows for further stricter matching.

adam-smnk: Zero factor is now allowed for `isInVnniLayout` to also double check the factor in one go. Zero blocking factor is rejected as invalid.

rengolin (approved):
Let's make sure we resolve all comments, but LGTM. Thanks!

adam-smnk:
The API has been improved as suggested. Well, we'll still see in practice if it's good but already works better after this cleanup.

I will follow up with further cleanup to reenable integration tests. That might require more relaxed verification and probably improvements to some test cases. But let's do it separately to weigh tradeoffs properly.

Otherwise, should be fine.

PR #1002, opened 2025-01-28
AVX2/SVE benchmarks for vector-to-kernels and user option design change

arun-thmn:
This PR is for two things:
1. Add new benchmarks to test `avx2` and `arm sve` instructions set for `vector-to-kernel` lowering,
2. User input option design change for register tiling. Changes the user options from `lhsTile=<x,y> and rhsTile=<y,z>` to `registerBlocking=<x,z>`

### DIFF
--- a/benchmarks/config/omp/mlir-fp32-vector-to-kernel.json
+++ b/benchmarks/config/omp/mlir-fp32-vector-to-kernel.json
@@ -6,28 +6,28 @@
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": { "OMP_NUM_THREADS": "2", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=8,16 --vector-to-kernels --registerBlocking=8,32 '" ],
-      "extensions": [ "(avx2)" ]
+      "extensions": [ "avx512" ]
     },
     "fp32_3x1024_omp_4_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": { "OMP_NUM_THREADS": "4", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=8,8 --vector-to-kernels --registerBlocking=8,32 '" ],
-      "extensions": [ "(avx2)" ]
+      "extensions": [ "avx512" ]
     },
     "fp32_3x1024_omp_8_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": { "OMP_NUM_THREADS": "8", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=4,8 --vector-to-kernels --registerBlocking=8,32 '" ],
-      "extensions": [ "(avx2)" ]
+      "extensions": [ "avx512" ]
     },
     "fp32_3x1024_omp_16_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": { "OMP_NUM_THREADS": "16", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=2,8 --vector-to-kernels --registerBlocking=8,32 '" ],
-      "extensions": [ "(avx2)" ]
+      "extensions": [ "avx512" ]
     }
   }},
   {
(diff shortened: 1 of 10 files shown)


lib/TPP/Transforms/BrgemmLinalgTiling.cpp
```diff
@@ -54,17 +54,17 @@ struct LinalgOpTiling : OpRewritePattern<linalg::BatchReduceMatmulOp> {
...
-    if (tileShapeM.size() != 2 || tileShapeN.size() != 2)
-           return failure();
+    SmallVector<int64_t> tileShapeN(2);
+    tileShapeN[0] = 1;
+    tileShapeN[1] = tileShapeM[1];
+    tileShapeM[1] = 1;
 
-    if (tileShapeM[1] != tileShapeN[0])
-            return failure();
+    if (tileShapeM.size() != 2)
```
rengolin, line 66: Move the check before usage to avoid segfaults

lib/TPP/Transforms/BrgemmLinalgTiling.cpp
```diff
@@ -114,8 +114,8 @@ struct LinalgOpTiling : OpRewritePattern<linalg::BatchReduceMatmulOp> {
       Value ubCstTiledLoop = rewriter.create<arith::ConstantIndexOp>(loc, upperBound);
       //Tile size should not be greater than the upperBound
       if ((*itrShapeM) > upperBound)
-	      return failure();
-      Value stepCstTiledLoop = rewriter.create<arith::ConstantIndexOp>(loc, upperBound/(*itrShapeM));
+              return failure();
+      Value stepCstTiledLoop = rewriter.create<arith::ConstantIndexOp>(loc, *itrShapeM);
```
rengolin, line 118: This change seems unrelated to the code movement, which should not change logic. Was that a bug in the code that you're fixing?

lib/TPP/Transforms/BrgemmLinalgTiling.cpp
```diff
@@ -184,8 +184,8 @@ struct LinalgOpTiling : OpRewritePattern<linalg::BatchReduceMatmulOp> {
             strides.push_back(rewriter.getIndexAttr(1));
           }
         } else {
-          shape.push_back(rewriter.getIndexAttr(tensorShape[j] / (*tileItr)));
-          indices.push_back(tensorShape[j] / (*tileItr));
+          shape.push_back(rewriter.getIndexAttr( (*tileItr)));
```
rengolin, line 187: Same as above, this changes logic.

lib/TPP/Transforms/BrgemmLinalgTiling.cpp
arun-thmn: For the earlier `lhsTile` and `rhsTile`, we give the blocking size as the input and we divide it with the dimensions to get the `M,K,and N` tile sizes. Sorry, this is a bad design I did at that time.

Now, we changed this. We directly give tile sizes to `registerBlocking`. So, removing the old division logic.

lib/TPP/Transforms/BrgemmLinalgTiling.cpp
arun-thmn: Yes. Moved.

rengolin:
Please rebase on top of `main` and after addressing the comments, add the `benchmark-all` tag.

lib/TPP/Transforms/BrgemmLinalgTiling.cpp
rengolin: Oh, I see, this is why now all the `4,32` became `8,32`?

arun-thmn: Yep.

arun-thmn:
@rengolin For benchmarks, (a) I have removed the extra parenthesis for only one extension, and (b) added correct extensions for `avx512`, `avx2`, and `asimd`.

benchmarks/config/omp/mlir-fp32-vector-to-kernel.json
```diff
@@ -1,128 +1,250 @@
...
+      "extensions": [ "avx512" ]
     },
     "fp32_3x1024_omp_16_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --bias --relu --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": { "OMP_NUM_THREADS": "16", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
-      "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=2,8 --vector-to-kernels --lhsTile=4,32 --rhsTile=32,1'" ],
-      "extensions": [ "(avx2|asimd)" ]
+      "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=2,8 --vector-to-kernels --registerBlocking=8,32 '" ],
+      "extensions": [ "avx512" ]
```
rengolin, line 61: Ah, no, old benchmarks should continue as before. This is because libxsmm emulates BF16 in AVX2, and we need to run these on Arm, too.

arun-thmn: This is `vector-to-kernel` FP32 benchmarks not the libxsmm one. For `sve`, we have added new benchmarks for `vector-to-kernels`.

rengolin: Fair, my mistake.

test/Passes/pass-tile-brgemm-linalg-matmul.mlir
```diff
@@ -47,12 +44,11 @@ module {
 
 // -----
 
-
-// RUN: tpp-opt %s --tile-brgemm-linalg="lhsTile=8,8 rhsTile=8,16" --split-input-file | FileCheck %s
+// RUN: tpp-opt %s --tile-brgemm-linalg="registerBlocking=8,32" --split-input-file | FileCheck %s
```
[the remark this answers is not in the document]
arun-thmn: Ahh ok. Removed.

arun-thmn:
@rengolin It couldn't parse the `avx512*` regular expression, So, left with `avx512f` and the benchmarks pass through.

arun-thmn:
`avx512.*` works.

rengolin (approved):
Thanks, LGTM!

PR #1009, opened 2025-02-08
Replace op->erase() with rewriter.eraseOp

rengolin:
Ref: https://mlir.llvm.org/docs/PatternRewriter/#pattern-rewriter


arun-thmn (approved):
@rengolin Ah okay. Thanks.
Is this option introduced newly to MLIR, or it does exists already and I missed it?

rengolin:
> Is this option introduced newly to MLIR, or it does exists already and I missed it?

It's old, but easy to miss, because both ways are _exposed_ but you only know you're not _supposed_ to use the other one by reading a document. Not great API design. :)

PR #1011, opened 2025-02-10
[CI] Fix benchmark triggers

rengolin:
This fixes both `workflow_dispatch`, now with check-boxes instead of integers and actually running correctly, as well as `pull_request` that now doesn't need to remove/add the `benchmark` label every time you update the PR.

### DIFF
--- a/.github/workflows/tpp-benchmark.yml
+++ b/.github/workflows/tpp-benchmark.yml
@@ -29,7 +29,8 @@ jobs:
   Check_LLVM:
     if: |
       (github.event_name == 'push') || (github.event_name == 'workflow_dispatch') ||
-      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark'))
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')) ||
+      (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark-full'))
     uses: ./.github/workflows/tpp-llvm.yml
     secrets: inherit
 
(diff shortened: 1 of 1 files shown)


adam-smnk:
What's the new behavior? Does it dispatch benchmarks on all PRs automatically or only after labeling but without the need to reapply label afterwards?

.github/workflows/tpp-benchmark.yml
```diff
@@ -25,18 +27,20 @@ env:
...
     secrets: inherit
 
   TPP-MLIR-EMR-BASE:
     runs-on: pcl-tiergarten
-    if: ${{ github.event_name }} == "push" || \
-        ${{ inputs.RUN_EMR_BENCH }} == 1 || \
-        (${{ github.event_name }} == "pull_request" && ${{ github.event.label.name == 'benchmark' }}) || \
-        (${{ github.event_name }} == "pull_request" && ${{ github.event.label.name == 'benchmark-all' }})
+    if: |
+      (github.event_name == 'push') ||
```
adam-smnk, line 40: Won't this trigger on any push?

rengolin: Any push to the main branch. There's a restriction on the trigger.

rengolin:
> What's the new behavior? Does it dispatch benchmarks on all PRs automatically or only after labeling but without the need to reapply label afterwards?

It always dispatches the benchmarks on every PR, but they don't run of the label is not there, so they show as "skipped".

Once applied, the label will trigger the check and when the benchmarks run again on update, the check will succeed and they won't be skipped.

the only problem now is that any update on a PR will run the benchmarks. I'm trying to find the right set of actions that will only update on rebase / reopen / label.

adam-smnk (approved):
Looks fine, let's take it for a spin and see how it goes in practice

PR #1019, opened 2025-03-04
[OpenMP] Rehash of #960 with GCC trick to get LLVM OMP

### LINKED ISSUE #959: Build issues on Ubuntu 24.04 - incorrect openmp linkage in CMake

rengolin:
This is now more robust than my previous trick, using @dbabokin's code and only hacking `openmp.cmake`, not the other ones. This works on my machine for GCC and Clang, let's see what the CI thinks about it.

Fixes #959


adam-smnk (approved):
Works on CI and locally, no benchmark regressions
Plenty of cmake magic but LGTM

cmake/modules/openmp.cmake
```diff
@@ -4,21 +4,40 @@ option(USE_OpenMP "Use OpenMP" ON)
...
 
 if(USE_OpenMP)
-  find_package(OpenMP)
-  if(OPENMP_FOUND)
+  # Only bypass for GCC
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # Only if we found an "llvm" path that we need to add
+    find_library(LLVM_OMP NAMES omp libomp libomp.so libomp.so.5)
+    find_library(PTHREAD NAMES pthread libpthread libpthread.a libpthread.so)
+    if (NOT LLVM_OMP)
```
adam-smnk, line 15: nit: shouldn't you check `PTHREAD` too?

rengolin: Very unlikely to not have pthreads, but yeah, doesn't hurt. Added.

dbabokin:
@rengolin Thank for pushing this forward.

PR #1022, opened 2025-03-14
Bump LLVM

adam-smnk:
Fixes after bump:
  - Use LLVM Triple struct
  - Use default memref.copy instead of linalg.copy during bufferization + deprecate related test
  - Lower memref.copy to XSMM unary
  - Minor syntax updates


adam-smnk:
ARM CI still not picking up after rebase - to be reopened.

PR #1024, opened 2025-03-25
Vector contract lowering to bf16dp for brgemm

arun-thmn:
To support lowering of `vector.contract` to `DotBF16Op` for gemm/mlp benchmarks with `bf16` type and `vnni=2` layout.

### DIFF
--- a/lib/TPP/DefaultPipeline.cpp
+++ b/lib/TPP/DefaultPipeline.cpp
@@ -190,7 +190,9 @@ private:
     // Lower to LLVM
     ConvertVectorToLLVMPassOptions options;
     options.amx = vnni::utils::hasAMX();
-    options.x86Vector = true;
+    #if defined(__x86_64__)
+    	options.x86Vector = true;
+    #endif
     pm.addPass(createConvertVectorToLLVMPass(options));
     pm.addPass(createFinalizeMemRefToLLVMConversionPass());
     pm.addPass(createSCFToControlFlowPass());
--- a/lib/TPP/DefaultPipeline.cpp
+++ b/lib/TPP/DefaultPipeline.cpp
@@ -190,6 +190,7 @@ private:
     // Lower to LLVM
     ConvertVectorToLLVMPassOptions options;
     options.amx = vnni::utils::hasAMX();
+    options.x86Vector = true;
     pm.addPass(createConvertVectorToLLVMPass(options));
     pm.addPass(createFinalizeMemRefToLLVMConversionPass());
     pm.addPass(createSCFToControlFlowPass());
--- a/lib/TPP/PassBundles/VectorToKernel.cpp
+++ b/lib/TPP/PassBundles/VectorToKernel.cpp
@@ -49,6 +49,7 @@ struct VectorToKernel : public tpp::impl::VectorToKernelBase<VectorToKernel>,
 
 private:
   void constructPipeline() override {
+    pm.addNestedPass<func::FuncOp>(createBF16DotProduct());
     pm.addNestedPass<func::FuncOp>(createHoistVectorTransfers());
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createVectorContractToFMA());
--- a/lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
+++ b/lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
@@ -60,6 +60,9 @@ static FailureOr<SmallVector<scf::ForOp>>
 getNestedLoop(vector::ContractionOp contractOp) {
   SmallVector<scf::ForOp> list;
   Operation *current = contractOp;
+  // It is register tiled loop structure on batch reduce matmul
+  // (M->N->Batch-reduce->K).
+  // TODO: support for matmul and batch matmul
   for (int i = 0; i < 4; i++) {
     Operation *parent = current->getParentOfType<scf::ForOp>();
     if (!parent)
@@ -70,32 +73,110 @@ getNestedLoop(vector::ContractionOp contractOp) {
   return list;
 }
 
+static bool isTransposedMatrix(vector::ContractionOp contractOp) {
+  SmallVector<AffineMap, 3> contractMaps = contractOp.getIndexingMapsArray();
+  AffineMap mapA = contractMaps[0];
+  AffineMap mapB = contractMaps[1];
+
+  auto resultsMapA = mapA.getNumResults();
+  auto resultsMapB = mapB.getNumResults();
+  assert(resultsMapA == 4 && resultsMapB == 4 &&
+         "Result dim map for A and B should be 4");
+
+  auto inputsMapA = mapA.getNumInputs();
+  auto inputsMapB = mapB.getNumInputs();
+  assert(inputsMapA == 5 && inputsMapB == 5 &&
+         "Input dim map for A and B should be 5");
+
+  auto vnniDim = dyn_cast<AffineDimExpr>(mapA.getResult(3));
+  auto dimBR = dyn_cast<AffineDimExpr>(mapA.getResult(0));
+
+  SmallVector<AffineDimExpr> listMxNxK;
+  for (unsigned int i = 0; i < inputsMapA; i++) {
+    auto affineExpr =
+        dyn_cast<AffineDimExpr>(mlir::getAffineDimExpr(i, mapA.getContext()));
+    if (affineExpr != vnniDim && affineExpr != dimBR)
+      listMxNxK.push_back(affineExpr);
+  }
+  auto dimM = listMxNxK[0];
+  auto dimN = listMxNxK[1];
+  auto dimK = listMxNxK[2];
+
+  // Transpose if the mapA is kxm
+  if (dyn_cast<AffineDimExpr>(mapA.getResult(1)) == dimK &&
+      dyn_cast<AffineDimExpr>(mapA.getResult(2)) == dimM)
+    return true;
+  // Transpose if the mapB is nxk
+  if (dyn_cast<AffineDimExpr>(mapB.getResult(1)) == dimN &&
+      dyn_cast<AffineDimExpr>(mapB.getResult(2)) == dimK)
+    return true;
+
+  return false;
+}
+
 static LogicalResult checkNestedLoop(SmallVector<scf::ForOp> loops,
                                      SmallVector<memref::SubViewOp> subviews) {
-  assert(loops.size() == 4 && subviews.size() == 3);
+  auto loopSize = loops.size();
+  assert(loopSize == 4 && subviews.size() == 3);
+
   auto subviewOpLhsOffsets = subviews[0].getOffsets();
   auto subviewOpRhsOffsets = subviews[1].getOffsets();
   auto subviewOpAccOffsets = subviews[2].getOffsets();
 
-  Value ivK = loops[0].getInductionVar();
-  if (ivK != subviewOpLhsOffsets[2] || ivK != subviewOpRhsOffsets[1])
+  SmallVector<Value> list;
+  for (size_t i = 0; i < loopSize; i++) {
+    list.push_back(loops[i].getInductionVar());
+  }
+
+  if (list[0] != subviewOpLhsOffsets[2] || list[0] != subviewOpRhsOffsets[1])
     return failure();
 
-  Value ivReduction = loops[1].getInductionVar();
-  if (ivReduction != subviewOpLhsOffsets[0] ||
-      ivReduction != subviewOpRhsOffsets[0])
+  if (list[1] != subviewOpLhsOffsets[0] || list[1] != subviewOpRhsOffsets[0])
     return failure();
 
-  Value ivN = loops[2].getInductionVar();
-  if (ivN != subviewOpAccOffsets[1] || ivN != subviewOpRhsOffsets[2])
+  if (list[2] != subviewOpAccOffsets[1] || list[2] != subviewOpRhsOffsets[2])
     return failure();
 
-  Value ivM = loops[3].getInductionVar();
-  if (ivM != subviewOpLhsOffsets[1] || ivM != subviewOpAccOffsets[0])
+  if (list[3] != subviewOpLhsOffsets[1] || list[3] != subviewOpAccOffsets[0])
     return failure();
 
   return success();
 }
+
+/// This pass lowers vector.contract (linalg.batch_reduce_matmul) for bf16
+/// (vnni=2) type into sequence of x86vector::DotBF16Op.
+///
+/// As an example, the following pseudo-code will be rewritten
+/// scf.for // m-tile
+///  scf.for // n-tile
+///   subview // C matrix
+///   scf.for // batch-reduce
+///   scf.for // k-tile
+///    subview // A and B matrix
+///    vector.read // A, B, and C matrix
+///    vector.contract
+///    vector.write // to C matrix
+///
+/// to:
+///
+/// scf.for // m-tile
+///  scf.for // n-tile
+///   vector.load // load C matrix in chunks of <16xbf16>
+///   arith.shli + vector.bitcast // upconvert to f32 and pass them as iterargs
+///   scf.for (iterargs = C matrix load as f32) // batch-reduce
+///    scf.for (iterargs = batch-reduce iterArgs) // k-tile
+///     vector.load // load 2 elements of A matrix and broadcast them into <32xbf16>
+///     vector.load // load elements of B matrix into <32xbf16>
+///     x86vector.avx512.dot %iterargs, %Ax, %Bx // accumulate in f32 (via iterargs)
+///     x86vector.avx512.dot %iterargs, %Ax, %By // accumulate in f32 (via iterargs)
+///     ..............
+///     ..............
+///    scf.yield // yield dpbf16 results
+///   scf.yield // yield results of scf.for k-tile
+///  arith.truncate // downconvert accumulator value from f32 to bf16
+///  vector.store // store back into C matrix
+///  .............
+///  ............
 
 struct BF16DotProductOp : OpRewritePattern<vector::ContractionOp> {
   using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
@@ -103,9 +184,12 @@ struct BF16DotProductOp : OpRewritePattern<vector::ContractionOp> {
   LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
                                 PatternRewriter &rewriter) const override {
 
+    if (contractOp.getKind() != vector::CombiningKind::ADD)
+      return rewriter.notifyMatchFailure(
+          contractOp, "Only combining kind 'ADD' is supported now.");
+
     // Check the vector contract operation satisfies the required pattern.
     // Check the Acc, Lhs, and Rhs of contract operation
-
     auto operands = getContractProducers(contractOp);
     if (failed(operands))
       return rewriter.notifyMatchFailure(contractOp,
@@ -137,8 +221,8 @@ struct BF16DotProductOp : OpRewritePattern<vector::ContractionOp> {
                    vector::IteratorType::reduction);
 
     if (reductionCount == 0)
-      return rewriter.notifyMatchFailure(contractOp,
-                                         "Matmul operation not supported yet");
+      return rewriter.notifyMatchFailure(
+          contractOp, "Element-wise operation not supported yet");
 
     if (reductionCount == 1)
       return rewriter.notifyMatchFailure(
@@ -180,6 +264,10 @@ struct BF16DotProductOp : OpRewritePattern<vector::ContractionOp> {
     if (vnni != 2)
       return rewriter.notifyMatchFailure(
           contractOp, "Only VNNI layout=2 is supported, now");
+
+    if (isTransposedMatrix(contractOp))
+      return rewriter.notifyMatchFailure(contractOp,
+                                         "Matrices shoudn't be transposed.");
 
     auto loops = getNestedLoop(contractOp);
     if (failed(loops))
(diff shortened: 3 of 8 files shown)


lib/TPP/DefaultPipeline.cpp
```diff
@@ -190,6 +190,7 @@ struct DefaultPipeline : public tpp::impl::DefaultPipelineBase<DefaultPipeline>,
     // Lower to LLVM
     ConvertVectorToLLVMPassOptions options;
     options.amx = vnni::utils::hasAMX();
+    options.x86Vector = true;
```
rengolin, line 193: this wont work on arm

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (!parent)
+      return failure();
+    list.push_back(dyn_cast<scf::ForOp>(parent));
+    current = parent;
+  }
+  return list;
+}
+
+static LogicalResult checkNestedLoop(SmallVector<scf::ForOp> loops,
+                                     SmallVector<memref::SubViewOp> subviews) {
```
rengolin, line 78:
```suggestion
                                     SmallVector<memref::SubViewOp> subviews) {
  assert(loops.size() == 4 && subviews.size() == 3);
```

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+  auto subviewOpLhsOffsets = subviews[0].getOffsets();
+  auto subviewOpRhsOffsets = subviews[1].getOffsets();
+  auto subviewOpAccOffsets = subviews[2].getOffsets();
+
+  Value ivK = loops[0].getInductionVar();
```
rengolin, line 83: if the only thing you do with the loops is get the induction variable, you could save as a list of indvars to begin with.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+      return rewriter.notifyMatchFailure(contractOp,
+                                         "Matmul operation not supported yet");
+
+    if (reductionCount == 1)
+      return rewriter.notifyMatchFailure(
+          contractOp, "Batch matmul operation not supported yet");
+
+    if (reductionCount == 2)
+      return rewriter.notifyMatchFailure(
+          contractOp, "Batch reduce matmul operation without vnni layout");
```
rengolin, line 147: you dont seems to be checking the type is actually bf16. a weird f32 contraction would have passed here

test/BF16/Integration/vector-contract-to-bf16dp.mlir
adam-smnk, line 1: Please make the matrix sizes small (e.g. use fewer tiles) for integration testing.
It might run fine on CI now but this slowly adds up and is more annoying for local runs.

test/Passes/pass-vector-contract-to-bf16dp.mlir
adam-smnk, line 1: Could these test cases be simplified?

I feel there's some ops are not really relevant to the test like the `scf.forall` parallelization loop.
Also, I'd suggest using different sizes for each memref dimension for both readability and better test coverage.

test/Passes/pass-vector-contract-to-bf16dp.mlir
```diff
@@ -0,0 +1,177 @@
+// RUN: tpp-opt %s --vector-contract-to-bf16dp --split-input-file  | FileCheck -check-prefix=CONF1 %s
+// RUN: tpp-opt %s --tile-brgemm-linalg="registerBlocking=2,32,2" --loop-invariant-code-motion --vectorization-pass --vector-contract-to-bf16dp --split-input-file  | FileCheck -check-prefix=CONF2 %s
+// RUN: tpp-opt %s --tile-brgemm-linalg="registerBlocking=4,32,2" --loop-invariant-code-motion --vectorization-pass --vector-contract-to-bf16dp --split-input-file  | FileCheck -check-prefix=CONF3 %s
+// RUN: tpp-opt %s --tile-brgemm-linalg="registerBlocking=2,16,2" --loop-invariant-code-motion --vectorization-pass --vector-contract-to-bf16dp --split-input-file  | FileCheck -check-prefix=CONF4 %s
+// RUN: tpp-opt %s --tile-brgemm-linalg="registerBlocking=2,32,4" --loop-invariant-code-motion --vectorization-pass --vector-contract-to-bf16dp --split-input-file  | FileCheck -check-prefix=CONF5 %s
```
adam-smnk, line 5: nit: it's more common to have one pass pipeline for testing and have more separate test cases that test for different outcomes - maybe some of these could be removed?

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+//-----------------------------------------*-
+// C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements tile configuration hoisting on parallel loops.
```
adam-smnk, line 11: copy pasta - imo, these descriptions can be omitted in general

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
+//===-VectorContractToBF16DotProduct.cpp
+//-----------------------------------------*-
+// C++-*-===//
```
adam-smnk, line 3: nit: header formatting

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+#define GEN_PASS_DEF_BF16DOTPRODUCT
+#include "TPP/Passes.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+namespace mlir {
+namespace tpp {
+
+static FailureOr<SmallVector<vector::TransferReadOp>>
+getContractOperands(vector::ContractionOp contractOp) {
```
adam-smnk, line 39: nit: it's more pf `getContractProducers`

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+  SmallVector<vector::TransferReadOp> list;
+  for (int i = 0; i < 3; i++) {
```
adam-smnk, line 41: nit: you could use foreach loop with `getOperands()`

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+    if (reductionCount == 3 &&
+        (lhsType.getRank() != 4 || rhsType.getRank() != 4))
+      return rewriter.notifyMatchFailure(
+          contractOp,
+          "Invalid rank for batch reduce operation with vnni layout");
+
+    int64_t M = accType.getDimSize(0);
+    int64_t N = accType.getDimSize(1);
+    int64_t K = lhsType.getDimSize(lhsType.getRank() - 2);
```
adam-smnk, line 165: Do you check there are no transposes etc.?

Side note, maybe `inferContractionDims` could help with some of those checks but not sure.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+    list.push_back(subViewOp);
+  }
+  return list;
+}
+
+static FailureOr<SmallVector<scf::ForOp>>
+getNestedLoop(vector::ContractionOp contractOp) {
+  SmallVector<scf::ForOp> list;
+  Operation *current = contractOp;
+  for (int i = 0; i < 4; i++) {
```
adam-smnk, line 67: That's extremely specific 😅
Probably fine for quick PoC - maybe just add some todo comment.

adam-smnk:
Btw, please mark it as `ready for review` if it's good to go.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Thanks. Added this assertion.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Added a checklist to validate `bf16` type or not.
`    if (!elementType.isBF16())
      return rewriter.notifyMatchFailure(contractOp, "The type is not BF16")`

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Sorry! I'm being too lazy :). Updated the description.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Correctly formatted.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Okay. Changed function name `getContractOperands` to `getContractProducers`

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Yep. Changed the loop with the usage of 'getOperands()'

test/BF16/Integration/vector-contract-to-bf16dp.mlir
arun-thmn: Sure. I have reduced the layers sizes from `1024` to `128 or 64`

test/Passes/pass-vector-contract-to-bf16dp.mlir
arun-thmn: Okay, Removed the un-relevant instructions from the test-cases.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+    auto vectorReadOpRhs = readOps[1];
+
+    // Check whether the operand of vector transfer read is a subview
+    auto readOpSubviews = getReadOperands(readOps);
+    if (failed(readOpSubviews))
+      return rewriter.notifyMatchFailure(
+          contractOp, "Vector read op operands are not a subview");
+
+    auto subviews = *readOpSubviews;
+    auto subviewOpAcc = subviews[2];
```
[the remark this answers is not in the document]
arun-thmn: Two reasons:
(1) To check whether it is `bf16` type,
(2) We need it later to create further new subviews from the producer subview. Used later here: https://github.com/libxsmm/tpp-mlir/blob/5896549c26eb150238ebb8f48e198eb20f30b5e8/lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp#L214

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+  if (ivM != subviewOpLhsOffsets[1] || ivM != subviewOpAccOffsets[0])
+    return failure();
+
+  return success();
+}
+
+struct BF16DotProductOp : OpRewritePattern<vector::ContractionOp> {
+  using OpRewritePattern<vector::ContractionOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(vector::ContractionOp contractOp,
```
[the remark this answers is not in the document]
arun-thmn: This you meant `tensor` or `memref` type check?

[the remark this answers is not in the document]
adam-smnk: The `vector.contract` attribute

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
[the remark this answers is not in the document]
adam-smnk: The type you should be able to check directly on contraction.

The number 2 makes me wonder how we'd compose it with GEMM prologue like eltwise ops before contract.
But that's probably a problem for later. If we move to linalg tiling then it probably not an issue anymore.

lib/TPP/DefaultPipeline.cpp
arun-thmn: Added a check to set `x86Vector = true` only on `x86` machines

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Okay. I saved all induction variables to a list and then begin with validation.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+    return failure();
+
+  Value ivM = loops[3].getInductionVar();
+  if (ivM != subviewOpLhsOffsets[1] || ivM != subviewOpAccOffsets[0])
+    return failure();
+
+  return success();
+}
+
+struct BF16DotProductOp : OpRewritePattern<vector::ContractionOp> {
```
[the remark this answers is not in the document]
arun-thmn: Sure. I have a simple rewrite example with comments.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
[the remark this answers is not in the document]
arun-thmn: Okay.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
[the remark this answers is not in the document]
arun-thmn: Got it. Added a `CombiningKind` check whether it is `ADD` or not.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,381 @@
...
+    // Check the operation type MatMul, B-MatMul, or BR-MatMul (FP32/BF16)
+    SmallVector<vector::IteratorType> contractIteratorTypes =
+        contractOp.getIteratorTypesArray();
+    int reductionCount =
+        std::count(contractIteratorTypes.begin(), contractIteratorTypes.end(),
+                   vector::IteratorType::reduction);
+
+    if (reductionCount == 0)
+      return rewriter.notifyMatchFailure(contractOp,
+                                         "Matmul operation not supported yet");
```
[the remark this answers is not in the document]
arun-thmn: Agreed. Updated the comment as `element-wise` oeration.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Added a new function `isTransposedMatrix` to check A and B matrix are transposed or not.

https://github.com/libxsmm/tpp-mlir/blob/edb647237744d5f8fcaff539e5db2a0aa302d0a4/lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp#L76

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Okay. Added a comment saying it support `brgemm` now, later will be extended to `matmul` and `brmatmul`

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
adam-smnk: Are possible permutations in affine maps supported?
If not, and we don't really need to support all these variant right now, it should be safer to have any explicit matcher to allow only specific set of maps and their order.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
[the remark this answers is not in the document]
adam-smnk: IMO, it's more generic to say `Expected at least one reduction`.

Anyway, doesn't `vector.contract` require to have at least one reduction? If it's an op invariant then we shouldn't revalidate it in the pass matcher.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: I think `permutations` are supported in `affine maps`.

adam-smnk: Correct, I meant if the pass expects and can handle them?

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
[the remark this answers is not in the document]
arun-thmn: Ahh okay. Changed the failure msg to `Expected at least one reduction`.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: I think, it should as we do check (i) the loop `indVar` with `subview` access, and (ii) map dimensions.

arun-thmn: Added function `permutationCheck` to validate certain pattern of affine maps.
We support pattern of:
MapA: `{Batch-reduction, vnni, M, N, K} or {Batch-reduction, M, N, K, vnni} -> {Batch-reduction, M, K, vnni}`
MapB: `{Batch-reduction, vnni, M, N, K} or {Batch-reduction, M, N, K, vnni} -> {Batch-reduction, K, N, vnni}`

https://github.com/libxsmm/tpp-mlir/blob/17352b935bf714dee76b13d9bee73e47860279d6/lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp#L117

adam-smnk: Looks pretty good 👍

It could be a bit more concise using `bindDims` and `inferFromExprList` pattern for matching explicit maps.
For example, see `ContractionOpToDotLowering::matchAndRewriteMaskableOp`.
Either way, seems fine.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,529 @@
+//===-------------- VectorContractToBF16DotProduct.cpp ----------*- C++-*-===//
```
adam-smnk, line 1: nit: this header formatting is a bit off

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
```diff
@@ -0,0 +1,529 @@
...
+      return rewriter.notifyMatchFailure(contractOp, "The type is not BF16");
+
+    // Check the operation type MatMul, B-MatMul, or BR-MatMul (FP32/BF16)
+    SmallVector<vector::IteratorType> contractIteratorTypes =
+        contractOp.getIteratorTypesArray();
+    int reductionCount =
+        std::count(contractIteratorTypes.begin(), contractIteratorTypes.end(),
+                   vector::IteratorType::reduction);
+
+    if (reductionCount == 0)
```
adam-smnk, line 279: Is this case needed at all? I think vector.contract can't have zero reductions but please double check.

adam-smnk (commented):
Small nits but looks pretty good already.
Could it be added somewhere to the default pipeline?

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Formatted correctly now, I think. 
Looked into other file's header but found different formatting. Not sure, which one is correct. If, still the formatting is not proper, please point me to a file with correctly formatted header. I will follow it.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
arun-thmn: Seems a valid point. I removed this case.

arun-thmn:
Thanks @adam-smnk for your comments. I have added this pass before the `vector hoisting pass` in the pipeline.
The input to this pass is after `vectorization` pass. This pass has a check on the `type`, this will not rewrite for `f32` type.

lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
adam-smnk: Looks fine now.
When in doubt, check random upstream files for reference. They're mostly consistent so cross-reference a few 😉

lib/TPP/PassBundles/VectorToKernel.cpp
```diff
@@ -49,6 +49,7 @@ struct VectorToKernel : public tpp::impl::VectorToKernelBase<VectorToKernel>,
 
 private:
   void constructPipeline() override {
+    pm.addNestedPass<func::FuncOp>(createBF16DotProduct());
```
adam-smnk, line 52: Won't this prevent hoisting and/or shouldn't the rewrite be at least after preprocessing?
Not sure so just double checking

adam-smnk (approved):
Looks fine - I think it's a solid specialized lowering
Let's see how we can generalize it next 😎

lib/TPP/PassBundles/VectorToKernel.cpp
arun-thmn: This pass takes IR after `vectorization` pass. It don't rely on the hoisting and has `type` check. So, it won't rewrite the `f32` type.

PR #1025, opened 2025-04-03
Benchmark files for avx512bf16 dot-product

arun-thmn:
Adding  `json` files to CI pipeline for bench-marking `avx512bf16` dot-product path. The performance should be:
 (1) on `zen5` machine it should be `95%` performance compared to `bf16` tpp-libxsmm run,
 (2) on `emr` machine it should be `~2x` slower compared to `f32` tpp-vector run.

### DIFF
--- a/benchmarks/config/base/base.json
+++ b/benchmarks/config/base/base.json
@@ -90,7 +90,7 @@
       "benchmark": [ "mlir-gen", "--kernel=const --bias --relu --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": {},
       "flags": [ "-n", "100",  "-run-args='--def-parallel --vector-to-kernels --registerBlocking=8,32,1'" ],
-      "extensions": ["avx512_bf16"]
+      "extensions": ["avx512.*"]
     },
     "mlp_fp32_mlir_vector_avx2": {
       "type": "IR-GEN",
--- a/benchmarks/config/base/base.json
+++ b/benchmarks/config/base/base.json
@@ -40,21 +40,21 @@
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=f32 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32" ],
       "environment": {},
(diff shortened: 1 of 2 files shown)


rengolin:
Weird, the OMP tests seem to be achieving speed 2x faster than libxsmm_dnn.

adam-smnk:
Is there even point in running these on EMR?

rengolin:
> Is there even point in running these on EMR?

They should not be running at all, since EMR doesn't have `avx512_bf16` and there's a filter for that on the JSON file. Maybe something's fishy with the `cpuid` detection?

adam-smnk:
> They should not be running at all, since EMR doesn't have `avx512_bf16`

That's the neat part it does just runs poorly.

arun-thmn:
> > Is there even point in running these on EMR?
> 
> They should not be running at all, since EMR doesn't have `avx512_bf16` and there's a filter for that on the JSON file. Maybe something's fishy with the `cpuid` detection?

The EMR has `avx512_bf16`.

`[athangam@pcl-spr01 ~]$ lscpu | grep -i bf16
Flags:                              .......cqm_mbm_total cqm_mbm_local split_lock_detect avx_vnni avx512_bf16 wbnoinvd dtherm ida arat pln pts hwp hwp_act_window hwp_epp hwp_pkg_req avx512vbmi umip pku ospke waitpkg avx512_vbmi2 gfni vaes vpclmulqdq avx512_vnni ...`

rengolin:
> > > Is there even point in running these on EMR?
> > 
> > 
> > They should not be running at all, since EMR doesn't have `avx512_bf16` and there's a filter for that on the JSON file. Maybe something's fishy with the `cpuid` detection?
> 
> The EMR has `avx512_bf16`.

I stand corrected! 😄

adam-smnk (approved):
Looks good on its own
It'd be nice to have finer control over where we run various benchmarks but that might require a separate solution - not a blocker here

rengolin:
> > Weird, the OMP tests seem to be achieving speed 2x faster than libxsmm_dnn.
> 
> Is this for libxsmm_dnn_bf16? In `zen5`, I see the OMP numbers are matching other than `16 threads`. ` libxsmm_dnn_bf16` 16 threads is ~1.5x slower, other numbers are fine I think.

I may have misread. LGTM!

rengolin (approved):

PR #1027, opened 2025-04-10
Add contract to mlir-gen

rengolin:
Adds `linalg.contract` to MLIR Gen so we can start testing contraction inputs.

### DIFF
--- a/test/Integration/mlir-gen.mlir
+++ b/test/Integration/mlir-gen.mlir
@@ -8,10 +8,18 @@
 // RUN: mlir-gen --kernel=const --bias --relu --seed=123 --batch=10 --layers=10,10,10 | tpp-run -e entry -entry-point-result=void
 
 // Matmul only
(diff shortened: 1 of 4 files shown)


rengolin:
Not bad co-pilot reviews...

rolfmorel:
Indeed, overall LGTM but was blanking on `keepGenericMatmul` is for (... and still am, TBH).

rengolin:
> Indeed, overall LGTM but was blanking on what `keepGenericMatmul` is for (... and still am, TBH).

It's because some of our patterns don't match `matmul`, just `generic`, so we allow printing named element-wise but not matmuls. With `contract`, this makes no sense. The idea is to eventually get rid of that restriction by matching _matmul-like_ operations on all forms (`generic`, `contract`, `matmul`).

rolfmorel (approved):
LGTM.

You could address _the machine_'s comments, though as I hope we will replace C++ mlir-gen soon (apologies for the delay on this 😢 ), one could argue it's not the best way to spend time.

PR #1034, opened 2025-04-29
Fix CI build issue for Graviton

shahidact:
CI was failing for the AMX specific test on ARM graviton. This patch disables the AMX specific
test on ARM graviton to fix the CI build issue.


adam-smnk (approved):
Looks fine plus fixes issues on my local machine

PR #1044, opened 2025-05-11
Fix missing CMake dependency for python script

rolfmorel:
Fixes issues https://github.com/libxsmm/tpp-mlir/issues/1043


alheinecke (approved):
reviewed, tested, works.

PR #1049, opened 2025-05-20
Support target feature propagation to llvm backend.

shahidact:
Add support to construct TargetMachine based on the target feature which can be propagated to llvm backend. Target feature can be specified from command line flag '--target-feature' to tpp-run tool. This flag can be used by passes which needs knowledge of cpu target feature.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -538,7 +538,7 @@ def TppRunnerWrapper : Pass<"tpp-runner-wrapper", "ModuleOp">{
            "Kernel target device backend (cpu, cuda, intel).">,
     Option<"wrapperCpuTargetFeature", "target-feature", "std::string",
             /*default=*/"",
-           "CPU target feature (avx, avx2, avx512).">,
+           "CPU target feature (avx, avx2, avx512f, avx512vnni, avx512bf16, amx, amx_bf16, amx_tile, neon, sve).">,
     Option<"offloadToDevice", "offload-on-device", "bool",
             /*default=*/"true",
            "Offload kernel arguments to the target device.">,
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -538,7 +538,7 @@ def TppRunnerWrapper : Pass<"tpp-runner-wrapper", "ModuleOp">{
            "Kernel target device backend (cpu, cuda, intel).">,
     Option<"wrapperCpuTargetFeature", "target-feature", "std::string",
             /*default=*/"",
-           "CPU target feature (avx512, avx2, avx, sse).">,
+           "CPU target feature (avx, avx2, avx512).">,
     Option<"offloadToDevice", "offload-on-device", "bool",
             /*default=*/"true",
            "Offload kernel arguments to the target device.">,
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -103,6 +103,10 @@ def VectorContractToFMA : Pass<
                            "tensor::TensorDialect",
                            "vector::VectorDialect",
                            "arith::ArithDialect"];
+  let options = [
+         Option<"targetFeature", "target-feature", "std::string", "",
+                "Target feature for a given architecture">,
+  ];
 }
 
 def VectorContractToAMX : Pass<
@@ -532,6 +536,9 @@ def TppRunnerWrapper : Pass<"tpp-runner-wrapper", "ModuleOp">{
     Option<"backend", "backend", "std::string",
             /*default=*/"\"cpu\"",
            "Kernel target device backend (cpu, cuda, intel).">,
+    Option<"wrapperCpuTargetFeature", "target-feature", "std::string",
+            /*default=*/"",
+           "CPU target feature (avx512, avx2, avx, sse).">,
     Option<"offloadToDevice", "offload-on-device", "bool",
             /*default=*/"true",
            "Offload kernel arguments to the target device.">,
--- a/lib/TPP/Transforms/VectorContractToFMA.cpp
+++ b/lib/TPP/Transforms/VectorContractToFMA.cpp
@@ -9,6 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "TPP/Passes.h"
 #include "TPP/Transforms/Transforms.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
@@ -21,7 +22,6 @@
 
 namespace mlir {
 namespace tpp {
-#define GEN_PASS_DECL_VECTORCONTRACTTOFMA
 #define GEN_PASS_DEF_VECTORCONTRACTTOFMA
 #include "TPP/Passes.h.inc"
 } // namespace tpp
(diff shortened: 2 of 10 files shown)


include/TPP/Passes.td
```diff
@@ -532,6 +536,9 @@ def TppRunnerWrapper : Pass<"tpp-runner-wrapper", "ModuleOp">{
     Option<"backend", "backend", "std::string",
             /*default=*/"\"cpu\"",
            "Kernel target device backend (cpu, cuda, intel).">,
+    Option<"wrapperCpuTargetFeature", "target-feature", "std::string",
+            /*default=*/"",
+           "CPU target feature (avx512, avx2, avx, sse).">,
```
rengolin, line 541: there's no `avx512` or `sse` in the actual list.

tools/tpp-run/tpp-run.cpp
```diff
@@ -141,13 +141,39 @@ llvm::cl::opt<std::string>
     defGpuBackend("gpu", llvm::cl::desc("Target GPU backend for lowering"),
                   llvm::cl::value_desc("cuda,intel"), llvm::cl::init(""));
 
+// Select target CPU feature for the pipeline.
+llvm::cl::opt<std::string> runnerCpuTargetFeature(
+    "target-feature", llvm::cl::desc("Specify CPU target feature for lowering"),
+    llvm::cl::value_desc("avx, avx2, avx512"), llvm::cl::init(""));
```
rengolin, line 147: I'd add the whole list here, and keep it up-to-date when we add more.

Should be safe to ignore the `testfeature`, though. :)

rengolin (commented):
Looks great, thanks!

How will this affect the existing `cpu`, `fpu` and `triple` options? Shall we get rid of them and set the `ifdef` defaults for the `target-feature` flag instead?

tools/tpp-run/tpp-run.cpp
```diff
@@ -141,13 +141,39 @@ llvm::cl::opt<std::string>
...
+
 // Kernel buffers - arguments and return values - are expected to be allocated
 // on GPU.
 llvm::cl::opt<bool>
     defGpuArgs("gpu-args",
                llvm::cl::desc("Kernel buffers are allocated on GPU"),
                llvm::cl::init(true));
 
+struct TargetMachineOptions {
+  StringRef triple;
```
adam-smnk, line 157: `StringRef` doesn't own the data, this might cause lifetime issues with the current getter.
I'd suggest using plain string storage like `std::string`.

lib/TPP/Transforms/VectorContractToFMA.cpp
```diff
@@ -21,6 +21,7 @@
 
 namespace mlir {
 namespace tpp {
+#define GEN_PASS_DECL_VECTORCONTRACTTOFMA
```
adam-smnk, line 24: Why is this needed?

tools/tpp-run/tpp-run.cpp
```diff
@@ -141,13 +141,39 @@ llvm::cl::opt<std::string>
...
+TargetMachineOptions getTargetMachineOptions(StringRef option) {
+  return StringSwitch<TargetMachineOptions>(option)
+      .Case("avx", {"x86_64-linux-gnu", "sandybridge", "+avx"})
+      .Case("avx2", {"x86_64-linux-gnu", "haswell", "+avx2"})
+      .Case("avx512f", {"x86_64-linux-gnu", "skylake", "+avx512f"})
+      .Case("avx512vnni", {"x86_64-linux-gnu", "znver4", "+avx512vnni"})
+      .Case("neon", {"armv8a-linux-gnu", "cortex-a53", "+neon"})
+      .Case("sve", {"armv8a-linux-gnu", "a64fx", "+sve"})
+      .Case("testfeature", {"x86_64-linux-gnu", "sandybridge", "+testfeature"})
+      .Default({"", "", ""});
```
adam-smnk, line 174: Don't we need the default to at least point to the current host machine?
I'd suggest checking benchmarks to see if we're not losing anything with this change.

rolfmorel:
Does this feature allow us to get rid of the following `#ifdef`:
https://github.com/libxsmm/tpp-mlir/blob/61536e44fbd27d5409cff411144a42a17185d10f/lib/TPP/DefaultPipeline.cpp#L193-L195

shahidact:
> Does this feature allow us to get rid of the following `#ifdef`:
> 
> https://github.com/libxsmm/tpp-mlir/blob/61536e44fbd27d5409cff411144a42a17185d10f/lib/TPP/DefaultPipeline.cpp#L193-L195

Yes, by having a small parsing routine to parse this option using **pipelineCpuTargetFeature** to distinguish among x86, arm etc

lib/TPP/Transforms/VectorContractToFMA.cpp
shahidact: Oh, this is to show the flow of `target-feature` option to client passes. I have added an option to this pass.

shahidact:
> Looks great, thanks!
> 
> How will this affect the existing `cpu`, `fpu` and `triple` options? Shall we get rid of them and set the `ifdef` defaults for the `target-feature` flag instead?

Existing options `cpu`, `fpu` and `triple` will become redundant. We shall get rid of them and move the default setting logic into the getTargetMachineOptions() API.

tools/tpp-run/tpp-run.cpp
shahidact: > I'd add the whole list here, and keep it up-to-date when we add more.

Is there any handy list of features with corresponding cpu name?

rengolin: I mean just the ones we support in the `StringSwitch`.

tools/tpp-run/tpp-run.cpp
shahidact: Done.

include/TPP/Passes.td
shahidact: Done.

tools/tpp-run/tpp-run.cpp
shahidact: Done.

lib/TPP/Transforms/VectorContractToFMA.cpp
shahidact: Thanks, updated accordingly.

test/Integration/tpp-run.mlir
```diff
@@ -159,3 +162,6 @@ func.func @entry(%A: tensor<4x8xf32>,
 // BENCH_STATS_10: call @_entry
 // BENCH_STATS_10-NOT: call @_entry
 // BENCH_STATS_10: {{[0-9]+}}{{.?}}{{[0-9e-]+}}
+
+// CPU_FEATURE-LABLE: @entry
+// CPU_FEATURE: '+testfeature' is not a recognized feature for this target (ignoring feature)
```
adam-smnk, line 167: Is this needed for anything?

adam-smnk (approved):
LGTM

test/Integration/tpp-run.mlir
shahidact: It is needed to show the propagation of target feature. There could have been other way to do that, but this felt easy to me.

adam-smnk: Ohhh I see, propagation as to confirm that the flag went correctly all the way down through passes.
Sounds good 👍

PR #1050, opened 2025-05-20
[tpp-sched] Fix scoping of convert-cf-to-llvm from func to module

rolfmorel:
With the improper scoping it kept around unrealized_casts.


adam-smnk:
Maybe most of these could be replaced by the newer `convert-to-llvm`?

rengolin (approved):

PR #1051, opened 2025-05-20
[tpp-sched] Make passes which operate on modules operate on modules

rolfmorel:
There was no reason for these passes to only get access to the contained func.func. In case of `-convert-cf-to-llvm` (#1050) it led to the incorrect behaviour.

This corresponds to what is happening in `DefaultPipeline.cpp`.

Also fixes typo "convert-omp-to-llvm" -> "convert-openmp-to-llvm".


rolfmorel:
Agreed, though this PR is just about getting `tpp-sched` to behave the same as the `tpp-opt` pipelines.

We can do the move to `-convert-to-llvm` in both the pipeline and the schedule in another PR (as it a quick look did not tell which passes can actually be replaced).

adam-smnk (approved):

PR #1055, opened 2025-06-06
Bump llvm - transform.apply_registered_pass options params

rolfmorel:


test/Passes/DefaultPipeline/amx-initialization.mlir
```diff
@@ -5,9 +5,10 @@
 
 
 // CHECK-AMX-BF16-LABEL:   llvm.func @entry
-// CHECK-AMX-BF16:         amx.tileloadd64
-// CHECK-AMX-BF16:         amx.tdpbf16ps
-// CHECK-AMX-BF16:         amx.tilestored64
+// CHECK-AMX-BF16:         llvm.call_intrinsic "llvm.x86.tileloadd64.internal"{{.*}} -> !llvm.x86_amx
```
rolfmorel, line 8: @adam-smnk , is it expected that these ops now lower to `llvm.call_intrinsic`s?

rengolin: The previous CHECK line was not complete. I think it was calling the intrinsic anyway, but those now changed to "x86.tileload64" instead of "amx.tileload64". Please check that this is true, and if it is, the CHECK lines just need changing.

arun-thmn:
@rolfmorel The below two tests should be removed as these tests targets `amx` and in wrong `avx512bf16dp` folder.
 ` TPP_OPT :: BF16/Integration/avx512bf16/vector-contract-to-amx-gemm.mlir`
  `TPP_OPT :: BF16/Integration/avx512bf16/vector-contract-to-amx-mlp.mlir`
  
  And, the below tests passes on `Zen5` (compile node) because it's target is `AMX` machine. On the login node `emr` it get invoked and on compile node `zen4` it get skipped.
  ` TPP_OPT :: BF16/Integration/amx/vector-contract-to-amx-gemm.mlir`
  `TPP_OPT :: BF16/Integration/amx/vector-contract-to-amx-mlp.mlir`

test/Passes/DefaultPipeline/amx-initialization.mlir
shahidact: > @adam-smnk , is it expected that these ops now lower to `llvm.call_intrinsic`s?

Yes, it seems the definition has changed, it is being generated through the intrinsic "llvm.x86.tileloadd64.internal".

shahidact: Ah so @adam-smnk has updated it :)

shahidact: So, the downstream needs to be updated accordingly.

adam-smnk: Yes, x86vector and amx now lower through llvmir dialect intrinsic ops instead of these low-level llvm-compatible ops.

For example, `amx.tile_load` used to lower to amx intrinsic op `amx.tileloadd64` which would later lower to the specific llvm intrinsic call `llvm.x86.tileload...`. Now it lowers directly to an intrinsic call.

rolfmorel: @adam-smnk , do you have a PR for me to link to in the description?

adam-smnk: https://github.com/llvm/llvm-project/pull/133692
https://github.com/llvm/llvm-project/pull/140559

adam-smnk:
It looks like type conversion for within non-amx ops is missing after I axed `AMXDialectLLVMIRTranslationInterface`.
I'll see if I can somehow add the type conversion without bringing the whole thing back.

rengolin:
> This should be fixed after [llvm/llvm-project#143871](https://github.com/llvm/llvm-project/pull/143871)

Indeed. Best to re-bump now than change the tests and bump again later.

test/Integration/tpp-run-splat-mlp.mlir
```diff
@@ -53,7 +53,7 @@ func.func @entry(%arg0: tensor<8x8xf32>, %output: tensor<8x8xf32>) -> tensor<8x8
 // CHECK-DAG: memref.global "private" constant @__constant_1x1x8x8xf32 : memref<1x1x8x8xf32>
 // CHECK-DAG: memref.global "private" constant @__constant_1x1x8x8xf32_0 : memref<1x1x8x8xf32>
 // CHECK-DAG: memref.global "private" constant @__constant_8xf32 : memref<8xf32>
-// CHECK-DAG: memref.global "private" constant @__constant_8xf32_0 : memref<8xf32>
+// CHECK-DAG: memref.global "private" constant @__constant_8xf32_1 : memref<8xf32>
```
adam-smnk, line 56: Any clue where this change is coming from or if there's any actual change other than name?
My guess is the bufferization state update caused this.

adam-smnk (approved):

test/Integration/tpp-run-splat-mlp.mlir
rolfmorel: No change besides the name. Other constants which share a type still have `_0` as the first repeated occurrence of a constant of that type. There's no `@__constant_8xf32_0` at all in the IR any more. My guess would be that some change upstream duplicated `@__constant_8xf32_0` so that there were `@__constant_8xf32_0` and `@__constant_8xf32_1` and some canonicalization/cse got rid of the duplicate constant. Just a guess though.

PR #1056, opened 2025-06-10
Vector contract to µKernel lowering f32 + bf16 + f16 type

arun-thmn:
This PR target µKernel lowering for `f32`, `bf16` and `f16`. The lowering is based on the target architecture and type:

1. f32 - lowering is decided based on avx512 (1st preference) or avx2 support by the machine
2. bf16 - We support three different lowerings (a) avx512bf16dp - machine that has `avx512_bf16` support, (b) machine that has `vcvtneebf162ps, vcvtneobf162ps,` and `vbcstnebf162ps` instructions, and (c) fallback case - where we up-convert bf16 to f32, do FMAs, down-convert the accumulation back to bf16.
3. f16 - we support lowering only for machine that has `vcvtneeph2ps,  vcvtneoph2ps,` and `vbcstnesh2ps`.

`TODO:` After this patch gets merged, the existing `f32-fma` and `avx512dp` pass will get removed and `.json` files will get added/updated to measure `perf`


lib/TPP/Transforms/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ add_mlir_library(TPPTransforms
   VectorContractToFMA.cpp
   VectorContractToBF16DotProduct.cpp
   VectorContractToAMX.cpp
+  VectorContractTouKernels.cpp
```
adam-smnk: nit: this naming breaks camel case, I'd suggest `ToUKernels` or `ToUkernels`

lib/TPP/Transforms/VectorContractTouKernels.cpp
```diff
@@ -0,0 +1,1342 @@
...
+      vnni = 1;
+
+      if (K != 1)
+        return rewriter.notifyMatchFailure(
+            contractOp, "K tile size should be equal to one");
+    }
+
+    if (avx512 && (N < 16))
+      return rewriter.notifyMatchFailure(
+          contractOp, "N tile size should be >= 16 for avx512 targets");
```
adam-smnk: Wouldn't smaller sizes also work through smaller registers (xmm, ymm)?
It might be less optimal but I'd hope it vectorizes fine too.

lib/TPP/Transforms/VectorContractTouKernels.cpp
```diff
@@ -0,0 +1,1342 @@
...
+    auto memrefMask = rewriter.create<mlir::memref::AllocOp>(
+        reductionForOp.getLoc(),
+        MemRefType::get({1}, VectorType::get(sizeFactor, i32Type)));
+    rewriter.create<memref::StoreOp>(reductionForOp.getLoc(), mBcst, memrefMask,
+                                     ValueRange{indexOp_c0});
+
+    rewriter.setInsertionPoint(reductionForOp);
+    llvm::SmallVector<Value, 8> loopItrArgs;
+
+    // C matrix load:
```
adam-smnk: It'd be nice to refactor these cases into a small helper func to improve readability

lib/TPP/Transforms/CMakeLists.txt
arun-thmn: Actually, the `u` is representation for `µ` (micro). It's wired to have `µ` on a file name, so lived with `u`.
I think, it would be better I re-name to `ToMicroKernels.cpp`.

adam-smnk: I know it's a bit silly but generally within the codebase camel case overrules everything else for file names.
For example, you'll find a lot of files with `X86` with capital `X` due to the formatting rules.

`MicroKernels` is good too.

arun-thmn: Renamed from `uKernels` to `MicroKernels`.

lib/TPP/Transforms/VectorContractTouKernels.cpp
arun-thmn: Sure. Moved to a small helper function.

arun-thmn:
@adam-smnk @rengolin As register blocking and linalg vectorizer on `tensor` type are now available in `tpp-mlir`, this pass can be extended to work on `tensor` type as well. Making it to support on `tensor` type will make it more generalized and it will be much easier during upstreaming.

I will now move this  PR back to `draft`, will work on `tensor` type support and move back to ready for your reviews/approval.

lib/TPP/Transforms/VectorContractTouKernels.cpp
arun-thmn: We may not need worry about it. 
The LLVM lowers the `vector.contract` to `xmm`, if this pass failed on contracts.

include/TPP/Passes.td
```diff
@@ -143,7 +143,16 @@ def BF16DotProduct : Pass<"vector-contract-to-bf16dp"> {
...
- }
+}
+
+
+def MicroKernels : Pass<"vector-contract-to-micro-kernels"> {
+   let summary = "Perform  lowering of vector contraction operation";
+   let description = [{Add commentMore actions
+     Perform  micro kernel lowering of vector contraction operation.
+   }];
+   let dependentDialects = [ "vector::VectorDialect", "scf::SCFDialect", "x86vector::X86VectorDialect" ];
```
adam-smnk: nit: needs at least arith extra

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
```diff
@@ -0,0 +1,1349 @@
+//===- VectorContractToMicroKernels.cpp -----------------------*- C++-*-===//
```
adam-smnk: nit: header formatting

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
```diff
@@ -0,0 +1,1349 @@
...
+    flag = flag && (c4 && c5);
+  }
+
+  return flag;
+}
+
+static memref::AllocOp createMask(Location loc, PatternRewriter &rewriter,
+                                  Value indexOp_c0, int64_t sizeFactor,
+                                  Type i32Type) {
+  auto intAttr = rewriter.getI32IntegerAttr(-65536);
```
adam-smnk: It is a bit mask right? I'd use hex representation in that case

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
```diff
@@ -0,0 +1,133 @@
...
+     linalg.batch_reduce_matmul ins(%arg0, %arg1 : memref<32x24x32xf32>, memref<32x32x64xf32>) outs(%arg2 : memref<24x64xf32>)
+   return 
+ }
+}
+
+// CHECK-LABEL:   func.func @optimal_register_allocation(
+// CHECK-SAME:                     %[[ARG0:.*]]: memref<32x24x32xf32>,
+// CHECK-SAME:                     %[[ARG1:.*]]: memref<32x32x64xf32>,
+// CHECK-SAME:                     %[[ARG2:.*]]: memref<24x64xf32>) {
+// CHECK:           %[[VAL_0:.*]] = arith.constant 16 : index
```
adam-smnk: Things like constants should use `CHECK-DAG` as they might get reordered randomly

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
```diff
@@ -0,0 +1,1349 @@
...
+                }
+
+                rewriterNewKForOp.create<scf::YieldOp>(locNewKForOp, evenFMAs);
+              });
+
+          rewriterNewReductionForOp.create<scf::YieldOp>(
+              locNewReductionForOp, newKForOp.getResults());
+        });
+
+    // Check the mlp pattern
```
adam-smnk, line 1089: A general question, is this fusion safe if add has multiple users etc.?

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
```diff
@@ -0,0 +1,133 @@
...
+// CHECK:               %[[VAL_24:.*]] = vector.load %[[VAL_14]]{{\[}}%[[VAL_3]], %[[VAL_4]]] : memref<3x32xf32, strided<[64, 1], offset: ?>>, vector<8xf32>
+// CHECK:               %[[VAL_25:.*]] = vector.load %[[VAL_14]]{{\[}}%[[VAL_8]], %[[VAL_4]]] : memref<3x32xf32, strided<[64, 1], offset: ?>>, vector<8xf32>
+// CHECK:               %[[VAL_26:.*]] = vector.load %[[VAL_14]]{{\[}}%[[VAL_2]], %[[VAL_4]]] : memref<3x32xf32, strided<[64, 1], offset: ?>>, vector<8xf32>
+// CHECK:               %[[VAL_27:.*]]:12 = scf.for %[[VAL_28:.*]] = %[[VAL_3]] to %[[VAL_7]] step %[[VAL_8]] iter_args(%[[VAL_29:.*]] = %[[VAL_15]], %[[VAL_30:.*]] = %[[VAL_16]], %[[VAL_31:.*]] = %[[VAL_17]], %[[VAL_32:.*]] = %[[VAL_18]], %[[VAL_33:.*]] = %[[VAL_19]], %[[VAL_34:.*]] = %[[VAL_20]], %[[VAL_35:.*]] = %[[VAL_21]], %[[VAL_36:.*]] = %[[VAL_22]], %[[VAL_37:.*]] = %[[VAL_23]], %[[VAL_38:.*]] = %[[VAL_24]], %[[VAL_39:.*]] = %[[VAL_25]], %[[VAL_40:.*]] = %[[VAL_26]]) -> (vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>, vector<8xf32>) {
```
adam-smnk: nit: line breaks with `CHECK-SAME` for readability

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
```diff
@@ -0,0 +1,133 @@
...
+// RUN: tpp-opt %s --tile-brgemm-linalg="registerBlocking=3,4,2" --loop-invariant-code-motion --vectorization-pass --hoist-vector-transfer --vector-contract-to-ukernels  --split-input-file  | FileCheck -check-prefix=CHECKBF162 %s
+
+module {
+ func.func @optimal_register_allocation(%arg0: memref<32x24x32xf32>, %arg1: memref<32x32x64xf32>, %arg2: memref<24x64xf32>) {
+     linalg.batch_reduce_matmul ins(%arg0, %arg1 : memref<32x24x32xf32>, memref<32x32x64xf32>) outs(%arg2 : memref<24x64xf32>)
+   return 
+ }
+}
+
+// CHECK-LABEL:   func.func @optimal_register_allocation(
```
adam-smnk: Do we really need super verbose testing like that?
Integration test already ensures correctness. This breaks and becomes unmaintainable after any minor change.
I'd suggest focusing on a small subset e.g., a few loads, their fmas, and writes.

include/TPP/Passes.td
arun-thmn: Okay. Added `arith` dialect.

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
arun-thmn: I hope corrected it now.

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
arun-thmn: Yes.
Thanks updated to hexadecimal (`0xFFFF0000`).

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: Now, simplified the unit-tests to check only for the `vector.fma` sequence.

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
arun-thmn: At the end of the pass, we don't remove the `addop`. We remove the `vector.transfer_write` to `C` matrix.
If there is any other use for `addOp`, it remains after the pass.

arun-thmn: Simplified the unit-test to check only for `vector.fma`

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: Simplified the unit-test to check only for `vector.fma`.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: Simplified the unit-test to check only for `vector.fma`.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
adam-smnk, line 1: These tests are closer to integration tests due to multiple passes and rewrites being performed.
`CHECK1` and `CHECK2` indicate that the rewrite should not happen but I have no clue why and/or due to which pass.

Could you add unit tests for the pass itself? For both gemm and mlp patterns.

test/BF16/Integration/uKernels/vector-contract-to-ukernels-bf16.mlir
adam-smnk, line 1: I think there's sth flaky in the rewrite.
If I take `@mlp_bf16` from this test and run:
```
tpp-opt --tile-brgemm-linalg="registerBlocking=4,32,2" --loop-invariant-code-motion --vectorization-pass --hoist-vector-transfer --vector-contract-to-micro-kernels -cleanup
```
I get different output compared to piping the last stage to a separate rewrite stage:
```
tpp-opt --tile-brgemm-linalg="registerBlocking=4,32,2" --loop-invariant-code-motion --vectorization-pass --hoist-vector-transfer | tpp-opt --vector-contract-to-micro-kernels -cleanup
```

arun-thmn: Interesting.
An unwanted check I did, `addOp` followed by `maxOp` for fusion. When we do via pipe `maxOp` comes first in the iterator followed by `addOp`.

Removed the check, now it's working fine.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: `CHECK1` and `CHECK2` are for checking the tile sizes. If the `k` tile size greater than `1` or `N` tile size less than `8` the rewriter will skip the lowering.

adam-smnk: I get that's the intention but the lowering might be skipped because: it didn't tile at all, sth else (like hoisting) failed and pattern doesn't match, something different in the matcher failed etc.

It'd be better to have minimal test cases with input expected by the pass (`vector.contract` + whatever else matcher needs) without the need for multiple passes to preprocess IR.
This both makes it more obvious what is tested and decouples tests and passes.

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
adam-smnk: I don't think that's safe due to aliasing etc. but let's ignore it for now.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: Sure. Have updated the unit-tests such that `tpp-opt` is passed only with `vector-contact-to-micro-kernels` pass.
Have added simple gemm and mlp benchmarks. For `mlp`, to check fusion i tried a way that should be no `vector.transfer_read` and `vector.transfer_write`.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
```diff
@@ -0,0 +1,149 @@
+// RUN: tpp-opt %s --vector-contract-to-micro-kernels  --split-input-file  | FileCheck -check-prefix=CHECK %s
+// RUN: tpp-opt %s --vector-contract-to-micro-kernels  --split-input-file  | FileCheck -check-prefix=CHECK1 %s
+// RUN: tpp-opt %s --vector-contract-to-micro-kernels  --split-input-file  | FileCheck -check-prefix=CHECKBF16 %s
```
adam-smnk: nit: now these can be collapsed and use default `CHECK`

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
```diff
@@ -0,0 +1,149 @@
...
+    %5 = vector.transfer_read %arg2[%c0, %c0], %cst_0 {in_bounds = [true, true]} : memref<32x32xbf16>, vector<32x32xbf16>
+    %6 = arith.maximumf %5, %cst : vector<32x32xbf16>
+    vector.transfer_write %6, %arg2[%c0, %c0] {in_bounds = [true, true]} : vector<32x32xbf16>, memref<32x32xbf16>
+    return
+  }
+}
+
+
+// CHECKBF16-LABEL: func.func @mlp_bf16
+// CHECKBF16-COUNT-16: vector.fma
```
adam-smnk: Won't it produce like 8 FMAs?

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
arun-thmn: Okay.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
adam-smnk: Does this test even run in CI?
If not, could we force codegen through pass option or DLTI? We should be able to validate rewrite logic regardless of the target as long as we don't need to execute it.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
```diff
@@ -0,0 +1,149 @@
...
+        vector.transfer_write %1, %subview[%c0, %c0] {in_bounds = [true, true]} : vector<3x32xf32>, memref<3x32xf32, strided<[64, 1], offset: ?>>
+      }
+    }
+    return
+  }
+}
+
+// CHECK-LABEL:   func.func @optimal_register_allocation_gemm
+// CHECK: vector.fma
+// CHECK-NEXT: vector.fma
```
adam-smnk: nit: I'd check also shape in one FMA as now these look the same for both avx2 and avx512 but I guess SIMD lengths differ

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: Ahh correct. Sorry for missing these minor things.
Now, we have only one `CHECK`

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: Okay, nice idea.
Added `vector.fma{{.*}}vector<16xf32>` for avx512 and  `vector.fma{{.*}}vector<8xf32>` for avx2.

test/Passes/uKernels/avx2/pass-vector-contract-to-FMAs.mlir
arun-thmn: As the type is `bf16`, for `avx2` we do 8 `odd` FMAs + 8 `even` FMAs. So, totally 16 FMAs.

The `avx2` test runs on machines like `ARL` or `ADL`, as this test is meant only for those machines. I checked this test on those machines and are passing. I'm not sure whether this test runs on `CI`. We can add only `avx2` supported machine to `CI` or I can do a very minor change to my `rewrite` so this test can run. Let me know please.

adam-smnk: Sorry I confused myself a bit.
Double checked the flow and seems good.

PR #1058, opened 2025-06-16
Vectorizer - register blocking

adam-smnk:
Tile and fuse driver pass designed primarily for register blocking that is splitting linalg ops into smaller workloads (nanokernels) suitable for vectorization and further target-aware code generation.

The pass uses DLTI specification for tile size selection:
  `DLTI = "CPU"<"reg_blocks" = [bm,bn,bk]">`
Currently, the pass focuses on tiling contractions and greedily fusing their elementwise consumers.

Static shapes are required to allow for later vectorization. The tiling driver ensures that through loop peeling which results in up to two nanokernels per tiled dimensions when the tiles do not perfectly divide operands' shapes.

### DIFF
--- a/lib/TPP/Transforms/RegisterBlocking.cpp
+++ b/lib/TPP/Transforms/RegisterBlocking.cpp
@@ -6,6 +6,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "TPP/Passes.h"
 #include "TPP/Transforms/Utils/VNNIUtils.h"
 
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
@@ -50,7 +51,7 @@ static SmallVector<IntType> extractVector(ArrayAttr arrayAttr) {
       [](IntegerAttr attr) { return static_cast<IntType>(attr.getInt()); }));
 }
 
-// Returns register blocks for innermost dims: [M, N, K]
+// Returns register blocks for the innermost dims: [M, N, K]
 static SmallVector<int64_t> getRegisterBlocks(Operation *op) {
   auto res = dlti::query(op, {"CPU", "reg_blocks"});
   if (failed(res))
@@ -127,6 +128,9 @@ static void peelTiledLoops(ArrayRef<Operation *> loops,
 struct TileAndFuseContraction : OpInterfaceRewritePattern<linalg::LinalgOp> {
   using OpInterfaceRewritePattern<linalg::LinalgOp>::OpInterfaceRewritePattern;
 
+  TileAndFuseContraction(MLIRContext *ctx, tpp::RegisterBlockingOptions options)
+      : OpInterfaceRewritePattern<linalg::LinalgOp>(ctx), options(options) {}
+
   LogicalResult matchAndRewrite(linalg::LinalgOp matmulOp,
                                 PatternRewriter &rewriter) const override {
     if (!matmulOp.hasPureTensorSemantics())
@@ -157,7 +161,9 @@ struct TileAndFuseContraction : OpInterfaceRewritePattern<linalg::LinalgOp> {
       return rewriter.notifyMatchFailure(
           matmulOp, "expects at only 2 parallel non-batch dimensions");
 
-    SmallVector<int64_t> regBlocks = getRegisterBlocks(matmulOp);
+    SmallVector<int64_t> regBlocks = options.blocks;
+    if (regBlocks.empty())
+      regBlocks = getRegisterBlocks(matmulOp);
     if (regBlocks.size() != 3)
       return rewriter.notifyMatchFailure(matmulOp, "invalid register blocking");
 
@@ -259,6 +265,9 @@ struct TileAndFuseContraction : OpInterfaceRewritePattern<linalg::LinalgOp> {
 
     return success();
   }
+
+private:
+  tpp::RegisterBlockingOptions options;
 };
 
 // Tile reduction dimensions of previously tile-and-fused contraction ops.
@@ -274,6 +283,10 @@ struct TileAndFuseContraction : OpInterfaceRewritePattern<linalg::LinalgOp> {
 struct TileContractionReductionDims
     : OpInterfaceRewritePattern<linalg::LinalgOp> {
   using OpInterfaceRewritePattern<linalg::LinalgOp>::OpInterfaceRewritePattern;
+
+  TileContractionReductionDims(MLIRContext *ctx,
+                               tpp::RegisterBlockingOptions options)
+      : OpInterfaceRewritePattern<linalg::LinalgOp>(ctx), options(options) {}
 
   LogicalResult matchAndRewrite(linalg::LinalgOp matmulOp,
                                 PatternRewriter &rewriter) const override {
@@ -299,7 +312,9 @@ struct TileContractionReductionDims
       return rewriter.notifyMatchFailure(
           matmulOp, "expects at only 2 parallel non-batch dimensions");
 
-    SmallVector<int64_t> regBlocks = getRegisterBlocks(matmulOp);
+    SmallVector<int64_t> regBlocks = options.blocks;
+    if (regBlocks.empty())
+      regBlocks = getRegisterBlocks(matmulOp);
     if (regBlocks.size() != 3)
       return rewriter.notifyMatchFailure(matmulOp, "invalid register blocking");
 
@@ -377,6 +392,9 @@ struct TileContractionReductionDims
 
     return success();
   }
+
+private:
+  tpp::RegisterBlockingOptions options;
 };
 
 struct RegisterBlocking
@@ -387,13 +405,16 @@ struct RegisterBlocking
     auto *ctx = &getContext();
     auto op = getOperation();
 
+    tpp::RegisterBlockingOptions options;
+    options.blocks = SmallVector<int64_t>{*blocks};
+
     GreedyRewriteConfig config;
     config.setStrictness(GreedyRewriteStrictness::ExistingOps);
 
     // First, tile and fuse contraction along parallel dimensions.
     {
       RewritePatternSet patterns(ctx);
-      patterns.add<TileAndFuseContraction>(ctx);
+      patterns.add<TileAndFuseContraction>(ctx, options);
       if (failed(applyPatternsGreedily(op, std::move(patterns), config)))
         return signalPassFailure();
     }
@@ -419,7 +440,7 @@ struct RegisterBlocking
     // Then tile reduction dimensions.
     {
       RewritePatternSet patterns(ctx);
-      patterns.add<TileContractionReductionDims>(ctx);
+      patterns.add<TileContractionReductionDims>(ctx, options);
       if (failed(applyPatternsGreedily(op, std::move(patterns), config)))
         return signalPassFailure();
     }
--- a/lib/TPP/Transforms/RegisterBlocking.cpp
+++ b/lib/TPP/Transforms/RegisterBlocking.cpp
@@ -160,7 +160,7 @@ struct TileAndFuseContraction : OpInterfaceRewritePattern<linalg::LinalgOp> {
     // BRGEMM cases.
     if (dims->m.size() != 1 || dims->n.size() != 1)
       return rewriter.notifyMatchFailure(
-          contractOp, "expects at only 2 parallel non-batch dimensions");
+          contractOp, "expects only 2 parallel (M and N) non-batch dimensions");
 
     SmallVector<int64_t> regBlocks = options.blocks;
     if (regBlocks.empty())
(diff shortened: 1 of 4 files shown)


adam-smnk:
The first of a few passes that will come together as a x86 vectorizer.
The pass targets linalg on tensors with intended lowering position somewhere after initial mapping (packing etc.), and tiling for parallelism/microkernel sizes.

Exact nanokernel size selection can be performed through different strategies.
The current simple approach of tiling and peeling enables tile size exploration. However, certain configurations will result in suboptimal nanokernels. For example, for a GEMM dimension `M=16` and a tile `m=5` will result in a split `5-5-5-1` instead of potentially better `5-5-6`.
Such optimizations are out-of-scope but the nanokernel partitioning can be improved in the future by extra pre-/post-processing steps around the core tile-and-fuse rewrite.

It is also assumed that the `N` (SIMD) dimension is generally divisible by the SIMD size (8, 16 etc.).
Non-divisible dimension are still tiled and peeled, however, such cases will require later postprocessing to enable SIMD operations e.g., padding or masked reads/writes. Otherwise, the peeled tail nanokernel will have to default to an alternative (naive) lowering.
Such shape improvements are expected to be performed independently before and/or after register blocking.

rolfmorel:
Just a general note as it came up when skimming through the PR: it might be nice to expose _configurables_ like blocking factors here as options on the pass as well.

In particular, DLTI and the transform ops now facilitate the following:
```mlir
//%block_factors = transform.dlti.query ["CPU","reg_blocks"] at %payload
%block_factors = transform.dlti.query ["CPU", "blocking_model", "reg_blocks"] at %payload // or commenting this line and uncommenting the previous line
transform.apply_registered_pass "register-blocking" with options = { "block-factors" = %block_factors }
```
This gives a bit more flexibility and quicker iteration, e.g. in case I want to switch between values provided as static attributes ( e.g. a `#dlti.map` that handles `["CPU", "reg_blocks"]`) and a query handler that is supposed to be a bit more involved, e.g. a `#imex.cpu_blocking_model` (in a map at `["CPU", "blocking_model"]`) responding to `["reg_blocks"]` queries.

It might still make sense for passes to do a `dlti::query(...)` themselves, at fixed keys, _if_ we know that those options are supposed to be payload IR dependent. And in particular if we expect the pass to _not_ be used in a schedule, as then it's the only way to make the options payload dependent.

Finally, I believe the pass options should take precedence over the `dlti::query`s. At least for this pass it would be necessary in order to allow the pattern of `%opt = transform.dlti.query ...; transform.apply_registered_pass ... with options = { ... = %opt}` to be useful.

shahidact:
A general note, If I understand correctly this patch does multiple rewrites to achieve register blocking namely tiling on parallel dimensions, peeling for non-perfectly divisible sizes, fusion of producer with elementwise fusible ops and lastly register blocking. Can the variables/comments be updated such that to distinguish between register blocking and tiling wherever applicable?

adam-smnk:
One question that might come up - why not reuse existing `tile-consumer-fuse-producers`?
It does a lot more pre- and post-processing and I've found it a bit messy to use here. They're both tiling driver but this one is leaner and simpler (plus has peeling).

shahidact:
That make sense now. I was wondering more from the perspective of having these rewrites as tools for our future experiments and eventual upstreaming. For example, here peeling is very useful tool in general.
Moreover, the separation/abstraction/representation of VNNI layout related logic is something we should consider too.
Do we have any plan for that?

test/Passes/register-blocking.mlir
```diff
@@ -0,0 +1,476 @@
...
+// CHECK-LABEL: @negative_tile_memref
+// CHECK-NOT: scf.for
+// CHECK:     linalg.matmul
+
+// -----
+
+!matA = memref<256x?xf32>
+!matB = memref<?x128xf32>
+!matC = memref<256x128xf32>
+func.func @negative_tile_dynamic_shapes(
```
arun-thmn, line 356: Does `tensor` type supports dynamic shapes?

adam-smnk:
> That make sense now. I was wondering more from the perspective of having these rewrites as tools for our future experiments and eventual upstreaming. For example, here peeling is very useful tool in general.

The rewrites themselves are all powered by existing upstream APIs (tile using interface, scf transforms). In that sense, it proves that available tools are sufficient and generic enough for us to steer them the way we want.

Majority of this pass is the “driver logic” defining when and where to apply these rewrites - picking sizes, controlling what to fuse and peel.
These decisions are heavily opinionated. Therefore, not very suitable for upstreaming on their own. However, there’s some value in showing how to use the upstream “batteries” and proving they ate good enough.

> Moreover, the separation/abstraction/representation of VNNI layout related logic is something we should consider too.

I think that’s sth we could contribute/define. An upstream definition and representation of what VNNI is with suitable utilities like ‘isVnni’ etc.

> Do we have any plan for that?

Overall plan would be to:
- Show within tpp-mlir that vectorization within mlir is possible - largely already done with all the performance results
- Drive vectorization in a progressive and reusable manner - goal of this x86 vectorizer concept - ideally it should reach parity with the above results
- Identify missing tools upstream - also part of this PoC, currently high-level rewrites seems to be covered (which is great), we’ll probably identify missing bits further in the lowering

I’d say we’re still at the investigation stage. But the driver itself (how to guide rewrites) would be suitable for the lighthouse projects once it kicks off.
And I think there will be opportunities to controbute upstream smaller bits: VNNI definition, extra x86 ops, maybe some patterns at vector ops level.

test/Passes/register-blocking.mlir
adam-smnk: Tensor type yes.
However, in general vector doesn’t and you can’t vectorize ops with dynamic shapes.

For now, I disabled dynamic shape support for simplicity.
Of course, it could be enabled and various cases can be handled. Dynamic reduction and batch dims are largely fine. I just don’t wanna worry about all these variants just yet.

If you have any use cases for them, dynamic shapes could be enabled.

arun-thmn: Okay. Shall we a add a negative test-case with `tensor` type dynamic shape and check the vectorizer doesn't vectorize it.

adam-smnk: Sure, I’ll add a test case to the vectorize pass.

test/Passes/register-blocking.mlir
```diff
@@ -0,0 +1,476 @@
...
+!matC = tensor<256x128xf32>
+func.func @batch_reduce_matmul(
+  %arg0: !matA, %arg1: !matB, %arg2: !matC) -> !matC
+  attributes {
+    dlti.target_system_spec = #dlti.target_system_spec<
+    "CPU" = #dlti.target_device_spec<
+      #dlti.dl_entry<"reg_blocks", [2, 32, 4]>
+  >>}
+{
+  %0 = linalg.batch_reduce_matmul ins(%arg0, %arg1 : !matA, !matB)
```
arun-thmn, line 83: What happens, if I give non-divisible tile sizes for input AxB matrix.
For example, my matrix are:
`matA = tensor<16x256x512xf32>`
`matB = tensor<16x512x128xf32>`
`matC = tensor<256x128xf32>`

and the `reg_blocks` is `[6, 64, 1]`.  `256`  is not divisible by `6`. Do we do making kind of stuff for that?

adam-smnk: It will compute as many `6` sized tiles as possible and a separate tail tile i.e., tiles 6x64 up to 252 and one tile 4x64 afterwards.

See `@peel_remainder_block` case.

test/Passes/register-blocking.mlir
adam-smnk: Added

test/Passes/register-blocking.mlir
arun-thmn: Ahh, nice. Got it.

shahidact:
> These decisions are heavily opinionated. Therefore, not very suitable for upstreaming on their own. However, there’s some value in showing how to use the upstream “batteries” and proving they ate good enough.

Yes, it is strong proof point, however, how do we show it without upstreaming the driver? Am I missing something? My assumption was that we are going to upstream it.

adam-smnk:
> how do we show it without upstreaming the driver? Am I missing something? My assumption was that we are going to upstream it.

How exactly is an open question. We can upstream mechanical bits like conversion patterns from vector to x86dialect etc.
The driver is useful for our downstream experiments and running various workloads.

For an upstream end-to-end demonstrator, I see a few options:
- a test pass/schedule - a simple input (most likely nicely divisible, static shapes etc.) with hardcoded decisions (tile sizes, unroll shapes) - this acts as a demo from linalg to x86 or llvmir
- mechanical last mile lowering (see `gpu-lower-to-nvvm-pipeline` and its tests) - we start with nicely shaped pre-tiled, potentially pre-unrolled, vector ops and lower them to llvmir - this shows what IR you need first (how to get there left as an exercise for the reader) and how to "drive it home"

Then maybe lighthouse could have an opinionated driver. Although, a schedule (akin to test schedule point above) might be better.

@rengolin any better ideas?

lib/TPP/Transforms/RegisterBlocking.cpp
```diff
@@ -0,0 +1,458 @@
...
+  }
+  return eltWiseTiles;
+}
+
+// Apply loop peeling to split tail iterations and allow for
+// canonicalization to ensure all blocked ops operate on static values.
+// Peeling is applied in reverse order from the innermost loop to ensure
+// that all tiling loops are affected.
+//
+// Result is ignored as peeling can fail when tiling cleanly divides
```
shahidact, line 107: Does it have potential to increase compile time significantly?

test/Passes/register-blocking.mlir
```diff
@@ -0,0 +1,476 @@
+// RUN: tpp-opt %s -register-blocking -split-input-file | FileCheck %s
```
shahidact, line 1: Hope this also work for `linalg.contract`. Pls add few tests for it.

lib/TPP/Transforms/RegisterBlocking.cpp
adam-smnk: You mean always invoking peeling?
Minimal, underlying logic has early exits if there's nothing to do.

test/Passes/register-blocking.mlir
adam-smnk: Added a case for completeness.

In general the matching logic is pretty universal and works on generic linalg op interface. Effectively, you can think of it as every linalg op is a `linalg.generic` - you have access to all its fields and can perform all usual checks.
The tiling again works on abstract `TilingInterface` which all linalg ops we care about here should be implementing.

But if we find that anything breaks in these generic tools then there's a coverage gap upstream.

lib/TPP/Transforms/RegisterBlocking.cpp
shahidact: > You mean always invoking peeling?

Yes, I thought if there could be a check if peeling is required before invoking it.

lib/TPP/Transforms/RegisterBlocking.cpp
```diff
@@ -0,0 +1,458 @@
...
+    FailureOr<linalg::ContractionDimensions> dims =
+        linalg::inferContractionDims(contractOp);
+    if (failed(dims))
+      return rewriter.notifyMatchFailure(contractOp, "not a contraction");
+
+    // Matching stil requires parallel dimensions to allow for VNNI detection.
+    // TODO: Relax constraints.
+    if (dims->m.size() != 1 || dims->n.size() != 1)
+      return rewriter.notifyMatchFailure(
+          contractOp, "expects at only 2 parallel non-batch dimensions");
```
shahidact: May be a typo in message here.

lib/TPP/Transforms/RegisterBlocking.cpp
```diff
@@ -0,0 +1,458 @@
...
+  auto vals = llvm::dyn_cast<ArrayAttr>(*res);
+  if (!vals)
+    return {};
+  return extractVector<int64_t>(vals);
+}
+
+// Returns position of a dimension corresponding to the given iteration map
+// and an iterator.
+static std::optional<unsigned>
+mapIteratorToDim(PatternRewriter &rewriter, AffineMap map, unsigned iterPos) {
```
shahidact: I would rename it to "mapIteratorToDimPos" to make it sound consistent with what it is returning.

arun-thmn:
Thanks @adam-smnk. I'm ok to get it merged.

adam-smnk:
Thanks for the feedback 👍 
If there are no more comments, could I get the green mark?

adam-smnk:
> @adam-smnk I don't have permission to approve a `PR`, I guess. Not able to locate the approve button.

You should be able to under `Files changed -> Review changes -> Approve (radio button) -> Submit review`.
If that doesn't work then we have an issue with repo configuration.

arun-thmn (approved):
Thanks, @adam-smnk. LGTM!.

PR #1060, opened 2025-06-17
Vectorizer - linalg vectorize

adam-smnk:
Vectorization driver tuned toward further vector ops lowering. It primarily targets linalg ops on tensors.

### DIFF
--- a/test/Passes/linalg-vectorize.mlir
+++ b/test/Passes/linalg-vectorize.mlir
@@ -35,7 +35,7 @@ func.func @vectorize_eltwise(
 #map = affine_map<(d0, d1, d2, d3) -> (d0, d2, d3)>
 #map1 = affine_map<(d0, d1, d2, d3) -> (d2, d1, d3)>
 #map2 = affine_map<(d0, d1, d2, d3) -> (d0, d1)>
-func.func @vectorize_contract_mixed_precision(
+func.func @vectorize_contract_mixed_precision_float(
     %arg0: tensor<256x128x2xbf16>, %arg1: tensor<128x256x2xbf16>,
     %arg2: tensor<256x256xf32>) -> tensor<256x256xf32> {
   %0 = linalg.contract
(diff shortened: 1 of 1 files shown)


arun-thmn:
@adam-smnk Thanks. It looks good and straight forward. Two doubts:

1. Does it vectorizes `memref` types? if not should we add a check on the `tensor` or `memref` type. if yes shall we add one test-case with `memref` type.
2. The existing vectorization pass: https://github.com/libxsmm/tpp-mlir/blob/main/lib/TPP/Transforms/Vectorization.cpp does `vector-contract` lowering for `memref` types. It does checks for the `Iterator type` sizes. Hope those check are not needed now.

lib/TPP/Transforms/LinalgVectorize.cpp
```diff
@@ -0,0 +1,86 @@
...
+    // Disable insert vectorization for now and allow bufferization to fold it
+    // into subview in many cases.
+    if (isa<tensor::InsertSliceOp>(op))
+      return rewriter.notifyMatchFailure(op,
+                                         "Insert slice vectorization disabled");
+    return linalg::vectorize(rewriter, op);
+  }
+};
+
+// Based on 'vectorize_children_and_apply_patterns' transform op.
```
rolfmorel, line 60: Looks like this pass is a candidate for what @rengolin has asked for a couple times: passes that just wrap around a small schedule. Might be worth discussing if it is straightforward in this case or not.

adam-smnk: Not sure how’d that look like but sounds like worth exploring.

test/Passes/linalg-vectorize.mlir
```diff
@@ -0,0 +1,134 @@
...
+    %arg0: tensor<256x128x2xbf16>, %arg1: tensor<128x256x2xbf16>,
+    %arg2: tensor<256x256xf32>) -> tensor<256x256xf32> {
+  %0 = linalg.contract
+    indexing_maps = [#map, #map1, #map2]
+    ins(%arg0, %arg1 : tensor<256x128x2xbf16>, tensor<128x256x2xbf16>)
+    outs(%arg2 : tensor<256x256xf32>) -> tensor<256x256xf32>
+  return %0 : tensor<256x256xf32>
+}
+
+// Ensure that mixed precision contraction vectorizes cleanly
```
shahidact, line 48: Could you Pls include i8 mixed precision test after tiling and vnni packing for linalg.generic matmul pattern.

shahidact (commented):
Overall, this looks good.

test/Passes/linalg-vectorize.mlir
adam-smnk: Added 👍

shahidact (approved):
Thanks, LGTM.

PR #1062, opened 2025-06-24
Vectorizer - register unroll

adam-smnk:
Vector unroll driver pass designed to split operations into smaller hardware-compatible shapes in preparation for target-specific code generation.
Additionally, VNNI utility is generalized to be applicable also to vector contractions.

The register unrolling is intended to target nanokernel-shaped ops and splitting them into smaller sub-ops that can fit into physical registers.
Currently, the driver focuses on unrolling contractions and their data trasfer ops when possible.

Note that at this point, ops might still be in higher abstraction than target-specific operations. Thus, further specialized lowering is expected to be applied afterwards.

### DIFF
--- a/lib/TPP/Transforms/RegisterUnroll.cpp
+++ b/lib/TPP/Transforms/RegisterUnroll.cpp
@@ -207,15 +207,16 @@ static std::optional<SmallVector<int64_t>> getVectorShape(Operation *op) {
 // hardware-compatible shapes in preparation for target-specific code
 // generation.
 //
-// First, unroll shapes are assigned and propagated thorughout the graph as
+// First, unroll shapes are assigned and propagated throughout the graph as
 // attribute annotations. Then the unroll driver consumes these annotations
 // to apply rewrites.
+//
 // For example - in prepation for further FMA lowering:
 // ```
 //   vector.contract {unroll_shape = [1, 16, 1]}
 //     : vector<4x1xf32>, vector<1x32xf32> into vector<4x32xf32>
 // ```
-// get split into eight smaller contractions:
+// the contractions is split into eight smaller ops:
 // ```
 //   vector.contract : vector<1x1xf32>, vector<1x16xf32> into vector<1x16xf32>
 // ```


adam-smnk:
To add more context, we start at vector ops at nanokernel level after register blocking and vectorization. These nanokernels define individual units of computation maximizing data reuse in registers.

Unrolling can now be used to split these computations into smaller ops that are register sized (more or less, some further unrolling can be left to LLVM backends) ops i.e., split <4x16xf32> outer products into separate <1x16xf32> `fma`-sized contractions or divide larger block into multiple `amx`-sized ops.

The advantages of this approach over 1-to-N conversion:
- better upstream tools reuse - no need to reimplement common logic like generic validation and offset computations for unrolled sub-ops
- separate and simpler further lowering - once operations have been preprocessed and are already shaped as their intended target ops further rewrites to intrinsics or amx operations can be done on 1-to-1 basis with simple matching logic

This PoC driver focuses on contractions as this is primary enabler for efficient codegen while simpler operations, like elementwise computations, can be left to LLVM to handle. Ideally, one would perform a separate layout assignment and propagation step to pick optimal sizes for various operations as well as resolve possible conflicts e.g., multi-user scenarios. As an example, upstream XeGPU layout propagation passes can serve as inspiration.

One current drawback of limited unrolling is leaving N-dim `vector.transfer` ops which have to lower through `vector-to-scf` and cause stack allocation. However, the performance hit should be limited and I would suggest focusing on enabling more lowering patterns before building complex vector unrolling framework.

include/TPP/Passes.td
```diff
@@ -664,4 +664,19 @@ def LinalgVectorize : Pass<"linalg-vectorize"> {
...
+  let summary = "Unroll ops for register sizes.";
+  let description = [{
+    Unroll vector ops to bring them to hardware-compatible shapes.
+  }];
+  let dependentDialects = ["vector::VectorDialect",
+                           "arith::ArithDialect",
+                           "math::MathDialect",
+                           "affine::AffineDialect"];
+  let options = [
+    ListOption<"gemmShape", "gemm-shape", "int64_t",
```
shahidact: Is it the shape of unroll factors for each dimension?

adam-smnk: It's only for the innermost dimensions same as in register blocking

adam-smnk: Any other dims like batch or batch-reduce are unrolled by 1. Although ideally they shouldn't be present at this point of lowering.

shahidact: What about this `reg_gemm_unroll`?

adam-smnk: It is equivalent to this option just as an DLTI attribute.
Maybe their names should be more similar 😄

shahidact (commented):
Pls add an example description about this pass as a comment.

include/TPP/Passes.td
adam-smnk: Renamed pass option to be closer to the DLTI attribute.
Ideally, all these unroll choices should be derived from target information and assigned dynamically e.g., some GEMMs go to AMX and need their shapes while FMA need different shapes yet.

For now, static choice it is 🙂

lib/TPP/Transforms/RegisterUnroll.cpp
```diff
@@ -203,6 +203,22 @@ static std::optional<SmallVector<int64_t>> getVectorShape(Operation *op) {
   return SmallVector<int64_t>(unrollAttr.asArrayRef());
 }
 
+// Vector unroll driver pass designed to split operations into smaller
+// hardware-compatible shapes in preparation for target-specific code
+// generation.
+//
+// First, unroll shapes are assigned and propagated thorughout the graph as
```
shahidact: So, it will unroll even the accumulator/result read-write. Are we not planning hoisting these reads-writes? I hope it is planned and there will not be any interference with the unrolling.

adam-smnk: In general, you can hoist either before or after unrolling. And ideally we would want to unroll reads and writes too.

Now, because the current unrolling strategy is really naive (only contraction and its immediate transfers), I've found it a bit better to first hoist to cancel out write-read pairs and unroll afterwards.
In the future with better unrolling driver, it might make no difference when you do which or unrolling before hoisting could be slightly better.

shahidact: > I've found it a bit better to first hoist to cancel out write-read pairs and unroll afterwards.

Oh that's what i was expecting, maybe I missed, where it is being done?

adam-smnk: It is independent of this pass.
The example of all these passes coming together can be found here:
https://github.com/libxsmm/tpp-mlir/blob/x86-vectorizer/lib/TPP/PassBundles/X86Vectorizer.cpp#L82

I'll probably add this bundle once I have all needed pieces. I still need to think how to integrate it into the whole pipeline but in general I've been using it after `tpp-mapping`.

adam-smnk:
Bump - any more comments?

test/Passes/register-unroll.mlir
```diff
@@ -0,0 +1,257 @@
...
+!vecC = vector<4x64xf32>
+#map = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+func.func @vector_matmul(
+  %arg0: !vecA, %arg1: !vecB, %arg2: !vecC) -> !vecC
+  attributes {
+    dlti.target_system_spec = #dlti.target_system_spec<
+    "CPU" = #dlti.target_device_spec<
+      #dlti.dl_entry<"reg_gemm_unroll", [2, 32, 1]>
```
arun-thmn, line 14: `"reg_gemm_unroll", [2, 32, 1]` - We have to divide the dims to get the number of unrolls. Does the `upsteam` pass requires this user interface? If yes, please ignore this comment.

Else, can we just directly give the number of unroll like, `[2, 2, 2]`.  Earlier in my `tiling` pass, I used the kind of earlier `[2, 32, 1]` option, where the user felt uncomfortable as they have to divide and get the tile sizes.

test/Passes/register-unroll.mlir
```diff
@@ -0,0 +1,257 @@
...
+!vecC = vector<3x4x64xf32>
+#map = affine_map<(d0, d1, d2, d3) -> (d0, d1, d3)>
+#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d3, d2)>
+#map2 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2)>
+func.func @vector_batch_matmul(
+  %arg0: !vecA, %arg1: !vecB, %arg2: !vecC) -> !vecC
+  attributes {
+    dlti.target_system_spec = #dlti.target_system_spec<
+    "CPU" = #dlti.target_device_spec<
+      #dlti.dl_entry<"reg_gemm_unroll", [4, 64, 1]>
```
arun-thmn, line 44: Suppose, If I give `[8,64,1]` option, `4/8` is the unroll for `m` dims. Does the upstream pass throws `failure()` or the `m` dims is not unrolled.

adam-smnk: In this case, `failure`.
Unlike tiling, unrolling requires ops to be exactly divisible by the selected unrolled shapes.
However, AFAIK it is in line with typical usage as unrolling splits the ops into physical register sizes which are typically static. Generally, shapes should've been preprocessed earlier by padding, masking etc.

test/Passes/register-unroll.mlir
adam-smnk: It's not like loop unrolling where you specify how many times you want to unroll.

Vector unrolling basically splits ops into smaller shapes and that's what you define here. The unroll shape is your desired op size regardless whether you start from `8x32` or `128x128` shape. An op will get unrolled as many times as needed to reach the target unroll shape.
Also, the chosen shape is usually derived by the hardware sizes, thus, it is more suitable to do it in this direction.

arun-thmn: Okay, fine then.

arun-thmn (approved):
Thanks @adam-smnk LGTM.

PR #1063, opened 2025-06-25
Vectorizer - convert vector to x86

adam-smnk:
A conversion pass that focuses on lowering hardware-sized vector ops into efficient target-aware x86 implementations e.g., sequences of intrinsic ops from x86 dialects.

At this point, the vector operations' shapes are expected to be hardware-compatible. However, their exact implementation might depend on available target extensions etc.
The pass is intended to be positioned at late vectorization stage and to target individual operations or small subgraphs for rewriting.


include/TPP/Passes.td
```diff
@@ -669,4 +669,16 @@ def VectorDropUnitDims : Pass<"vector-drop-unit-dims"> {
...
 
+def ConvertVectorToX86 : Pass<"convert-vector-to-x86"> {
+  let summary = "Convert vector ops to x86 sequences.";
+  let description = [{
+    Lowers generic vector ops into specialized x86
+    sequences of operations.
+  }];
+  let dependentDialects = ["vector::VectorDialect",
+                           "x86vector::X86VectorDialect",
+                           "amx::AMXDialect",
```
arun-thmn, line 680: Why we needed `amx` Dialect.

adam-smnk: The end goal is to collect here patterns for all different lowerings through plain vector, x86vector, and amx.
Just added here for conceptual completness but it's not really needed right now.

arun-thmn:
Seems ok for me. 
Targeting the lowering of `contract` to  `fma` for unit dims. With `TODO` for later support.

arun-thmn (approved):

PR #1064, opened 2025-06-25
[Transform][Python] Introduce transform.ffi.callback

rolfmorel:
This transform op allows for invoking an external handler routine which can, for example, implement an analysis or transform in whichever way - and programming language - is convenient.

This PR provides full facilities for registering callbacks from Python and has test cases to show how MLIR's Python bindings can be used to easily implement at least some transforms.


python/TPPDialects.cpp
```diff
@@ -66,6 +74,82 @@ NB_MODULE(_tppDialects, m) {
       },
       "registry");
 
+  transformTuneModule.def(
+      "register_callback_handler", [&](nb::callable callable) {
+        callback_handler =
+            nb::steal(callable); // TODO: should we ever release this?
```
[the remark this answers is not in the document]
rolfmorel: Yeah, yeah.

rolfmorel:
```
Failed Tests (4):
  TPP_OPT :: BF16/matmul-vnni.mlir
  TPP_OPT :: Conversion/VectorToXsmm/vector-to-transpose.mlir
  TPP_OPT :: Passes/DefaultPipeline/default-tpp-passes.mlir
  TPP_OPT :: Python/transform_callback.py
```

Hmm - all fail with segfaults. These tests all succeed for me on the compile server. 😕 

Will need to think about it. Might be to do with enabling some Python environment vars in the `lit.cfg.py` as otherwise `tpp-sched` (which is implicated in the segfaults) is not touched _at all_ by this PR.

test/Python/transform_callback.py
```diff
@@ -0,0 +1,128 @@
...
+@run_schedule
+def testTransformCallbackPyTransform(target):
+    func_ops = match(target, ops={"func.func"})
+    matmul_ops = match(target, ops={"linalg.matmul"})
+    td_param = transform.param_constant(
+        transform.AnyParamType.get(), ir.StringAttr.get("ABC")
+    )
+
+    @call_with(func_ops, matmul_ops, td_param)
+    def py_transform(funcs, matmuls, params) -> (transform.AnyValueType.get(),):
```
rolfmorel, line 111: Here's an example of writing transforms in Python in a way that integrates with the Transform dialect.

ssilva-encharge: This is really cool. Looking forward to see more in this direction.

rolfmorel:
> Will need to think about it. Might be to do with enabling some Python environment vars in the `lit.cfg.py` as otherwise `tpp-sched` (which is implicated in the segfaults) is not touched _at all_ by this PR.

Issue was reference counting of the callback object. This is now fixed. 

IMO this PR is ready to go (e.g. to be used for nano-tile selection algorithm prototyping).

For reference purposes, another instance of passing callbacks around: https://reviews.llvm.org/D116568

rolfmorel:
> I take that the callback signature is reified into IR. 

Yes - automatically converted from the signature on the Python callback. There's a test case with an example: https://github.com/libxsmm/tpp-mlir/pull/1064/files#diff-7aa62724b21b998da9cf032da6e6b77bbdb664258a5dca850f9509a5459646f6R95

> How/Where does the runtime find the actual callback definition to call?

A global func pointer gets set when a Python callback handler is provided. When the op executes it just reaches for the global and calls the handler.

The main thing that a proper solution should have is that this global gets moved to hang of the MlirContext. Though in general I think of how the FFI mechanism works as an implementation detail (e.g. the user could not use the Py bindings and use something else to set the global or just swap out the entire mechanism).

test/Python/transform_callback.py
rolfmorel: @JosseVanDelm, as discussed on Friday, here's an example of a transform op that's essentially defined in Python (by having an actual op do FFI callbacks) doing arbitrary stuff in Python.

If you have use for it, I can make it into an upstream branch/PR. Just let me know.

PR #1065, opened 2025-06-30
[μKernels]: vector-to-kernel pipeline changes + new .json files for bf16 type

arun-thmn:
This PR cleans-up the `vector-to-kernel` pipeline and add new .json file for `bf16` benchmarking.

### DIFF
--- a/benchmarks/config/omp/mlir-bf16.json
+++ b/benchmarks/config/omp/mlir-bf16.json
@@ -31,34 +31,65 @@
     }
   }},
   {
-  "gemm_bf16_dp2_mlir_vector_kernel_avx512bf16": {
+  "gemm_bf16_dp2_mlir_vector_kernel_avx512": {
     "bf16_dp2_3x1024_omp_2_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
       "environment": { "OMP_NUM_THREADS": "2", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=8,16 --vector-to-kernels --registerBlocking=8,32,2'" ],
-      "extensions": [ "avx512_bf16" ]
+      "extensions": [ "(avx512.*)" ]
     },
     "bf16_dp2_3x1024_omp_4_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
       "environment": { "OMP_NUM_THREADS": "4", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=8,8 --vector-to-kernels --registerBlocking=8,32,2'" ],
-      "extensions": [ "avx512_bf16" ]
+      "extensions": [ "(avx512.*)" ]
     },
     "bf16_dp2_3x1024_omp_8_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
       "environment": { "OMP_NUM_THREADS": "8", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=4,8 --vector-to-kernels --registerBlocking=8,32,2'" ],
-      "extensions": [ "avx512_bf16" ]
+      "extensions": [ "(avx512.*)" ]
     },
     "bf16_dp2_3x1024_omp_16_mlir": {
       "type": "IR-GEN",
       "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
       "environment": { "OMP_NUM_THREADS": "16", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
       "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=2,8 --vector-to-kernels --registerBlocking=8,32,2'" ],
-      "extensions": [ "avx512_bf16" ]
+      "extensions": [ "(avx512.*)" ]
+    }
+  }},
+  {
+  "gemm_bf16_dp2_mlir_vector_kernel_avx2": {
+    "bf16_dp2_3x1024_omp_2_mlir": {
+      "type": "IR-GEN",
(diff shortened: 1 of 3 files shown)


lib/TPP/PassBundles/VectorToKernel.cpp
```diff
@@ -53,17 +53,14 @@ struct VectorToKernel : public tpp::impl::VectorToKernelBase<VectorToKernel>,
...
-    // avx512 -> avx2 to subset needs to be improved by moving out some logic of
-    // Bf16DotProduct related to iterarg creation and let hoistvectorTransfer
-    // pass address it.
-    pm.addNestedPass<func::FuncOp>(createBF16DotProduct());
+    // avx512 -> avx2 to subset needs to be improved by updating the `k`
+    // tile size check for AMX lowering. With k = 1 (or vnni size) AMX fails
+    // lowering to micro-kernels on EMR. Bf16DotProduct tests with k = 1 
+    // and those tests gets lowered by AMX pass on EMR machine.
     pm.addNestedPass<func::FuncOp>(createHoistVectorTransfers());
+    pm.addNestedPass<func::FuncOp>(createMicroKernels());
```
adam-smnk: Just to double check - this order doesn't mess up lowering to amx?

benchmarks/config/omp/mlir-bf16.json
```diff
@@ -31,34 +31,65 @@
...
+    }
+  }},
+  {
+  "gemm_bf16_dp2_mlir_vector_kernel_avx2": {
+    "bf16_dp2_3x1024_omp_2_mlir": {
+      "type": "IR-GEN",
+      "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=256 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
+      "environment": { "OMP_NUM_THREADS": "2", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
+      "flags": [ "-n", "100", "-run-args='--def-parallel --parallel-task-grid=8,16 --vector-to-kernels --registerBlocking=2,32,2'" ],
+      "extensions": [ "(avx2)" ]
```
adam-smnk, line 71: On targets that support both avx2 and avx512, won't it lower to avx512 anyway but just as a smaller kernel (due to smaller register blocking)?

lib/TPP/PassBundles/VectorToKernel.cpp
```diff
@@ -53,17 +53,14 @@ struct VectorToKernel : public tpp::impl::VectorToKernelBase<VectorToKernel>,
 private:
   void constructPipeline() override {
     // TODO: Pass ordering based on target architecture starting from AMX ->
-    // avx512 -> avx2 to subset needs to be improved by moving out some logic of
-    // Bf16DotProduct related to iterarg creation and let hoistvectorTransfer
-    // pass address it.
-    pm.addNestedPass<func::FuncOp>(createBF16DotProduct());
+    // avx512 -> avx2 to subset needs to be improved by updating the `k`
```
shahidact, line 56: Older comments does not apply now, right?

lib/TPP/PassBundles/VectorToKernel.cpp
shahidact: Are we not propagating `targetFeature` option?

arun-thmn: No it doesn't mess-up. The `micro-kernels` pass needs `k` tile to be always `1`, where as `amx` pass needs `k` to be `>` than 1 (k = 32).

The `amx` tests have  `--registerBlocking=32,32,32`, which the `micro-kernels` pass skips rewriting.

benchmarks/config/omp/mlir-bf16.json
arun-thmn: Yes, if it has both it lowers to `avx512` with smaller kernels. These tests are intension for `ARL` and `ADL` machines.
We have tests both for `avx512` and `avx2` separately.

adam-smnk: My main concern here is that it's pretty misleading. They still run on EMR etc. named as avx2 but completely different code is benchmarked in reality.

adam-smnk: Imo, we need a flag or dlti attribute that can force specific variant, or to not run these benchmarks at all (as in no avx2 case in presence of avx512).

lib/TPP/PassBundles/VectorToKernel.cpp
arun-thmn: These comments  are updated to address some needed checks in `amx` pass. With `--registerBlocking=8,32,vnni`, the amx pass lowers test-cases and it results in `segmentation` fault.  If I remember correctly, `amx` pass needs `k` tile size = 32.

We need `amx` pass to check the `k` tile size or the `segmentation` issue needs to be fixed for `k` = 1.

shahidact: Ok, but "// TODO: Pass ordering based on target architecture starting from ..." seems irrelevant.

lib/TPP/PassBundles/VectorToKernel.cpp
arun-thmn: We do lowering based on the target machine, if the target machine has both `avx512` and `avx2`, `avx512` has the higher preference.

lib/TPP/PassBundles/VectorToKernel.cpp
arun-thmn: I try to mean if machine supports `amx`, `avx512`, and `avx2`, the preference of lowering is `amx`->`avx512`->`avx2`. 
Because of the `k` tile size issue, the `micro-kernel` pass is called before `amx` pass.

The `TODO` is to fix the issue with `amx` pass and then move the `micro-kernel` pass after `amx` pass.

benchmarks/config/omp/mlir-bf16.json
arun-thmn: True, Let me try to find out a way.

arun-thmn: This `"extensions": [ "!(avx512.*) & avx2" ]` triggers those tests only on `avx2` machines. I think, this helps, let me changes the .json file and check it completely.

lib/TPP/PassBundles/VectorToKernel.cpp
shahidact: Yes, in case of both avx512 and avx2 is present `targetFeature` gives the flexibility to choose between them if user wants. This way we can even handle the concern raised by @adam-smnk regarding the avx2 benchmark execution as avx512. This feature was added for better control on testing and benchmarking.

shahidact: Moreover, we would want to maintain the existing behaviour of command line flags.

adam-smnk: +1 it's also nice to decouple code generation capabilities from the host machine
Then we can verify lowering (without execution) on any target

arun-thmn: Added the `targetFeature` option to choose the target instruction set `avx512` or `avx2`.

benchmarks/config/omp/mlir-bf16.json
arun-thmn: This `extension` doesn't work for all case. So, added a `targetFeature` to let the user to choose `avx2` or `avx512` lowering.

PR #1067, opened 2025-07-08
[Linalg-Vectorizer] Add pattern for mixed precision vector.contract.

shahidact:
This pattern tries to fold arith.ext* ops present in high level mixed precision linalg contraction ops into mixed precision vector.contract.


test/Passes/linalg-vectorize.mlir
```diff
@@ -54,7 +54,6 @@ func.func @vectorize_contract_mixed_precision_float(
 // CHECK-NOT: vector.transpose
 // CHECK: vector.transfer_read{{.*}}: tensor<128x256x2xbf16>, vector<128x256x2xbf16>
 // CHECK: vector.transfer_read{{.*}}: tensor<256x256xf32>, vector<256x256xf32>
-// CHECK-COUNT-2: arith.extf
```
adam-smnk, line 57: nit: I'd replace with `CHECK-NOT` to make sure they're not here

adam-smnk (approved):

test/Passes/linalg-vectorize.mlir
shahidact: Done.

PR #1068, opened 2025-07-08
[μKernels]: lowering based on m and n tile size

arun-thmn:
This `PR` update a small logic in the micro-kernels lowering based on `m` and `n` tile size. If:

-  `m >= n` - first load all B matrix elements, then broadcast A one-by-one + do `fma`.
- `n > m` - do the opposite. First broadcast all A matrix elements, then load B one-by-one + do `fma`

The logic is updated for `fp32` (both avx512 & avx2) and `bf16` (only avx512). 
`Bf16` avx2 will be done later after fixing the `llvm` pattern matching problem on `ADL` machine.

### DIFF
--- a/lib/TPP/Transforms/VectorContractToMicroKernels.cpp
+++ b/lib/TPP/Transforms/VectorContractToMicroKernels.cpp
@@ -412,6 +412,10 @@ struct MicroKernelsOp : OpRewritePattern<vector::ContractionOp> {
     // matrix then broadcast A ony-by-one + FMA.
     // If N > M: perform opposite. Broadcast A matrix then load B one-by-
     // one + FMA.
+    // Following this kind of lowering, we reduce the register loads by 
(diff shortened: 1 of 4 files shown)


lib/TPP/Transforms/VectorContractToMicroKernels.cpp
```diff
@@ -409,6 +408,16 @@ struct MicroKernelsOp : OpRewritePattern<vector::ContractionOp> {
       return rewriter.notifyMatchFailure(
           contractOp, "Affine map permutation not supported.");
 
+    // Lowering is done based on M and N tile sizes. If M >= N: load all B
```
adam-smnk, line 411: nit: maybe add a small explanation why we do that

arun-thmn: Okay. Added an explanation.

PR #1069, opened 2025-07-08
Vectorizer - linalg to x86 vectorization bundle

adam-smnk:
x86 vectorizer bundle combines individual vectorization passes to drive linalg ops lowering into sequences of small vector ops.

The bundle is compatible with `tpp-mapping` to allow for initial packing and parallelization tiling before further lowering.


test/Integration/vectorizer-gemm-fma.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: tpp-opt %s | \
+// RUN:  tpp-run -e entry --entry-point-result=void -seed 123 -print > %t.1
```
arun-thmn: `tpp-run -e entry --entry-point-result=void %s -seed 123 -print > %t.1` - should be a straight run as we are not invoking any pass from `tpp-opt`. Why we are going via `tpp-opt`.

adam-smnk: Good point, cleaned up

lib/TPP/PassBundles/X86Vectorizer.cpp
```diff
@@ -0,0 +1,98 @@
...
+    //
+    // Hoisting allows for more opportunities to fold write-read pairs which
+    // results in fewer transfers after unrolling.
+    pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
+    pm.addNestedPass<func::FuncOp>(createLoopInvariantSubsetHoistingPass());
+    pm.addPass(createCleanup());
+
+    // Split vectors into register shapes.
+    //
+    // Current unrolling only targets contractions and relies on LLVM backend
```
shahidact, line 78: Do we have a test to see what is going on in the llvm backend or what this pipeline produces once it goes through llvm backend?

shahidact (commented):
In general, the pipeline looks good.

lib/TPP/PassBundles/X86Vectorizer.cpp
adam-smnk: Not in repo, do we have such tests for any passes?
It should definitely be added once we align on testing strategy at large.

For correctness, I have tested and inspected it manually. As such the pipeline produces tight inner loop without spills (for reasonable parameters). The lack of parameter selection (all the tiling sizes per target) currently blocks it from running automated benchmarking for comparison.
Loads/stores are a bit suboptimal as mentioned in the unrolling PR but the penalty is small enough to ignore it for now. Otherwise, improving unrolling driver is relatively large effort.

adam-smnk:
> > Also, there's no tool for selecting appropriate (default) register shapes depending on target/workload.
> 
> This is sth we should brainstorm. I think the workload input/output data type and underlying HW register width can be used to infer the default register shape. (May be I am missing sth?)

We'd need sth that integrates into pipeline and enables running benchmarks.
We could start with a simple hardcoded choices depending on the target either within the individual passes themselves or (perhaps preferably) attach module-level DLTI attributes with global decisions.

Long-term we'd need to make different choices for each workloads e.g., F32 tiling toward outer product but BF16 toward AMX sizes. Attaching attributes directly to ops could be one way.

> I would suggest that we prototype this with the Python callbacks.

Would be really cool to explore.
It could be good opportunity to move some infrastructure to python as well to enable execution.

shahidact:
Can we use this `--x86-vectorizer` with `tpp-run` to experiment with mixed precision case?

arun-thmn:
> Bump, any more comments?

Looks fine for me, to get merged.

arun-thmn (approved):
No more comments, LTGM!
Thanks @adam-smnk

PR #1070, opened 2025-07-11
Support lowering of mixed precision vector.contract to AMX.

shahidact:
This patch support mixed precision lowering of vector.contract with bf16/i8 input and fp32/i32 output type respectively
while also refactoring the code for readability/modularity. Moreover, it also prepares for other possible mixed precision
operation.

It was observed that the 'REQUIRES' directive was not working due to a lit config gap. This also fixes the observed issue.

### DIFF
--- a/lib/TPP/Transforms/VectorContractToAMX.cpp
+++ b/lib/TPP/Transforms/VectorContractToAMX.cpp
@@ -36,6 +36,113 @@ using namespace mlir;
 using namespace mlir::tpp;
 
 namespace {
+// Helper function to down convert and copy the result back to the output
+// matrix.
+static void downConvertAndCopyResult(OpBuilder &rewriter, Location loc,
+                                     Value accBuffer, Value accSubview,
+                                     MemRefType bufferType, ShapedType accType,
+                                     Value c0, Value mBound, Value nBound,
+                                     Value one, Value sixteen) {
+  rewriter.create<scf::ForOp>(
+      loc, c0, mBound, one, ValueRange{},
+      [&](OpBuilder &nestedBuilder, Location loc, Value iv,
+          ValueRange iterArgs) {
+        nestedBuilder.create<scf::ForOp>(
+            loc, c0, nBound, sixteen, iterArgs,
+            [&](OpBuilder &innerBuilder, Location loc, Value innerIv,
+                ValueRange innerIterArgs) {
+              auto elementType = bufferType.getElementType();
+              FloatType floatType = cast<FloatType>(elementType);
+              Value f0 = rewriter.create<arith::ConstantFloatOp>(
+                  loc, APFloat::getZero(floatType.getFloatSemantics()),
+                  floatType);
+              // Read
+              auto readC = rewriter.create<vector::TransferReadOp>(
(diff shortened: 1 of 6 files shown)


lib/TPP/Transforms/VectorContractToAMX.cpp
```diff
@@ -36,6 +36,113 @@ using namespace mlir;
 using namespace mlir::tpp;
 
 namespace {
+// Helper function to down convert and copy the result back to the output
+// matrix.
+static void downConvertAndCopyResult(OpBuilder &rewriter, Location loc,
+                                     Value accBuffer, Value accSubview,
+                                     MemRefType bufferType, ShapedType accType,
+                                     Value c0, Value mBound, Value nBound,
+                                     Value one, Value sixteen) {
```
adam-smnk: Could you generalize these parameters a bit?
I guess `accBuffer` is more of source and `accSubview` destination?
If you need arith constants just create them directly within this function.

test/Passes/pass-vector-contract-to-amx.mlir
```diff
@@ -1,4 +1,4 @@
-// REQUIRES: target=x86{{.*}}
```
adam-smnk, line 1: In which case the old matcher fails?
Maybe we could have a different regex pattern instead of manual `cfg` hack?

lib/TPP/Transforms/VectorContractToAMX.cpp
shahidact: Sure.

test/Passes/pass-vector-contract-to-amx.mlir
shahidact: > In which case the old matcher fails?

What I observed is `// REQUIRES: target=x86{{.*}}` did not have expected effect, i.e on all other target except x86 the test should have been marked `unsupported`, but it was being marked `unsupported `even for x86 target.

shahidact: > Maybe we could have a different regex pattern instead of manual `cfg` hack?

I followed the approach LLVM is using in general.

adam-smnk: Oh cool, now I found other examples too.
Then it should be fine.

adam-smnk (approved):

lib/TPP/Transforms/VectorContractToAMX.cpp
```diff
@@ -399,73 +529,44 @@ struct VectorContractToAMXPattern
...
-                    loc, bitcast, accBuffer, ValueRange{iv, innerIv},
-                    ArrayRef{true});
-                innerBuilder.create<scf::YieldOp>(loc);
-              });
-
-          // Yield results from inner loop to outer loop
-          nestedBuilder.create<scf::YieldOp>(loc);
-        });
+    // auto one = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+    // auto sixteen = rewriter.create<arith::ConstantIndexOp>(loc, 16);
```
adam-smnk: nit: remove unused code

shahidact: Done.

PR #1072, opened 2025-07-16
[Benchmarking] update signed integer bit -128 to 127

arun-thmn:


include/TPP/Transforms/Utils/TensorInitInt.h
```diff
@@ -124,15 +124,15 @@ struct RandomTensorInitInt : TensorInitInt {
...
+      : TensorInitInt(type), generator(seed), distribution(127) {}
 
   // Next random number.
   float next() {
     auto value = distribution(generator);
     return value;
   }
 
-  // Return a dense<normal(0, 255)> throughout the shape.
+  // Return a dense<normal(0, 127)> throughout the shape.
```
adam-smnk: What about the negative half?

test/Integration/tpp-run-splat-tensor.mlir
adam-smnk, line 1: Please add `i8` case as well

include/TPP/Transforms/Utils/TensorInitInt.h
arun-thmn: If I understood correctly, we don't initialize with `-ve` values. So, we start from `0` and kind of ignoring the `-ve` half.

arun-thmn:
> Can you also change `ContinuousTensorInitInt ` and `RandomTensorInitInt `?

Added.

include/TPP/Transforms/Utils/TensorInitInt.h
```diff
@@ -100,18 +100,18 @@ struct ContinuousTensorInitInt : TensorInitInt {
...
   // Upper bound for quantization.
-  int upperBound = 255;
+  int upperBound = 127;
 };
 
 // Random init (uniform).
 struct RandomTensorInitInt : TensorInitInt {
   RandomTensorInitInt(DataType type, int seed)
-      : TensorInitInt(type), generator(seed), distribution(0, 255) {}
+      : TensorInitInt(type), generator(seed), distribution(0, 127) {}
```
rengolin: This could be -128, 127

include/TPP/Transforms/Utils/TensorInitInt.h
adam-smnk: Probably but is that what we really want for integers?

include/TPP/Transforms/Utils/TensorInitInt.h
arun-thmn: This leads to an `LLVM` error `Assertion '_M_a <= _M_b' failed`

rengolin: oh, maybe the distribution can't be negative. We can then fix both issues with the same fix: revert the distribution ack to 255 on signed int and subtract 128 and cast down to signed char.

arun-thmn: https://github.com/libxsmm/tpp-mlir/blob/b4efad44547cb152b4c9f331065ad713a12d462e/include/TPP/Transforms/Utils/TensorInitInt.h#L121

The distribution is `uint64_t`

rengolin: should be fine. It's guaranteed to not exceed 255, so you can easily cast to `int32_t` subtract then cast to `int8_t`.

arun-thmn: This distribution can be changed to `int64_t` and by checking the `isSigned` flag we can set the distribution either `0 to 255` or `-127 t0 128`

arun-thmn:
@rengolin @adam-smnk 
Have, set a `flag` that check the type and updates `distribution` within `127` for `i8`. I hope, this should cover the case.

adam-smnk (commented):
Please add tests to cover these initializers and maybe update PR title

arun-thmn:
> Please add tests to cover these initializers and maybe update PR title

Changed the title and update the unit-test to cover `i8` type as well. Now, should be good to be merged.

PR #1075, opened 2025-07-22
[mlir-gen] Add mixed precision type support(1/N).

shahidact:
Adds creation of gemm kernels with mixed types such as bf16 -> fp32, f16 -> f32, i8 -> i32.

### DIFF
--- a/tools/mlir-gen/MLIRGen.cpp
+++ b/tools/mlir-gen/MLIRGen.cpp
@@ -114,16 +114,27 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
          "Must have 3 tile sizes (or none)");
 
   // Pick data type
-  auto elementType = llvm::StringSwitch<std::optional<Type>>(targetType)
-                         .CaseLower("f32", builder.getF32Type())
-                         .CaseLower("f16", builder.getF16Type())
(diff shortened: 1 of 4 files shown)


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -114,16 +114,27 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
          "Must have 3 tile sizes (or none)");
 
   // Pick data type
-  auto elementType = llvm::StringSwitch<std::optional<Type>>(targetType)
-                         .CaseLower("f32", builder.getF32Type())
-                         .CaseLower("f16", builder.getF16Type())
-                         .CaseLower("bf16", builder.getBF16Type())
-                         .Default(std::nullopt);
+  auto elementType =
+      llvm::StringSwitch<std::optional<SmallVector<mlir::Type, 3>>>(targetType)
```
adam-smnk: nit: any reason for the vector to have capacity of `3`? I think it could be just plain `SmallVector<Type>`

shahidact: Actually.. no reason, just intuition, but we may need even more when we add scale arguments. I will update.

shahidact: Done.

PR #1076, opened 2025-07-23
[μKernels]: AMX - optimal register allocation

arun-thmn:
Changes/Relaxation in `AMX` pass to support optimal register allocation.


adam-smnk (approved):
Good on its own

What exactly has changed here?

arun-thmn:
@adam-smnk  Like other `avx-*` lowering,  `amx` optimal register allocation doesn't depend on `m` and `n` sizes. So, the pass itself competent to do the optimal blocking.
I just relaxed few condition + corrected a typo in the `loop` condition.

PR #1079, opened 2025-07-31
[Init] Add identity initializer

rengolin:
This is a step towards allowing tests like A * I = A.

As is, it still doesn't do what we want, since like the others, it tried to initialize all tensors with identity, so any non square tensor will fail.

The next step is to add options like "--identity-b" together with "--splat-to-random", so that only the B matrix is replaced by identity.

Also remove some unused initializers.

### DIFF
--- a/include/TPP/Runner/MLIRBench.h
+++ b/include/TPP/Runner/MLIRBench.h
@@ -47,7 +47,7 @@ struct MLIRBenchConfig {
 
   int seed = 0;
   TensorInitType initType = TensorInitType::Auto;
-  int identity = 0;
+  int identity = -1;
   std::string backend = "cpu";
   bool offloadToDevice = true;
 };
--- a/include/TPP/Runner/MLIRBench.h
+++ b/include/TPP/Runner/MLIRBench.h
@@ -40,13 +40,14 @@ class FuncOp;
 // pipeline.
 struct MLIRBenchConfig {
   MLIRBenchConfig() = default;
-  MLIRBenchConfig(int seed, TensorInitType initType, std::string backend,
+  MLIRBenchConfig(int seed, TensorInitType initType, int identity, std::string backend,
                   bool offloadToDevice)
-      : seed(seed), initType(initType), backend(backend),
+      : seed(seed), initType(initType), identity(identity), backend(backend),
         offloadToDevice(offloadToDevice) {}
 
   int seed = 0;
   TensorInitType initType = TensorInitType::Auto;
+  int identity = 0;
   std::string backend = "cpu";
   bool offloadToDevice = true;
 };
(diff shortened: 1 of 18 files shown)


shahidact:
I would also like to have the capability to generate A/B identity matrix in `mlir-gen` which will allow to create correctness test easily.

rengolin:
@shahidact now with an option in `mlir-gen` to generate identity weights and constant biases, so that you can check the operations for errors when testing vectorization.

lib/TPP/Transforms/Utils/TensorInit.cpp
```diff
@@ -147,3 +140,14 @@ TensorInitPtr getTensorInit(StringRef type, mlir::Type elmType, int seed) {
...
+
+FailureOr<int> getIdentityOption(StringRef identity) {
+  int option = StringSwitch<int>(identity)
+                   .Case("", -1)
+                   .CaseLower("a", 0)
+                   .CaseLower("b", 1)
+                   .Default(-2);
+  if (option == -2)
+    return failure();
+  return option;
```
adam-smnk: nit: collapse into `return StringSwitch...`

tools/tpp-run/tpp-run.cpp
```diff
@@ -85,18 +85,26 @@ llvm::cl::opt<int> seed("seed",
...
-// Speed optimization level
-llvm::cl::opt<unsigned>
-    optLevel("O", llvm::cl::desc("Speed optimization level (O0, O1, O2, O3)"),
-             llvm::cl::value_desc("0-3"), llvm::cl::init(2));
-
 // Initializer type
 // Default const if seed == 0, and normal otherwise
 llvm::cl::opt<std::string> initType(
     "init-type",
     llvm::cl::desc("Initializer type (const, simple, cont, rand, normal)"),
```
adam-smnk: Desc could use an update too

tools/tpp-run/tpp-run.cpp
```diff
@@ -85,18 +85,26 @@ llvm::cl::opt<int> seed("seed",
...
     llvm::cl::init(""));
 
+// Identity matrix
+// Replace single square argument with identity matrix
+// Note: Must have two arguments and the selected must be square
```
adam-smnk, line 97: I don't think it really follows any `a`/`b` "semantics".
Maybe the flag could just take `int` argument position directly?

include/TPP/Runner/MLIRBench.h
```diff
@@ -40,13 +40,14 @@ class FuncOp;
...
-  MLIRBenchConfig(int seed, TensorInitType initType, std::string backend,
+  MLIRBenchConfig(int seed, TensorInitType initType, int identity, std::string backend,
                   bool offloadToDevice)
-      : seed(seed), initType(initType), backend(backend),
+      : seed(seed), initType(initType), identity(identity), backend(backend),
         offloadToDevice(offloadToDevice) {}
 
   int seed = 0;
   TensorInitType initType = TensorInitType::Auto;
+  int identity = 0;
```
adam-smnk: `-1` default?

PR #1080, opened 2025-07-31
[Script] Helper for dumping assembly with target-feature

rengolin:


arun-thmn:
Look good for generating assembly dump via `tpp-run`. But, we should make sure whether we need to pass the `-m` option.
FYI, the options is pass to `objdump` as below:
x86 - `objdump -D -m i386:x86-64:intel -b binary`
arm - `objdump -D -b binary -m aarch64`

PR #1082, opened 2025-08-01
[μKernels]: Micro-kernel lowering for splat bf16 - ARL

arun-thmn:
This `patch` supports micro-kernel lowering for `bf16` splat layouts on `ARL` kind of machines.

`TODO`: Once all the lowering for `bf16 splat` layout on different machines completed. We will factor them under one `if` branch.

### DIFF
--- a/lib/TPP/Transforms/VectorContractToMicroKernels.cpp
+++ b/lib/TPP/Transforms/VectorContractToMicroKernels.cpp
@@ -1334,7 +1334,6 @@ struct MicroKernelsOp : OpRewritePattern<vector::ContractionOp> {
                 }
 
                 if (isSplat && srf) {
-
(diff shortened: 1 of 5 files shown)


lib/TPP/Transforms/VectorContractToMicroKernels.cpp
```diff
@@ -1406,7 +1528,7 @@ struct MicroKernelsOp : OpRewritePattern<vector::ContractionOp> {
                     }
 
                   } else { // N -> M
-	            // Load A matrix
+                           // Load A matrix
```
adam-smnk: nit: I guess indent is a bit off or bring `// N -> M` comment below `else` to be within its body

arun-thmn: Updated, Thanks.

PR #1083, opened 2025-08-01
[Benchmarking]: Changes to bench files w.r.t scaling factor

arun-thmn:


benchmarks/driver.py
```diff
@@ -459,6 +460,12 @@ def scanBenchmarks(self):
                     else:
                         # List of possible supported (ex. avx512 OR sve2)
                         for ext in run["extensions"]:
+                            if ext == "arl":
+                                env = os.environ.copy()
+                                compile_result = subprocess.run(["./check-cpuid"], env=env, capture_output=True, text=True)
```
rengolin: what is this tool? If it isn't in this repo, there's no way we can depend on it.

Also remember, the `extensions` list is of regular expressions, so it could be `arl*` and still be valid.

A simpler thing would be to add more fields from `lscpu` into the check and regexp-match "Core Ultra 9 285K" instead.

tools/check-cpuid/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ target_compile_options(check-cpuid PRIVATE -O2)
 # Optional: ensure it outputs to a known build dir
 set_target_properties(check-cpuid PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin
+  RUNTIME_OUTPUT_DIRECTORY ${CMAKE_SOURCE_DIR}/benchmarks
```
rengolin: What's the purpose of this change?

benchmarks/driver.py
arun-thmn: `./check-cpuid` is an executable file generated as part of `tpp-mlir` build for `c` file `https://github.com/libxsmm/tpp-mlir/blob/main/tools/check-cpuid/check-cpuid.c. 

For `i8` lowering, we added this `c` file as part of `lit` test to check whether `VPDPBSSD` is present in a machine or not as `lscpu` on `ARL` doesn't list the flag for `VPDPBSSD`. 

So, kind of re-using this existing check for `ARL` in benchmarking.

tools/check-cpuid/CMakeLists.txt
arun-thmn: This generates the executable `check-cpuid` in the `benchmarks` folder, which is used by `driver.py` script to check whether the machine is `ARL` or not.

benchmarks/driver.py
rengolin: Oh, I see, I thought you were adding this tool here and missing the file.

tools/check-cpuid/CMakeLists.txt
rengolin: The benchmark tool already has a way to figure out where the `bin` directory is (for all other tools), so you don't need this change, just follow the pattern for `mlir-gen`, `tpp-run`, etc.

arun-thmn: Removed this change. Followed the existing options available in the benchmarks script to get the `build/bin` path.

arun-thmn:
@adam-smnk any suggestions+comments

adam-smnk (approved):
Looks fine

PR #1087, opened 2025-08-07
[μKernels]: Fix the broken ARM FP32 micro-kernel lowering

arun-thmn:
This `patch` fixes the Issue: https://github.com/libxsmm/tpp-mlir/issues/1086 broken ARM FP32 micro-kernel lowering.

### DIFF
--- a/include/TPP/Transforms/Utils/VNNIUtils.h
+++ b/include/TPP/Transforms/Utils/VNNIUtils.h
@@ -36,14 +36,18 @@ enum class VnniOperandRank {
   BRGEMM_INS = 4,
   BRGEMM_OUTS = 3
 };
-// Returns True if the current architecture supports AVX2 instructions.
+
+// Returns true if the current architecture supports AVX2 instructions.
 bool hasAVX2();
 
-// Returns True if the current architecture supports AVX512 instructions.
+// Returns true if the current architecture supports AVX512 instructions.
 bool hasAVX512();
 
-// Returns True if the current architecture supports AMX instructions.
+// Returns true if the current architecture supports AMX instructions.
 bool hasAMX();
+
+// Returns true if the current architecture is not x86.
+bool isNotX86();
 
 // Returns the current target architecture name
 std::string getTargetArchName();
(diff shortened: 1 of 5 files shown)


include/TPP/Transforms/Utils/VNNIUtils.h
```diff
@@ -36,15 +36,19 @@ enum class VnniOperandRank {
...
-// Returns True if the current architecture supports AVX512 instructions.
+// Returns true if the current architecture supports AVX512 instructions.
 bool hasAVX512();
 
-// Returns True if the current architecture supports AMX instructions.
+// Returns true if the current architecture supports AMX instructions.
 bool hasAMX();
 
+// Returns true if the current architecture is not x86.
+bool isNotX86();
```
rengolin: This is not a good idea. Just create one for SVE like the others.

Also note that Armv8.1 does not mean SVE, even though that's probably what `LIBXSMM_AARCH64_V81` assumes.

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
```diff
@@ -354,16 +354,27 @@ struct MicroKernelsOp : OpRewritePattern<vector::ContractionOp> {
...
-
     bool isF32 = elementType.isF32();
     bool isF16 = elementType.isF16();
     bool isBF16 = elementType.isBF16();
     bool isI8 = elementType.isSignlessInteger(8);
 
     bool isPackedType = isF16 || isBF16 || isI8;
+
+    if (vnni::utils::isNotX86()) {
+      sizeFactor = 8;
```
rengolin: This will be impossible to upstream. It's ok for this PR, but remember, we'll have to write all of that from scratch when trying to send into LLVM.

include/TPP/Transforms/Utils/VNNIUtils.h
arun-thmn: Added `sve256` and `sve512` helper functions, similar to `avx` and `avx512`.

lib/TPP/Transforms/VectorContractToMicroKernels.cpp
arun-thmn: Okay. Removed this code.
Now, the `vector length` is decided based on the `sve` machine. For, `sve256` vector length is set to `8` and `sve512` set to `16`.

PR #1089, opened 2025-08-26
[mlir-gen] Introduce basic support for quantization ops (2/n).

shahidact:
This patch tries to extend the mixed precision isupport and add basic infra to define or create quantization kernel as we get more clarity on kind of primitive operations required.

Addition of run time test is not in scope of this patch.


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -175,21 +199,39 @@ void MLIRGenerator::getKernelTypes(KernelArgs &args) {
     LayerArgs arg;
     arg.index = i;
     arg.input.type = currentType;
+    // Scale inputs are only needed for dequantization.
+    if (isQuantKernel && quantType == QuantizationType::Dequant)
```
rengolin: Could just use `quantType != QuantizationType::None` to mean `isQuantKernel` and avoid the extra flag.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -284,17 +338,23 @@ void MLIRGenerator::createKernel() {
       arg.output.value = getZeroInitTensor(arg.output.type);
     }
 
-    // Now pass the input through all layers
-    lastOutput = createLayer(arg);
+    // Now pass the input through all layers.Separated the quantization layer
+    // creation to simplify the design and reduce code complxity as there will
+    // be more ways to introduce quantization ops in the future.
+    if (isQuantKernel)
+      lastOutput = createQuantLayer(arg);
```
rengolin: Can we just add quantization on top of GEMM generator instead of creating a whole new `createLayer`?

test/Integration/mlir-gen-matmul.mlir
```diff
@@ -135,3 +140,101 @@
 // MXF16-CONTRACT:           %[[VAL_0:.*]] = linalg.contract indexing_maps = [#[[$ATTR_0]], #[[$ATTR_1]], #[[$ATTR_2]]] ins(%[[ARG0]], %[[ARG1]] : tensor<2x36x64x64xf16>, tensor<16x36x64x48xf16>) outs(%[[ARG2]] : tensor<2x16x64x48xf32>) -> tensor<2x16x64x48xf32>
 // MXF16-CONTRACT:           return %[[VAL_0]] : tensor<2x16x64x48xf32>
 // MXF16-CONTRACT:         }
+
+// MXBF16-DEQUANT: #[[$ATTR_0:.+]] = affine_map<(d0, d1, d2) -> (d0, d2)>
```
adam-smnk: Could we omit some of the details?
IMO, It's too verbose to debug anything.

rolfmorel:
> General question, shouldn't we migrate to lighthouse's gen?

That's indeed a good question. Might be a good example to dogfood how we facilitate downstream projects interacting with lighthouse.

tools/mlir-gen/MLIRGen.cpp
shahidact: I hope so.

adam-smnk (approved):
General structure seems fine
I assume that the sequence of quant ops has been tested and is correct

PR #1091, opened 2025-09-16
[CI] Bump LLVM +  CI partition changes

arun-thmn:
Bump are mostly related to below  LLVM patches:
https://github.com/llvm/llvm-project/pull/151150 - `InitAll` replaced by static library. So, added necessary header to our pipeline.
https://github.com/llvm/llvm-project/pull/147961 - Remove or update lit tests related to `matmul_transpose`

For `CI`:
- Removed `v100` and `clx` partition + removed benchmarking related to `CLX` base and omp runs,
- Updated the `gcc` src to system src + cuda src to `12.9`,


lib/TPP/DefaultPipeline.cpp
```diff
@@ -6,24 +6,49 @@
...
 
-#include "TPP/PassBundles.h"
-
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/InitAllDialects.h"
 #include "mlir/InitAllPasses.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Pass/PassManager.h"
 #include "llvm/Support/CommandLine.h"
+#include "mlir/Conversion/MathToLLVM/MathToLLVM.h"
```
adam-smnk: Can we use `mlir/Conversion/Passes.h` to pull all of these in one go?

lib/TPP/Transforms/ToBlockLayoutAndBack.cpp
```diff
@@ -459,14 +459,15 @@ namespace {
 static SmallVector<int64_t>
 getDefaultBlockingFactors(linalg::LinalgOp linalgOp) {
   assert(linalgOp && "expect a valid linalgOp");
-  if (isa<linalg::Conv2DNchwFchwOp>(linalgOp) ||
-      isa<linalg::Conv2DNhwcHwcfOp>(linalgOp)) {
+  auto *op = linalgOp.getOperation();
```
adam-smnk, line 462: Any particular reason to change?
It's fine overall.

arun-thmn: With the new bump: We need to pass the `operation` for `ISA` not the wrapper.

Error if we just pass linalgOp: `cannot initialize a parameter of type Operation * with an rvalue of type const mlir::linalg::LinalgOp *`

lib/TPP/DefaultPipeline.cpp
arun-thmn: Thanks. It's is the most simplest way.
Updated with `mlir/Conversion/Passes.h`

PR #1092, opened 2025-09-16
[CI] Update the CUDA salloc

arun-thmn:
With the recent changes in `pcl-tiergarten` the `v100` support is no more available. So, we update the `CI` salloc partition.

### DIFF
--- a/.github/workflows/tpp-llvm.yml
+++ b/.github/workflows/tpp-llvm.yml
@@ -27,6 +27,6 @@ jobs:
       - name: LLVM CUDA
         run: |-
               GPU=cuda scripts/github/check_llvm.sh || \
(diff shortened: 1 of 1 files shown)


adam-smnk:
The two `a100` nodes should be fine.
We'll revise it in the future if we ever have problems with availability.

adam-smnk:
Not exactly sure why it fails.
Maybe sth clashes with the old cached LLVM build? You could try to add this change to the LLVM bump PR and see if it runs.

Also, now CUDA 13 being available, we should be able to use default system's gcc.
You could try to tweak `setup_gpu_env.sh` to stop sourcing older version.

arun-thmn:
Merged the changes with PR: https://github.com/libxsmm/tpp-mlir/pull/1091. Hence, cancelling this.

PR #1116, opened 2026-04-03
nano-kernels upstream rebase

arun-thmn:
This patch adds a pipeline named `nano-kernels` in the TPP to use the MLIR upstream patches to lower a `vector.contract` to a target-specific CPU kernels that proves to be a peak performance for a `gemm` operation.

TODOs:

1. The existing `vector-to-kernels` pipeline will be replaced with this pipeline and all the existing `integration-test` + benchmarking `.json` files will use the new upstream pipeline.
2. TPP-MLIR default pipeline using `libxsmm` will be shifted to `nano-kernels` pipeline.


lib/TPP/DefaultTppPasses.cpp
```diff
@@ -130,19 +130,52 @@ struct DefaultTppPasses
...
-      pm.addPass(createBufferize());
+      if (!nanoKernel) {
+        pm.addPass(createBufferize());
+      }
 
       // Lower Linalg to XSMM.
       pm.addNestedPass<func::FuncOp>(
           createLinalgLowering(LinalgLoweringOptions{skipOperations}));
 
+      if (nanoKernel) {
```
rengolin: This should be a bundle inside linalg lowering, no?

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -130,19 +130,52 @@ struct DefaultTppPasses
...
+
+        pm.addNestedPass<func::FuncOp>(createHoistLoopInvariantSubsets());
+
+        pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+
+        pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
+
+        pm.addPass(createBufferize());
+
+        pm.addNestedPass<func::FuncOp>(createVectorContractToNanoKernels());
```
rengolin: and these split into a low-level bundle for the nano-kernel lowering.

rengolin (commented):
we also should get rid of the hack on which ops to lower to xsmm. I can do that after this PR.

lib/TPP/DefaultTppPasses.cpp
arun-thmn: Now. kind of aligned it correctly.

`linalg` to ` vector` within the `linalgTovector` block, followed by the passes to further lower the `vector` code to `nanoKernels`.

lib/TPP/DefaultTppPasses.cpp
arun-thmn: Okay, split these passes into two groups. linalg -> vector one bundle and vector -> nanoKernels another bundle.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -130,33 +130,58 @@ struct DefaultTppPasses
       // bufferization.
       pm.addNestedPass<func::FuncOp>(createDecomposeAggregatedOps());
 
+      // pm.addPass(createBufferize());
```
rengolin: no commented out code

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -130,33 +130,58 @@ struct DefaultTppPasses
...
+      }
 
       // Lower Linalg to XSMM.
       pm.addNestedPass<func::FuncOp>(
           createLinalgLowering(LinalgLoweringOptions{skipOperations}));
 
       if (linalgToVector || forceLinalgToVector) {
+
+        if (nanoKernel) {
+	  pm.addNestedPass<func::FuncOp>(createLinalgGeneralizeNamedOpsPass());
```
rengolin: alignment

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -130,33 +130,58 @@ struct DefaultTppPasses
...
+
         // Please note, canonicalizer should be after hoisting pass because
         // it fuses outer tiling loops and it results in no pattern
         // matching for hoisting pass. Moved inside VectorToKernel Path.
 
         if (vectorToXSMM) {
           pm.addPass(createVectorToXSMM());
         }
+
+        // This path will be soon replaced by the nanoKernel path.
```
rengolin, line 182: why can't we just replace it now?

lib/TPP/DefaultTppPasses.cpp
arun-thmn: Yep, removed it.

lib/TPP/DefaultTppPasses.cpp
arun-thmn: Properly aligned in `vim`, bot not in webpage. Anyways corrected it now.

lib/TPP/DefaultTppPasses.cpp
arun-thmn: Wanted to compared the GFLOPS between these two paths + also have to update the multiple, unit-test, integration tests and benchmarking files to `nanoKernel` path istead of `vector-to-kernels`. So, first I just wanted to integrate the `nanokernel` path, check for similar GFLOPS perf, and  then retire the local one.

rengolin: We can take the last `main` benchmark run and compare with this PR's run. If the perf is similar, we don't need to keep the old code.

lib/TPP/Transforms/HoistLoopInvariantSubsets.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+    return success();
+  }
+};
+
+void populateHoistLoopInvariantSubsetPatterns(RewritePatternSet &patterns) {
+  patterns.add<HoistLISubsetOp>(patterns.getContext());
+}
+
+struct HoistLoopInvariantSubsets
```
adam-smnk, line 57: Any reason we can't use upstream `loop-invariant-subset-hoisting` pass directly?

lib/TPP/Transforms/HoistLoopInvariantSubsets.cpp
```diff
@@ -0,0 +1,70 @@
...
+    : public impl::HoistLoopInvariantSubsetsBase<HoistLoopInvariantSubsets> {
+  using HoistLoopInvariantSubsetsBase::HoistLoopInvariantSubsetsBase;
+
+  void runOnOperation() override {
+    RewritePatternSet patterns(&getContext());
+    populateHoistLoopInvariantSubsetPatterns(patterns);
+    GreedyRewriteConfig config;
+    config.setStrictness(GreedyRewriteStrictness::ExistingOps);
```
adam-smnk, line 65: nit: it should be safe to let it run in iterations to hoist multiple times if possible

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -136,27 +136,50 @@ struct DefaultTppPasses
...
         pm.addNestedPass<func::FuncOp>(createVectorizationPass());
 
+        if (nanoKernel) {
+          tpp::RegisterUnrollOptions unrollOpts;
+          unrollOpts.gemmUnroll = SmallVector<int64_t>{*gemmUnroll};
+          pm.addNestedPass<func::FuncOp>(createRegisterUnroll(unrollOpts));
+          pm.addNestedPass<func::FuncOp>(createHoistLoopInvariantSubsets());
+          pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+          pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
+          pm.addPass(createBufferize());
```
adam-smnk, line 168: The default flow is already complicated.
I'd suggest to split these passes into two groups: before and after bufferization, and reenable the above bufferization registration. It's easier to reason about the pipeline when there's only one clear tensor-to-memref conversion point.

adam-smnk (commented):
Minor nits, overall pretty clean replacement.
Before merging, please add a small summary about benchmarks and if there are any perf differences.

rengolin (approved):
We can do the clean ups later. LGTM, thanks!

lib/TPP/Transforms/HoistLoopInvariantSubsets.cpp
arun-thmn: For now, want to apply for the `scf.for` associated with `vector.contract`.

lib/TPP/DefaultTppPasses.cpp
arun-thmn: True. We will do that in a clean-up pass.

lib/TPP/Transforms/HoistLoopInvariantSubsets.cpp
arun-thmn: Okay, modified the code to run in iterations.

PR #1123, opened 2026-05-01
Support simple elemntwise tiling on linalg.generic.

shahidact:
Element wise operations on bigger tensor shapes results in bad code generation with inserts/extracts and spill. This pass tries to tile such operation so that backend could produce nice compact code without spills.

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -363,6 +363,17 @@ def ElementWiseFusion : Pass<"element-wise-fusion", "func::FuncOp"> {
   let summary = "Run linalg element-wise fusion";
 }
 
+def TileElementWiseOps : Pass<"tile-elementwise-ops", "func::FuncOp"> {
+  let summary = "Tile elementwise linalg.generic operations";
+  let description = [{
+    Tile elementwise linalg.generic operations to enable better vectorization.
+  }];
+  let dependentDialects = ["linalg::LinalgDialect", "scf::SCFDialect"];
+  let options = [
+    ListOption<"tileSizes", "tile-sizes", "int64_t", "Tile sizes">,
+  ];
+}
+
 def Bufferize : Pass<"bufferize", "ModuleOp"> {
   let summary = "Bufferize tensor to memref for the entire module";
   let options = [
--- a/lib/TPP/DefaultTppPasses.cpp
+++ b/lib/TPP/DefaultTppPasses.cpp
@@ -155,7 +155,9 @@ private:
         pm.addNestedPass<func::FuncOp>(createBrgemmLinalgTiling(
             BrgemmLinalgTilingOptions{SmallVector<unsigned>{*registerBlocking}}));
         pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
-        pm.addNestedPass<func::FuncOp>(createTileDequantElementwiseOps());
+        // pm.addNestedPass<func::FuncOp>(createTileDequantElementwiseOps());
+        if (!disableTileElementwiseOps)
+          pm.addNestedPass<func::FuncOp>(createTileElementWiseOps());
         pm.addNestedPass<func::FuncOp>(createVectorizationPass());
 
         if (nanoKernel) {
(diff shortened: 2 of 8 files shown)


rengolin:
@adam-smnk Do we have any element-wise tiling inside generics upstream?

adam-smnk:
> @adam-smnk Do we have any element-wise tiling inside generics upstream?

Looks like this pass is just a wrapper around upstream API to control tiling.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -796,4 +798,89 @@ struct ElementWiseFusion : tpp::impl::ElementWiseFusionBase<ElementWiseFusion> {
...
+    : tpp::impl::TileElementWiseOpsBase<TileElementWiseOps> {
+  using TileElementWiseOpsBase::TileElementWiseOpsBase;
+
+  void runOnOperation() override {
+    auto &ctx = getContext();
+    RewritePatternSet patterns(&ctx);
+
+    // Define a pattern of linalg.generic op with parallel iterators on which we
+    // want to apply tiling.
+    struct TileElementwiseGenericPattern
```
adam-smnk, line 813: Maybe `linalg::isElementwise` could be useful for some matcher validation

include/TPP/Passes.td
```diff
@@ -363,6 +363,17 @@ def ElementWiseFusion : Pass<"element-wise-fusion", "func::FuncOp"> {
   let summary = "Run linalg element-wise fusion";
 }
 
+def TileElementWiseOps : Pass<"tile-elementwise-ops", "func::FuncOp"> {
```
adam-smnk, line 366: Could you add a few small test cases for the pass itself?
It'll be easier to quickly grasp scope of rewrites and have a small regression guard.

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
```diff
@@ -796,4 +798,89 @@ struct ElementWiseFusion : tpp::impl::ElementWiseFusionBase<ElementWiseFusion> {
...
+        if (body.without_terminator().empty())
+          return rewriter.notifyMatchFailure(
+              op, "No non-terminating operations in the body");
+
+        auto bodyOps = body.without_terminator();
+        if (llvm::hasSingleElement(bodyOps) &&
+            isa<arith::AddFOp>(*bodyOps.begin()))
+          return rewriter.notifyMatchFailure(
+              op, "Dont tile linalg.generic with single addf in the body, as "
+                  "it is likely a reduction");
```
adam-smnk: nit: It can't be a reduction if it only has parallel iterators

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -155,7 +155,9 @@ struct DefaultTppPasses
         pm.addNestedPass<func::FuncOp>(createBrgemmLinalgTiling(
             BrgemmLinalgTilingOptions{SmallVector<unsigned>{*registerBlocking}}));
         pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
-        pm.addNestedPass<func::FuncOp>(createTileDequantElementwiseOps());
+        // pm.addNestedPass<func::FuncOp>(createTileDequantElementwiseOps());
```
adam-smnk: nit: remove dead code

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
shahidact: ok

include/TPP/Passes.td
shahidact: Sure

lib/TPP/Transforms/TileConsumerAndFuseProducers.cpp
shahidact: Ah right, will see why is it here?

lib/TPP/DefaultTppPasses.cpp
shahidact: ok

adam-smnk (approved):

PR #1126, opened 2026-06-08
[mlir-gen]: Use output argument tensor instead of empty tensor.

shahidact:
For 16-bit float, we don't need temporary empty tensor accumulator, instead we have to use the output tensor argument. This patch removes the generation of empty tensor and utilize output tensor argument.

### DIFF
--- a/tools/mlir-gen/MLIRGen.cpp
+++ b/tools/mlir-gen/MLIRGen.cpp
@@ -605,10 +605,6 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args, bool hasMixedType = false) {
       TensorType intAccumulatorType =
           RankedTensorType::get(shape, builder.getIntegerType(32));
       output = getZeroInitTensor(intAccumulatorType);
-    } else if (elementType.isBF16() || elementType.isF16()) {
-      output = getZeroInitTensor(cast<TensorType>(outputType));
-    } else {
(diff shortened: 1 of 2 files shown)


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -605,10 +605,6 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args, bool hasMixedType = false) {
       TensorType intAccumulatorType =
           RankedTensorType::get(shape, builder.getIntegerType(32));
       output = getZeroInitTensor(intAccumulatorType);
```
alheinecke, line 607: There seems to be still some empty tensor for int8, also for int8 there should not be a tensor allocated.

alheinecke (commented):
it seems i8 still allocates a tensor as well.

tools/mlir-gen/MLIRGen.cpp
shahidact: As discussed, For Int8, we do need to introduce i32 type empty tensor for accumulator, which will not be materialised as **alloc** by promoting it to a 32x32xTy alloca(). Hence, I have just removed this empty tensor for bf16 which is not required at all.

alheinecke: lgtm

rengolin (approved):

PR #1127, opened 2026-06-22
[tpp-run]: Support cache nuke using buffer replication.

shahidact:
For consistent and accurate benchmarking cold cache is desired. This patch implements the cache nuking by replicating the actual kernel calls into layers such that complete cache is invalidated. Layers are unrolled and each layer uses separate memory allocation for all of its arguments.

This patch also integrates this cache nuking in the middle of the pass pipeline by bifurcating the actual one into two and scheduling this pass after DefaultTppPasses to ensure bufferization is done to avoid allocation and copies inside benchmark loop. This is disabled by default and can be enabled with "--cache-nuke" flag.

### DIFF
--- a/include/TPP/Transforms/Utils/TensorInit.h
+++ b/include/TPP/Transforms/Utils/TensorInit.h
@@ -51,9 +51,17 @@ template <typename T> struct TensorInit : public ITensorInit {
 
     // For some reason, memref global op needs dense tensor type
     // See: lib/Dialect/MemRef/IR/MemRefOps.cpp :: GlobalOp::verify
-    auto tensorType =
-        mlir::RankedTensorType::get(shape.getShape(), shape.getElementType());
-    return mlir::DenseElementsAttr::get(tensorType, buffer);
+    mlir::ShapedType attrType;
+    if (mlir::isa<mlir::MemRefType, mlir::TensorType>(shape)) {
+      attrType =
+          mlir::RankedTensorType::get(shape.getShape(), shape.getElementType());
+    } else {
+      // Fixes this error "'arith.constant' op failed to verify that all of
+      // {value, result} have same type" as there is no need to convert it to
+      // ranked TensorType.
+      attrType = shape;
+    }
+    return mlir::DenseElementsAttr::get(attrType, buffer);
   }
 
 protected:
--- a/lib/TPP/LowerToLLVMPipeline.cpp
+++ b/lib/TPP/LowerToLLVMPipeline.cpp
@@ -1,0 +1,126 @@
+//===- LowerToLLVMPipeline.cpp -----------------------------------*- C++-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements the LowerToLLVMPipeline pass bundle, which performs
+// partial lowering and LLVM dialect conversion. This pass is designed to run
+// after bufferization (e.g., after DefaultTppPasses), allowing wrapper passes
+// to be inserted between tensor-to-memref conversion and final LLVM lowering.
+//
+//===----------------------------------------------------------------------===//
+
+#include "TPP/PassBundles.h"
+#include "TPP/PassUtils.h"
+#include "TPP/Passes.h"
+
+#include "mlir/Conversion/Passes.h"
+#include "mlir/Dialect/Arith/Transforms/Passes.h"
+#include "mlir/Dialect/Async/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/GPU/Transforms/Passes.h"
+#include "mlir/Dialect/Linalg/Passes.h"
+#include "mlir/Dialect/MemRef/Transforms/Passes.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/InitAllDialects.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/Passes.h"
+
+using namespace mlir;
+using namespace mlir::tpp;
+
+namespace mlir {
+namespace tpp {
+#define GEN_PASS_DEF_LOWERTOLLVMPIPELINE
+#include "TPP/PassBundles.h.inc"
+} // namespace tpp
+} // namespace mlir
+
+namespace {
+
+// The lowering pipeline from memref-level IR to LLVM dialect.
+struct LowerToLLVMPipeline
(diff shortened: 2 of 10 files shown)


lib/TPP/Runner/MLIRBench.cpp
```diff
@@ -243,7 +243,8 @@ LogicalResult MLIRBench::createKernelArgs() {
 
     // For quantized kernels output arguments, make the init type normal.
     auto funcType = kernel.getFunctionType();
-    auto numOutputs = funcType.getNumResults();
+    // For memref arguments in cache-nuke path, set numOutputs to 1.
```
rengolin, line 246: This will not just run on the cache-nuke path, but every where. Are you sure this semantics is correct for every usage of memref?

include/TPP/Transforms/Utils/TensorInit.h
```diff
@@ -51,9 +51,17 @@ template <typename T> struct TensorInit : public ITensorInit {
...
     // See: lib/Dialect/MemRef/IR/MemRefOps.cpp :: GlobalOp::verify
-    auto tensorType =
-        mlir::RankedTensorType::get(shape.getShape(), shape.getElementType());
-    return mlir::DenseElementsAttr::get(tensorType, buffer);
+    mlir::ShapedType attrType;
+    if (mlir::isa<mlir::MemRefType, mlir::TensorType>(shape)) {
+      attrType =
+          mlir::RankedTensorType::get(shape.getShape(), shape.getElementType());
+    } else {
+      // Fixes this error "'arith.constant' op failed to verify that all of
```
rengolin, line 59: I assume you are getting numeric globals here, in which case there's not "error", just that you can't convert a number to a tensor this way. I'd just remove this comment, as it may be confusing.

tools/tpp-run/tpp-run.cpp
```diff
@@ -205,11 +238,51 @@ static LogicalResult prepareMLIRKernel(Operation *op,
...
   wrapperOpts.identity = identity;
-  passManager.addPass(tpp::createTppRunnerWrapper(wrapperOpts));
 
-  tpp::DefaultPipelineOptions defPipelineOpts{defGpuBackend,
-                                              runnerCpuTargetFeature};
-  passManager.addPass(tpp::createDefaultPipeline(defPipelineOpts));
+  // Benchmarking with cache-nuking requires wrapper passes to run after
+  // bufferization to avoid memref.copy being inserted inside the timed region.
+  // Use split pipeline.
+  if (enableCacheNuking && benchNumLoops > 1) {
```
rengolin, line 245: We cannot have a special pipeline for benchmarking, as that'll skew the numbers and make it really hard to reproduce when we're not benchmarking. The order and the passes here are completely different and hard-coded. If I want to benchmark something different, this will override and do one thing only.

lib/TPP/LowerToLLVMPipeline.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // Otherwise, just run the cached one.
+    if (pm.empty())
+      constructPipeline();
+
+    if (failed(runPipeline(pm, module)))
+      return signalPassFailure();
+  }
+
+private:
+  void constructPipeline() override {
```
rengolin, line 69: Unclear why you need a whole new pipeline just for this.

rengolin (changes requested):

lib/TPP/Runner/MLIRBench.cpp
shahidact: As we have only one output in general, this will be semantically correct. However, I will see if I can guard with cache-nuke flag.

include/TPP/Transforms/Utils/TensorInit.h
shahidact: I was getting a dense initializer with vector type, that's why I got this error.

tools/tpp-run/tpp-run.cpp
shahidact: As such there is no change in pipeline semantically. Neither `TppRunWrapper` nor `ReplicateBuffersForBenchmark` does anything to the actual kernel to be benchmarked.

Why it is required?
In the beginning, `TppRunWrapper` pass in existing `tpp-run` driver pipeline was transforming the arguments of kernel `entry(A)` in this way:
tensor_A -> memref_A -> to_tensor(memref_A) and producing

func.func @entry() {
    %0 = memref.get_global @__wrapper_0 : memref<4x4xf32>
    %1 = bufferization.to_tensor %0 restrict writable : memref<4x4xf32> to tensor<4x4xf32>
    %2 = memref.get_global @__wrapper_1 : memref<4x4xf32>
    %3 = bufferization.to_tensor %2 restrict writable : memref<4x4xf32> to tensor<4x4xf32>
    %4 = memref.get_global @__wrapper_2 : memref<4x4xf32>
    %5 = bufferization.to_tensor %4 restrict writable : memref<4x4xf32> to tensor<4x4xf32>
    %c1_i64 = arith.constant 1 : i64
    **%6 = perf.bench(%c1_i64 : i64) -> f64 {
      %10 = func.call @_entry(%1, %3, %5) : (tensor<4x4xf32>, tensor<4x4xf32>, tensor<4x4xf32>) -> tensor<4x4xf32>**
    }
    %c10_i64 = arith.constant 10 : i64
    %7 = perf.bench(%c10_i64 : i64) -> f64 {
      %10 = func.call @_entry(%1, %3, %5) : (tensor<4x4xf32>, tensor<4x4xf32>, tensor<4x4xf32>) -> tensor<4x4xf32>
    }
    ...
    return
  }
perf.bench(n_iters)
 call _entry(to_tensor(memref_A)).

**However, subsequently after bufferization it gets transformed into:**

func.func @entry() {
    %cst = arith.constant 1.000000e+01 : f64
    %c10_i64 = arith.constant 10 : i64
    %c1_i64 = arith.constant 1 : i64
    %0 = memref.get_global @__wrapper_0 : memref<4x4xf32>
    %1 = memref.get_global @__wrapper_1 : memref<4x4xf32>
    %2 = memref.get_global @__wrapper_2 : memref<4x4xf32>
    %3 = perf.bench(%c1_i64 : i64) -> f64 {
      **%alloc = memref.alloc() {alignment = 64 : i64} : memref<4x4xf32>
      memref.copy %2, %alloc : memref<4x4xf32> to memref<4x4xf32>**
      **%6 = func.call @_entry(%0, %1, %alloc) : (memref<4x4xf32>, memref<4x4xf32>, memref<4x4xf32>) -> memref<4x4xf32>**
    }
    %4 = perf.bench(%c10_i64 : i64) -> f64 {
      %6 = func.call @_entry(%0, %1, %2) : (memref<4x4xf32>, memref<4x4xf32>, memref<4x4xf32>) -> memref<4x4xf32>
    }
    ...
    return
  }

To avoid these allocs() and copies(), we have to insert the TppRunnerWrapper and replication after DefaultTppPasses.

I think we can even make this new pipeline as default for both benchmarking and non-benchmarking cases.

rengolin: Any new pipeline needs to be run over all previous cases and make sure that the performance across all benchmarks are same or better before we can accept such a change.

PR #1130, opened 2026-06-25
Cold-cache Buffer Replication & Benchmark scripts for tracking sfc_ca_gemm

alheinecke:
### Summary
Adds the ability to benchmark kernels out of cold caches in tpp-run. Instead of timing the same buffers repeatedly (which keeps everything hot in L2/L3 and inflates throughput), the kernel's tensor arguments are replicated so their combined footprint exceeds the cache size, and each timed iteration runs on a distinct copy. This mirrors the "n_layers" replication used by libxsmm's sfc_ca_gemm cold-cache GEMM benchmark.

### Motivation
For memory-bound or large-GEMM workloads, repeatedly hitting the same hot buffers reports unrealistically high performance. Cold-cache replication produces timings representative of real workloads where weights/activations are not already resident in cache.

### How it works
- New -bench-replication-gb=<GiB> flag on tpp-run (0 = disabled). Only active on the benchmarking path (-n > 1); single runs keep original argument shapes.
- MLIRBench computes a replicationFactor so that factor × argument_footprint ≈ target GiB, stamps it onto the module as tpp.bench_replication_factor, and normalizes the reported per-call time by the factor in getTimerStats() (so throughput stays correct).
-  New replicate-bench-args pass (runs after bufferization, on memrefs) rewrites each perf.bench region:
-- Each kernel argument is backed by a flat, zero-initialized i8 global holding factor contiguous copies.
-- The kernel call is wrapped in an scf.for over the replica dimension; each iteration feeds the kernel a distinct memref.view into the buffer (pure pointer arithmetic, no per-iteration allocation or copy).
-- memref.view (identity layout, offset 0) is used instead of memref.subview so layout-sensitive ops in the kernel (e.g. the memref.expand_shape from bf16 VNNI packing) keep verifying.

### Random initialization
All-zero inputs let FMA units run at an unrealistically high clock. With random-init (auto-enabled via randomSplat), each float buffer is filled once before any timed region using a cheap counter-based MurmurHash3-style PRNG producing values in [1, 2) (guaranteed normal/finite). f32 and bf16 are handled explicitly; integer buffers keep the safe zero init.

### CAUTION
This PR should be merged after #1129 as otherwise the following plot cannot be created as the compiler crashes.  

image

### DIFF
--- a/include/TPP/Passes.td
+++ b/include/TPP/Passes.td
@@ -397,22 +397,35 @@ def ReplicateBenchArgs : Pass<"replicate-bench-args", "ModuleOp"> {
   let description = [{
     Operates after bufferization on the benchmark wrapper produced by tpp-run.
     For each `perf.bench` region, the (single) kernel call is wrapped in an
-    `scf.for` loop that iterates over a new outer "replica" dimension. Each
-    kernel argument is backed by a freshly allocated, value-initialized global
-    memref whose leading dimension is the replication factor; every iteration
-    feeds the kernel a distinct `memref.subview` (pure pointer arithmetic, no
-    allocation or copy). This mirrors the "n_layers" replication used by
-    libxsmm's cold-cache GEMM benchmark: the same problem is run on different
-    memory so the caches stay cold.
+    `scf.for` loop that iterates over a new "replica" dimension. Each kernel
+    argument is backed by a flat, zero-initialized `i8` global holding
+    `replication-factor` contiguous copies of the argument; every iteration
+    feeds the kernel a distinct `memref.view` into that buffer (pure pointer
+    arithmetic, no allocation or copy). `memref.view` is used instead of
+    `memref.subview` so that each replica keeps the original identity layout and
+    zero offset, which avoids breaking layout-sensitive ops in the kernel (such
+    as the `memref.expand_shape` introduced by bf16 VNNI packing). This mirrors
+    the "n_layers" replication used by libxsmm's cold-cache GEMM benchmark: the
+    same problem is run on different memory so the caches stay cold.
 
     The replication factor is taken from the `replication-factor` option when
     positive, otherwise from the `tpp.bench_replication_factor` module attribute
     set by the benchmark producer. A factor <= 1 disables the pass.
+
+    By default the buffers are zero initialized. When the `random-init` option
+    or the `tpp.bench_replication_random_init` module attribute is set, each
+    float buffer is filled once (before any timed region) with PRNG-generated
+    values in [1, 2); this avoids the unrealistically high clock that the FMA
+    units reach when fed all-zero inputs.
   }];
   let options = [
     Option<"replicationFactor", "replication-factor", "int64_t",
            /*default=*/"0",
            "Number of replicas (0 = read from module attribute).">,
+    Option<"randomInit", "random-init", "bool",
+           /*default=*/"false",
+           "Initialize the replicated buffers at runtime with random "
+           "floating-point values instead of zeros.">,
   ];
   let dependentDialects = ["memref::MemRefDialect",
                            "scf::SCFDialect",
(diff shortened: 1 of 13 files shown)


include/TPP/Passes.td
```diff
@@ -392,6 +392,46 @@ def Bufferize : Pass<"bufferize", "ModuleOp"> {
...
+    zero offset, which avoids breaking layout-sensitive ops in the kernel (such
+    as the `memref.expand_shape` introduced by bf16 VNNI packing). This mirrors
+    the "n_layers" replication used by libxsmm's cold-cache GEMM benchmark: the
+    same problem is run on different memory so the caches stay cold.
+
+    The replication factor is taken from the `replication-factor` option when
+    positive, otherwise from the `tpp.bench_replication_factor` module attribute
+    set by the benchmark producer. A factor <= 1 disables the pass.
+
+    By default the buffers are zero initialized. When the `random-init` option
```
rengolin: Can we make the default 1 instead of 0? Avoids the hardware shortcut problem and it's easy to represent as a splat. Even if the results are bogus and hard to verify (all the same), it's better than zero.

lib/TPP/Runner/MLIRBench.cpp
```diff
@@ -226,6 +230,56 @@ LogicalResult MLIRBench::createKernelArgs() {
...
+  // used by libxsmm's sfc_ca_gemm benchmark to defeat large L3 caches and
+  // measure cold-cache GEMM performance.
+  replicationFactor = 1;
+  if (replicationTargetGiB > 0.0) {
+    double footprintBytes = 0.0;
+    for (auto argTy : argTypes) {
+      auto shaped = dyn_cast<ShapedType>(argTy);
+      if (!shaped)
+        continue;
+      if (!isa<TensorType>(argTy))
```
rengolin: I'd put all these three errors into one if statement and `emitError` like "Replication benchmarking requires statically tensor arguments".

lib/TPP/Transforms/ReplicateBenchArgs.cpp
```diff
@@ -0,0 +1,295 @@
...
+// for any other float type it falls back to a constant 1.0.
+static Value emitRandomFloat(OpBuilder &b, Location loc, Value linearIdx,
+                             Type elemTy) {
+  Type i32Ty = b.getI32Type();
+  auto c32 = [&](uint32_t v) -> Value {
+    return arith::ConstantOp::create(
+        b, loc, b.getIntegerAttr(i32Ty, static_cast<int32_t>(v)));
+  };
+
+  // MurmurHash3-style finalizer using modulo-2^32 arithmetic.
```
rengolin, line 76: We already use C++ new random library in the initializer, could we use it here, too?

https://en.cppreference.com/cpp/numeric/random

rengolin (commented):
Some comments inline, otherwise, looks good to me.

lib/TPP/Transforms/ReplicateBenchArgs.cpp
alheinecke: This is happening at runtime (as we increase the array size) hence I did something simple we can do directly in IR. So it's different than the static init for correctness checks. Of course I can look into functional calls to the C++ standard library but I felt this is not worth the effort, as we will never validate the outputs. It's simply to make sure all the mantissa bits are full and the FMA burn high power.

lib/TPP/Runner/MLIRBench.cpp
alheinecke: done

include/TPP/Passes.td
alheinecke: 1.0 for float now. I had 0 as this is much easier for the i8 cast. Hope this is now correct :-)

rengolin (approved):

PR #1131, opened 2026-06-26
Revert "Support simple elemntwise tiling on linalg.generic. (#1123)"

rengolin:
This reverts commit 1c702d59052f521e6fff169d157ca35b9c8443b8. #1123 

The base benchmarks are broken since 18th May and I have bisected to this commit.
https://github.com/libxsmm/tpp-mlir/actions/runs/28251961972/job/83705478358

This is getting in the way of other developers, so I'll revert for now and we can re-merge again with a fixed version.


alheinecke (approved):
yepp let's focus on stable CI

Issue #1121, opened 2026-04-22
Merge linalg-to-vector and nano-kernel pipelines


rengolin:
We need to clean up the vectorizer pipeline, merge the similar paths into one and make it the default pass, while making the xsmm the legacy pipeline for performance conparison.

We should also remove the instruction list to avoid between xsmm and vector.

rengolin:
Here are the two lists of passes with each flag:

# vector-to-kernels

## Pre-bufferization

* FoldAddIntoDest
* FoldIntoEltwise
* ConvertLinalgToInplace (f)
* RewriteBatchMatmulToMatmul
* TppMapping
* LowerPacksAndUnPacks
* Cleanup
* DecomposeAggregatedOps (f)
* SCFForAllLoopFlattenSFC

## Bufferize ← runs here

* ReplicateBenchArgs
* LinalgLowering (f)
* BrgemmLinalgTiling (f)
* LoopInvariantCodeMotion (f)
* TileElementWiseOps (f)
* VectorizationPass (f)

## Vector To Kernel Bundle

* HoistVectorTransfers
* VectorContractToAMX (f)
* MicroKernels (f)
* MicroKernelsAMX (f)
* Canonicalizer (f)

## Post-processing

* Cleanup
* ConvertForAllToParallelOp
* ConvertVectorToSCF
* IntelAMXTileConfigInsertion (f)
* Canonicalizer (f)
* LoopInvariantCodeMotion (f)
* Canonicalizer (f)
* IntelAMXTileConfigHoisting (f)
* ConvertXsmmToFunc
* LocalDialectsLowering
* Postprocessing (f)

## Partial lowering

Identical in both cases

# Nano Kernel

## Pre-bufferization

* FoldAddIntoDest
* FoldIntoEltwise
* ConvertLinalgToInplace (f)
* RewriteBatchMatmulToMatmul
* TppMapping
* LowerPacksAndUnPacks
* Cleanup
* DecomposeAggregatedOps (f)
* SCFForAllLoopFlattenSFC

### Bufferization skipped here!

* ReplicateBenchArgs
* LinalgLowering (f)
* LinalgGeneralizeNamedOps (f)
* ConvertLinalgGenericTo32BitAccumulation (f)
* BrgemmLinalgTiling (f)
* LoopInvariantCodeMotion (f)
* TileElementWiseOps (f)
* VectorizationPass (f)

## Nano Kernel bundle

* RegisterUnroll (f)
[...]

PR #1133, opened 2026-06-28
Retry to support simple elemntwise tiling on linalg.generic.

shahidact:
Element wise operations on bigger tensor shapes results in bad code
generation with inserts/extracts and spill. This pass tries to tile such
operation so that backend could produce nice compact code without
spills.

Revert a couple of tests to use vector-to-kernels bundle to fix CI.


benchmarks/config/base/base.json
```diff
@@ -66,9 +66,9 @@
     },
     "gemm_bf16_vnni_dp2_mlir_nano_kernel_avx512": {
       "type": "IR-GEN",
-      "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=512 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2 --registerUnroll=1,16,1" ],
+      "benchmark": [ "mlir-gen", "--kernel=const --float-type=bf16 --batch=512 --layers=1024,1024,1024,1024 --tiles=32,32,32 --vnni=2" ],
       "environment": {},
-      "flags": [ "-n", "100", "-run-args='--nano-kernels --registerBlocking=8,32,2'" ],
+      "flags": [ "-n", "100", "-run-args='--vector-to-kernels --registerBlocking=8,32,2'" ],
```
rengolin, line 71: Talking to Arun about this, we really need to remove all those variants and leave one. The fact that this is getting in the way of an unrelated patch makes it even more clear. Let's do this before we add new features.

shahidact: > Talking to Arun about this, we really need to remove all those variants and leave one. The fact that this is getting in the way of an unrelated patch makes it even more clear. Let's do this before we add new features.

Do you mean that I should completely remove nano-kernel path from all benchmarks in a separate PR?

rengolin:
@alheinecke make sure this doesn't break your work

alheinecke (commented):
This doesn't seem to provide support in a clean-way as the future default (nano-kernels) is disabled. Do we have an issue with detailed items that actually need to happen to make this right? If not, please create one.

shahidact:
> This doesn't seem to provide support in a clean-way as the future default (nano-kernels) is disabled. Do we have an issue with detailed items that actually need to happen to make this right? If not, please create one.

Created an issue.

rengolin:
We already have an issue in this repository: #1121

rengolin:
Benchmark numbers are good in CI.

rengolin (approved):

PR #1134, opened 2026-06-29
[Cleanup] Remove old vector-to-xsmm path and skip ops

rengolin:
This partially implements #1121, removing the old vector-to-xsmm path that is no longer used or properly tested.

It also removes the skip-operations list that was no longer used and just useful for the vector-to-xsmm path.


alheinecke (approved):
LGTM, I assume you validated LinAlg->XSMM and LinAlg->Vector->Nanokernel remain both functional and performant after we have merged this.

rengolin:
Benchmark numbers look sane

PR #1135, opened 2026-07-02
Fix replication and initialization for perf.

shahidact:
Static global argument replication and intialization could not simulate the cold cache for int8 kernel for bigger M,N sizes. This patch fixes it by using dynamic allocation alongwith copying the data from the initialized wrapper globals corresponding to kernel arguments for every replication.

Adds additional perf and correctness scripts.


lib/TPP/Transforms/ReplicateBenchArgs.cpp
```diff
@@ -148,26 +104,48 @@ struct ReplicateBenchArgs
...
-    // init still serves as a safe default for any element type the runtime
-    // fill does not cover (e.g. integers).
-    OpBuilder globalBuilder(ctx);
-    globalBuilder.setInsertionPointToStart(module.getBody());
-    Type i8Ty = globalBuilder.getI8Type();
-    SmallVector<StringRef> globalNames(origInputs.size());
-    SmallVector<int64_t> replicaByteSizes(origInputs.size());
-    SmallVector<int64_t> totalElemCounts(origInputs.size());
-    auto alignment = globalBuilder.getI64IntegerAttr(128);
+    // Trace the original kernel call's operands back to their source globals.
```
rengolin, line 110: Doesn't that mean you'll be running the cache nuke with a repeat of the benchmark data (over and over) and then not really nuking the cache?

benchmarks/mlir/sfc_ca_gemm/run_tpp_vector_gemm_bf16.sh
```diff
@@ -96,7 +96,7 @@ for M in "${SIZES[@]}"; do
 
       echo "=== Running   M=${M} N=${N} K=${K} ==="
       run_out="$("${TPP_RUN}" "${mlir_file}" \
-        -e entry -entry-point-result=void \
+        -e entry -entry-point-result=void --vector-to-kernels --registerBlocking=32,32,32 \
```
rengolin: Do we know if we have a good lowering for SFC on vectors for now (for any M,N,K shape)?

lib/TPP/Transforms/ReplicateBenchArgs.cpp
shahidact: No, data is same, but memory is different.

benchmarks/mlir/sfc_ca_gemm/run_tpp_vector_gemm_bf16.sh
shahidact: As this is based on our vectorization work, whatever shape it can handle well is good.

rengolin: My point is that the original test was lowering to XSMM, and you changed it. This changes the baseline.

shahidact: Ah, I got your point. I will update.

shahidact:
This PR is no longer required as existing cold cache feature works well with proper K-blocking value for int8 i.e K=64.

PR #1141, opened 2026-07-16
[Cleanup] Decompose pipeline passes (NFC)

rengolin:
Simplify pipeline by decomposing into different functions, joining diversion flags.

Merging vector-to-kernels and nano-kernels are next.

### DIFF
--- a/lib/TPP/DefaultPipeline.cpp
+++ b/lib/TPP/DefaultPipeline.cpp
@@ -169,7 +169,26 @@ private:
     if (print == PrintStage::Early)
       pm.addPass(createPrintIRPass());
 
-    // Apply the default preprocessing pass
+    addDefaultTppPasses();
+
+    if (print == PrintStage::Mid)
+      pm.addPass(createPrintIRPass());
+
+    addPartialLoweringPasses();
+
+    // Print IR of optimized kernel and main
+    if (print == PrintStage::Late)
+      pm.addPass(createPrintIRPass());
+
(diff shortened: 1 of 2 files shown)


lib/TPP/DefaultPipeline.cpp
```diff
@@ -202,26 +219,22 @@ struct DefaultPipeline : public tpp::impl::DefaultPipelineBase<DefaultPipeline>,
...
     pm.addPass(createLowerAffinePass());
+  }
 
-    // Print IR of optimized kernel and main
-    if (print == PrintStage::Late)
-      pm.addPass(createPrintIRPass());
-
-    // Lower to LLVM
+  // Lower the remaining dialects to the LLVM dialect.
+  void addLowerToLLVMPasses() {
```
adam-smnk, line 225: To need for GPU to llvm might be just some hidden cross dependency.

I'd try removing most of these individual passes and using `createConvertToLLVMPass`.
A few might be needed like vector to LLVM with extra options but it should greatly simplify the pass ordering.

rengolin: I'll leave that for after I remove the vector-to-kernel option, as I believe it's broken and we should all use the new nano-kernel one instead. This bug only occurs in two identical tests added recently, which I'm not sure are correct to begin with.

rengolin: FYI, these are the tests: #1140

PR #1142, opened 2026-07-16
[Cleanup] Steps moving into nano-kernel direction

rengolin:
These are tree separate and unrelated commits that moves us closer to unifying, but there are still some blockers.

The main problem is that lowering the same test through `--nano-kernel` (even when using the appropriate `--gemm-unroll` factors), I get unrealized cast problems.

Claude analysis says:
* `vector-contract-to-amx-flat.mlir` — non-VNNI bf16; needs the x86-fork bf16 inline-packing patterns (enabling them for CPX_SPR only lowered ~⅓ of the contracts).
* The `mlp_splat` case in the bf16-uKernels file and both ARL cases — same non-VNNI limitation.
* `vector-contract-to-ukernels-i8.mlir` and the two I8/Integration/amx/tpp-run-gemm-dequantize-* tests — nano emits X86ISD::VPDPBSSD ("Cannot select").

Benchmarks either get significantly slower (~20%) or crash with unrealized casts, so I have not moved them over.

I have added some more context on #1121.


rengolin:
@arun-thmn after we merge this PR, I'll need your help to untangle some assumptions that were not clear to me (or to Claude). The main thing is the difference between the old way (bufferize then vectorize) to the new way (vectorize, then bufferize, then vectorize).

Some of the old tests may be just broken and we need to check if they're still valid before we judge if they need to be moved over or not. @shahidact your quantization tests break with nano-kernels, and they need a GPU pass for some reason. I think it's an artifact that we picked up by not properly splitting the GPU pipeline form the CPU one. @adam-smnk.

My plan is to look at the tests one more time and scan what we want to keep and what we want to just throw away, and then try to move the rest to nano-kernels. 

Then we look at benchmarks...

arun-thmn:
> @arun-thmn after we merge this PR, I'll need your help to untangle some assumptions that were not clear to me (or to Claude). The main thing is the difference between the old way (bufferize then vectorize) to the new way (vectorize, then bufferize, then vectorize).

The main reason is the upstream `hoisting` pass works only on `tensors`. So, after tiling we `vectorize->hoisting->bufferize`.

PR #1144, opened 2026-07-20
[Benchmarks] Major renaming

rengolin:
Renaming files, benchmarks, re-organizing things, but should not make any difference in the numbers themselves.

Also deleted the GPU benchmarks (unused), packing benchmarks (irrelevant), increased AMX coverage.

This is in preparation for moving the argument `--vector-to-kernel` to `--nano-kernel`, which some benchmarks are already using (FP32).

Ran full benchmarks on EMR, ARL, Zen and Graviton. All green.


rengolin:
Arm benchmarks: https://github.com/libxsmm/tpp-mlir/actions/runs/29731363764

Full benchmarks: https://github.com/libxsmm/tpp-mlir/actions/runs/29729753681

PR #1145, opened 2026-07-20
[nK] Fix broken XSMM generation for nano-kernels

rengolin:
The XSMM lowering was unguarded and always running, but somehow in the previous pipeline it wasn't generating calls to xsmm on the nano-kernel path. Since this path isn't well tested (especially IR checks), my previous PR allowed XSMM calls to be generated in the nano-kernel path and the Integration tests "passed" because they were not generating nano-kernels at all.

I also had to revert one test that moved on to nano-kernels (actually, xsmm calls) back to `vector-to-kernels`.

### DIFF
--- a/lib/TPP/DefaultTppPasses.cpp
+++ b/lib/TPP/DefaultTppPasses.cpp
@@ -69,13 +69,6 @@ struct DefaultTppPasses
   }
 
 private:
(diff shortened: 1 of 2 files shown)


rengolin:
Ref: #1121

@arun-thmn FYI

rengolin:
Benchmark numbers back to normal

PR #1146, opened 2026-07-20
[nK] Separating more the XSMM/Loops pipeline from nano-kernel

rengolin:
Also, moving some benchmarks that already worked with nano-kernels.


rengolin:
Benchmark numbers look sane

PR #1151, opened 2026-07-25
bump libxsmm to release 2.1.0

alheinecke:


alheinecke:
@rengolin : Is this exposing the same bug we discussed before?

rengolin:
I believe so. Though, this test is weird, not sure why it runs `tpp-opt --pack-vnni` before the default (XSMM) path, which should be the default behaviour anyway.

I'm also seeing the precision problem when trying to fix #1121, so there's something troubling somewhere.

alheinecke:
This was a real bug, as tpp-mlir doesn't use libxsmm enums, but their values and those changed in libxsmm version 2+.

rengolin (approved):

PR #1159, opened 2026-07-28
[Cleanup] Change sweep benchmark scripts to use nano-kernels

rengolin:


rengolin:
This is NOT tested. Please look carefully and do local testing to make sure it is correct.

@shahidact make sure this still runs to completion on your setup.

@arun-thmn check that the `gemm-unroll` parameters are correct for `i8` and `bf16` on AMX.

arun-thmn:
> This is NOT tested. Please look carefully and do local testing to make sure it is correct.
> 
> @shahidact make sure this still runs to completion on your setup.
> 
> @arun-thmn check that the `gemm-unroll` parameters are correct for `i8` and `bf16` on AMX.

I can confirm the `gemm-unroll` and `blocking` parameters are correct. Just make sure the `M` and `N` values are greater than `blocking` size. I am making a patch in the upstream to fix the case where `canonization` results in pattern mis-match.

arun-thmn (approved):
I'm good. Lets wait for @shahidact to make sure he could run the bench.

scripts/benchmarks/run-sweep.sh
```diff
@@ -132,8 +132,8 @@ if [[ "${SKIP_CORRECTNESS}" -eq 0 ]]; then
   # --seed=123 makes both baseline and test runs see identical inputs.
   gen_args_i8="--kernel=args --data-type=mx-i8-f32 --tiles=32,32,64 --vnni=4 --quant-type=dequantize --seed=123"
   gen_args_bf16="--kernel=args --data-type=mx-bf16 --tiles=32,32,32 --vnni=2 --seed=123"
-  vk_args_i8="--def-parallel --vector-to-kernels --registerBlocking=32,32,64 --sfc-order=true"
-  vk_args_bf16="--def-parallel --vector-to-kernels --registerBlocking=32,32,32 --sfc-order=true"
+  vk_args_i8="--def-parallel --nano-kernels --registerBlocking=32,32,64 --gemm-unroll=16,16,16 --sfc-order=true"
```
shahidact, line 135: Why are we adding this flag, "--gemm-unroll=16,16,16", here?

scripts/benchmarks/gen-sweep-config.py
```diff
@@ -31,7 +31,7 @@
             "--quant-type=dequantize"
         ),
         "run_args": (
-            "--def-parallel --vector-to-kernels "
+            "--def-parallel --nano-kernels --gemm-unrolling=16,16,16 "
```
shahidact, line 34: Why are we adding this flag, "--gemm-unroll=16,16,16", here?

shahidact:
> @shahidact make sure this still runs to completion on your setup.

Yes, it runs successfully.

scripts/benchmarks/run-sweep.sh
arun-thmn: It is to `unroll` the contract operation. Earlier we did with the help of `dlti` attribute through `mlir-gen`. As the IR from other inputs like `openVINO` or `torch-mlir` not necessarily add the `dlti` attribute we pass those values in a separate flag value for now.

shahidact: Do you mean this flag was implied with "vector-to-kernels" path?  

For this script, I had not chosen this flag or attribute for my benchmark generation and sweep.

rengolin: It's necessary for now. @arun-thmn will fix this soon.

shahidact: Ok

PR #1160, opened 2026-07-31
Large refactoring of MLIR-Gen, benchmarks and tests to move vector-to-kernel to nano-kernel

rengolin:
Basically, I need to support output types different than accumulation type to enable bf16->fp32 gemms. So I had to add this lowering to XSMM, and change how mlir-gen generates the IR in the first place, and also how the default tpp passes pack VNNI, which needs to be done in generics.

Along the way I found out mlir-gen was completely broken, saving `args` state in local stack variables and then having to redo the work later, so I had to completely rewrite how we handle the arguments throughout the module. Now it's all in the same struct, and changes are persisted on a layer per layer basis.

I would have stopped there, but this work has made the `--vector-to-kernels` option unstable and benchmarks would crash. The new `--nano-kernels` path was stable, so I had to move all benchmarks, which was the original piece of work in the first place. Right now, there are no tests or benchmarks that use it, but the code is still there. I'll leave it as is for now, and we can remove later.

Some benchmarks still crash with nano-kernel options, which may be related to the upstream implementation (at least that's where copilot went when hallucinating solutions), so I decided to stop for now and persist the current changes. We need to continue working towards nano-kernels.

#1121


rengolin:
Baseline to compare to:
https://github.com/libxsmm/tpp-mlir/actions/runs/30435634775/job/90522632518

This benchmark:
https://github.com/libxsmm/tpp-mlir/actions/runs/30656042521/job/91240602867

rengolin:
ARL BASE:
* Time out after 30 minutes.
* That's too long, something is stuck in a loop.
* https://github.com/libxsmm/tpp-mlir/actions/runs/30656042521/job/91240602923#step:3:415

EMR OMP:
* ` error: LLVM Translation failed for operation: builtin.unrealized_conversion_cast`
* Test: `config/omp/mlir-vector-bf16.json : gemm_bf16_dp2_mlir_vector_amx`
* https://github.com/libxsmm/tpp-mlir/actions/runs/30656042521/job/91240602915#step:3:731

Everything else passes.

rengolin:
Performance comparison of the benchmarks that did run.

## EMR Base

Benchmark | Old | New | Change
-- | -- | -- | --
mlir_gemm_bf16_vnni_dp2_xsmm | 966.287 | 2.357 | -99.76%
mlir_mlp_bf16_vnni_dp2_xsmm | 925.886 | 2.364 | -99.74%
mlir_mlp_bf16_splat_dp2_xsmm | 812.907 | 632.249 | -22.22%

This has definitely something to do with the new XSMM lowering, will fix.

## EMR OMP

The ones that ran, look good.

Crash on `config/omp/mlir-vector-bf16.json : gemm_bf16_dp2_mlir_vector_amx`.

Probably a similar problem.

## Zen Base

Using an older run for Zen:

Benchmark | Old | New | Change
-- | -- | -- | --
mlir_gemm_bf16_vnni_dp2_xsmm | 681.252 | 6.803 | -99.0%
mlir_mlp_bf16_vnni_dp2_xsmm | 675.744 | 6.810 | -99.0%

Similar to EMR, so good news.

## Zen OMP

Benchmark | Old | New | Change
-- | -- | -- | --
TPP-MLIR XSMM BF16 VNNI DP2 (2 threads) | 1351.735 | 10.451 | -99.2%
TPP-MLIR XSMM BF16 VNNI DP2 (4 threads) | 2663.402 | 20.527 | -99.2%
TPP-MLIR XSMM BF16 VNNI DP2 (8 threads) | 5143.320 | 40.071 | -99.2%
TPP-MLIR XSMM BF16 VNNI DP2 (16 threads) | 7899.749 | 73.384 | -99.1%
TPP-MLIR XSMM BF16 MLP DP2 (2 threads) | 1347.981 | 10.430 | -99.2%
TPP-MLIR XSMM BF16 MLP DP2 (4 threads) | 2657.697 | 20.518 | -99.2%
TPP-MLIR XSMM BF16 MLP DP2 (8 threads) | 5156.247 | 39.987 | -99.2%
TPP-MLIR XSMM BF16 MLP DP2 (16 threads) | 8053.177 | 73.116 | -99.1%

Same as the others, seems to be a single problem with XSMM lowering.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -559,77 +604,56 @@ void MLIRGenerator::computeElementwiseScalingFlops(ShapedType outputShape) {
...
-      // Get integer tensor accumulator type for dequantization.
-      TensorType intAccumulatorType =
-          RankedTensorType::get(shape, builder.getIntegerType(32));
-      output = getZeroInitTensor(intAccumulatorType);
-    }
+    // Quantization casts the result later; accumulate in the input type.
+    args.accumulator.value = getZeroInitTensor(
+        RankedTensorType::get(shape, inputType.getElementType()));
+  } else if ((quantType == QuantizationType::Dequant ||
+              quantType == QuantizationType::Mixed) &&
```
shahidact: https://github.com/libxsmm/tpp-mlir/pull/1137 

Above PR removes this condition to use output argument instead of creating empty tensor.

rengolin: merged your PR, will continue investigating, thanks!

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -112,6 +112,44 @@ static Value createCastToFloat(OpBuilder &builder, Location loc, Value value,
...
+  auto accTy = cast<ShapedType>(accumulator.getType());
+  Type accElem = accTy.getElementType();
+  Type outElem = outputType.getElementType();
+  bool sameFloat = outElem.isFloat() && accElem.isFloat();
+  bool sameInt = outElem.isInteger() && accElem.isInteger();
+  if (accElem == outElem || (!sameFloat && !sameInt))
+    return accumulator;
+
+  auto resultTy = RankedTensorType::get(accTy.getShape(), outElem);
+  Value dest = tensor::EmptyOp::create(builder, loc, resultTy, ValueRange{});
```
shahidact, line 133: Why do we need to create new tensor, can't we use output.value as the type is same?

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -559,77 +604,56 @@ void MLIRGenerator::computeElementwiseScalingFlops(ShapedType outputShape) {
...
+         "Unsupported Lowering for VNNI/input rank > 2. "
+         "Try 'generic' or 'contract' lowering");
+
+  Value output = linalg::MatmulOp::create(
+                     builder, loc, TypeRange{args.accumulator.value.getType()},
+                     ValueRange{chain, args.weight.value},
+                     ValueRange{args.accumulator.value})
+                     .getResult(0);
+
+  return downcastToOutput(builder, loc, output, args.output.type);
```
shahidact: Which test covers this?

lib/TPP/PassBundles/TppMapping.cpp
```diff
@@ -56,6 +56,12 @@ struct TppMapping : public tpp::impl::TppMappingBase<TppMapping>,
     // Canonicalize.
     pm.addPass(createCleanup());
 
+    // Packing only works on generic matmuls
+    LinalgMorphOpsPassOptions options;
+    options.namedToGeneric = true;
+    options.categoryToGeneric = true;
+    pm.addPass(createLinalgMorphOpsPass(options));
```
shahidact, line 63: Why do we need this pass?

shahidact (commented):
Few minor comments. Mostly looks ok.

lib/TPP/PassBundles/TppMapping.cpp
rengolin: Because the packing passes don't work with contract for some cases. Instead of trying to fix them in this huge PR, I decided to convert all to generic for now.

tools/mlir-gen/MLIRGen.cpp
rengolin: Benchmarks in `base.json`. I can add a quick test to `mlir-gen.mlir` too.

tools/mlir-gen/MLIRGen.cpp
rengolin: Good point, let me change this.

lib/TPP/PassBundles/TppMapping.cpp
rengolin: More to this point, TileAndFuse pass has a whole back and forth with named vs generic forms from 3 years ago before we had this upstream. This needs to be cleaned, but again, not for this PR.

Created a new issue for this #1162

tools/mlir-gen/MLIRGen.cpp
rengolin: Oh, because in the new impl, the `args.output.value` hasn't been created yet. At this point, only `args.accumulator.value` was created in `lowerMatmul`. What I can do to make it simpler is to move that logic back to the outer function (`lowerMatmul`) and so you can see the reasoning.

rengolin:
The recent state seems to revert all crashes and performance problems. It does not, however, move _every_ benchmark to the `nano-kernel` path.

@arun-thmn, Copilot hinted at problematic support for the bf16 -> fp32 case in the vector to AMX nano kernel pass upstream. We need to revisit this soon.

@shahidact, any work that you're doing against `--vector-to-kernel` need to move to `--nano-kernels`. We'll also need to move the previous quantization examples / benchmarks to nano-kernel, which may break things. Copilot has identified a number of quantization changes in VectorToAMX that did not exist in VectorToNanoKernel. We need to bridge that gap before completely moving.

adam-smnk (approved):

PR #1166, opened 2026-08-04
[Cleanup] Remove vector-to-kernel path

### LINKED ISSUE #1121: Merge linalg-to-vector and nano-kernel pipelines

rengolin:
Completely remove everything to do with the old vector-to-kernel pass. We should all be using the new nano-kernel path from now on.

Fixes #1121


arun-thmn:
The below files can also be removed along with their declaration in passes.td file:

1. https://github.com/libxsmm/tpp-mlir/blob/main/lib/TPP/Transforms/VectorContractToAMX.cpp
2. https://github.com/libxsmm/tpp-mlir/blob/main/lib/TPP/Transforms/VectorContractToBF16DotProduct.cpp
3. https://github.com/libxsmm/tpp-mlir/blob/main/lib/TPP/Transforms/VectorContractToFMA.cpp

arun-thmn (approved):
Thanks for the cleanup.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -99,28 +96,16 @@ struct DefaultTppPasses
...
-      options.vecBundleCpuTargetFeature = defBundleCpuTargetFeature;
-      pm.addPass(createVectorToKernel(options));
-    }
+    tpp::RegisterUnrollOptions unrollOpts;
+    unrollOpts.gemmUnroll = SmallVector<int64_t>{*gemmUnroll};
+    pm.addNestedPass<func::FuncOp>(createRegisterUnroll(unrollOpts));
+    pm.addNestedPass<func::FuncOp>(createHoistLoopInvariantSubsets());
+    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+    pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
+    pm.addPass(createBufferize());
```
shahidact, line 105: ReplicateBenchArgs pass also needs to be moved here.

rengolin: please submit a PR

PR #1168, opened 2026-08-05
[mlir-gen] Support for multiple matmul variant

arun-thmn:
This patch adds `mlir-gen` to generate below variants of `matmul` for `bf16`, `i8`, `hf8`, and `bf8`.

-  A VNNI+Transpose, B VNNI
-  A Flat + Transpose, B Flat
-  A Flat, B Flat + Transpose
-  A Flat + Transpose, B Flat + Transpose

ps: `mlir-gen` already supports below variant:

- A Flat, B VNNI
- A Flat, B Flat.


tools/mlir-gen/MLIRGen.h
```diff
@@ -113,6 +113,21 @@ class MLIRGenerator {
...
+  /// Store the A (input) operand transposed
+  bool transposeA;
+
+  /// Store the B (weight) operand transposed
+  bool transposeB;
+
+  /// Whether the current layer's input operand is stored transposed. Only the
+  /// first layer consumes the external (transposed) A argument; intermediate
+  /// activations are produced in normal layout.
+  bool transposeActiveInput = false;
```
rengolin: This needs to be a property of the `args` struct and populated by `getKernelTypes`. We should keep struct-level arguments to whole-graph properties only.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -612,6 +637,20 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args) {
   auto outputType = cast<ShapedType>(args.output.value.getType());
   auto shape = outputType.getShape();
 
+  // Every layer reads a transposed A. For the plain GEMM each producing layer
+  // (except the last) stores its output transposed so the next layer can read
+  // it as a transposed input. VNNI transposed A uses a pre-packed 5D argument
+  // that is not reachable by reshaping an activation, so only the first layer
+  // reads a transposed input there.
+  bool isLastMatmul = args.index == layers.size() - 1;
```
rengolin: see my other comment about layer properties

tools/mlir-gen/MLIRGen.h
arun-thmn: Okay, moved these into the `getKernelTypes`

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Yep, moved these logc into `getKernelTypes`

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -46,6 +46,15 @@ void parseStringList(StringRef str, SmallVector<int64_t> &list) {
 }
 
 /// Returns the vector of boolean for the required broadcast dimensions
+// Swap outer/inner tile pairs of a packed 4D type: [a,b,c,d] -> [b,a,d,c].
```
rengolin: left comment from the next function above

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -630,7 +668,7 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args) {
     args.accumulator.value = getZeroInitTensor(args.accumulator.type);
   }
 
-  if (vnniPacked) {
+  if (vnniPacked && !args.transposeInput) {
```
rengolin: should we emit an error here? Where is the code that handles vnni + transposed input?

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -336,6 +355,20 @@ void MLIRGenerator::getKernelTypes(KernelArgs &args) {
     // type AND we want to propagate the truncation through the element-wise ops.
     arg.bias.type = getShape({outputSize}, PACK_OUTPUT);
```
rengolin: bias will also need to be transposed on `transposeOutput`?

tools/mlir-gen/MLIRGen.cpp
rengolin, line 311: `transposeInput` will also need to change this, at least for the first layer.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -1265,6 +1314,14 @@ TensorType MLIRGenerator::getShape(ArrayRef<int64_t> dims, PackingType type) {
   case PACK_INPUT:
     assert(x % n == 0 && "Invalid tile size for N dim");
     assert(y % c == 0 && "Invalid tile size for C dim");
+    if (transposeA) {
```
rengolin, line 1254: can you not just get the type and then transpose it before returning?

rengolin (commented):
How did you test this? Right now there's only a LIT check against some IR that is hard to parse.

It would be good to verify correctness comparing output with libxsmm / pytorch (whatever is easier and yields the same variations).

arun-thmn:
> How did you test this? Right now there's only a LIT check against some IR that is hard to parse.
> 
> It would be good to verify correctness comparing output with libxsmm / pytorch (whatever is easier and yields the same variations).

True. I did now added `tpp-run` tests (like the earlier ones) for the `bf16`, `hf8`, and `bf8`.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: I think, `bias` need not to be transposed. It is an `mxn` matrix.
And the output is not `transposeOutput`. Both `libxsmm` and `nano-kernels (ACE)` should handle the transpose/non-transpose inputs accordingly and the output is always non-transpose (`mxn`).

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Fixed the comment correctly. Moved the 1st comment to the earlier function.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: The refactor cleaned-up few. The code for vnni + transpose starts at : 675.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Okay. Did it.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: We are handling in the later code.

alheinecke:
how does MLP work? E.g. when have have A in VNNI^T, B in VNNI and C is flat then for a chain of GEMMs/MLPs we need to store C in VNNI^T as this (in our current flow) becomes the next A. Same is true for all other pre-format combinations, I just took the first one of you list.

arun-thmn:
> how does MLP work? E.g. when have have A in VNNI^T, B in VNNI and C is flat then for a chain of GEMMs/MLPs we need to store C in VNNI^T as this (in our current flow) becomes the next A. Same is true for all other pre-format combinations, I just took the first one of you list.

For the Multi-gemm, I am following the below approach, please let me know whether it is correct. I can update if you need modification.

For example, Let us take `A Flat+transpose` and `B Flat+transpose` case (`row-major order`). The `mlir-gen` will emit below IR for a two-layer.

`func @gemm(arg0{A Flat+transpose}, arg1{B1 Flat+transpose}, arg2{C1 Mat}, arg3{B1 Flat+transpose}, arg4{C2 Mat}) {`
  `// 1st gemm falls into A Flat+transpose and B Flat+transpose case`
 ` gemm1 = ( A Flat+transpose x B1 Flat+transpose ) + C1 Mat.`
  `// 2nd gemm falls into A Flat and B Flat+transpose case. As the gemm1 result is not transposed.`
  `gemm2 = ( gemm1 Flat x B1 Flat+transpose ) + C1 Mat.`
  `return gemm2;`
`}`
Subsequently all gemm from `2nd layer` onwards will be a `A Flat x B Flat+transpose` case.

This similar for `A VNNI+Transpose` case. The first gemm will be in case `A VNNI_Transpose` from the next gemm onwards it is `A VNNI` case.

arun-thmn:
New multi-gemm IR using `mlir-gen`
1. Like `torch-mlir` `A` Flat+Transpose and `B` Flat+Transpose: [TT-pytorch.txt](https://github.com/user-attachments/files/30984038/TT-pytorch.txt)
2.  Like `libxsmm_dnn` `A` Flat+Transpose and `B` Flat+Transpose: [TT-libxsmm_dnn.txt](https://github.com/user-attachments/files/30984046/TT-libxsmm_dnn.txt)
3. `A` VNNI+Transpose and `B` VNNI: [TN-vnni-libxsmm_dnn.txt](https://github.com/user-attachments/files/30984065/TN-vnni-libxsmm_dnn.txt)

test/Integration/MLIRGen/mlir-gen-matmul.mlir
```diff
@@ -289,3 +289,275 @@
...
+// RUN: mlir-gen --kernel=args --seed=0 --data-type=bf16 --batch=256 --layers=512,512,512 --tiles=32,32,32 --vnni=2 --transpose-a=1 --transpose-b=0 2>&1 | FileCheck %s --check-prefix=VNNI-TA-CHAIN
+
+
+// ===== bf16 flat combinations =====
+
+// NN-BF16-DAG: #map = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d2, d3, d5)>
+// NN-BF16-DAG: #map1 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d1, d2, d5, d4)>
+// NN-BF16-DAG: #map2 = affine_map<(d0, d1, d2, d3, d4, d5) -> (d0, d1, d3, d4)>
+// NN-BF16: func.func @entry(%arg0: tensor<8x16x32x32xbf16>, %arg1: tensor<16x16x32x32xbf16>, %arg2: tensor<8x16x32x32xbf16>) -> tensor<8x16x32x32xbf16>
+// NN-BF16: linalg.generic {{.*}}iterator_types = ["parallel", "parallel", "reduction", "parallel", "parallel", "reduction"]
```
rengolin, line 249: there are two of these in each kernel, right? Perhaps you can use `COUNT-2` to make sure they're identical

tools/mlir-gen/MLIRGen.h
```diff
@@ -178,13 +186,22 @@ class MLIRGenerator {
     Arg intermediate; // For quantdequant validation
     Arg output;
     Arg accumulator;
+    // Whether this layer reads its input / weight or writes its output
+    // transposed.
+    bool transposeInput = false;
```
rengolin: nit: bundle all input/weigth/output args with each other, and follow existing naming, ex. `inputTranspose`

tools/mlir-gen/MLIRGen.h
```diff
@@ -178,13 +186,22 @@ class MLIRGenerator {
...
+    bool transposeOutput = false;
+    bool transposeWeight = false;
+    // Whether the (packed) input activation is relaid out with an explicit
+    // transpose before the contraction (tiled transposed-A hidden layers).
+    bool materializeInputTranspose = false;
```
rengolin: what's the difference between this and `transposeInput`?

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -45,7 +45,23 @@ void parseStringList(StringRef str, SmallVector<int64_t> &list) {
...
+/// 5D [a,b,c,d,v] -> [b,a,d,c,v].
+static TensorType transposePackedType(TensorType type) {
+  auto rt = cast<RankedTensorType>(type);
+  auto shape = rt.getShape();
+  assert((shape.size() == 2 || shape.size() == 4 || shape.size() == 5) &&
+         "Expected a flat 2D, packed 4D or VNNI 5D type");
+  if (shape.size() == 2)
+    return RankedTensorType::get({shape[1], shape[0]}, rt.getElementType());
+  SmallVector<int64_t> transposed{shape[1], shape[0], shape[3], shape[2]};
+  if (shape.size() == 5)
```
rengolin: looks like you can repeat the process once more:
```
SmallVector<int64_t> transposed{shape[1], shape[0]};
if (shape.size() >= 4) {
   transposed.push_back(shape[3]);
   transposed.push_back(shape[2]);
}
if (shape.size() == 5)
    transposed.push_back(shape[4]);
return RankedTensorType::get(transposed, rt.getElementType());

**Review by rengolin — COMMENTED** [review]
initial comments, mostly nits. Will look deeper into MLIRGen.cpp now

File: tools/mlir-gen/MLIRGen.cpp [hunk 11]
```diff
@@ -326,24 +362,95 @@ void MLIRGenerator::getKernelTypes(KernelArgs &args) {
     LayerArgs arg;
     arg.index = i;
     arg.input.type = currentType;
+
+    // In a multi-layer MLP the activation produced by a hidden layer is the
+    // next layer's reduction (A) input, so it must be tiled by the contraction
+    // tile (tiles[2]) instead of the output tile (tiles[1]). The final layer's
+    // output is never reduced again, so it keeps the output tile.
+    bool isLast = (i == max - 1);
```
rengolin: This whole logic exploded to a complexity that will be impossible to maintain. I don't understand why it has to be so complicated.

There's a validation part: can I do transpose X VNNI, which you're already doing it elsewhere. Here,  you just need to know if you need to transpose between each layer or not.

For inputs and outputs, you don't need to guess the types at this stage even. You have the input shape, the output shape and if they don't match, you transform. This can be a simple transpose, a VNNI transpose or both. For weights, they're all the same, so it's a single model-wide flag on which form it takes.

The GEMMs then just follow those forms.

test/Integration/MLIRGen/mlir-gen-matmul.mlir
arun-thmn: No, That is a single layer tests.
The test for `multi-gemms` are added at: https://github.com/arun-thmn/tpp-mlir/blob/f1537073933304bf3754ba1013dcd8138f3970ea/test/Integration/MLIRGen/mlir-gen-matmul.mlir#L514 and 550.

tools/mlir-gen/MLIRGen.h
arun-thmn: Okay. re-named the variables names.

tools/mlir-gen/MLIRGen.h
arun-thmn: The `materializeInputTranspose` removed as part of logic simplification.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Okay, done as you mentioned.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Okay. Simplified the logic now.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -1249,24 +1470,44 @@ TensorType MLIRGenerator::getShape(ArrayRef<int64_t> dims, PackingType type) {
       }
     }
     // Unpacked type, just return 2D tensor
+    // Transposed A (N x C -> C x N) or B (C x K -> K x C) swaps the two dims.
+    if ((type == PACK_INPUT && transposeA) ||
+        (type == PACK_WEIGHT && transposeB)) {
+      assert(dims.size() == 2 && "Transpose requires a 2D operand");
+      return RankedTensorType::get({dims[1], dims[0]}, dataTypes.input);
```
rengolin: Cant we have a method that transposes the dims, and then just apply if the flag is set?

Keeping all these constant swaps in code makes harder to review and avoid bugs.

arun-thmn: Sure. Re-factored the code, so that we have "a method that transposes the dims, and then just apply if the flag is set"

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -312,8 +342,18 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
 }
 
 void MLIRGenerator::getKernelTypes(KernelArgs &args) {
+  // A flat (untiled) pytorch-style transposed-A MLP transposes every layer's A
```
rengolin: This comment makes no sense. Becareful with LLM generated comment as they go on and on about random things and don't help.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -312,8 +342,18 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
...
+  // activation (matching the named-matmul TN chain emitted by pytorch-mlir), so
+  // its stored input keeps the natural NN layout and the leading (M) dim flips
+  // per layer. See the flatTransposeChain branch below.
+  bool flatTransposeChain = !tiles.size() && vnniFactor == 0 &&
```
rengolin: Why do we need a special case for this particular combination?

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -326,6 +366,34 @@ void MLIRGenerator::getKernelTypes(KernelArgs &args) {
     LayerArgs arg;
     arg.index = i;
     arg.input.type = currentType;
+
+    if (flatTransposeChain) {
```
rengolin: This is just LLM focusing on one particular case and ignoring everything else by isolating the case it cares about inside an if statement. This is unmiaintainable and why we needed to refactor the code this month.

rengolin (commented):
There are still too many special cases created by LLM that "pass the tests we have". I'd recommend giving up on LLM help here and just think how to implement the actual logic: when to transpose A, B, etc. and add small changes to the generation.

If the generation is broken and making it harder, then fix the generation first, then apply small fixes.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: For the different `matmul` variants (mentioned in the earlier thread), the kernel generation differs between `pytorch` vs `libxsmm_dnn`. 

- The `pytorch` style of `gemm` follows only the `NN layout` (for `multi-gemm` as well) and it transposes (`linalg.transpose`) the input `A/B` and output `acc` matrices accordingly to other variants (`TN`, `TT`, `NT`). A `unpacked` inputs with `2D` matrices.
- The `libxsmm_dnn` style represents multiple variants of `gemm` with the help of `affine_map`. It fixes one layout (`NN` or `TN` or `TT` or `NT`) across the multiple gemms for both `packed` and `vnni` formats.

I have re-factored the code with simplified comments easy to understand. The code:
`bool unpacked = !tiles.size() && vnniFactor == 0 && !quant;` - is not a special-case, used to different between the `pytorch` or `libxsmm_dnn` style of gemm creation.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Please see the above response.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Please see the above response.

The `if` condition needed to separate `pytorch` or 'libxsmm_dnn`. Hope, re-factoring will convey it's need properly.

tools/mlir-gen/MLIRGen.h
```diff
@@ -201,6 +220,12 @@ class MLIRGenerator {
   /// Returns the chain value to be used in the next op
   Value lowerMatmul(LayerArgs &args);
 
+  /// Re-tile the reduction (feature) dimension of a standard packed activation
+  /// {BN, Bk, bn, k} to {BN, Bc, bn, reduceTile}, keeping the batch blocking.
```
rengolin: This is  weird comment. Is it just talking about VNNI packing?

tools/mlir-gen/MLIRGen.h
```diff
@@ -129,8 +135,10 @@ class MLIRGenerator {
     OUTPUT_SCALE
   };
 
-  /// Return shaped type (packed if requested)
-  TensorType getShape(ArrayRef<int64_t>, PackingType);
+  /// Return shaped type (packed if requested). \p nTile overrides the output
+  /// (N) tile size when non-zero; used to tile hidden-layer activations by the
+  /// contraction dim so multi-layer chains stay consistent.
```
rengolin: Is this also about VNNI? Can this all just be checked via the `vnni` option?

tools/mlir-gen/MLIRGen.h
arun-thmn: Yes, for the multi-gemms. The Function `retilePackedActivation` is to unpack and pack again the output of a gemm for different `M, N, K tile` sizes (eg., `--tiles=32,32,64`) before being fed as an input to the following or next gemm.

The `mlir-gen` doesn't handle this case ( `--tiles=32,32,64`) for any matmul variant.

tools/mlir-gen/MLIRGen.h
arun-thmn: Okay. Removed the new parameter in the functions. It now tries to `getShape` with the earlier code.

tools/mlir-gen/MLIRGen.h
```diff
@@ -129,7 +135,7 @@ class MLIRGenerator {
     OUTPUT_SCALE
   };
 
-  /// Return shaped type (packed if requested)
+  /// Return shaped type (packed if requested).
```
rengolin: unnecessary comment change left over

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -285,8 +311,28 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
...
+  // True: pytorch like gemm kernels
+  bool unpacked = !tiles.size() && vnniFactor == 0 && !quant;
+
+  // Input type, also first layer's input. The unpacked GEMM (pytorch) stores A in its
+  // natural NN layout (any transpose is later materialized as an explicit
+  // linalg.transpose); the packed GEMM (libxsmm_dnn)stores A in the canonical packed layout.
+  TensorType currentType =
+      unpacked
+          ? RankedTensorType::get({batch, layers.front()}, dataTypes.input)
+          : getShape({batch, layers.front()}, PACK_INPUT);
```
rengolin: this logic should already be covered by `getShape`. If not, the fix is there, not in every call to it.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -552,6 +632,40 @@ void MLIRGenerator::computeElementwiseScalingFlops(ShapedType outputShape) {
   flops += 2 * scalingFlops;
 }
 
+Value MLIRGenerator::retilePackedActivation(Value packed, int64_t reduceTile) {
+  auto packedType = cast<ShapedType>(packed.getType());
+  auto shape = packedType.getShape(); // {BN, Bk, bn, k}
+  auto elemTy = packedType.getElementType();
+  int64_t bnBlocks = shape[0], kBlocks = shape[1], bn = shape[2], k = shape[3];
```
rengolin: assert(shape.size() >= 4)

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -552,6 +632,40 @@ void MLIRGenerator::computeElementwiseScalingFlops(ShapedType outputShape) {
...
+  int64_t sizeDim = kBlocks * k;
+  int64_t newKBlocks = sizeDim / reduceTile;
+
+  // {BN, Bk, bn, k} -> {BN, bn, Bk, k} so the batch and feature sub-blocks are
+  // each contiguous and can be collapsed into a flat {batch, size} tensor.
+  Value init0 = tensor::EmptyOp::create(
+      builder, loc, ArrayRef<int64_t>{bnBlocks, bn, kBlocks, k}, elemTy);
+  Value t0 = linalg::TransposeOp::create(builder, loc, packed, init0,
```
rengolin: If this is the VNNI pack, can't we just call `pack`?

arun-thmn:
The different matmul/mlp variant generated by `mlir-gen` as aligned with the `libxsmm_dnn`: [libxsmm_IR_dump.zip](https://github.com/user-attachments/files/31407271/libxsmm_IR_dump.zip)
`pytorch`: [pytorch_IR_MLP_dump.zip](https://github.com/user-attachments/files/31407291/pytorch_IR_MLP_dump.zip)

For example: The IR with name `lib_bf16_NT_no-vnni_tiles_32_32_32.mlir` denotes `lib` - libxsmm_dnn; `NT` - `A` non-tranpose `B` transposed; `-tiles_32_32_32` is the packing.

alheinecke:
I suggest to put a table here on what is supported: for which datatype: ATrans, BTrans, Avnni, Bvnni, Apacked, Bpacked. For BF16, FP16, BF8 and HF8 all should be support. For FP32 VNNI is not needed.

Additionally, when we chain, we need to bring C into the right form and this we want to do when storing it such that it can be easily consumed. 

Perhaps it's easier to split the generator into gemm only and mlp, as many options don't exists for MLP / make no sense?!

arun-thmn:
Below `table` represents the kernels `mlir-gen` can emit after this patch (libxsmm_dnn resemble). 
Notations: 
- `NT` - the two letters represents matrix transpose. Here `N` - matrix `A` is non-transposed and `T` - matrix `B` is transposed.
- `+vnni` - The matrices are `vnni` packed. Not mentioned - `Flat` layout.
- `in-support` - Kernel generation already in-support.
- `yes` - supported from this patch.

| Case<br>(multi-gemm/mlp)| F32<br>`tiles=32,32,32` | F16<br>`tiles=32,32,32` | BF16<br>`tiles=32,32,32` | BF16<br>`tiles=32,32,64` | BF8<br>`tiles=32,32,64` | HF8<br>`tiles=32,32,64` |
|------|-----|-----|--------------------|--------------------|-----|-----|
| NN + vnni=2/4 | NA | in-support | in-support | yes | yes | yes |
| TN + vnni=2/4 | NA | yes | yes | yes | yes | yes |
| NN | in-support | in-support | in-support | yes | yes | yes |
| TN | yes | yes | yes | yes | yes | yes |
| NT | yes | yes | yes | yes | yes | yes |
| TT | yes | yes | yes | yes | yes | yes |

The IR for all these cases are generated and attached as dump in the earlier thread/response.

For `pytorch` the table resembles the same, with the `vnni` rows be `NA` and no `tiles`.

ps: For multi-gemm we have to unpack and pack again the output of a gemm for different `M, N, K tile` sizes (eg., `--tiles=32,32,64`) before being fed as an input to the following or next gemm.
The `mlir-gen` doesn't handle this case ( `--tiles=32,32,64`) for any matmul variant.

tools/mlir-gen/MLIRGen.h
arun-thmn: Removed it.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Code removed as part of reduced gemm variant clean-up.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Code removed as part of reduced gemm variant clean-up.

tools/mlir-gen/MLIRGen.cpp
arun-thmn: Code removed as part of reduced gemm variant clean-up.

arun-thmn:
@rengolin Please re-review this patch.

Now, the patch is trimmed/cleaned-up to have only the below following changes for `mlir-gen`:

- Support `transpose` matmul variant only for packed (`libxsmm_dnn`) single gemm. The unpacked (`pytorch`) will have no new additions, but the existing support will be maintained.
- For multi-gemm case the `pack` tiles should be same. For example `tiles=32,32,32`/`tiles=64,64,64` are allowed. No more `tiles=32,32,64`.
- Supported types are `f32`,`bf16`, `f16`, `bf8`, and `hf8`.

rengolin (approved):

PR #1169, opened 2026-08-07
[mlir-gen]: add i8xi8->i32->i8 requantization support

shahidact:
Add a requantization GEMM kernel to mlir-gen that rescales an i32 accumulator back to i8 using a per-output-channel output scale argument. Scale arguments (input/weight dequant and requant output) are now marked with the tpp.quant_scale argument attribute, and MLIRBench detects scale operands from that attribute instead of hard-coded argument positions.


lib/TPP/Runner/MLIRBench.cpp
```diff
@@ -269,9 +269,12 @@ LogicalResult MLIRBench::createKernelArgs() {
       argInitType = TensorInitType::Normal;
     }
 
-    // Scale arguments are paired with corresponding input and weight matrices.
+    // Scale arguments are marked by the generator with the `tpp.quant_scale`
+    // argument attribute (used for input/weight dequant scales and the requant
+    // output scale).
     bool isScaleArgument =
-        argInitType == TensorInitType::Quant && (argNum == 1 || argNum == 3);
+        argInitType == TensorInitType::Quant &&
```
rengolin, line 277: No special treatment in mlir-gen, please. This will never come from PyTorch, and if we write our compiler to only work with out special generation, it won't work with any other input.

shahidact: The idea was to do it cleanly without relying on indices check as in future we may have arg type which may implicitly or explicitly identify if a given arg is scale kind or not.

rengolin: Attributes are just a different kind of problem, but has the same effect: the compiler only works when they're there, and someone (not PyTorch, not Torch-MLIR) has to put them there. Discardable attributes are also dangerous in that they can be removed by canonicalizations.

shahidact: Ok, then I will fall back to the existing approach until we have a type with associated scale for such arguments.

shahidact: Updated accordingly.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -311,6 +312,10 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
   builder.setInsertionPoint(module);
 }
 
+bool MLIRGenerator::isRequantization() const {
+  return quantType == QuantizationType::Quant && dataTypes.input.isInteger(8);
```
rengolin: I think the quantization types are too complex. This should be a boolean `quant` that generates scaling factors, whatever the datatype.

I would remove the test quant and the dequant types, and just have one flag that does one thing, which is the same as whatever pytorch does with quantized models.

shahidact: Updated accordingly, keeping the testquant with separate flag for now so that removing that would be trivial.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -256,30 +256,19 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
...
-  quantType = *optQuantType;
-
-  // If the target type contains "mx", it is a mixed precision type. If
-  // quantization type is not explicitly specified, we will default to Mixed
-  // quantization type for mixed precision target types.
-  if (quantType == QuantizationType::None && !dataType.empty() && dataType.contains("mx"))
-    quantType = QuantizationType::Mixed;
+  // A mixed-precision ("mx") data type has differing input and output element
+  // types. Quantization is only meaningful for such types.
+  mixedType = !dataType.empty() && dataType.contains("mx");
```
rengolin: Actually, this is broken, and it's not really MX.

The correct semantics is to _always_ accumulate into a larger type and _always_ downcast after. 

So in reality, we don't need both `i8` and `mx-i8`, or both `bf16` and `mx-bf16`, etc. There's only one per type, and it's not MX (that's a different thing and the name confuses people).

This is _orthogonal_ to quantization, which has an extra step to downcast (C scale multiplication, accumulation), _in addition_ to the final downcast.

An error should be emitter by the command line argument validation (well before this code), if the user specifies `-quant` AND the type is 16/32-bit or larger (need to double check which).

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -311,6 +300,18 @@ MLIRGenerator::MLIRGenerator(StringRef outputOpKindStr, StringRef kernelStr,
...
+bool MLIRGenerator::isDequantization() const {
+  return quant && dataTypes.output.isFloat();
+}
+
+bool MLIRGenerator::isQuantization() const {
+  return quant && dataTypes.output.isInteger(8) && !dataTypes.input.isInteger(8);
+}
+
+bool MLIRGenerator::isRequantization() const {
+  return quant && dataTypes.output.isInteger() && dataTypes.input.isInteger(8);
```
rengolin: I'm not sure what we're trying to do here. Quantization should not be restricted by type combinations at this point. We may not have support for certain combinations, but that's an error message in the command line validation, not here.

Once quantization is enabled, the IR needs to be valid:
* Low precision input types
* High precision accumulation types
* Scaling factors with the right types
* Add quantization epilogue based on available types

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -435,12 +451,14 @@ void MLIRGenerator::createKernel() {
   SmallVector<Type, 1> inputTypes{firstArg.input.type};
   if (kernelType == KernelType::Args) {
     for (auto &layer : args) {
-      if (quantType == QuantizationType::Dequant)
+      if (isDequantization() || isRequantization())
```
rengolin: this bunch of `if () { ... }` blocks is impossible to follow.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -921,6 +954,146 @@ Value MLIRGenerator::quantizeGemm(LayerArgs &args, Value chain) {
   return dquantRes;
 }
 
+Value MLIRGenerator::requantizeGemm(LayerArgs &args, Value chain) {
+  // Chain is the i32 GEMM accumulator to be requantized back to i8.
+  // Requantization first dequantizes the accumulator using the per-row input
+  // scale and the per-output-channel weight scale, then rescales it with the
+  // per-output-channel output scale and saturates to the signed i8 range:
+  //   out_i8 = clamp(round(acc_i32 * S_input * S_weight * S_output), -128, 127)
```
rengolin: let's not assume this method will only ever work with `i8` and `i32`.

The validation here is:
* Can I cast low-precision input to the high-precision accumulator type?
* Can I multiply the scales together, and then with the temporary value above?
* Can I downcast the second value to the output type?

If the answer is YES to all three, the lowering becomes the same for all types (modulo using different arith/math ops for int/float).

Invalid combinations can be caught early in the argument validation

test/Integration/I8/AMX/tpp-run-gemm-dequantize-correctness-i8-f32-diffscale.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: tpp-run -e entry --entry-point-result=void --linalg-to-loops -print --splat-to-random --init-type quant -seed 123  %s > %t.1
+// RUN: tpp-run -e entry --entry-point-result=void -print --splat-to-random --init-type quant -seed 123  %s > %t.1
```
rengolin: why change from loops to xsmm?

shahidact: I thought, after the cleanup, the default path is the way for validation. I will change it back to existing -linalg-to-loop.

tools/mlir-gen/MLIRGen.cpp
shahidact: > The correct semantics is to _always_ accumulate into a larger type and _always_ downcast after.
Agree about mx-i8/mx-bf16, we can just use i8/bf16 along with quant flag. What about the following two cases:
1. mx-i8-f32: Low precision gemm + dequantization to FP32.
2. mx-f32-i8: High precision gemm + quantization to I8.

Don't we want to support these?

rengolin: No, those were only for tests.

We may also get away with the test quant and just validate the output against loops and xsmm like we do for everything else.

shahidact: No, it was not only for test. I had it, as we designed those API for future reference when we work on dynamic quantization by computing the scales.

shahidact: However, I won't insist on keeping it and remove it completely if you wish.

shahidact: Let me know what you think about it in my latest update.

tools/mlir-gen/MLIRGen.cpp
shahidact: Updated accordingly.

tools/mlir-gen/MLIRGen.cpp
shahidact: Updated accordingly, hope this make sense.

lib/TPP/Transforms/Utils/TensorInitFloat.cpp
```diff
@@ -109,8 +109,20 @@ void IdentityTensorInitFloat::fillData() {
 
 // Update internal buffer with rescale values.
 void QuantScaleTensorInitFloat::fillData() {
-  assert(scaleBuffer.size() > 0 && "scaleBuffer is empty");
-  for (size_t i = 0; i < scaleBuffer.size(); i++) {
-    push(scaleBuffer[i]);
+  // When paired with a quantized data matrix (dequantization), the scale buffer
```
rengolin: This is tricky. If the scales are populated externally, why would you call this method?

tools/mlir-gen/mlir-gen.cpp
```diff
@@ -68,7 +68,7 @@ llvm::cl::opt<std::string>
 llvm::cl::opt<std::string>
     dataType("data-type", llvm::cl::desc("Data type for arguments"),
               llvm::cl::value_desc(
-                  "f32|f16|bf16|bf8|hf8|mx-bf16|mx-f16|mx-i8|mx-i8-f32|mx-f32-i8"),
+                  "f32|f16|bf16|bf8|hf8|bf16-f32|f16-f32|i8|i8-f32|f32-i8"),
```
rengolin: Every `bf16` will accumulate to `f32`, you don't need to specify that. This is what the `bf16` case above should be, no need for a new one. Same for `f16`. 

We don't need the `f32-i8` case either. IIRC, that was just for the quantization test, right?

tools/mlir-gen/MLIRGen.h
```diff
@@ -101,11 +102,17 @@ class MLIRGenerator {
...
+  bool quant;
 
-  /// Type of quantization ops to be generated
-  QuantizationType quantType;
+  /// When true, generate a quantize-then-dequantize validation kernel.
+  bool testQuant;
+
+  /// True when the selected data type is mixed-precision, i.e. the input and
+  /// output element types differ.
+  bool mixedType;
```
rengolin: You don't need this flag. You can just validate that the types are different when selecting `--quant` options and bail if not.

tools/mlir-gen/MLIRGen.h
```diff
@@ -214,9 +223,17 @@ class MLIRGenerator {
...
+  /// Applies the quantization epilogue to the wide GEMM accumulator, choosing
+  /// the concrete lowering from the available input/output types.
+  Value quantizeEpilogue(LayerArgs &args, Value chain);
+
   /// Creates a matmul quantization kernel
   Value quantizeGemm(LayerArgs &args, Value chain);
 
+  /// Requantizes an i32 GEMM accumulator back to i8 using a per-output-channel
+  /// output scale argument.
+  Value requantizeGemm(LayerArgs &args, Value chain);
```
rengolin, line 215: Not sure what's the difference between the three methods above. There's only one functionality we need: quantize 32-bit results into low-precision (`i8`, `f8`, ...). If that's after a GEMM, or at the end of a function makes no difference.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -668,6 +674,11 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args) {
       break;
   }
 
+  // The quantization epilogue consumes the raw wide accumulator and performs
+  // the final cast, so skip the same-domain downcast here.
+  if (quant)
```
rengolin: why not just do:
```
if (quant)
  return requantize(...);
else
  return downscast(...)
```

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -612,13 +617,14 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args) {
   auto outputType = cast<ShapedType>(args.output.value.getType());
   auto shape = outputType.getShape();
 
-  // Select the matmul accumulator tensor.
-  if (quantType == QuantizationType::Quant) {
-    // Quantization casts the result later; accumulate in the input type.
+  // Select the matmul accumulator tensor. Quantized kernels always accumulate
+  // into a wide type: a float input quantized to an integer output accumulates
+  // in the (float) input type, while every other quantized kernel accumulates
+  // into i32.
```
rengolin: Let's avoid this complexity and only keep integer-integer / float-float for now.

lib/TPP/Transforms/Utils/TensorInitFloat.cpp
shahidact: Ah.. the comment is not proper. By externally, I meant initialization of scale argument. I will fix the comment.

tools/mlir-gen/mlir-gen.cpp
shahidact: > Every `bf16` will accumulate to `f32`, you don't need to specify that. This is what the `bf16` case above should be, no need for a new one. Same for `f16`.

Yes, accumulator is always f32, here *-f32", specify output type. 

For above "bf16" case, the output type is 'bf16'.

If we don't want 'bf16-f32' variant kernel I will remove it completely (same for f16-f32).

shahidact: > We don't need the `f32-i8` case either. IIRC, that was just for the quantization test, right?

Ok I will remove it completly.

rengolin: > For above "bf16" case, the output type is 'bf16'.
> 
> If we don't want 'bf16-f32' variant kernel I will remove it completely (same for f16-f32).

Right, we don't "need" output f32 or we'd also have to cope with accumulator typed epilogue, etc. Let's simplify this as much as possible (accumulate in f32 & downcast to bf16) then move any intelligence to the compiler, not the generator.

lib/TPP/Transforms/Utils/TensorInitFloat.cpp
rengolin: It's still unclear what is the difference between the two generation logics in the same method

shahidact: Input scale arguments are with respect to corresponding input metrices, while output scale is not. Output scale is from oracle (say).

tools/mlir-gen/MLIRGen.cpp
shahidact: > why not just do:
> 
> ```
> if (quant)
>   return requantize(...);
> else
>   return downscast(...)
> ```

Because "quant" has two paths based on output type. One goes to requantize and other to dequantize.

tools/mlir-gen/MLIRGen.h
shahidact: Updated, now only two methods, one for dequantization and other for requantization of output.

tools/mlir-gen/MLIRGen.h
shahidact: Updated accordingly.

tools/mlir-gen/mlir-gen.cpp
shahidact: Updated accordingly.

test/Integration/I8/AMX/tpp-run-gemm-dequantize-correctness-i8-f32-diffscale.mlir
shahidact: Done.

benchmarks/config/omp/mlir-vector-amx.json
```diff
@@ -1,64 +1,82 @@
...
-      "flags": [ "-n", "100", "-run-args='--def-parallel --nano-kernels --gemm-unroll=16,16,16 --registerBlocking=32,32,64 --init-type=quant --splat-to-random --seed=123'" ],
-      "extensions": [ "amx_int8" ]
-    },
-    "i8_f32_dequant_4096x4096_omp_64_mlir": {
-      "type": "IR-GEN",
-      "benchmark": [ "mlir-gen", "--kernel=args --data-type=mx-i8-f32 --batch=4096 --layers=8192,4096 --tiles=32,32,64 --vnni=4 --quant-type=dequantize" ],
-      "environment": { "OMP_NUM_THREADS": "64", "KMP_AFFINITY": "granularity=fine,verbose,compact,1,0" },
-      "flags": [ "-n", "100", "-run-args='--def-parallel --nano-kernels --gemm-unroll=16,16,16 --registerBlocking=32,32,64 --init-type=quant --splat-to-random --seed=123'" ],
-      "extensions": [ "amx_int8" ]
+    "gemm_i8_f32_dequant_mlir_vector_large_amx": {
```
rengolin: doesn't seem to have any changes, please revert this change.

test/Integration/I8/AMX/tpp-run-gemm-dequantize-correctness-i8-f32-i8-scale.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: tpp-opt --default-tpp-passes="nano-kernel registerBlocking=32,32,64 gemm-unroll=16,16,16" %s | FileCheck %s -check-prefix=IR
 
-// RUN: tpp-run -e entry --entry-point-result=void --linalg-to-loops -print --splat-to-random --init-type quant -seed 123  %s > %t.1
+// RUN: tpp-run -e entry --entry-point-result=void -print --splat-to-random --init-type quant -seed 123  %s > %t.1
```
rengolin: revert this change also, please.

tools/mlir-gen/MLIRGen.cpp
rengolin: You can insert `quantizeEpilogue` here.

lib/TPP/Transforms/Utils/TensorInitFloat.cpp
```diff
@@ -109,8 +109,20 @@ void IdentityTensorInitFloat::fillData() {
...
+  // is populated externally. Otherwise (e.g. the requantization output scale)
+  // self-generate random positive per-channel scale factors.
+  if (!scaleBuffer.empty()) {
+    for (size_t i = 0; i < scaleBuffer.size(); i++) {
+      push(scaleBuffer[i]);
+    }
+    return;
+  }
+  for (size_t i = 0; i < size; i++) {
```
rengolin: please move this logic inside `QuantScaleTensorInitFloat::setScaleBuffer`, for example, if the `newBuffer` is null or empty, randomize.

tools/mlir-gen/MLIRGen.cpp
shahidact: Done.

benchmarks/config/omp/mlir-vector-amx.json
shahidact: Done.

test/Integration/I8/AMX/tpp-run-gemm-dequantize-correctness-i8-f32-i8-scale.mlir
shahidact: Done.

lib/TPP/Transforms/Utils/TensorInitFloat.cpp
shahidact: Done.

include/TPP/Transforms/Utils/TensorInitFloat.h
```diff
@@ -184,8 +184,26 @@ struct QuantScaleTensorInitFloat : TensorInitFloat {
...
+  //      integer initializer picks per-channel quantization scales while
+  //      quantizing the data and hands them over via `setScaleBuffer` so that
+  //      `data_int * scale == original_float`.
+  //
+  //   2. Standalone mode: The scale has no paired data matrix.
   void setScaleBuffer(const std::vector<llvm::APFloat> &newBuffer) {
-    scaleBuffer = newBuffer;
+    // Paired mode
+    if (!newBuffer.empty())
+      scaleBuffer = newBuffer;
```
rengolin, line 198: without `return`, this will always randomize.

benchmarks/config/omp/mlir-vector-amx.json
rengolin: still here?

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -326,21 +299,29 @@ void MLIRGenerator::getKernelTypes(KernelArgs &args) {
...
     arg.input.type = currentType;
-    // Scale inputs are only needed for dequantization.
-    if (quantType == QuantizationType::Dequant)
+    // Quantized kernels carry a per-row input scale and a per-output-channel
+    // weight scale to dequantize the wide accumulator.
+    if (quant)
       arg.inputScale.type = getShape({batch}, INPUT_SCALE);
     arg.weight.type = getShape({inputSize, outputSize}, PACK_WEIGHT);
-    if (quantType == QuantizationType::Dequant)
+    if (quant)
```
rengolin: put all scales under the same `if (quant)`

tools/mlir-gen/MLIRGen.cpp
rengolin: That function is so small and does so little that you can just inline it here, no need for an extra function

include/TPP/Transforms/Utils/TensorInitFloat.h
shahidact: Oh my bad.

tools/mlir-gen/MLIRGen.cpp
shahidact: Done.

tools/mlir-gen/MLIRGen.cpp
```diff
@@ -627,7 +568,7 @@ Value MLIRGenerator::lowerMatmul(LayerArgs &args) {
     // Accumulator matches the output type; accumulate directly into it.
     args.accumulator.value = args.output.value;
   } else {
-    args.accumulator.value = getZeroInitTensor(args.accumulator.type);
+      args.accumulator.value = getZeroInitTensor(args.accumulator.type);
```
rengolin: unrelated indentation change?

rengolin (approved):
Looks good with a small indentation change.

tools/mlir-gen/MLIRGen.cpp
shahidact: Done.

PR #1170, opened 2026-08-07
Feat/streaming store

shahidact:
[TPP] Stream write-only stores via nontemporal vector.store

    Add a ConvertToStreamingStore pass that rewrites a write-only
    vector.transfer_write (e.g. the GEMM output C matrix, never read back)
    into a vector.store carrying nontemporal = true, avoiding cache pollution
    for data that is not reused. The pass runs after FlattenVectorOps so the
    transfer is already 1-D. Also reposition ReplicateBenchArgs to run on the
    bufferized nano-kernel path.

### DIFF
--- a/lib/TPP/Runner/MLIRBench.cpp
+++ b/lib/TPP/Runner/MLIRBench.cpp
@@ -269,9 +269,12 @@ LogicalResult MLIRBench::createKernelArgs() {
       argInitType = TensorInitType::Normal;
     }
 
-    // Scale arguments are paired with corresponding input and weight matrices.
+    // Scale arguments are marked by the generator with the `tpp.quant_scale`
+    // argument attribute (used for input/weight dequant scales and the requant
+    // output scale).
     bool isScaleArgument =
-        argInitType == TensorInitType::Quant && (argNum == 1 || argNum == 3);
+        argInitType == TensorInitType::Quant &&
+        kernel.getArgAttr(argNum, "tpp.quant_scale") != nullptr;
     auto arg =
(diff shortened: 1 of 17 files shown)


lib/TPP/Runner/MLIRBench.cpp
```diff
@@ -269,9 +269,12 @@ LogicalResult MLIRBench::createKernelArgs() {
...
     }
 
-    // Scale arguments are paired with corresponding input and weight matrices.
+    // Scale arguments are marked by the generator with the `tpp.quant_scale`
+    // argument attribute (used for input/weight dequant scales and the requant
+    // output scale).
     bool isScaleArgument =
-        argInitType == TensorInitType::Quant && (argNum == 1 || argNum == 3);
+        argInitType == TensorInitType::Quant &&
+        kernel.getArgAttr(argNum, "tpp.quant_scale") != nullptr;
```
rengolin, line 277: No special treatment in mlir-gen, please. This will never come from PyTorch, and if we write our compiler to only work with out special generation, it won't work with any other input.

rengolin: Also, what does this have to do with streaming stores?

rengolin: I see this is on top of another branch. In this case, please do stacked PRs.
https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

Basically, add each PR to a different branch in the libxsmm/tpp-mlir repo, then make the first apply to `main` and the second to the first.

PR #1172, opened 2026-08-12
[mlir-gen] Remove redundant empty tensor.

shahidact:
Currently truncated result is being written into a new tensor instead of output argument. This patch fixes it.

### DIFF
--- a/test/Integration/MLIRGen/mlir-gen-matmul.mlir
+++ b/test/Integration/MLIRGen/mlir-gen-matmul.mlir
@@ -1,5 +1,6 @@
 // RUN: mlir-gen --kernel=args --seed=0 --data-type=f32 --batch=128 --layers=2304,768 --tiles=64,48,64 2>&1 | FileCheck %s --check-prefix=FP32
 // RUN: mlir-gen --kernel=args --seed=0 --data-type=bf16 --batch=128 --layers=2304,768 --tiles=64,48,64 2>&1 | FileCheck %s --check-prefix=BF16
+// RUN: mlir-gen --kernel=args --data-type=bf16 --batch=8192 --layers=1024,4096 --tiles=32,32,32 --vnni=2 | FileCheck %s --check-prefix=BF16-TRUNCATE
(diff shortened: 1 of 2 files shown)


test/Integration/MLIRGen/mlir-gen-matmul.mlir
```diff
@@ -40,6 +41,12 @@
 // BF16:         arith.addf
 // BF16-NOT: dealloc
 
+// Check that tensor.empty() not created to store the result of arith.truncf in the generic op.
```
adam-smnk, line 44: I know it works to steer bufferization but it's not realistic IR.
Empty is pretty much the default for write-only output tensor. `torch-mlir` also emits empty as expected.

shahidact: Ok, I see your point. Then how the result would be written to output argument?

adam-smnk: You could use `bufferization.materialize_in_destination` to steer bufferization directly.
For example, see upstream test case `test_materialize_in_destination_op`.

shahidact: My understanding is that `mlir-gen` would generate Ops at higher level dialect i.e Linalg. Is it otherwise?

adam-smnk: It's orthogonal to payload itself, nothing about computation itself changes.

If you really need to ensure that the result is placed in args then bufferization hint would the most neutral approach.
It's also sth we add as a pre-processing step when moving results to args in `torch.compile` backend.
Note that it might still result in a separate copy if bufferization decides so.

That being said, I don't think it fully belongs to `mlir-gen` as ingress paths generally don't provide such hints.
But it's an option that could be feasibly done by an early stage compiler pass.

I'd stick with `tensor.empty` as is and try to address any problems it might be causing deeper in the compiler pipeline.

rengolin: `mlir-gen` should generate PyTorch level IR, yes. Any intelligence in bufferization needs to be done in the compiler side.

shahidact: Ok.

rengolin (approved):

PR #1175, opened 2026-08-21
[TPP] Stream write-only stores via nontemporal vector.store

shahidact:
Add a ConvertToStreamingStore pass that rewrites a write-only vector.transfer_write (e.g. the GEMM output C matrix, never read back) into a vector.store carrying nontemporal = true, avoiding cache pollution for data that is not reused. The pass runs after FlattenVectorOps so the transfer is already 1-D. Also reposition ReplicateBenchArgs to run on the bufferized nano-kernel path.

### DIFF
--- a/lib/TPP/DefaultTppPasses.cpp
+++ b/lib/TPP/DefaultTppPasses.cpp
@@ -103,9 +103,16 @@ private:
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
     pm.addPass(createBufferize());
+    // Replicate benchmark kernel arguments for cold-cache timing. Runs on
+    // bufferized memrefs so replicas are plain subviews (no allocs/copies).
+    // No-op unless the benchmark producer requested replication.
+    pm.addPass(createReplicateBenchArgs());
     pm.addNestedPass<func::FuncOp>(createVectorContractToNanoKernels());
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createFlattenVectorOps());
+    // Stream write-only stores (e.g. the output C matrix) with a nontemporal
+    // hint. Runs after flattening so the transfer_write is already 1-D.
+    pm.addNestedPass<func::FuncOp>(createConvertToStreamingStore());
   }
 
   // Default TPP lowering: map and pack at the linalg level, bufferize, lower to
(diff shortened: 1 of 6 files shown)


lib/TPP/DefaultTppPasses.cpp
```diff
@@ -103,9 +103,16 @@ struct DefaultTppPasses
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
     pm.addPass(createBufferize());
+    // Replicate benchmark kernel arguments for cold-cache timing. Runs on
+    // bufferized memrefs so replicas are plain subviews (no allocs/copies).
+    // No-op unless the benchmark producer requested replication.
+    pm.addPass(createReplicateBenchArgs());
```
rengolin: Why is this also affected?

lib/TPP/Transforms/ConvertToStreamingStore.cpp
```diff
@@ -0,0 +1,128 @@
...
+      if (isa<ViewLikeOpInterface>(user)) {
+        for (Value res : user->getResults())
+          if (isa<MemRefType>(res.getType()))
+            worklist.push_back(res);
+        continue;
+      }
+      // Any user with unknown effects might read the buffer.
+      auto memEffects = dyn_cast<MemoryEffectOpInterface>(user);
+      if (!memEffects)
+        return false;
```
rengolin: If the op doesn't implement the memory effect interface, can we assume it doesn't have any? Then it would be safe to `continue` rather than `return`?

lib/TPP/DefaultTppPasses.cpp
shahidact: It is not affected as such. It is a decision we need to make if we want to keep replication in default path too?

lib/TPP/Transforms/ConvertToStreamingStore.cpp
shahidact: Good point! I will check.

lib/TPP/DefaultTppPasses.cpp
rengolin: Can this be a separate PR, then?

shahidact: Yes, it can.

lib/TPP/Transforms/ConvertToStreamingStore.cpp
shahidact: Updated accordingly.

lib/TPP/DefaultTppPasses.cpp
shahidact: Updated accordingly.

lib/TPP/DefaultTppPasses.cpp
```diff
@@ -135,13 +138,13 @@ struct DefaultTppPasses
...
       pm.addPass(createBufferize());
-
-    // Replicate benchmark kernel arguments for cold-cache timing. Runs on
-    // bufferized memrefs so replicas are plain subviews (no allocs/copies).
-    // No-op unless the benchmark producer requested replication.
-    pm.addPass(createReplicateBenchArgs());
+      // Replicate benchmark kernel arguments for cold-cache timing. Runs on
+      // bufferized memrefs so replicas are plain subviews (no allocs/copies).
+      // No-op unless the benchmark producer requested replication.
+      pm.addPass(createReplicateBenchArgs());
```
rengolin: This is also unrelated. For xsmm /loop lowering, the bench args won't be replicated.

shahidact: This was existing outside of if block, I will keep it as it was.

shahidact: Done.

rengolin (approved):

PR #1176, opened 2026-08-21
Wire ReplicateBenchArgs pass into the nano-kernel lowering path.

shahidact:
This pass is required to do cold-cache benchmarking for realistic throughput results using the nano-kernel pass pipeline.

### DIFF
--- a/lib/TPP/DefaultTppPasses.cpp
+++ b/lib/TPP/DefaultTppPasses.cpp
@@ -142,8 +142,13 @@ private:
       pm.addPass(createSCFForAllLoopFlattenSFC());
 
     // Bufferize: tensor->memref.
-    if (!nanoKernel)
+    // Replicate benchmark kernel arguments for cold-cache timing. Runs on
+    // bufferized memrefs so replicas are plain subviews (no allocs/copies).
+    // No-op unless the benchmark producer requested replication.
+    if (!nanoKernel) {
       pm.addPass(createBufferize());
+      pm.addPass(createReplicateBenchArgs());
+    }
 
     if (nanoKernel)
       // Lower Linalg to Vector.
--- a/lib/TPP/DefaultTppPasses.cpp
+++ b/lib/TPP/DefaultTppPasses.cpp
@@ -103,6 +103,10 @@ private:
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
     pm.addPass(createBufferize());
+    // Replicate benchmark kernel arguments for cold-cache timing. Runs on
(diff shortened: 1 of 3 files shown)


lib/TPP/DefaultTppPasses.cpp
```diff
@@ -103,6 +103,10 @@ struct DefaultTppPasses
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
     pm.addNestedPass<func::FuncOp>(createLoopInvariantCodeMotionPass());
     pm.addPass(createBufferize());
+    // Replicate benchmark kernel arguments for cold-cache timing. Runs on
+    // bufferized memrefs so replicas are plain subviews (no allocs/copies).
+    // No-op unless the benchmark producer requested replication.
+    pm.addPass(createReplicateBenchArgs());
```
rengolin, line 109: This function is only called for nano-kernels, not linalg-to-loops or xsmm. We need cold-cache for all. Why do you need to move this?

shahidact: It was my wrong interpretation of your query in last PR. I will reinstate it, thanks.

shahidact: Done.

test/Integration/TPP/tpp-run-replicate-bench-args.mlir
```diff
@@ -0,0 +1,46 @@
...
+
+// Replication loop feeds distinct memref.view slices to the kernel call.
+// CHECK: scf.for
+// CHECK:   %[[OFF0:.+]] = arith.muli
+// CHECK:   %[[V0:.+]] = memref.view %{{.+}}[%[[OFF0]]][] : memref<{{[0-9]+}}xi8> to memref<4x8xf32>
+// CHECK:   memref.view %{{.+}}[%{{.+}}][] : memref<{{[0-9]+}}xi8> to memref<8x4xf32>
+// CHECK:   memref.view %{{.+}}[%{{.+}}][] : memref<{{[0-9]+}}xi8> to memref<4x4xf32>
+// CHECK:   func.call @_entry(%[[V0]]
+
+// With -splat-to-random the fill loop emits a counter-based PRNG instead of 1.0.
```
rengolin, line 38: what's the benefit of testing random here?

shahidact: > what's the benefit of testing random here?

The RANDOM checks cover a different pass branch: instead of a hoisted arith.constant 1.0, emitRandomFloat emits the PRNG sequence `index_cast→muli→xori→bitcast`. This matters because all-zero/constant inputs let the FMA units run at an unrealistic clock, and the random path is what produces realistic [1,2) operands.

rengolin (approved):

PR #1178, opened 2026-08-24
[ReplicateBenchArgs] Fix crash on value-returning kernels

shahidact:
The pass erased the original kernel call unconditionally, but for kernels that return a result (e.g. an internally-allocated output buffer) the call was still consumed by trailing cleanup ops (extract_strided_metadata/dealloc) in the bench body, triggering 'operation destroyed but still has uses'.Rewire those uses to the replicated call.

### DIFF
--- a/lib/TPP/Transforms/ReplicateBenchArgs.cpp
+++ b/lib/TPP/Transforms/ReplicateBenchArgs.cpp
@@ -292,7 +292,23 @@ struct ReplicateBenchArgs
               /*sizes=*/ValueRange{});
         }
 
-        func::CallOp::create(bodyBuilder, loc, kernel, viewArgs);
+        auto newCall = func::CallOp::create(bodyBuilder, loc, kernel, viewArgs);
+
+        // When the kernel returns results (e.g. an internally-allocated output
+        // buffer), the original call is consumed by trailing cleanup ops
+        // (extract_strided_metadata/dealloc) in the bench body. Rewire those
+        // uses to the replicated call.
+        call.replaceAllUsesWith(newCall);
+        if (call->getNumResults() != 0) {
+          Operation *benchTerminator = call->getBlock()->getTerminator();
+          Operation *loopTerminator = loop.getBody()->getTerminator();
+          SmallVector<Operation *> cleanup;
+          for (Operation *op = call->getNextNode(); op && op != benchTerminator;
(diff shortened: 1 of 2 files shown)


lib/TPP/Transforms/ReplicateBenchArgs.cpp
```diff
@@ -292,7 +292,23 @@ struct ReplicateBenchArgs
...
-        func::CallOp::create(bodyBuilder, loc, kernel, viewArgs);
+        auto newCall = func::CallOp::create(bodyBuilder, loc, kernel, viewArgs);
+
+        // When the kernel returns results (e.g. an internally-allocated output
+        // buffer), the original call is consumed by trailing cleanup ops
+        // (extract_strided_metadata/dealloc) in the bench body. Rewire those
+        // uses to the replicated call.
+        call.replaceAllUsesWith(newCall);
+        if (call->getNumResults() != 0) {
+          Operation *benchTerminator = call->getBlock()->getTerminator();
```
rengolin, line 304: Isn't the call's block the same as the loop body?

shahidact: No, call->getBlock() is the block inside the perf.bench region (where the original kernel call still lives). For a perf.bench body, the terminator is perf.yield. So benchTerminator is the perf.yield op that closes the timed region. 

On the other hand, loop.getBody(), returns block of newly created scf::ForOp.

rengolin: I get it now, the `call` is the old call that was inside the perf loop, and the `newCall` is inside this new loop, which will need all its users to be moved inside the new loop.

rengolin: Can we guarantee that it's safe to move every user of the call to a new loop?

shahidact: Safety depends on the invariant implied by the perf.bench Op that it wraps around only the actual kernel to be timed as it ensures there is no other user except "extract_strided_metadata/dealloc" before it terminates with perf.yeild. However, I can harden this check in the pass.

shahidact: Done.

rengolin (approved):

PR #1180, opened 2026-08-26
Allow tiling of single-op elementwise linalg.generic body.

shahidact:
Earlier single op tling was not allowed conservatively. This causes
 sub-optimal code generation while writing the final output.This patch
 removes this restriction.

### DIFF
--- a/test/Passes/tile-elementwise-ops.mlir
+++ b/test/Passes/tile-elementwise-ops.mlir
@@ -41,12 +41,13 @@ func.func @tile_basic_elementwise(%arg0: tensor<32x64xf32>,
 
 // -----
 
-// Test 2: Single-op body (e.g., just yield of an arg) should NOT be tiled.
+// Test 2: Copy body (just a yield of an input) is tiled so it can fuse with
+// surrounding tiled ops.
 
 #map = affine_map<(d0, d1) -> (d0, d1)>
 
-func.func @no_tile_single_op_body(%arg0: tensor<32x64xf32>,
-                                   %arg1: tensor<32x64xf32>) -> tensor<32x64xf32> {
+func.func @tile_copy_body(%arg0: tensor<32x64xf32>,
+                          %arg1: tensor<32x64xf32>) -> tensor<32x64xf32> {
   %0 = linalg.generic {
     indexing_maps = [#map, #map],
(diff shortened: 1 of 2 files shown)


test/Passes/tile-elementwise-ops.mlir
```diff
@@ -0,0 +1,206 @@
...
+
+// CUSTOM-LABEL: func.func @tile_basic_elementwise
+// CUSTOM:         scf.for %{{.*}} step %c4
+// CUSTOM:           scf.for %{{.*}} step %c8
+// CUSTOM:             linalg.generic
+// CUSTOM-SAME:          ins({{.*}} : tensor<4x8xf32>, tensor<4x8xf32>)
+
+// -----
+
+// Test 2: Single-op body (e.g., just yield of an arg) should NOT be tiled.
```
rengolin: Why not? This is a copy, and if tiled, can fuse with the other surrounding tiled ops.

test/Passes/tile-elementwise-ops.mlir
```diff
@@ -0,0 +1,206 @@
...
+  ^bb0(%a: f32, %c: f32):
+    linalg.yield %a : f32
+  } -> tensor<32x64xf32>
+  return %0 : tensor<32x64xf32>
+}
+
+// CHECK-LABEL: func.func @no_tile_single_op_body
+// CHECK-NOT:     scf.for
+// CHECK:         linalg.generic
+// CHECK-SAME:      ins({{.*}} : tensor<32x64xf32>) outs({{.*}} : tensor<32x64xf32>)
```
rengolin: what about CUSTOM?

test/Passes/tile-elementwise-ops.mlir
shahidact: I was being conservative. It is a good point. I will include it.

shahidact: Updated accordingly.

test/Passes/tile-elementwise-ops.mlir
shahidact: Updated accordingly.

rengolin (approved):

PR #1186, opened 2026-08-31
[μKernels]: Extend register tiling to support tiling of matmul and batch matmul.

arun-thmn:
This patch rename `BrgemmLinalgTiling.cpp` pass to `RegisterTiling.cpp` and extends the `registerBlocking` to `matmul` and `batch_matmul` variants. 

Assisted by: Claude.


lib/TPP/Transforms/RegisterTiling.cpp
```diff
@@ -0,0 +1,216 @@
...
+    int reductionCount =
+        std::count(gemmIteratorTypes.begin(), gemmIteratorTypes.end(),
+                   utils::IteratorType::reduction);
+
+    int parallelCount =
+        std::count(gemmIteratorTypes.begin(), gemmIteratorTypes.end(),
+                   utils::IteratorType::parallel);
+
+    // Reject anything that is not a GEMM-like contraction with two inputs.
+    if (reductionCount == 0 || parallelCount < 2 ||
```
rengolin: This is not robust. We know three things: gemm, batch gemm, batch reduce gemm. These three have specific requirements (may even have upstream matchers in the morphism code). Anything else should bail.

lib/TPP/Transforms/RegisterTiling.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+    // Check dimensions are aligned with the iterator types. The batch
+    // dimensions may be parallel (batch matmul) or reduction (batch reduce).
+    if (gemmIteratorTypes[dimM] != mlir::utils::IteratorType::parallel ||
+        gemmIteratorTypes[dimN] != mlir::utils::IteratorType::parallel ||
+        gemmIteratorTypes[dimK] != mlir::utils::IteratorType::reduction)
+      return rewriter.notifyMatchFailure(
+          gemmOp, "Failed matching with iterator types and dimension");
+
+    // Every batch/batch-reduce dimension is tiled by 1.
```
rengolin: Comment and code in the wrong place. This comment is for the `for` loop. Always double check LLM generated comments as they're never good enough.

lib/TPP/Transforms/RegisterTiling.cpp
```diff
@@ -0,0 +1,216 @@
...
+    SmallVector<int64_t> tileSizes(gemmIteratorTypes.size(), 0);
+    tileSizes[dimM] = options.registerTileShape[0];
+    tileSizes[dimN] = options.registerTileShape[1];
+    for (unsigned dim : batchDims)
+      tileSizes[dim] = 1;
+
+    // Interchange: any extra (outer) batch dimensions come first, followed by
+    // the M, N, innermost-batch, K (and vnni) loop ordering.
```
rengolin: The M,N are above, as mentioned. Bad comments

test/Passes/pass-tile-brgemm-linalg-matmul.mlir
```diff
@@ -0,0 +1,501 @@
+// RUN: tpp-opt %s --tile-gemm="registerBlocking=8,32,1" --split-input-file  | FileCheck -check-prefix=CONF1 %s
+// RUN: tpp-opt %s --tile-gemm="registerBlocking=32,32,32" --split-input-file  | FileCheck -check-prefix=CONF2 %s
+
+module {
+  func.func @gemm_do_register_tiling(%arg0: memref<16x32x16x32xf32>, %arg1: memref<32x32x32x32xf32>, %arg2: memref<16x32x16x32xf32>) {
```
rengolin:
```suggestion
  func.func @brgemm_do_register_tiling(%arg0: memref<16x32x16x32xf32>, %arg1: memref<32x32x32x32xf32>, %arg2: memref<16x32x16x32xf32>) {
```

test/Passes/Linalg/pass-tile-linalg-matmul.mlir
```diff
@@ -0,0 +1,501 @@
+// RUN: tpp-opt %s --tile-gemm="registerBlocking=8,32,1" --split-input-file  | FileCheck -check-prefix=CONF1 %s
```
rengolin, line 1: there are no tests that run on both confs, so split this into two tests

lib/TPP/Transforms/RegisterTiling.cpp
arun-thmn: Okay. Cleaned-up the code now.
Using `linalg::isaContractionOpInterface` upstream to check whether is a contraction or not. 

So, the count on the `parallel`/`reduction` dim count is removed.

lib/TPP/Transforms/RegisterTiling.cpp
arun-thmn: Okay. I kind of re-phrased the comments. Thre are two parts:

- Set the tiling sizes
- Order of the loop structure M->N->batch->K->vnni.

lib/TPP/Transforms/RegisterTiling.cpp
arun-thmn: Same comments as above. This `M`, `N` is for loop order, earlier one is for `tile` size.

test/Passes/pass-tile-brgemm-linalg-matmul.mlir
arun-thmn: Fixed it.

test/Passes/Linalg/pass-tile-linalg-matmul.mlir
arun-thmn: Last three tests uses both the `CONFIG1/2`.

rengolin (approved):

adam-smnk (approved):

PR #1189, opened 2026-09-06
[TensorInit] Replace std::random with a reproducible PRNG

alheinecke:
Seeded tensor initialization produced different values depending on the C++ standard library it was built against, which made three integration tests fail on Fedora 44 while passing on Rocky Linux with the exact same sources, the same clang 22.1.8 and the same seed:

  TPP_OPT :: Integration/TPP/tpp-run-quant-initialize.mlir
  TPP_OPT :: Integration/TPP/tpp-run-splat-tensor.mlir
  TPP_OPT :: Integration/XSMM/xsmm-fusion.mlir

Root cause
----------
The C++ standard fixes only the statistical properties of the std::*_distribution templates, not the value sequence they produce for a given seed, and std::default_random_engine is implementation defined as well. The seed -> value mapping is therefore owned by the standard library, not by the compiler. clang ships no C++ runtime on Linux and picks up the system libstdc++ (Fedora 44 selects GCC 16), so the same sources built with the same compiler yield different "random" data on different distros.

The failures were not memory corruption, stack limits or uninitialized reads: the generated values were all in range and plausible, just not bit-identical, e.g. for tpp-run-splat-tensor.mlir

  expected  9.62642952E-4, 0.179147944, 0.939454615
  actual    9.70184803E-4, 0.180558503, 0.946851909

and every affected test was one that hardcodes exact RNG-derived values.

Change
------
Add TensorInitRNG, a self-contained per-instance xoshiro128+ engine that replicates libxsmm's generator (libxsmm/src/libxsmm_rng.c): same 16-lane state, same seed table, same jump and same float conversion, so uniformFloat() reproduces libxsmm_rng_f32_seq() element for element.

libxsmm's RNG is reused algorithmically rather than by calling into the library because libxsmm_rng_set_seed/libxsmm_rng_f32_seq operate on global state. Keeping the state per instance preserves the existing semantics of one independently seeded generator per initializer and avoids introducing a shared mutable global into the pass pipeline.

The distributions layered on top only use exactly rounded IEEE-754 operations:

  std::uniform_real_distribution<float>  -> uniformFloat()
  std::uniform_int_distribution<uint64_t>-> uniformInt(lo, hi)
  std::normal_distribution<float>        -> normalFloat(mean, stddev)
  std::binomial_distribution<uint64_t>   -> binomial(n, p)

uniformInt() uses rejection sampling to stay unbiased. normalFloat() uses the Irwin-Hall construction (sum of 12 uniforms) rather than Box-Muller or Marsaglia polar on purpose: those need log()/cos() from libm, which is not bit-reproducible across libm versions and would have moved the same class of bug from libstdc++ to libm instead of removing it. Irwin-Hall needs only exact float additions. It is a truncated (+/-6 sigma) approximation, which is adequate here because this data only has to be a reasonable distribution for layout testing, and the float path clamps to [0, 1] with sigma = 0.2 anyway.

Verification
------------
A standalone harness linked against the built libxsmm.a compared 4096 draws per seed and confirmed uniformFloat() is bit-identical to libxsmm_rng_f32_seq() for seeds 0, 1, 42, 123 and 999983. libxsmm's scalar and AVX-512 paths agree by construction (the vector path round-trips its state through the scalar lanes), so the sequence is also independent of the host CPU.

Golden values regenerated for the three tests above. No other test needed updating, which confirms these were the only ones asserting on exact generated values; the remaining seeded tests compare structurally or with tolerance. check-tpp is green: 194 passed, 17 unsupported, 2 expectedly failed, 0 failed.

Note
----
tools/mlir-gen/MLIRGen.cpp still seeds rand() via srand(). glibc's rand() happens to be stable across versions so it does not trigger this particular Fedora/Rocky divergence, but it is the same class of portability hazard and is left for a follow-up.


rengolin (approved):
We don't care about stability, security or homogeneous distribution of our RNGs anyway, this is more stable, for sure.

PR #1190, opened 2026-09-09
Add nano-gemm K-cache-blocking pass.

shahidact:
Introduce a post-nano AMX K-cache-blocking transformation that hoists a K-block loop outside the parallel forall so the reused B panel stays L2-resident across the reduction:

- NanoGemmKCacheBlocking pass: bf16 path threads the C tile in place, i8 quant path threads the wide i32 accumulator and runs the requant epilogue once after the K-loop.
- Expose --k-cache-blocking through the tpp-run CLI, forwarded into the nano vectorization path. Opt-in (default 0) i.e. disabled by default.


shahidact:
Closing this as this is not required.

PR #1191, opened 2026-09-13
Add tensor-level GEMM K-cache-blocking pass

shahidact:
Introduce a two-level cache-blocking scheme for a batch-reduce GEMM, enabled by --k-cache-blocking (off by default):

- Level 1 (M/N cache panels): TileConsumerAndFuseProducers tiles the spatial M/N dims by --m-cache-panel/--n-cache-panel so each spatial scf.forall iteration owns a CM x CN group of register tiles and fuses the down-convert epilogue.
- Level 2 (K cache blocks): the new GemmKCacheBlocking pass tiles the reduction (K / batch-reduce) dimension inside that panel and threads a small high-precision (f32/i32) accumulator across the K-blocks, so the reused A/B panels stay L2-resident instead of thrashing L3.


shahidact:
Closing this PR as it is not required now.

PR #1194, opened 2026-09-22
[mlir-gen]: Fix i8 requantization output scale(oS).

shahidact:
In requantization epilogue, multiply the dequantized i32 accumulator by reciprocal(oS) instead of (oS) so the per-output-channel output scale is applied correctly. Update the mlir-gen-matmul requant CHECK lines to match the new divf sequence. The intent is to lower this sequence to a reciprocal instruction.


tools/mlir-gen/MLIRGen.cpp
```diff
@@ -885,7 +885,16 @@ Value MLIRGenerator::requantizeGemm(LayerArgs &args, Value chain) {
             Value scale =
                 arith::MulFOp::create(nestedBuilder, nestedLoc, inS, wS)
                     ->getResult(0);
-            scale = arith::MulFOp::create(nestedBuilder, nestedLoc, scale, oS)
+            // Apply the output scale after reciprocating it.
```
alheinecke, line 888: either we divide directly or we use reciprocal (one 1 instruction). can you elaborate on why using div to create a reciprocal and then multiplying with it?

alheinecke (changes requested):
please clarify the intend of this change, e.g. this new 2 operand sequence lower the the reciprocal instructions?

tools/mlir-gen/MLIRGen.cpp
shahidact: ReciprocalOp is not available in Arith dialect.

